Issue #1, opened 2017-12-05
Config module should support overrides and default values


JGulbronson:
Currently the `ConfigModule` only supports loading one file, and does not respect default values of data classes. There should be a simple naming scheme for overriding configurations, i.e. `foo-production.yaml` overrides `foo-common.yaml`, and a config class similar to the following:

```kotlin
data class FooConfig : Config (
  val port : Int = 8080,
  val max_connections : Int = 300
)
```

Should by default have `port` and `max_connections` set without explicit configuration, to reduce the verbosity of config yaml files.

JGulbronson:
Upon local testing, it appears the library we're currently using already supports default values. As such, we just need to support overrides, which needs #2 to land first.

PR #17, opened 2017-12-05
Add testing, development, staging, production Environment enums

### LINKED ISSUE #2: Misk should support some concept of environments
This will be useful when determining which yaml files for parse for config, as mentioned in #1. Examples of environments are testing, staging, production, etc.

A simple implementation could just be a String that gets bound so you can inject it like
```java
@Inject @Env String environment
```

Misk can provide a module that binds based of an environment variable. Alternatively a module can statically provide an environment, and then an app can select which to install.

JGulbronson:
@maniksurtani this is super simple, but should get the job done and will unblock #1 

This has the potential to be pretty barebones, or more constrictive, so curious on your thoughts.

Fixes #2


maniksurtani:
I think this works for now. Let's look at making it an interface later.

PR #18, opened 2017-12-06
Add Dockerfile and instructions

maniksurtani:
@JGulbronson the update to `ConfigModule.kt` is to allow for reading the YAML file from inside a JAR.


JGulbronson:
Hit merge for you @maniksurtani to avoid any more merge conflicts.

PR #20, opened 2017-12-06
Allow Environment to be set using an environment variable of key ENV

JGulbronson:


misk/src/main/kotlin/misk/environment/EnvironmentModule.kt
```diff
@@ -1,10 +1,26 @@
...
     override fun configure() {
         bind(Environment::class.java).toInstance(environment)
     }
 
+    companion object {
+        @JvmStatic
+        fun fromEnvironmentVariable(): EnvironmentModule {
+            val environmentName = System.getenv("ENV")
+            val environment = if (environmentName != null) {
+                Environment.valueOf(environmentName)
```
maniksurtani: We'd want to ensure `environmentName` is upper-cased since `Enum#valueOf()` isn't case-insensitive.

JGulbronson: Ah gotcha, so rather than forcing the variable value to be uppercased, we do it for them? Sounds good, will update.

misk/src/main/kotlin/misk/environment/EnvironmentModule.kt
```diff
@@ -1,10 +1,26 @@
...
     override fun configure() {
         bind(Environment::class.java).toInstance(environment)
     }
 
+    companion object {
+        @JvmStatic
+        fun fromEnvironmentVariable(): EnvironmentModule {
+            val environmentName = System.getenv("ENV")
+            val environment = if (environmentName != null) {
+                Environment.valueOf(environmentName.toUpperCase())
```
swankjesse: Not a fan of converting it to upper case here actually. We want our users to just use the correct case don’t we? I think fewer clever behaviors is better.

Independently, you never want `toUpperCase()`; that API is very toxic and should never be used. Instead use `toUpperCase(Locale.US)`. Otherwise `staging` is converted to `STAGİNG` in Turkey.

misk/src/main/kotlin/misk/environment/EnvironmentModule.kt
```diff
@@ -1,10 +1,26 @@
...
+
+class EnvironmentModule(val environment: Environment) : AbstractModule() {
     override fun configure() {
         bind(Environment::class.java).toInstance(environment)
     }
 
+    companion object {
+        @JvmStatic
+        fun fromEnvironmentVariable(): EnvironmentModule {
+            val environmentName = System.getenv("ENV")
```
swankjesse: Should we call this `ENV` or `ENVIRONMENT` ? I’m partial to `ENVIRONMENT` because I hate having multiple different names for the same thing.

misk/src/main/kotlin/misk/environment/EnvironmentModule.kt
JGulbronson: On second thought, I think you're right. We shouldn't try to be super clever, since that's what confuses people when things don't work. We want to enforce the correct usage rather than trying to consider all possible edge cases. Will fix in a followup.

misk/src/main/kotlin/misk/environment/EnvironmentModule.kt
JGulbronson: Let's go environment. I tried to keep everything as `Environment` instead of `ENV` in the code, but missed it here. Will fix.

PR #21, opened 2017-12-07
Rename ENV to ENVIRONMENT to keep naming consistent

JGulbronson:
Also doesn't upper case the value of the key, to make sure we don't do anything "magical" for the user. The less layers the better.

Followup to #20


swankjesse:
Thanks much.

PR #22, opened 2017-12-07
Add health check and liveness endpoints

JGulbronson:
This borrows from the Kubernetes notion of [liveness and readiness](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-probes/), to help distinguish between when a service should stop serving traffic vs needs to be restarted.

### DIFF
--- a/exemplar/src/main/kotlin/com/squareup/exemplar/ExemplarModule.kt
+++ b/exemplar/src/main/kotlin/com/squareup/exemplar/ExemplarModule.kt
@@ -2,7 +2,9 @@ package com.squareup.exemplar
 
 import com.google.inject.AbstractModule
 import misk.web.WebActionModule
+import misk.web.actions.ReadinessCheckAction
 import misk.web.actions.InternalErrorAction
+import misk.web.actions.LivenessCheckAction
 import misk.web.actions.NotFoundAction
 
 class ExemplarModule : AbstractModule() {
@@ -10,6 +12,8 @@ class ExemplarModule : AbstractModule() {
         install(WebActionModule.create<HelloWebAction>())
         install(WebActionModule.create<HelloWebPostAction>())
         install(WebActionModule.create<InternalErrorAction>())
+        install(WebActionModule.create<ReadinessCheckAction>())
+        install(WebActionModule.create<LivenessCheckAction>())
         install(WebActionModule.create<NotFoundAction>())
     }
 }
--- a/misk/src/main/kotlin/misk/healthchecks/HealthCheck.kt
+++ b/misk/src/main/kotlin/misk/healthchecks/HealthCheck.kt
@@ -1,0 +1,26 @@
+package misk.healthchecks
+
+import misk.web.actions.ReadinessCheckAction
+
+/**
+ * Allows users to define custom health checks. An app with a failing HealthCheck will fail the
+ * readiness check in [ReadinessCheckAction], indicating that the app should not accept traffic.
+ */
+interface HealthCheck {
+    /**
+     * Computes whether a component of an application is healthy. For example, an implementing class
+     * can check database connectivity.
+     */
+    fun status(): HealthStatus
+}
+
+data class HealthStatus(
+    val isHealthy: Boolean,
+    val messages: List<String>
+) {
+    companion object {
+        fun healthy() = HealthStatus(true, listOf())
+
+        fun unhealthy(vararg messages: String) = HealthStatus(false, messages.asList())
+    }
+}
--- a/misk/src/main/kotlin/misk/web/actions/LivenessCheckAction.kt
+++ b/misk/src/main/kotlin/misk/web/actions/LivenessCheckAction.kt
@@ -1,0 +1,27 @@
+package misk.web.actions
+
+import com.google.common.util.concurrent.Service
+import com.google.common.util.concurrent.Service.State
+import misk.web.Get
+import misk.web.PlaintextResponseBody
+import misk.web.Response
+import javax.inject.Inject
+import javax.inject.Singleton
+
+@Singleton
+class LivenessCheckAction : WebAction {
+    @Inject private lateinit var services: MutableSet<Service>
+
+    @Get("/_liveness")
+    @PlaintextResponseBody
+    fun livenessCheck(): Response<String> {
+        val isAlive = services.all {
+            it.state() != State.FAILED && it.state() != State.TERMINATED
+        }
+        // TODO(jgulbronson) - Should return an empty body
+        return Response(
+                "",
+                statusCode = if (isAlive) 200 else 503
+        )
+    }
+}
--- a/misk/src/main/kotlin/misk/web/actions/ReadinessCheckAction.kt
+++ b/misk/src/main/kotlin/misk/web/actions/ReadinessCheckAction.kt
@@ -1,0 +1,28 @@
+package misk.web.actions
+
+import com.google.common.util.concurrent.Service
+import misk.healthchecks.HealthCheck
+import misk.web.Get
+import misk.web.JsonResponseBody
+import misk.web.Response
+import javax.inject.Inject
+import javax.inject.Singleton
+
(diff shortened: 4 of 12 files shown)


misk/src/main/kotlin/misk/config/HealthCheck.kt
```diff
@@ -0,0 +1,39 @@
...
+     * the ability to accept traffic, for a database service this would indicate a valid connection
+     * that can be used to query the database.
+     */
+    fun isReady(): Boolean
+
+    /**
+     * Returns true if the service has the ability to operate normally now or in the future. Returns
+     * false is the service has failed and requires a restart.
+     */
+    fun isAlive(): Boolean
```
mmihic: strongly prefer having this return a message with the check, similar to the way our internal status checks and dropwizard healthchecks work. that way tools and human observers can easily see exactly which checks are failing. also need to allow anyone to install health checks, not just base them on the state of a single running service.

misk/src/main/kotlin/misk/config/HealthChecksModule.kt
```diff
@@ -0,0 +1,16 @@
+package misk.config
+
+import com.google.inject.AbstractModule
+import com.google.inject.multibindings.Multibinder
+import misk.inject.newMultibinder
+
+class HttpHealthChecks(
```
mmihic: This should be decoupled from HTTP exposure. Also, should allow sub-components to register their own healthchecks, rather than the requiring the application install them all manually via this module

JGulbronson: Sub-components can install their own healthchecks, since they're just adding to the Multibinder. Though this class name is wrong, I'll rename to `HealthChecksModule`.

JGulbronson:
Will add a Status endpoint in a followup, which will contain human readable information.

misk/src/main/kotlin/misk/web/actions/ReadinessCheckAction.kt
```diff
@@ -0,0 +1,30 @@
...
+import com.google.common.util.concurrent.Service
+import misk.web.Get
+import misk.web.JsonResponseBody
+import misk.web.Response
+import javax.inject.Inject
+import javax.inject.Singleton
+
+@Singleton
+class ReadinessCheckAction : WebAction {
+    @Inject lateinit var services: MutableSet<Service>
```
swankjesse: This wants the HealthCheck interface

JGulbronson: Added

exemplar/src/main/kotlin/com/squareup/exemplar/ExemplarModule.kt
```diff
@@ -2,14 +2,18 @@ package com.squareup.exemplar
...
 import misk.web.actions.InternalErrorAction
+import misk.web.actions.LivenessCheckAction
 import misk.web.actions.NotFoundAction
 
 class ExemplarModule : AbstractModule() {
     override fun configure() {
         install(WebActionModule.create<HelloWebAction>())
         install(WebActionModule.create<HelloWebPostAction>())
         install(WebActionModule.create<InternalErrorAction>())
+        install(WebActionModule.create<ReadinessCheckAction>())
```
mmihic: Do we want these added by default (I think so)

misk/src/main/kotlin/misk/web/actions/ReadinessCheckAction.kt
```diff
@@ -0,0 +1,42 @@
...
+    @Inject lateinit var healthChecks: MutableSet<HealthCheck>
+
+    @Get("/_readiness")
+    @JsonResponseBody
+    fun readinessCheck(): Response<String> {
+        var isReady = true
+        for (service in services) {
```
mmihic: fwiw you could just do:

``
val isReady = services.all { it.isRunning } && healthChecks.all { it.status().isHealthy }
``

short-circuiting will work properly

misk/src/main/kotlin/misk/web/actions/LivenessCheckAction.kt
```diff
@@ -0,0 +1,32 @@
...
+import javax.inject.Singleton
+
+@Singleton
+class LivenessCheckAction : WebAction {
+    @Inject private lateinit var services: MutableSet<Service>
+
+    @Get("/_liveness")
+    @PlaintextResponseBody
+    fun livenessCheck(): Response<String> {
+        var isAlive = true
```
mmihic: ditto here, can eliminate the mutable var by doing

``
val isAlive = services.any { it.state() == Status.FAILED || it.state() == State.TERMINATED }
``

misk/src/main/kotlin/misk/healthchecks/HealthCheck.kt
```diff
@@ -0,0 +1,21 @@
...
+     * Computes whether a component of an application is healthy. For example, an implementing class
+     * can check database connectivity. As this method may involve a network call, callers should
+     * cache the response appropriately.
+     */
+    fun status(): HealthStatus
+}
+
+data class HealthStatus(
+    val isHealthy: Boolean,
+    val messages: List<String>
```
mmihic: It'd be nice to have companion functions for creating healthy and unhealthy status, so you don't have to remember whether the boolean means healthy or not when you see it written in code. e.g.

````
companion object {
   fun unhealthy(vararg messages: String) = HealthStatus(false, listOf(messages))
   fun healthy() = HealthStatus(true, listOf())
}
````

File: exemplar/src/main/kotlin/com/squareup/exemplar/ExemplarModule.kt (hunk 4 above)
Comment (JGulbronson) [reply exemplar/src/main/kotlin/com/squareup/exemplar/ExemplarModule.kt]: Maybe, I don't know if we have too much in the way of defaults so far. Even `InternalErrorAction` isn't a default. I think in a followup, we can define a `DefaultWebActionsModule` which installs some sane defaults for users.

Comment (swankjesse) [reply exemplar/src/main/kotlin/com/squareup/exemplar/ExemplarModule.kt]: yeah!

File: misk/src/main/kotlin/misk/healthchecks/HealthCheck.kt [hunk 8]
```diff
@@ -0,0 +1,21 @@
...
+import misk.web.actions.ReadinessCheckAction
+
+/**
+ * Allows users to define custom health checks. An app with a failing HealthCheck will fail the
+ * readiness check in [ReadinessCheckAction], indicating that the app should not accept traffic.
+ */
+interface HealthCheck {
+    /**
+     * Computes whether a component of an application is healthy. For example, an implementing class
+     * can check database connectivity. As this method may involve a network call, callers should
```
swankjesse: Remove the “As this method...” sentence; it’s prescriptive in the wrong way.

misk/src/test/kotlin/misk/healthchecks/FakeHealthCheck.kt
```diff
@@ -0,0 +1,13 @@
+package misk.healthchecks
+
+import javax.inject.Singleton
+
+@Singleton
+class FakeHealthCheck : HealthCheck {
+    var isHealthy = true
+    val messages = listOf<String>()
+
+    override fun status(): HealthStatus {
```
swankjesse: expression function?

JGulbronson: Expressed.

misk/src/main/kotlin/misk/healthchecks/HealthCheck.kt
JGulbronson: Removed.

misk/src/main/kotlin/misk/healthchecks/HealthCheck.kt
JGulbronson: Added

PR #25, opened 2017-12-08
Add metrics interceptor for web requests

mmihic:
Tracks incoming requests, response codes by family and individual code,
and total application level response times

### DIFF
--- a/misk/src/main/kotlin/misk/config/AppName.kt
+++ b/misk/src/main/kotlin/misk/config/AppName.kt
@@ -1,8 +1,8 @@
 package misk.config
 
-import com.google.inject.BindingAnnotation
+import javax.inject.Qualifier
 
-@BindingAnnotation
+@Qualifier
 @Target(
         AnnotationTarget.FIELD,
         AnnotationTarget.PROPERTY,
--- a/misk/src/main/kotlin/misk/web/interceptors/MetricsInterceptor.kt
+++ b/misk/src/main/kotlin/misk/web/interceptors/MetricsInterceptor.kt
@@ -1,0 +1,30 @@
+package misk.web.interceptors
+
+import misk.Action
+import misk.Chain
+import misk.Interceptor
+import misk.metrics.Metrics
+import misk.metrics.MetricsScope
+import misk.web.Response
+import javax.inject.Inject
+import javax.inject.Singleton
+
+internal class MetricsInterceptor internal constructor(val scope: MetricsScope) : Interceptor {
+
+    @Singleton
+    class Factory @Inject constructor(val m: Metrics) : Interceptor.Factory {
+        override fun create(action: Action): Interceptor? =
+                MetricsInterceptor(m.scope("web.${action.name}"))
+    }
+
+    override fun intercept(chain: Chain): Any? {
+        scope.counter("requests").inc()
+        val result = scope.timer("timing").time { chain.proceed(chain.args) }
+        if (result is Response<*>) {
+            scope.counter("responses.${result.statusCode / 100}xx").inc()
+            scope.counter("responses.${result.statusCode}").inc()
+        }
+
+        return result
+    }
+}
--- a/misk/src/test/kotlin/misk/web/interceptors/MetricsInterceptorTest.kt
+++ b/misk/src/test/kotlin/misk/web/interceptors/MetricsInterceptorTest.kt
@@ -1,0 +1,70 @@
+package misk.web.interceptors
+
+import com.google.common.truth.Truth.assertThat
+import misk.asAction
+import misk.metrics.Metrics
+import misk.metrics.MetricsModule
+import misk.testing.MiskTestRule
+import misk.web.Get
+import misk.web.Response
+import misk.web.actions.WebAction
+import misk.web.actions.asChain
+import org.junit.Before
+import org.junit.Rule
+import org.junit.Test
+import javax.inject.Inject
+
+class MetricsInterceptorTest {
+    @get:Rule
+    val miskTestRule = MiskTestRule(MetricsModule())
+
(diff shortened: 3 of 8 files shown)


misk/src/main/kotlin/misk/config/AppName.kt
```diff
@@ -3,5 +3,10 @@ package misk.config
 import com.google.inject.BindingAnnotation
 
 @BindingAnnotation
```
swankjesse: We later wanna switch to `@Qualifier`

misk/src/main/kotlin/misk/metrics/Metrics.kt
```diff
@@ -0,0 +1,26 @@
+package misk.metrics
+
+import com.codahale.metrics.Counter
+import com.codahale.metrics.Gauge
+import com.codahale.metrics.Histogram
+import com.codahale.metrics.MetricRegistry
+import com.codahale.metrics.Timer
+import java.util.*
```
swankjesse: Wildcard!

misk/src/test/kotlin/misk/web/interceptors/MetricsInterceptorTest.kt
```diff
@@ -0,0 +1,71 @@
...
+    assertThat(invoke(200).statusCode).isEqualTo(200)
+    assertThat(invoke(202).statusCode).isEqualTo(202)
+    assertThat(invoke(404).statusCode).isEqualTo(404)
+    assertThat(invoke(403).statusCode).isEqualTo(403)
+    assertThat(invoke(403).statusCode).isEqualTo(403)
+  }
+
+  @Test
+  fun requests() {
+    assertThat(metrics.counters["web.TestAction.requests"]!!.count).isEqualTo(6)
```
swankjesse: Nice

misk/src/test/kotlin/misk/web/interceptors/MetricsInterceptorTest.kt
```diff
@@ -0,0 +1,71 @@
...
+    val chain = testAction.asChain(actionFn, listOf(desiredStatusCode), metricsInterceptor)
+
+    @Suppress("UNCHECKED_CAST")
+    return chain.proceed(chain.args) as Response<String>
+  }
+}
+
+internal class TestAction : WebAction {
+  @Get("/call/{result}")
+  fun call(desuredStatusCode: Int): Response<String> {
```
swankjesse: desired

misk/src/main/kotlin/misk/metrics/Metrics.kt
mmihic: argh. I finally found where it is automatically using wildcards for java.util.

misk/src/main/kotlin/misk/web/interceptors/MetricsInterceptor.kt
```diff
@@ -0,0 +1,30 @@
...
+import misk.metrics.Metrics
+import misk.metrics.MetricsScope
+import misk.web.Response
+import javax.inject.Inject
+import javax.inject.Singleton
+
+internal class MetricsInterceptor internal constructor(val scope: MetricsScope) : Interceptor {
+
+  @Singleton
+  class Factory @Inject constructor(val m: Metrics) : Interceptor.Factory {
```
JGulbronson: In general I'm opposed to single letter variable names, though this is such a small usage it's not a big deal. My vote would be `m` => `metrics` though.

misk/src/main/kotlin/misk/web/interceptors/MetricsInterceptor.kt
```diff
@@ -0,0 +1,30 @@
...
+    override fun create(action: Action): Interceptor? =
+        MetricsInterceptor(m.scope("web.${action.name}"))
+  }
+
+  override fun intercept(chain: Chain): Any? {
+    scope.counter("requests").inc()
+    val result = scope.timer("timing").time { chain.proceed(chain.args) }
+    if (result is Response<*>) {
```
JGulbronson: Why do we only want to record the status code if the result is a `Response<*>`?

I think we might want this interceptor to run after the `BoxResponseInterceptor`, so we don't ignore requests where we just return a value. See the `HelloWebPostAction` in the exemplar Kotlin app for a case we don't return a `Response` from an action.

misk/src/main/kotlin/misk/web/interceptors/MetricsInterceptor.kt
```diff
@@ -0,0 +1,30 @@
...
+      scope.counter("responses.${result.statusCode / 100}xx").inc()
+      scope.counter("responses.${result.statusCode}").inc()
+    }
+
+    return result
+  }
+}
```
JGulbronson: nit: newline

misk/src/test/kotlin/misk/web/interceptors/ProtobufInterceptorFactoryTest.kt
```diff
@@ -18,50 +17,45 @@ import okio.Pipe
...
-    @Test
-    fun test() {
-        val action = protobufAction::hello.asAction()
+  @Test
+  fun test() {
+    val action = ProtobufAction::hello.asAction()
 
-        val boxResponseInterceptor: Interceptor = boxResponseInterceptorFactory.create(action)!!
-        val protobufInterceptor: ProtobufInterceptor<*> = protobufInterceptorFactory.create(action)!!
+    val boxResponseInterceptor: Interceptor = boxResponseInterceptorFactory.create(action)!!
```
JGulbronson: This file's diff seems to be because the formatting changed from 4 to 2 spaces. I _think_ 4 is what we have elsewhere, so we probably want to stay consistent.

PR #28, opened 2017-12-08
Remove slf4j implementation from Misk classpath

JGulbronson:
The implementation dependency should come from the end-user, not the library, otherwise it defeats the purpose of using a logging facade.


maniksurtani:
Although it's arguable that a default (that logs to stdout) should be in Misk, to help with tests, etc.

JGulbronson:
Yep, I put it as a `testCompile` dependency instead, so that we still have logging in tests.

PR #29, opened 2017-12-10
Add metrics backends

mmihic:
Graphite + StackDriver + web endpoint exposing metrics as JSON over HTTP
for manual expection and pull-based tooling such as Prometheus

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -24,6 +24,8 @@ ext {
           jacksonKotlin: 'com.fasterxml.jackson.module:jackson-module-kotlin:2.9.2',
           metricsParent: 'io.dropwizard.metrics:metrics-parent:3.1.0',
           metricsCore: 'io.dropwizard.metrics:metrics-core:3.1.0',
+          metricsGraphite: 'io.dropwizard.metrics:metrics-graphite:3.1.0',
+          metricsStackDriver: 'com.google.apis:google-api-services-monitoring:v3-rev425-1.23.0',
   ]
 
   isCi = "true".equals(System.getenv('CI'))
--- a/exemplar/src/main/java/com/squareup/exemplar/ExemplarJavaModule.java
+++ b/exemplar/src/main/java/com/squareup/exemplar/ExemplarJavaModule.java
@@ -1,6 +1,7 @@
 package com.squareup.exemplar;
 
 import com.google.inject.AbstractModule;
+import misk.metrics.web.MetricsJsonAction;
 import misk.web.WebActionModule;
 import misk.web.actions.InternalErrorAction;
 import misk.web.actions.NotFoundAction;
@@ -8,6 +9,7 @@ import misk.web.actions.NotFoundAction;
 public class ExemplarJavaModule extends AbstractModule {
   @Override protected void configure() {
     install(WebActionModule.create(HelloJavaAction.class));
+    install(WebActionModule.create(MetricsJsonAction.class));
     install(WebActionModule.create(InternalErrorAction.class));
     install(WebActionModule.create(NotFoundAction.class));
   }
--- a/exemplar/src/main/kotlin/com/squareup/exemplar/ExemplarModule.kt
+++ b/exemplar/src/main/kotlin/com/squareup/exemplar/ExemplarModule.kt
@@ -1,6 +1,7 @@
 package com.squareup.exemplar
 
 import com.google.inject.AbstractModule
+import misk.metrics.web.MetricsJsonAction
 import misk.web.WebActionModule
 import misk.web.actions.ReadinessCheckAction
 import misk.web.actions.InternalErrorAction
@@ -11,6 +12,7 @@ class ExemplarModule : AbstractModule() {
     override fun configure() {
         install(WebActionModule.create<HelloWebAction>())
         install(WebActionModule.create<HelloWebPostAction>())
+        install(WebActionModule.create<MetricsJsonAction>())
         install(WebActionModule.create<InternalErrorAction>())
         install(WebActionModule.create<ReadinessCheckAction>())
         install(WebActionModule.create<LivenessCheckAction>())
--- a/misk/build.gradle
+++ b/misk/build.gradle
@@ -34,6 +34,8 @@ dependencies {
   compile dep.jacksonKotlin
   compile dep.metricsCore
   compile dep.metricsParent
+  compile dep.metricsGraphite
+  compile dep.metricsStackDriver
 
   testCompile dep.junit
   testCompile dep.truth
--- a/misk/src/main/kotlin/misk/metrics/MetricsConfig.kt
+++ b/misk/src/main/kotlin/misk/metrics/MetricsConfig.kt
@@ -1,0 +1,7 @@
+package misk.metrics
+
+import misk.config.Config
+import misk.metrics.backends.MetricsBackendConfig
+
+data class MetricsConfig(val backends: MetricsBackendConfig) : Config
+
--- a/misk/src/main/kotlin/misk/metrics/backends/MetricsBackendConfig.kt
+++ b/misk/src/main/kotlin/misk/metrics/backends/MetricsBackendConfig.kt
@@ -1,0 +1,10 @@
+package misk.metrics.backends
+
+import misk.config.Config
+import misk.metrics.backends.graphite.GraphiteBackendConfig
+import misk.metrics.backends.stackdriver.StackDriverBackendConfig
+
+data class MetricsBackendConfig(
+        val graphite: GraphiteBackendConfig?,
+        val stack_driver: StackDriverBackendConfig?
+) : Config
--- a/misk/src/main/kotlin/misk/metrics/backends/graphite/GraphiteReporterService.kt
+++ b/misk/src/main/kotlin/misk/metrics/backends/graphite/GraphiteReporterService.kt
@@ -1,0 +1,14 @@
+package misk.metrics.backends.graphite
+
+import com.codahale.metrics.graphite.GraphiteReporter
+import com.google.common.util.concurrent.AbstractIdleService
+import java.util.concurrent.TimeUnit
+import javax.inject.Inject
+
+internal class GraphiteReporterService @Inject internal constructor(
+        private val config: GraphiteBackendConfig,
+        private val reporter: GraphiteReporter
+) : AbstractIdleService() {
+    override fun startUp() = reporter.start(config.interval.toMillis(), TimeUnit.MILLISECONDS)
+    override fun shutDown() = reporter.stop()
+}
--- a/misk/src/main/kotlin/misk/metrics/backends/stackdriver/StackDriverBackendConfig.kt
+++ b/misk/src/main/kotlin/misk/metrics/backends/stackdriver/StackDriverBackendConfig.kt
@@ -1,0 +1,12 @@
+package misk.metrics.backends.stackdriver
+
+import misk.config.Config
+import java.time.Duration
+
+data class StackDriverBackendConfig(
+        val project_id: String,
+        val service_path: String,
+        val root_url: String,
+        val interval: Duration,
+        val batch_size: Int = 200
+) : Config
--- a/misk/src/main/kotlin/misk/metrics/backends/stackdriver/StackDriverBackendModule.kt
+++ b/misk/src/main/kotlin/misk/metrics/backends/stackdriver/StackDriverBackendModule.kt
@@ -1,0 +1,33 @@
+package misk.metrics.backends.stackdriver
+
+import com.google.api.client.googleapis.javanet.GoogleNetHttpTransport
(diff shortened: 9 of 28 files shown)


misk/src/main/kotlin/misk/metrics/backends/MetricsBackendConfig.kt
```diff
@@ -0,0 +1,9 @@
+package misk.metrics.backends
+
+import misk.config.Config
+import misk.metrics.backends.graphite.GraphiteBackendConfig
+import misk.metrics.backends.stackdriver.StackDriverBackendConfig
+
+data class MetricsBackendConfig(
+        val graphite: GraphiteBackendConfig?,
+        val stack_driver: StackDriverBackendConfig?) : Config
```
JGulbronson: nit: move `): Config` to a newline, to give each field its own line.

misk/src/main/kotlin/misk/metrics/backends/graphite/GraphiteReporterService.kt
```diff
@@ -0,0 +1,14 @@
...
+import com.google.common.util.concurrent.AbstractIdleService
+import java.util.concurrent.TimeUnit
+import javax.inject.Inject
+
+internal class GraphiteReporterService @Inject internal constructor(
+        private val config: GraphiteBackendConfig,
+        private val reporter: GraphiteReporter
+) : AbstractIdleService() {
+    override fun startUp() = reporter.start(config.interval.toMillis(), TimeUnit.MILLISECONDS)
+    override fun shutDown() = reporter.stop()
```
JGulbronson: This is awesome how short these are.

misk/src/main/kotlin/misk/metrics/backends/stackdriver/StackDriverBackendConfig.kt
```diff
@@ -0,0 +1,11 @@
...
+
+import misk.config.Config
+import java.time.Duration
+
+data class StackDriverBackendConfig(
+        val project_id: String,
+        val service_path: String,
+        val root_url: String,
+        val interval: Duration,
+        val batch_size: Int = 200) : Config
```
JGulbronson: nit: give each field its own line and move the `) : Config` bit to its own line.

misk/src/main/kotlin/misk/metrics/backends/stackdriver/StackDriverBackendModule.kt
```diff
@@ -0,0 +1,32 @@
...
+import com.google.inject.Singleton
+import misk.config.AppName
+import misk.inject.addMultibinderBinding
+import misk.inject.to
+
+class StackDriverBackendModule : AbstractModule() {
+    override fun configure() {
+        bind(StackDriverSender::class.java)
+                .to(StackDriverBatchedSender::class.java)
+                .`in`(Singleton::class.java)
```
JGulbronson: You can use `inSingleton()` which we define in `Guice.kt` to avoid having the ugly quotes.

misk/src/main/kotlin/misk/metrics/backends/stackdriver/StackDriverBatchedSender.kt
```diff
@@ -0,0 +1,51 @@
...
+    }
+
+    class ResponseCallback : JsonBatchCallback<Empty>() {
+        override fun onFailure(e: GoogleJsonError?, responseHeaders: HttpHeaders?) {
+            log.warn("failed to send timeseries to stack driver ${e?.toPrettyString()}")
+        }
+
+        override fun onSuccess(t: Empty?, responseHeaders: HttpHeaders?) {}
+    }
+
```
JGulbronson: nit: extra line.

build.gradle
```diff
@@ -24,6 +24,9 @@ ext {
           jacksonKotlin: 'com.fasterxml.jackson.module:jackson-module-kotlin:2.9.2',
           metricsParent: 'io.dropwizard.metrics:metrics-parent:3.1.0',
           metricsCore: 'io.dropwizard.metrics:metrics-core:3.1.0',
+          metricsGraphite: 'io.dropwizard.metrics:metrics-graphite:3.1.0',
+          metricsStackDriver: 'com.google.apis:google-api-services-monitoring:v3-rev425-1.23.0',
+          mockitoCore: "org.mockito:mockito-core:2.+",
```
JGulbronson: Is this actually being used? I don't see any `mockito` imports in the code.

misk/build.gradle
```diff
@@ -34,9 +34,12 @@ dependencies {
   compile dep.jacksonKotlin
   compile dep.metricsCore
   compile dep.metricsParent
+  compile dep.metricsGraphite
+  compile dep.metricsStackDriver
 
   testCompile dep.junit
   testCompile dep.truth
   testCompile dep.loggingImpl
+  testCompile dep.mockitoCore
```
JGulbronson: Doesn't seem to be used.

misk/src/test/kotlin/misk/metrics/backends/stackdriver/StackDriverReporterTest.kt
```diff
@@ -0,0 +1,290 @@
...
+import java.math.BigDecimal
+import java.math.BigInteger
+import java.util.concurrent.TimeUnit
+import javax.inject.Inject
+import javax.inject.Singleton
+
+internal class StackDriverReporterTest {
+    companion object {
+        @JvmStatic
+        val APP_NAME = "my_app"
```
JGulbronson: If these are just being used to test the `StackDriverReporter` class, they shouldn't need `@JvmStatic`.

misk/src/main/kotlin/misk/metrics/backends/stackdriver/StackDriverReporter.kt
```diff
@@ -0,0 +1,218 @@
...
+        sender.send(timeSeries)
+    }
+
+    enum class MetricKind {
+        GAUGE,
+        CUMULATIVE
+    }
+
+    companion object {
+        @JvmStatic
```
JGulbronson: Since this is an internal class, I don't think you need `@JvmStatic`.

misk/src/main/kotlin/misk/metrics/backends/stackdriver/StackDriverBatchedSender.kt
```diff
@@ -0,0 +1,51 @@
...
+            val request = CreateTimeSeriesRequest().setTimeSeries(timeSeries)
+            val create = timeSeriesApi.create(config.project_id, request).execute()
+            if (!create.isEmpty()) {
+                log.warn(create.toPrettyString())
+            }
+        }
+    }
+
+    class ResponseCallback : JsonBatchCallback<Empty>() {
+        override fun onFailure(e: GoogleJsonError?, responseHeaders: HttpHeaders?) {
```
JGulbronson: Will `e` ever be nullable? I think sometimes Intellij marks things as nullable to be extra safe, but I don't think you'll get a failure with a null error here.

misk/src/test/kotlin/misk/metrics/backends/stackdriver/StackDriverReporterTest.kt
```diff
@@ -0,0 +1,290 @@
...
+                metrics.gauges,
+                metrics.counters,
+                metrics.histograms,
+                null,
+                metrics.timers,
+                START_TIME, NOW)
+
+        val names = timeSeries.map { it.metric.type to it }.toMap()
+        assertThat(names.keys).containsExactly(
+                "custom.googleapis.com/dw/my_app/us-west2/165.33.45.25/im-basic/value",
```
JGulbronson: Are these coming from an actual HTTP request?

exemplar/src/main/kotlin/com/squareup/exemplar/ExemplarModule.kt
```diff
@@ -15,5 +16,6 @@ class ExemplarModule : AbstractModule() {
         install(WebActionModule.create<ReadinessCheckAction>())
         install(WebActionModule.create<LivenessCheckAction>())
         install(WebActionModule.create<NotFoundAction>())
+        install(WebActionModule.create<MetricsAction>())
```
JGulbronson: This should be before `NotFoundAction`. Maybe we should warn if `NotFoundAction` isn't last?

exemplar/src/main/java/com/squareup/exemplar/ExemplarJavaModule.java
```diff
@@ -10,5 +11,6 @@
     install(WebActionModule.create(HelloJavaAction.class));
     install(WebActionModule.create(InternalErrorAction.class));
     install(WebActionModule.create(NotFoundAction.class));
+    install(WebActionModule.create(MetricsAction.class));
```
JGulbronson: Same here, should be one line up.

JGulbronson:
Outside the scope of this PR, but we might want to look at allowing endpoints to be exposed using one of the Prometheus exposition formats (probably the text format to start), since I don't think it will understand the JSON.

misk/src/main/kotlin/misk/metrics/MetricsConfig.kt
```diff
@@ -0,0 +1,7 @@
+package misk.metrics
+
+import misk.config.Config
+import misk.metrics.backends.MetricsBackendConfig
+
+data class MetricsConfig(val backends: MetricsBackendConfig) : Config
```
JGulbronson: This isn't used right now, should config be set up in a different PR?

misk/src/main/kotlin/misk/metrics/backends/MetricsBackendConfig.kt
```diff
@@ -0,0 +1,9 @@
+package misk.metrics.backends
+
+import misk.config.Config
+import misk.metrics.backends.graphite.GraphiteBackendConfig
+import misk.metrics.backends.stackdriver.StackDriverBackendConfig
+
+data class MetricsBackendConfig(
```
JGulbronson: This only seems to be used in `MetricsConfig`, which isn't used. Should this be moved to a separate PR?

misk/src/test/kotlin/misk/metrics/backends/stackdriver/StackDriverReporterTest.kt
mmihic: Stackdriver uses path format to identify metrics. The custom.googleapis.com/dw prefix is required to identify these as custom metrics.

exemplar/src/main/kotlin/com/squareup/exemplar/ExemplarModule.kt
mmihic: Yes, seem fragile for binding statements to have an implicit required ordering. I'd hard error if the required ordering is not correct. Same for interceptors.

misk/src/main/kotlin/misk/metrics/web/MetricsAction.kt
```diff
@@ -0,0 +1,14 @@
+package misk.metrics.web
+
+import misk.metrics.Metrics
+import misk.web.Get
+import misk.web.JsonResponseBody
+import misk.web.actions.WebAction
+import javax.inject.Inject
+
+/** Exposes metrics as JSON over HTTP */
+class MetricsAction @Inject internal constructor(val metrics: Metrics) : WebAction {
```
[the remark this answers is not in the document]
mmihic: Sure, although doesn't really matter.

misk/src/main/kotlin/misk/metrics/MetricsConfig.kt
mmihic: The top-level config is the top-level container for the back-end specific configs introduced in this PR.  IMO it makes the most sense to add all of that configuration in this PR.  We don't use the metrics config in exemplar because I don't think we want to have an exemplar configuration specifying a specific metrics backend, with associated port, project, etc.

misk/src/main/kotlin/misk/metrics/backends/MetricsBackendConfig.kt
mmihic: Same comment as above

mmihic:
I'd like to have content-type dispatching to add a plain-text endpoint for metrics.

exemplar/src/main/java/com/squareup/exemplar/ExemplarJavaModule.java
mmihic: Done. BTW the java module is missing the liveness and readiness actions.

build.gradle
mmihic: Removed. I was originally trying to mock stackdriver to test the actual batch sending code, but that turns out to be impossible to do a meaningful way given the structure of the client.

PR #30, opened 2017-12-11
Improve the documentation for classes in the web.extractors package

JGulbronson:
Part of an ongoing effort to shore up the documentation for classes that likely won't change a whole lot, like the JSON and path pattern parameter extractors.

### DIFF
--- a/misk/src/main/kotlin/misk/web/extractors/HeadersParameterExtractorFactory.kt
+++ b/misk/src/main/kotlin/misk/web/extractors/HeadersParameterExtractorFactory.kt
@@ -8,7 +8,11 @@ import java.util.regex.Matcher
 import kotlin.reflect.KParameter
 import kotlin.reflect.full.findAnnotation
 
-val HeadersParameterExtractorFactory = object : ParameterExtractor.Factory {
+/**
+ * Creates a [ParameterExtractor] that returns [Request.headers] from a [Request] if the parameter
+ * is annotated with [RequestHeaders].
+ */
+object HeadersParameterExtractorFactory : ParameterExtractor.Factory {
     private val extractor = object : ParameterExtractor {
         override fun extract(webAction: WebAction, request: Request, pathMatcher: Matcher): Any? {
             return request.headers
--- a/misk/src/main/kotlin/misk/web/extractors/JsonBodyParameterExtractorFactory.kt
+++ b/misk/src/main/kotlin/misk/web/extractors/JsonBodyParameterExtractorFactory.kt
@@ -7,13 +7,19 @@ import misk.web.actions.WebAction
 import com.squareup.moshi.Moshi
 import java.util.regex.Matcher
 import javax.inject.Inject
+import javax.inject.Singleton
 import kotlin.reflect.KParameter
 import kotlin.reflect.full.findAnnotation
 import kotlin.reflect.jvm.javaType
 
+@Singleton
 internal class JsonBodyParameterExtractorFactory @Inject constructor(
     private val moshi: Moshi
 ) : ParameterExtractor.Factory {
+    /**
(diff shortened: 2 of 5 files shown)


misk/src/main/kotlin/misk/web/extractors/HeadersParameterExtractorFactory.kt
```diff
@@ -5,16 +5,22 @@ import misk.web.Request
 import misk.web.RequestHeaders
 import misk.web.actions.WebAction
 import java.util.regex.Matcher
+import javax.inject.Singleton
 import kotlin.reflect.KParameter
 import kotlin.reflect.full.findAnnotation
 
+@Singleton
```
swankjesse: Does this do anything?

misk/src/main/kotlin/misk/web/extractors/JsonBodyParameterExtractorFactory.kt
```diff
@@ -7,13 +7,19 @@ import misk.web.actions.WebAction
 import com.squareup.moshi.Moshi
 import java.util.regex.Matcher
 import javax.inject.Inject
+import javax.inject.Singleton
 import kotlin.reflect.KParameter
 import kotlin.reflect.full.findAnnotation
 import kotlin.reflect.jvm.javaType
 
+@Singleton
```
swankjesse: (this does do something)

misk/src/main/kotlin/misk/web/extractors/JsonBodyParameterExtractorFactory.kt
```diff
@@ -7,13 +7,19 @@ import misk.web.actions.WebAction
...
 internal class JsonBodyParameterExtractorFactory @Inject constructor(
     private val moshi: Moshi
 ) : ParameterExtractor.Factory {
+    /**
+     * Creates a [ParameterExtractor] that returns the JSON body of a request if the parameter
```
swankjesse: I feel like these comments should go on the type, not on the method. The documentation is for the consumer of the class, not the caller of the method.

misk/src/main/kotlin/misk/web/extractors/PathPatternParameterExtractor.kt
```diff
@@ -8,7 +8,11 @@ import java.util.regex.Matcher
 import kotlin.reflect.KParameter
 import kotlin.reflect.full.findAnnotation
 
-val PathPatternParameterExtractor = object : ParameterExtractor.Factory {
+val PathPatternParameterExtractorFactory = object : ParameterExtractor.Factory {
+    /**
+     * Creates a [ParameterExtractor] that extracts the URL parameter with the same name as
```
swankjesse: Maybe mention it must be a string?

misk/src/main/kotlin/misk/web/extractors/HeadersParameterExtractorFactory.kt
JGulbronson: Nope, apparently that's what happens when I write code at the end of the day. I'm surprised you can even annotate an instance as a singleton.

JGulbronson: Updated to be an object instance of an object assigned to a value after talking to @AlecStrong. This keeps behavior the same while making it searchable in Intellij.

misk/src/main/kotlin/misk/web/extractors/JsonBodyParameterExtractorFactory.kt
JGulbronson: Makes sense, when thinking about it from that perspective. Fixing.

PR #31, opened 2017-12-11
Split default parameter extractors and interceptor factories in to their own modules

JGulbronson:


exemplar/src/main/java/com/squareup/exemplar/ExemplarJavaApp.java
```diff
@@ -5,16 +5,14 @@
 import misk.environment.EnvironmentModule;
 import misk.hibernate.HibernateModule;
 import misk.moshi.MoshiModule;
-import misk.web.WebActionsModule;
-import misk.web.jetty.JettyModule;
+import misk.web.jetty.WebModule;
```
maniksurtani: `misk.web.WebModule`?

maniksurtani: Perhaps maybe then it is best to also have a `misk.web.jetty.JettyModule` that is installed by `misk.web.WebModule`...

JGulbronson: Good call, fixed.

misk/src/main/kotlin/misk/MiskDefault.kt
```diff
@@ -0,0 +1,9 @@
+package misk
+
+import com.google.inject.BindingAnnotation
+import kotlin.annotation.AnnotationRetention.RUNTIME
+
+@BindingAnnotation
+@Retention(RUNTIME)
+@Target(AnnotationTarget.CLASS, AnnotationTarget.VALUE_PARAMETER, AnnotationTarget.TYPE)
+annotation class MiskDefault
```
swankjesse: hmmm. . . without Javadoc I have no idea what this does

misk/src/main/kotlin/misk/inject/Guice.kt
```diff
@@ -12,12 +12,21 @@ import javax.inject.Singleton
...
 
-@Suppress("UNCHECKED_CAST")
 internal inline fun <reified T : Any> Binder.addMultibinderBinding(): LinkedBindingBuilder<T> {
+    return addMultibinderBinding(null)
+}
+
+internal inline fun <reified T : Any, reified A : Annotation> Binder.addMultibinderBindingWithAnnotation() = addMultibinderBinding<T>(A::class.java)
+
+@Suppress("UNCHECKED_CAST")
+private inline fun <reified T : Any> Binder.addMultibinderBinding(annotation: Class<out Annotation>?): LinkedBindingBuilder<T> {
```
swankjesse: can you make `annotation` default to null and save some overloads?

misk/src/main/kotlin/misk/web/WebActionModule.kt
```diff
@@ -26,7 +27,11 @@ class WebActionModule<A : WebAction> private constructor(
         val provider = getProvider(webActionClass.java)
         @Suppress("UNCHECKED_CAST")
         val binder: Multibinder<BoundAction<A, *>> = Multibinder.newSetBinder(
-                binder(), TypeLiteral.get(Types.newParameterizedType(BoundAction::class.java, Types.subtypeOf(WebAction::class.java), Types.subtypeOf(Object::class.java)))) as Multibinder<BoundAction<A, *>>
+                binder(),
+                TypeLiteral.get(Types.newParameterizedType(BoundAction::class.java,
```
swankjesse: we should make some Kotlin sugar for these

swankjesse:
```
  parameterizedType<BoundAction>(subtypeOf<WebAction>, subtypeOf<Object>).typeLiteral();
```

misk/src/main/kotlin/misk/web/WebActionModule.kt
```diff
@@ -67,14 +72,17 @@ internal class BoundActionProvider<A : WebAction, R>(
     val httpMethod: HttpMethod
 ) : Provider<BoundAction<A, *>> {
 
-    @Inject lateinit var interceptorFactories: List<Interceptor.Factory>
+    @Inject lateinit var userProvidedInterceptorFactories: List<Interceptor.Factory>
+    @Inject lateinit var defaultInterceptorFactories: @MiskDefault List<Interceptor.Factory>
```
swankjesse: I don’t think I like this name

misk/src/main/kotlin/misk/web/WebActionModule.kt
```diff
@@ -67,14 +72,17 @@ internal class BoundActionProvider<A : WebAction, R>(
...
     @Inject lateinit var parameterExtractorFactories: List<ParameterExtractor.Factory>
 
     override fun get(): BoundAction<A, *> {
         val action = function.asAction()
 
         val interceptors = ArrayList<Interceptor>()
-        interceptorFactories.mapNotNullTo(interceptors) { it.create(action) }
+        // Ensure that default interceptors are called before any user provided interceptors
+        defaultInterceptorFactories.mapNotNullTo(interceptors) { it.create(action) }
```
swankjesse: I find `mapNotNullTo` to be a very awkward method

misk/src/main/kotlin/misk/web/extractors/DefaultParameterExtractorsModule.kt
```diff
@@ -0,0 +1,13 @@
+package misk.web.extractors
+
+import com.google.inject.AbstractModule
+import misk.inject.addMultibinderBinding
+import misk.inject.to
+
+class DefaultParameterExtractorsModule : AbstractModule() {
```
swankjesse: Would this ever be used standalone? Why not just put this in WebModule ?

misk/src/main/kotlin/misk/web/extractors/JsonBodyParameterExtractorFactory.kt
```diff
@@ -1,25 +1,25 @@
...
-import com.squareup.moshi.Moshi
 import java.util.regex.Matcher
 import javax.inject.Inject
 import javax.inject.Singleton
 import kotlin.reflect.KParameter
 import kotlin.reflect.full.findAnnotation
 import kotlin.reflect.jvm.javaType
 
+/**
+ * Creates a [ParameterExtractor] that returns the JSON body of a request if the parameter
```
swankjesse: Nice

misk/src/main/kotlin/misk/web/interceptors/DefaultInterceptorsModule.kt
```diff
@@ -0,0 +1,22 @@
+package misk.web.interceptors
+
+import com.google.inject.AbstractModule
+import com.google.inject.multibindings.Multibinder
+import misk.Interceptor
+import misk.MiskDefault
+import misk.inject.addMultibinderBindingWithAnnotation
+import misk.inject.to
+
+class DefaultInterceptorsModule : AbstractModule() {
```
swankjesse: Same. Does this need to be its own module?

misk/src/main/kotlin/misk/inject/Guice.kt
JGulbronson: Right, that's the Java coming through when I forget about default params.

misk/src/main/kotlin/misk/web/WebActionModule.kt
JGulbronson: Agreed, I think that's the last "ugly" usage we haven't cleaned up yet. I'll follow up with that in a separate PR tomorrow.

misk/src/main/kotlin/misk/web/WebActionModule.kt
JGulbronson: `miskInterceptorFactories`? I think we'll end up with a few layers of Interceptors like OkHTTP, so we'll probably have to play around with it then.

misk/src/main/kotlin/misk/web/WebActionModule.kt
JGulbronson: Yeah, it doesn't really map well coming from other languages. I'll see if I can make it more grok-able.

misk/src/main/kotlin/misk/web/extractors/DefaultParameterExtractorsModule.kt
JGulbronson: Yeah I'll move it to `WebModule`. After talking to @maniksurtani I moved some modules around, and these don't make sense on their own anymore.

PR #35, opened 2017-12-12
Add Guice extension functions

JGulbronson:
Use Kotlin extension functions to allow interaction with Guice without converting types from Kotlin to Java.

### DIFF
--- a/misk/src/main/kotlin/misk/MiskApplication.kt
+++ b/misk/src/main/kotlin/misk/MiskApplication.kt
@@ -3,13 +3,12 @@ package misk
 import com.google.common.util.concurrent.ServiceManager
 import com.google.inject.Guice
 import com.google.inject.Module
+import misk.inject.getInstance
 
 class MiskApplication(private vararg val modules: Module) {
     fun startAndAwaitStopped() {
-        val injector = Guice.createInjector(
-               modules.asList()
-        )
-        val serviceManager = injector.getInstance(ServiceManager::class.java)
+        val injector = Guice.createInjector(*modules)
+        val serviceManager = injector.getInstance<ServiceManager>()
 
         Runtime.getRuntime().addShutdownHook(object : Thread() {
             override fun run() {
--- a/misk/src/main/kotlin/misk/inject/Guice.kt
+++ b/misk/src/main/kotlin/misk/inject/Guice.kt
@@ -1,14 +1,20 @@
 package misk.inject
 
 import com.google.inject.Binder
+import com.google.inject.Injector
 import com.google.inject.TypeLiteral
 import com.google.inject.binder.LinkedBindingBuilder
 import com.google.inject.binder.ScopedBindingBuilder
 import com.google.inject.multibindings.Multibinder
 import com.google.inject.util.Types
+import java.lang.reflect.ParameterizedType
+import java.lang.reflect.Type
+import java.lang.reflect.WildcardType
 import javax.inject.Inject
 import javax.inject.Provider
 import javax.inject.Singleton
+import kotlin.reflect.KType
+import kotlin.reflect.jvm.javaType
 
 internal inline fun <reified T : Any> LinkedBindingBuilder<in T>.to() = to(T::class.java)
 
@@ -18,8 +24,8 @@ internal inline fun <reified T : Any, reified A : Annotation> Binder.addMultibin
 internal inline fun <reified T : Any> Binder.addMultibinderBinding(
     annotation: Class<out Annotation>? = null
 ): LinkedBindingBuilder<T> {
-    val typeLiteral = TypeLiteral.get(Types.newParameterizedType(List::class.java, Types.subtypeOf(T::class.java))) as TypeLiteral<List<T>>
-    bind(typeLiteral).toProvider(TypeLiteral.get(Types.newParameterizedType(ListProvider::class.java, T::class.java)) as TypeLiteral<Provider<List<T>>>)
+    val typeLiteral = parameterizedType<List<*>>(subtypeOf<T>()).typeLiteral() as TypeLiteral<List<T>>
+    bind(typeLiteral).toProvider(parameterizedType<ListProvider<*>>(T::class.java).typeLiteral() as TypeLiteral<Provider<List<T>>>)
 
     return when (annotation) {
         null -> Multibinder.newSetBinder(this, T::class.java).addBinding()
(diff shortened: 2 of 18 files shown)


misk/src/main/kotlin/misk/MiskApplication.kt
```diff
@@ -3,13 +3,14 @@ package misk
 import com.google.common.util.concurrent.ServiceManager
 import com.google.inject.Guice
 import com.google.inject.Module
+import misk.inject.getInstance
 
 class MiskApplication(private vararg val modules: Module) {
     fun startAndAwaitStopped() {
         val injector = Guice.createInjector(
-               modules.asList()
+                modules.asList()
```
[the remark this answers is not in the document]
JGulbronson: Even better, I can change it to
```kotlin
val injector = Guice.createInjector(*modules)
```

misk/src/main/kotlin/misk/inject/Guice.kt
```diff
@@ -37,6 +42,22 @@ internal class ListProvider<T> : Provider<List<T>> {
     override fun get(): List<T> = ArrayList(set)
 }
 
+internal inline fun <reified T : Any> subtypeOf(): WildcardType {
```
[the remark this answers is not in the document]
JGulbronson: Added to some of the less obvious functions.

misk/src/main/kotlin/misk/inject/KAbstractModule.kt
```diff
@@ -0,0 +1,35 @@
...
+abstract class KAbstractModule : AbstractModule() {
+    protected class KotlinAnnotatedBindingBuilder<X>(
+        private val annotatedBuilder: AnnotatedBindingBuilder<X>
+    ) : AnnotatedBindingBuilder<X> by annotatedBuilder {
+        inline fun <reified T : Annotation> annotatedWith(): LinkedBindingBuilder<X> {
+            return this.annotatedWith(T::class.java)
+        }
+    }
+
+    protected inline fun <reified T : Any> KAbstractModule.bind(): KotlinAnnotatedBindingBuilder<in T> {
```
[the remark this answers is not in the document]
JGulbronson: Added Javadoc on the class.

misk/src/main/kotlin/misk/inject/KAbstractModule.kt
```diff
@@ -0,0 +1,35 @@
...
+        return KotlinAnnotatedBindingBuilder<T>(binder().bind(T::class.java))
+    }
+
+    protected inline fun <reified T : Any> AnnotatedBindingBuilder<in T>.to(): ScopedBindingBuilder {
```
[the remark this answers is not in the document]
JGulbronson: Added Javadoc on the class.

misk/src/main/kotlin/misk/inject/KAbstractModule.kt
```diff
@@ -0,0 +1,35 @@
...
+        return this.to(T::class.java)
+    }
+
+    protected inline fun <reified T : Any> newSetBinder(): Multibinder<T> {
```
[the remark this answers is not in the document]
JGulbronson: Added Javadoc on the class.

misk/src/main/kotlin/misk/inject/Guice.kt
```diff
@@ -37,6 +43,25 @@ internal class ListProvider<T> : Provider<List<T>> {
...
+    return Types.subtypeOf(T::class.java)
+}
+
+/**
+ * Returns the [TypeLiteral] of a [ParameterizedType], type parameterized on [typeParameters]
+ */
+internal inline fun <reified T : Any> parameterizedTypeLiteral(vararg typeParameters: Type): TypeLiteral<*> {
```
swankjesse: I don’t think I like functions that accept _Type_ and return _TypeLiteral_ because they don’t compose

swankjesse: easier to stay in one abstraction or the other

swankjesse:
```
parameterizedType(...).typeLiteral()
```

misk/src/main/kotlin/misk/inject/KAbstractModule.kt
```diff
@@ -0,0 +1,46 @@
...
+            return this.annotatedWith(T::class.java)
+        }
+    }
+
+    protected inline fun <reified T : Any> KAbstractModule.bind(): KotlinAnnotatedBindingBuilder<in T> {
+        return KotlinAnnotatedBindingBuilder<T>(binder().bind(T::class.java))
+    }
+
+    protected inline fun <reified T : Any> AnnotatedBindingBuilder<in T>.to(): ScopedBindingBuilder {
+        return this.to(T::class.java)
```
swankjesse: `this` isn’t necessary?

misk/src/main/kotlin/misk/inject/KAbstractModule.kt
```diff
@@ -0,0 +1,46 @@
...
+abstract class KAbstractModule : AbstractModule() {
+    protected class KotlinAnnotatedBindingBuilder<X>(
+        private val annotatedBuilder: AnnotatedBindingBuilder<X>
+    ) : AnnotatedBindingBuilder<X> by annotatedBuilder {
+        inline fun <reified T : Annotation> annotatedWith(): LinkedBindingBuilder<X> {
+            return this.annotatedWith(T::class.java)
+        }
+    }
+
+    protected inline fun <reified T : Any> KAbstractModule.bind(): KotlinAnnotatedBindingBuilder<in T> {
```
swankjesse: this declaration is weird. Why does it start with `KAbstractModule` ?

misk/src/main/kotlin/misk/inject/KAbstractModule.kt
```diff
@@ -0,0 +1,46 @@
...
+ * bind(Foo::class.java).to(RealFoo::class.java)
+ * ```
+ *
+ * To be rewritten as:
+ * ```
+ * bind<Foo>().to<RealFoo>()
+ * ```
+ */
+abstract class KAbstractModule : AbstractModule() {
+    protected class KotlinAnnotatedBindingBuilder<X>(
```
swankjesse: I don’t think this needs to be a type

misk/src/main/kotlin/misk/inject/KAbstractModule.kt
```diff
@@ -0,0 +1,46 @@
...
+ * ```
+ * bind(Foo::class.java).to(RealFoo::class.java)
+ * ```
+ *
+ * To be rewritten as:
+ * ```
+ * bind<Foo>().to<RealFoo>()
+ * ```
+ */
+abstract class KAbstractModule : AbstractModule() {
```
swankjesse: I don’t think this needs to be a type, just some extension functions? Ie. it doesn’t need state?

JGulbronson: It needs to extend `AbstractModule` because `binder` is protected.

misk/src/main/kotlin/misk/inject/KAbstractModule.kt
JGulbronson: So you can only call it in `KAbstractModule`.

misk/src/main/kotlin/misk/inject/KAbstractModule.kt
JGulbronson: It does so X is the same between the annotated binding builder and the linked binding builder.

misk/src/main/kotlin/misk/inject/KAbstractModule.kt
JGulbronson: Good call, removed.

misk/src/main/kotlin/misk/inject/Guice.kt
JGulbronson: Yeah I wasn't thrilled with it, moving back.

Issue #3, opened 2017-12-05
Update to use JUnit 5


maniksurtani:
This repo uses JUnit 4 for now, but we should be using the latest and greatest here.

swankjesse:
Is JUnit 5 good? I don’t think we should do this too eagerly. We definitely should wait until IDE support is solid.

![](http://junit.org/junit5/docs/current/user-guide/images/component-diagram.svg)

maniksurtani:
There are some nice new features in the way parameterized tests are written, exceptions and timeouts are tested for, etc. IDE support seems solid - IntelliJ [has supported it](https://blog.jetbrains.com/idea/2016/08/using-junit-5-in-intellij-idea/) for almost a year and a half now.

JGulbronson:
@maniksurtani because Junit 5 lacks Rules, it doesn't seem like we can customize our injector per test. Because of that I'm going to put the upgrade on ice for now, and we can revisit later.

maniksurtani:
Reading thru http://www.codeaffine.com/2016/04/06/replace-rules-in-junit5/ there are equivalent techniques. I'll take a stab at it if I find some free time this week.

PR #38, opened 2017-12-14
JUnit 4 --> Junit 5

maniksurtani:
Moves from JUnit 4 to JUnit 5 - see issue #3 

Some notes
* Replaced `MiskTestRule` with `MiskTestExtension` and a `@MiskTest` annotation which achieves the same thing. See `LivenessCheckActionTest` and `ConfigTest` as exemplars of usage.
* Replaced uses of `@Test(expected = ... )` with `assertThrows()`
* Moved test harness code (test extensions, etc) along with test dependencies into a sub-module under `/misk` (we should consider moving stuff in `misk.cloud.gcp.testing` here too)
* JUnit 4 is still dragged in as a dependency of Google Truth (see [this open issue](https://github.com/google/truth/issues/333)). Until this is resolved, we should be careful to use the `org.junit.jupiter.api.Test` annotation and not `org.junit.Test` by mistake.
* IntelliJ and gradle integration works out of the box.


misk/src/test/kotlin/misk/config/ConfigTest.kt
```diff
@@ -1,21 +1,23 @@
...
-    @get:Rule
-    val miskTestRule = MiskTestRule(
-            ConfigModule.create<TestConfig>("test_app"),
-            EnvironmentModule(TESTING)
-    )
+    class ModuleProvider: misk.testing.ModuleProvider {
+        override fun modules(): Array<out Module> {
+            return arrayOf(ConfigModule.create<TestConfig>("test_app"), EnvironmentModule(TESTING))
+        }
+
```
mmihic: super nit: extra blank line

misk/testing/build.gradle
```diff
@@ -0,0 +1,27 @@
+apply plugin: 'kotlin'
```
mmihic: +1 to putting test helper code in its own module separate from misk

maniksurtani:
Closing this in favour of #40

PR #39, opened 2017-12-14
WIP - Basic support for serving admin pages

### LINKED ISSUE #13: Add support for admin pages
Each service should be able to provide resources such as HTML, CSS, etc, which Misk can combine to form an admin dashboard. The data displayed would be used for debugging and administrative purposes, such as setting dynamic config.

JGulbronson:
Allows serving static css, html, js files located at resources/misk/admin without caching.

Closes #13


misk/src/main/resources/misk/admin/html/home.html
```diff
@@ -0,0 +1,10 @@
+<!DOCTYPE html>
```
mmihic: We almost definitely need to allow this to be customized by the application.

We also likely need some sort of templating support for HTML content. That could be a separate PR with some sort of resource resolver that takes the name of an admin resource and a data class/bean/map containing the template parameters, and writes to an output stream (i.e. a library class not an action)

JGulbronson: Yep, I was looking at templating engines. I think with this I mostly want to get the serving of static files set up, and then focus on building out the admin side of things.

PR #40, opened 2017-12-14
JUnit 4 --> Junit 5 Proposal 2

maniksurtani:
Moves from JUnit 4 to JUnit 5 - see issue #3

Some notes

* Replaced `MiskTestRule` with `MiskTestExtension` and a `@MiskTest` annotation which achieves the same thing. See `LivenessCheckActionTest` and `ConfigTest` as exemplars of usage.
* Replaced uses of `@Test(expected = ... )` with `assertThrows()`
* Moved test harness code (test extensions, etc) along with test dependencies into a sub-module under `/misk` (we should consider moving stuff in `misk.cloud.gcp.testing` here too)
* JUnit 4 is still dragged in as a dependency of Google Truth (see this open issue). Until this is resolved, we should be careful to use the `@org.junit.jupiter.api.Test` annotation and not `@org.junit.Test` by mistake.

IntelliJ and gradle integration works out of the box.


misk/src/test/kotlin/misk/config/ConfigTest.kt
```diff
@@ -3,19 +3,17 @@ package misk.config
...
 import javax.inject.Named
 
+@MiskTest
 class ConfigTest {
-    @get:Rule
-    val miskTestRule = MiskTestRule(
-            ConfigModule.create<TestConfig>("test_app"),
-            EnvironmentModule(TESTING)
-    )
+    @Modules val modules = ModuleProvider(ConfigModule.create<TestConfig>("test_app"), EnvironmentModule(TESTING))
```
JGulbronson: nit: Maybe preserve the one line for module pattern from above?

misk/testing/src/main/kotlin/misk/testing/MiskTest.kt
```diff
@@ -0,0 +1,23 @@
...
+ * Annotate your test classes with `@MiskTest` to have fields annotated with `@Inject` initialized.
+ * Provide zero or more [Module] classes or [Module] instances to be installed by annotating a
+ * [ModuleProvider] field in your test class with [Modules].
+ */
+annotation class MiskTest
+
+@Target(AnnotationTarget.FIELD)
+annotation class Modules
+
+data class ModuleProvider(val moduleList: List<Module> = ArrayList(), val moduleClassList: List<KClass<out Module>> = ArrayList()) {
```
JGulbronson: This signature is kind of odd, do we want to just accept `Module` instances, so we don't split between classes and objects?

misk/testing/src/main/kotlin/misk/testing/MiskTestExtension.kt
```diff
@@ -0,0 +1,48 @@
...
+import com.google.inject.Guice
+import com.google.inject.Module
+import com.sun.org.apache.xpath.internal.operations.Mod
+import misk.inject.uninject
+import org.junit.jupiter.api.extension.*
+import org.junit.platform.commons.util.AnnotationUtils
+import kotlin.reflect.full.createInstance
+
+internal class MiskTestExtension: BeforeEachCallback, AfterEachCallback {
+    override fun beforeEach(context: ExtensionContext?) {
```
JGulbronson: You Might be able to make `context` non-nullable here, sometimes Intellij defaults to assuming something is nullable.

misk/testing/src/main/kotlin/misk/testing/MiskTest.kt
maniksurtani: This allows for a combination to be injected. Happy to remove the signature for now and just allow `ModuleProvider` to accept multiple classes _or_ multiple instances, and add the ability to do both later if needed.

maniksurtani: But also just as happy to keep it even simpler and just accept instances rather than classes.

maniksurtani: PTAL.

misk/src/test/kotlin/misk/ActionsTest.kt
```diff
@@ -13,15 +14,19 @@ internal class ActionsTest {
         assertThat(action.returnType.type).isEqualTo(String::class.java)
     }
 
-    @Test(expected = IllegalArgumentException::class)
+    @Test
     fun methodReferenceNotAllowedAsAction() {
-        val t = TestAction()
-        t::myActionMethod.asAction()
+        assertThrows(IllegalArgumentException::class.java) {
```
swankjesse: Nice

misk/src/test/kotlin/misk/config/ConfigTest.kt
```diff
@@ -3,16 +3,17 @@ package misk.config
...
+import misk.testing.ModuleProvider
+import misk.testing.Modules
 import misk.web.WebConfig
-import org.junit.Rule
-import org.junit.Test
+import org.junit.jupiter.api.Test
 import javax.inject.Inject
 import javax.inject.Named
 
+@MiskTest
```
swankjesse: Should this be renamed to ActionTest?

misk/src/test/kotlin/misk/metrics/backends/stackdriver/StackDriverReporterTest.kt
```diff
@@ -55,8 +57,11 @@ internal class StackDriverReporterTest {
         }
     }
 
-    @get:Rule
-    val miskTestRule = MiskTestRule(MetricsModule(), FakeClockModule(), TestModule())
+    @Modules val modules = ModuleProvider(
```
swankjesse: Since modules themselves are composable I wonder if its easier to just do this:

```
@ActionTestModule val actionTestModule = Modules.combine(
        MetricsModule(),
        FakeClockModule(),
        TestModule()
);
```

misk/src/test/kotlin/misk/metrics/web/MetricsJsonActionTest.kt
```diff
@@ -3,15 +3,18 @@ package misk.metrics.web
...
+import misk.testing.Modules
+import org.junit.jupiter.api.Test
 import java.util.concurrent.TimeUnit
 import javax.inject.Inject
 
+@MiskTest
 internal class MetricsJsonActionTest {
-    @get:Rule
-    val miskTestRule = MiskTestRule(MetricsModule())
+    @Modules val modules = ModuleProvider(
```
swankjesse: ... in which case this wouldn’t need a ModuleProvider at all

misk/testing/src/main/kotlin/misk/testing/MiskTestExtension.kt
```diff
@@ -0,0 +1,40 @@
...
+        val namespace = ExtensionContext.Namespace.create(context.requiredTestClass)
+        // First check the context cache
+        @Suppress("UNCHECKED_CAST")
+        return context.getStore(namespace).getOrComputeIfAbsent("modules",
+                { _ -> modulesFromProviders(context) }) as Iterable<Module>
+    }
+
+    private fun modulesFromProviders(context: ExtensionContext): Iterable<Module> {
+        return context.requiredTestClass.declaredFields
+                .filter { f -> f.isAnnotationPresent(Modules::class.java) }
```
swankjesse: does `it` work here?

misk/src/test/kotlin/misk/config/ConfigTest.kt
maniksurtani: Could do. Will look at it in another PR.

misk/src/test/kotlin/misk/metrics/backends/stackdriver/StackDriverReporterTest.kt
maniksurtani: Also something I'll look into in another PR. And yes, good point.

misk/src/test/kotlin/misk/metrics/web/MetricsJsonActionTest.kt
maniksurtani: ... yup.

misk/testing/src/main/kotlin/misk/testing/MiskTestExtension.kt
maniksurtani: Possibly. I need another PR here anyway so will look there.

misk/src/test/kotlin/misk/config/ConfigTest.kt
maniksurtani: Done in this PR since I had to rebase anyway.

misk/src/test/kotlin/misk/metrics/backends/stackdriver/StackDriverReporterTest.kt
maniksurtani: Done in this PR since I had to rebase anyway.

misk/src/test/kotlin/misk/metrics/web/MetricsJsonActionTest.kt
maniksurtani: Done in this PR since I had to rebase anyway.

misk/testing/src/main/kotlin/misk/testing/MiskTestExtension.kt
maniksurtani: Done in this PR since I had to rebase anyway.

PR #43, opened 2017-12-16
Adds ActionScoped and relating bindings

mmihic:
Allows components and actions to define context objects that are scoped to the current action execution. Dispatch code initializes the action scope with initial seed data from the request; other components can provide additional context objects based on the seed data, or on other action scoped data. Scopes can be propagated onto other threads; an ActionScopedExecutorService is provided which automatically propagates the current scope to any callables / runnables schedule from the current thread.

### DIFF
--- a/misk/src/main/kotlin/misk/concurrent/WrappingListeningExecutorService.kt
+++ b/misk/src/main/kotlin/misk/concurrent/WrappingListeningExecutorService.kt
@@ -1,0 +1,56 @@
+package misk.concurrent
+
+import com.google.common.util.concurrent.ForwardingListeningExecutorService
+import com.google.common.util.concurrent.ListenableFuture
+import com.google.common.util.concurrent.ListeningExecutorService
+import java.util.concurrent.Callable
+import java.util.concurrent.ExecutionException
+import java.util.concurrent.Future
+import java.util.concurrent.TimeUnit
+import java.util.concurrent.TimeoutException
+
+/** [ListeningExecutorService] which wraps all calls */
+abstract class WrappingListeningExecutorService() : ForwardingListeningExecutorService() {
+    /** Wraps the specified callable and returns the new wrapped one. */
+    protected abstract fun <T> wrap(callable: Callable<T>): Callable<T>
+
+    override fun <T> submit(callable: Callable<T>): ListenableFuture<T> {
+        val wrapped = wrap(callable)
+        return delegate().submit<T>(wrapped)
+    }
+
+    override fun <T> submit(runnable: Runnable, result: T): ListenableFuture<T> {
+        return submit<T>({ runnable.run(); result })
+    }
+
+    override fun submit(runnable: Runnable): ListenableFuture<*> {
+        val wrapped = wrap<Unit>(Callable { runnable.run() })
+        return delegate().submit(wrapped)
+    }
+
+    @Throws(InterruptedException::class)
+    override fun <T> invokeAll(callables: Collection<Callable<T>>): List<Future<T>> {
+        return delegate().invokeAll(callables.map { wrap(it) })
+    }
+
+    @Throws(InterruptedException::class)
+    override fun <T> invokeAll(callables: Collection<Callable<T>>, timeout: Long,
+            timeUnit: TimeUnit): List<Future<T>> {
+        return delegate().invokeAll(callables.map { wrap(it) }, timeout, timeUnit)
+    }
+
+    @Throws(InterruptedException::class, ExecutionException::class)
+    override fun <T> invokeAny(callables: Collection<Callable<T>>): T {
+        return delegate().invokeAny(callables.map { wrap(it) })
+    }
+
+    @Throws(InterruptedException::class, ExecutionException::class, TimeoutException::class)
+    override fun <T> invokeAny(callables: Collection<Callable<T>>, timeout: Long,
+            timeUnit: TimeUnit): T {
+        return delegate().invokeAny(callables.map { wrap(it) }, timeout, timeUnit)
+    }
+
+    override fun execute(runnable: Runnable) {
+        delegate().submit(wrap<Unit>(Callable { runnable.run() }))
+    }
+}
--- a/misk/src/main/kotlin/misk/scope/ActionScope.kt
+++ b/misk/src/main/kotlin/misk/scope/ActionScope.kt
@@ -1,0 +1,127 @@
+package misk.scope
+
+import com.google.inject.Key
+import com.google.inject.Provider
+import java.util.concurrent.Callable
+import javax.inject.Inject
+import javax.inject.Singleton
+import kotlin.reflect.KFunction
+import kotlin.reflect.KParameter
+import kotlin.reflect.KType
+import kotlin.reflect.KTypeParameter
+import kotlin.reflect.KVisibility
+
+@Singleton
+class ActionScope @Inject internal constructor(
+        // NB(mmihic): ActionScoped depends on ActionScope depends on
+        // on ActionScopedProviders, which might depend on other ActionScopeds. We break
+        // this circular dependency by injecting a map of Provider<ActionScopedProvider>
+        // rather than the map of ActionScopedProvider directly
+        private val providers: @JvmSuppressWildcards Map<Key<*>, Provider<ActionScopedProvider<*>>>
+) : AutoCloseable {
+    companion object {
+        private val tls = ThreadLocal<LinkedHashMap<Key<*>, Any>>()
+    }
+
+    /** Starts the scope on a thread with the provided seed data */
+    fun enter(seedData: Map<Key<*>, Any>): ActionScope {
+        check(tls.get() == null) {
+            "cannot begin an ActionScope on a thread that is already running in an action scope"
+        }
+
+        tls.set(LinkedHashMap(seedData))
+        return this
+    }
+
+    override fun close() {
+        tls.remove()
+    }
+
+    /** Returns true if currently in the scope */
+    fun inScope(): Boolean = tls.get() != null
+
+    /**
+     * Wraps a [Callable] that will be called on another thread, propagating the current
+     * scoped data onto that thread
+     */
+    fun <T> propagate(c: Callable<T>): Callable<T> {
+        check(tls.get() != null) { "not running within an ActionScope" }
+
+        val currentScopedData = tls.get().toMap()
+        return Callable {
+            enter(currentScopedData).use {
+                c.call()
+            }
+        }
+    }
+
+    /**
+     * Wraps a [KFunction] that will be called on another thread, propagating the current
+     * scoped data onto that thread
+     */
+    fun <T> propagate(f: KFunction<T>): KFunction<T> {
+        check(tls.get() != null) { "not running within an ActionScope" }
+
+        val currentScopedData = tls.get().toMap()
+        return WrappedKFunction(currentScopedData, this, f)
+    }
+
+    /**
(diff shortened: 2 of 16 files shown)


misk/src/main/kotlin/misk/scope/ActionScope.kt
```diff
@@ -0,0 +1,127 @@
...
+import kotlin.reflect.KCallable
+import kotlin.reflect.KFunction
+import kotlin.reflect.KParameter
+import kotlin.reflect.KType
+import kotlin.reflect.KTypeParameter
+import kotlin.reflect.KVisibility
+
+@Singleton
+class ActionScope @Inject internal constructor(
+        // NB(mmihic): ActionScoped depends on ActionScope depends on
```
swankjesse: somewhat off topic: with TODOs we put a username in to volunteer who will do it. For `NB` notes is there any reason to have a username?

misk/src/main/kotlin/misk/scope/ActionScope.kt
```diff
@@ -0,0 +1,127 @@
...
+import javax.inject.Singleton
+import kotlin.reflect.KCallable
+import kotlin.reflect.KFunction
+import kotlin.reflect.KParameter
+import kotlin.reflect.KType
+import kotlin.reflect.KTypeParameter
+import kotlin.reflect.KVisibility
+
+@Singleton
+class ActionScope @Inject internal constructor(
```
swankjesse: Should we offer guidance that scoped values must be immutable values? Otherwise propagate is sketchy

swankjesse: (and franklin’s Op is deliberately mutable and not thread safe)

misk/src/main/kotlin/misk/scope/ActionScope.kt
```diff
@@ -0,0 +1,127 @@
...
+    fun <T> get(key: Key<T>): T {
+        check(tls.get() != null) { "not running within an ActionScope" }
+
+        @Suppress("UNCHECKED_CAST")
+        return tls.get().computeIfAbsent(key) { providerFor(key as Key<Any>).get() as Any } as T
+    }
+
+    @Suppress("UNCHECKED_CAST")
+    private fun providerFor(key: Key<*>): ActionScopedProvider<*> {
+        check(providers.containsKey(key)) { "no ActionScopedProvider available for $key" }
```
swankjesse: Stylish complaint: doing 2x as many map lookups is ick

swankjesse: I think that like 99% of calls to `containsKey()` are performance bugs

misk/src/main/kotlin/misk/scope/ActionScope.kt
```diff
@@ -0,0 +1,127 @@
...
+        return providers[key]!!.get()
+    }
+
+    private class WrappedKFunction<T>(
```
swankjesse: (Is there a ForwardingFunction?)

misk/src/main/kotlin/misk/scope/ActionScopedProvider.kt
```diff
@@ -0,0 +1,9 @@
+package misk.scope
+
+/**
+ * An [ActionScopedProvider] is implemented by components and application code that wants
+ * provide contextual information based on an incoming request, job data, etc.
+ */
+interface ActionScopedProvider<out T> {
```
swankjesse: I think our names or abstractions aren’t simple enough yet. I’ll have a tough time explaining the difference between this and ActionScoped!

misk/src/main/kotlin/misk/scope/ActionScopedProviderModule.kt
```diff
@@ -0,0 +1,112 @@
...
+        bindProvider(
+                Key.get(kclass.java),
+                Key.get(actionScopedType(kclass)),
+                binder().getProvider(providerClass.java))
+    }
+
+    /** Binds an annotation qualified [ActionScoped] along with its provider */
+    fun <T : Any> bindProvider(
+            kclass: KClass<T>,
+            a: Annotation,
```
swankjesse: I think we can make nicer APIs using default values instead of overloads.

Here it’d be an `Annotation?` defaulted to null.

misk/src/main/kotlin/misk/scope/ActionScopedProviderModule.kt
```diff
@@ -0,0 +1,112 @@
...
+            providerClass: KClass<out ActionScopedProvider<T>>) {
+        bindProvider(
+                Key.get(kclass.java, a),
```
swankjesse: ... with some helper function somewhere that makes a key with an nullable annotation

misk/src/main/kotlin/misk/scope/ActionScopedProviderModule.kt
```diff
@@ -0,0 +1,112 @@
...
+        private fun <T : Any> actionScopedType(kclass: KClass<T>) =
+                parameterizedType<ActionScoped<T>>(kclass.java).typeLiteral()
+                        as TypeLiteral<ActionScoped<T>>
+
+        private val KEY_TYPE = object : TypeLiteral<Key<*>>() {}
+        private val ACTION_SCOPED_PROVIDER_TYPE = object : TypeLiteral<ActionScopedProvider<*>>() {}
+
+        private class SeedDataActionScopedProvider<T>(private val key: Key<T>) :
+                ActionScopedProvider<T> {
+            override fun get(): T {
```
swankjesse: T -> Nothing

misk/src/main/kotlin/misk/scope/executor/ActionScopedExecutorService.kt
```diff
@@ -0,0 +1,24 @@
...
+import misk.concurrent.WrappingListeningExecutorService
+import misk.scope.ActionScope
+import java.util.concurrent.Callable
+import java.util.concurrent.ExecutorService
+
+/**
+ * Wraps a [ListeningExecutorService] to propagate the current action scope to any tasks
+ * submitted by the current thread
+ */
+class ActionScopedExecutorService(
```
swankjesse: I love how the abstractions layer for things like this.

That said, I’m still anxious about the application-layer consequences of scoped objects being shared between threads

swankjesse: ... would we be better off encouraging more of an Actors approach – if your action wants to do work in another thread you can‘t share state. 

I think the most valuable state that we transition with this is trace IDs. But I think trace IDs might be the exception and not the rule. Instead of trampolining all possible action-scoped stuff into other threads; maybe we should special case trace IDs.

Trace IDs are also the only things that effortlessly trampoline between processes. They’re not like application-layer things like HTTP request headers.

misk/src/main/kotlin/misk/web/WebModule.kt
```diff
@@ -17,10 +18,17 @@ import misk.web.interceptors.MetricsInterceptor
...
 import misk.web.jetty.JettyModule
+import javax.servlet.http.HttpServletRequest
 
 class WebModule : KAbstractModule() {
     override fun configure() {
         install(JettyModule())
+        install(object: ActionScopedProviderModule() {
+            override fun configureProviders() {
+                bindSeedData(Request::class)
+                bindSeedData(HttpServletRequest::class)
```
swankjesse: Yeah I think this might be a mistake. Lets discuss?

swankjesse: (I trust like 5 people to use this successfully; for everyone else I think it’s too dangerous)

misk/src/test/kotlin/misk/concurrent/WrappingListeningExecutorServiceTest.kt
```diff
@@ -0,0 +1,78 @@
...
+
+internal class WrappingListeningExecutorServiceTest {
+
+    private val wrapCounter = AtomicInteger()
+    private val executor = object : WrappingListeningExecutorService() {
+        val wrapped = MoreExecutors.listeningDecorator(Executors.newSingleThreadExecutor())
+
+        override fun <T> wrap(callable: Callable<T>): Callable<T> {
+            return Callable {
+                wrapCounter.incrementAndGet()
```
swankjesse: Sweet. Nice test

misk/src/main/kotlin/misk/concurrent/WrappingListeningExecutorService.kt
```diff
@@ -0,0 +1,58 @@
...
+        return super.submit(wrapped)
+    }
+
+    @Throws(InterruptedException::class)
+    override fun <T> invokeAll(callables: Collection<Callable<T>>): List<Future<T>> {
+        return super.invokeAll(callables.map { wrap(it) })
+    }
+
+    @Throws(InterruptedException::class)
+    override fun <T> invokeAll(callables: Collection<Callable<T>>, l: Long,
```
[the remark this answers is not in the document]
mmihic: ha - autogenerated by intellij. Will fix

misk/src/main/kotlin/misk/scope/ActionScope.kt
mmihic: It's the name of the person that has context for the NB. Could get this from git blame but that might not survive refactorings.

misk/src/main/kotlin/misk/scope/ActionScopedProvider.kt
mmihic: yeah i didn't have a good name for this - opinions welcome

misk/src/main/kotlin/misk/scope/ActionScopedProviderModule.kt
mmihic: ?

misk/src/main/kotlin/misk/scope/executor/ActionScopedExecutorService.kt
mmihic: I think many things propagate - calling principal for example should be inherited across threads spawned by the action, and should propagate across process boundaries as a principal chain (principal acting on-behalf of another principal). Not sure where the tipping point is though.

misk/src/main/kotlin/misk/web/WebModule.kt
mmihic: If we don't do this, most interceptors can't function - they need access to e.g. certs, headers, etc. We can selectively expose those things onto Request though

mmihic:
Thinking about this - one thing we could do is keep propagation but make it opt-in. Basically when you bind an ``ActionScoped`` you can specify whether it should propagate to threads started from the action, with the default being no propagation, and the constraints being that you can only depend on other propagated ``ActionScoped`` contexts.

As for immutability, I think we want to unequivocally state that ``ActionScoped`` context objects are immutable - they are essentially programmer-friendly wrappers around contextual information passed in the request. Once you make them mutable, you are basically introducing implicit "global" variables.

misk/src/main/kotlin/misk/scope/ActionScopedProviderModule.kt
swankjesse: `Nothing` is a special feature in Kotlin’s type system. It's a subtype of `T` that is unreachable.

swankjesse:
Immutable works for me.

Though I still think the application layer wants something more like actors than threads. In practice that might actually look like job queues. In particular when an application thread wants to do something on a thread I'd like the option of making that happen in a separate process.

PR #44, opened 2017-12-16
Fix broken web dispatch pipeline and add tests

mmihic:
The core dispatch pipeline was broken - none of the default interceptors
were being installed due errors in the way the MiskDefault annotation
and interceptor factory multibindings were being specified. Fixed the
bug, and added a basic test for core web dispatching.

### DIFF
--- a/misk/src/main/kotlin/misk/MiskDefault.kt
+++ b/misk/src/main/kotlin/misk/MiskDefault.kt
@@ -1,7 +1,7 @@
 package misk
 
-import com.google.inject.BindingAnnotation
 import misk.web.interceptors.JsonInterceptorFactory
+import javax.inject.Qualifier
 import kotlin.annotation.AnnotationRetention.RUNTIME
 
 /**
@@ -9,7 +9,11 @@ import kotlin.annotation.AnnotationRetention.RUNTIME
  * For example, the [JsonInterceptorFactory] is bound with [MiskDefault] and is automatically
  * installed.
  */
-@BindingAnnotation
+@Qualifier
 @Retention(RUNTIME)
-@Target(AnnotationTarget.CLASS, AnnotationTarget.VALUE_PARAMETER, AnnotationTarget.TYPE)
+@Target(
+        AnnotationTarget.CLASS,
+        AnnotationTarget.VALUE_PARAMETER,
+        AnnotationTarget.FIELD,
+        AnnotationTarget.TYPE)
 internal annotation class MiskDefault
--- a/misk/src/main/kotlin/misk/web/WebActionModule.kt
+++ b/misk/src/main/kotlin/misk/web/WebActionModule.kt
@@ -70,7 +70,7 @@ internal class BoundActionProvider<A : WebAction, R>(
 ) : Provider<BoundAction<A, *>> {
 
     @Inject lateinit var userProvidedInterceptorFactories: List<Interceptor.Factory>
-    @Inject lateinit var miskInterceptorFactories: @MiskDefault List<Interceptor.Factory>
+    @Inject @JvmSuppressWildcards @MiskDefault lateinit var miskInterceptorFactories: Set<Interceptor.Factory>
     @Inject lateinit var parameterExtractorFactories: List<ParameterExtractor.Factory>
 
     override fun get(): BoundAction<A, *> {
--- a/misk/src/test/kotlin/misk/web/WebDispatchTest.kt
+++ b/misk/src/test/kotlin/misk/web/WebDispatchTest.kt
@@ -1,0 +1,96 @@
+package misk.web
+
+import com.google.common.truth.Truth.assertThat
+import com.squareup.moshi.Moshi
+import misk.inject.KAbstractModule
+import misk.web.actions.WebAction
+import okhttp3.MediaType
+import okhttp3.OkHttpClient
+import okhttp3.Request
+import okhttp3.RequestBody
+import org.junit.Rule
+import org.junit.Test
+import javax.inject.Inject
+
+internal class WebDispatchTest {
+    data class HelloBye(val message: String)
+
+    val jsonMediaType = MediaType.parse("application/json")
+
+    @Rule
+    @JvmField
+    val misk = MiskTestRule(TestModule())
+
+    @Inject lateinit var moshi: Moshi
+    private val helloByeJsonAdapter get() = moshi.adapter(HelloBye::class.java)
+
+    @Test
+    fun post() {
+        val requestContent = helloByeJsonAdapter.toJson(HelloBye("my friend"))
+        val httpClient = OkHttpClient()
+        val request = Request.Builder()
+                .post(RequestBody.create(jsonMediaType, requestContent))
+                .url(misk.serverUrl().encodedPath("/hello").build())
+                .build()
+
+        val response = httpClient.newCall(request).execute()
+        assertThat(response.code()).isEqualTo(200)
+        val responseContent = response.body()!!.source()
+        assertThat(helloByeJsonAdapter.fromJson(responseContent)!!.message)
+                .isEqualTo("post hello my friend")
+    }
+
+    @Test
+    fun get() {
+        val httpClient = OkHttpClient()
+        val request = Request.Builder()
+                .get()
+                .url(misk.serverUrl().encodedPath("/hello/my_friend").build())
+                .build()
+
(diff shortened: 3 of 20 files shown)


misk/src/main/kotlin/misk/MiskDefault.kt
```diff
@@ -1,15 +1,19 @@
...
+import javax.inject.Qualifier
 import kotlin.annotation.AnnotationRetention.RUNTIME
 
 /**
  * Denotes a target as a Misk provided default, which will be automatically installed.
  * For example, the [JsonInterceptorFactory] is bound with [MiskDefault] and is automatically
  * installed.
  */
-@BindingAnnotation
+@Qualifier
```
swankjesse: thanks!

misk/src/main/kotlin/misk/web/WebActionModule.kt
```diff
@@ -70,7 +70,7 @@ internal class BoundActionProvider<A : WebAction, R>(
 ) : Provider<BoundAction<A, *>> {
 
     @Inject lateinit var userProvidedInterceptorFactories: List<Interceptor.Factory>
-    @Inject lateinit var miskInterceptorFactories: @MiskDefault List<Interceptor.Factory>
+    @Inject @JvmSuppressWildcards @MiskDefault lateinit var miskInterceptorFactories: Set<Interceptor.Factory>
```
swankjesse: Oooh, some hard-earned Kotlin knowledge

misk/src/test/kotlin/misk/testing/PortPicker.kt
```diff
@@ -0,0 +1,8 @@
+package misk.testing
+
+import java.net.ServerSocket
+
+/** Finds a free port for testing */
```
swankjesse: Ick. MockWebServer’s approach to this is much simpler. It does exactly this; but it actually uses the port it reserved for its server socket

swankjesse: (at a minimum this is racy)

swankjesse: ... and now I’m wondering whether Jetty has a similar facility

misk/src/test/kotlin/misk/web/WebDispatchTest.kt
```diff
@@ -0,0 +1,134 @@
...
+
+    val jsonMediaType = MediaType.parse("application/json")
+
+    @Inject lateinit var serviceManager: ServiceManager
+    @Inject lateinit var webConfig: WebConfig
+    @Inject lateinit var moshi: Moshi
+
+    @Before
+    fun before() {
+        uninject(this)
```
swankjesse: this is an `@After` thing, isn’t it?

misk/src/test/kotlin/misk/web/WebDispatchTest.kt
```diff
@@ -0,0 +1,134 @@
...
+
+        val injector = Guice.createInjector(
+                WebModule(),
+                TestModule(),
+                MiskModule()
+        )
+
+        injector.injectMembers(this)
+        serviceManager.startAsync()
```
swankjesse: I wanna make this part of our `@MiskTest` thing . . . it does injections and runs services and that’s it

misk/src/test/kotlin/misk/web/WebDispatchTest.kt
```diff
@@ -0,0 +1,134 @@
...
+        @JsonResponseBody
+        fun getBye(@PathParam("message") message: String) =
+                HelloBye("get bye $message")
+    }
+
+    private val serverUrl: HttpUrl
+        get() = HttpUrl.Builder()
+                .scheme("http")
+                .host("127.0.0.1")
+                .port(webConfig.port)
```
swankjesse: What happens if we set the port to say, `0` and just interrogate Jetty on which port it actually picked?

The way it works in MockWebServer is slightly neater than that even, since we rely on it to tell us both the port it picked and also the loopback IP address. This turns out to be relevant in weird network configs sadly.

misk/src/test/kotlin/misk/web/WebDispatchTest.kt
```diff
@@ -0,0 +1,134 @@
...
+                .build()
+
+    private val helloByeJsonAdapter get() = moshi.adapter(HelloBye::class.java)
```
swankjesse: should this be `by lazy` instead of `get()` ?

misk/src/test/kotlin/misk/web/WebDispatchTest.kt
```diff
@@ -0,0 +1,134 @@
...
+        val httpClient = OkHttpClient()
+        val request = Request.Builder()
+                .post(RequestBody.create(jsonMediaType, requestContent))
+                .url("${serverUrl}hello")
+                .build()
+
+        val response = httpClient.newCall(request).execute()
+        assertThat(response.code()).isEqualTo(200)
+        val responseContent  = response.body()!!.source()
+        assertThat(helloByeJsonAdapter.fromJson(responseContent)!!.message).isEqualTo("post hello my friend")
```
swankjesse: Wanna actually assert exactly the exact JSON? Moshi is deterministic (whitespace etc.) and then there’s one fewer things in the test infrastructure.

swankjesse: (and likewise for the request body)

misk/src/test/kotlin/misk/web/WebDispatchTest.kt
```diff
@@ -0,0 +1,134 @@
...
+    }
+
+    @Test
+    fun get() {
+        val httpClient = OkHttpClient()
```
swankjesse: For what it’s worth, `new OkHttpClient()` is frustratingly not efficient in use cases like this. It’s due to some behind-the-scenes thread pools that take a minute to time out.

https://square.github.io/okhttp/3.x/okhttp/okhttp3/OkHttpClient.html

This is how OkHttp tests itself. It’s not completely satisfying.
https://github.com/square/okhttp/blob/master/okhttp-tests/src/test/java/okhttp3/TestUtil.java#L29

Don’t take any action on this here; I’ll do something in a follow-up.

misk/src/test/kotlin/misk/web/WebDispatchTest.kt
```diff
@@ -0,0 +1,134 @@
...
+import okhttp3.Request
+import okhttp3.RequestBody
+import org.junit.After
+import org.junit.Before
+import org.junit.Test
+import java.util.concurrent.TimeUnit
+import javax.inject.Inject
+import javax.inject.Singleton
+
+internal class WebDispatchTest {
```
swankjesse: For what it’s worth, I think this test actually makes Misk look pretty awesome.

misk/src/test/kotlin/misk/testing/PortPicker.kt
mmihic: Jetty does not - there's no way to interrogate it for the bound port (I looked for this before adding PortPicker)

misk/src/test/kotlin/misk/web/WebDispatchTest.kt
mmihic: Yes. Also it goes away entirely in the content based routing branch.

misk/src/test/kotlin/misk/web/WebDispatchTest.kt
mmihic: Yes, see mmihic/content-based-routing which does exactly this.

I do think we want separate layering for the test rules though - one that just does injection, and one that spins up a full Misk environment (the former probably shouldn;t be called MiskTest though)

misk/src/test/kotlin/misk/web/WebDispatchTest.kt
mmihic: You can't interrogate Jetty, sadly. Well, I guess maybe you could peek into the connectors list and find the right connector, lemme see.

misk/src/test/kotlin/misk/testing/PortPicker.kt
mmihic: Also we don't directly expose the bound jetty instance or jetty service, but I can probably change that with some Guice-fu

PR #49, opened 2017-12-18
Truth -> AssertJ

maniksurtani:
How strongly do you feel about Truth over AssertJ? AssertJ supports JUnit 5; Truth pulls in JUnit 4 as a transitive dependency. Feature-wise, they're near-identical.


mmihic:
Personally I have no particular attachment to any assertion library.

swankjesse:
I really don't like how Truth isn't 1.0 yet.

AssertJ is nice. Some annoying API decisions but it mostly just works.

I dislike Hamcrest.

PR #50, opened 2017-12-18
Tests ActionScoped within the web pipeline

mmihic:


misk/src/test/kotlin/misk/web/ActionScopedWebDispatchTest.kt
```diff
@@ -0,0 +1,73 @@
...
+            TestWebModule(),
+            TestModule())
+
+    private @Inject lateinit var jettyService: JettyService
+
+    @Test
+    fun exposesActionScopedInInterceptors() {
+        val httpClient = OkHttpClient()
+        val response = httpClient.newCall(okhttp3.Request.Builder()
+                .url(jettyService.serverUrl.newBuilder().encodedPath("/hello").build())
```
maniksurtani: `WebDispatchTest` has a `serverUrlBuilder()` function that does some of this. Maybe it makes sense for `TestWebModule` to bind a `JettyTestSupport` that could be injected here to do some of this? Granted, `jettyService.serverUrl.newBuilder()` is hardly onerous, but I can see more functionality being added to such a class.

mmihic: Yeah, been thinking the same - there's a bunch of boilerplate around okhttp clients in these tests as well, might be good to pull these into helper TestSupport help functions

PR #61, opened 2018-01-05
Update dependencies

maniksurtani:


misk/src/main/kotlin/misk/web/interceptors/MetricsInterceptor.kt
```diff
@@ -19,7 +20,7 @@ internal class MetricsInterceptor internal constructor(val scope: MetricsScope)
 
     override fun intercept(chain: Chain): Any? {
         scope.counter("requests").inc()
-        val result = scope.timer("timing").time { chain.proceed(chain.args) }
+        val result = scope.timer("timing").time(Callable { chain.proceed(chain.args) })
```
maniksurtani: To ensure the appropriate version of `time()` is called. Dropwizard metrics 4.x introduced a new, overloaded version that caused the wrong one to be picked.

PR #62, opened 2018-01-08
Add API for injecting dynamically sourced flags

mmihic:


misk/src/main/kotlin/misk/flags/DynamicFlagModule.kt
```diff
@@ -0,0 +1,163 @@
...
+    }
+
+    protected abstract fun configureFlags()
+
+    inline fun <reified T : Flags> bindFlags(prefix: String = "", qualifier: Annotation? = null) =
+            bindFlags(T::class, prefix, qualifier)
+
+    fun <T : Flags> bindFlags(
+            kclass: KClass<T>,
+            prefix: String = "",
```
swankjesse: should this have a default?

misk/src/main/kotlin/misk/flags/DynamicFlagModule.kt
```diff
@@ -0,0 +1,163 @@
...
+    inline fun <reified T : Any> bindJsonFlag(
+            name: String,
+            desc: String,
+            qualifier: Annotation = Names.named(name)
+    ) = bindJsonFlag(T::class, name, desc, qualifier)
+
+    fun <T : Any> bindJsonFlag(
+            kclass: KClass<T>,
+            name: String,
+            desc: String,
```
swankjesse: nit: I like the greppability of avoiding abbreviations (description vs. desc)

misk/src/main/kotlin/misk/flags/DynamicFlagModule.kt
```diff
@@ -0,0 +1,163 @@
...
+                .annotatedWith(a)
+                .toProvider(JsonFlagProvider(kclass, name, desc))
+                .asEagerSingleton()
+    }
+
+    private class BooleanFlagProvider(val name: String, val desc: String) : Provider<BooleanFlag> {
+        @Inject
+        lateinit var flagStore: DynamicFlagStore
+
+        override fun get(): BooleanFlag = flagStore.registerBooleanFlag(name, desc)
```
swankjesse: `register` is an interesting word because that’s lazy, right?

Should we publish the description before the flag’s value is requested?

misk/src/main/kotlin/misk/flags/DynamicFlagModule.kt
```diff
@@ -0,0 +1,163 @@
...
+        lateinit var moshi: Moshi
+
+        override fun get(): JsonFlag<T> {
+            val stringFlag = flagStore.registerStringFlag(name, desc)
+            val adapter = moshi.adapter(kclass.java)
+            return JsonFlag(stringFlag, adapter)
+        }
+    }
+
+    private class FlagsProvider<T : Any>(
```
swankjesse: Should this have the word “Constructor” in it?

swankjesse: Would it be simpler/safer to just have the user provide a function rather than trying to convert the constructor into one?

misk/src/main/kotlin/misk/flags/DynamicFlagStore.kt
```diff
@@ -0,0 +1,22 @@
...
+/**
+ * Applications register their interest in monitoring flags through one of the registerXXXFlag
+ * calls. If registration and watching requires contacting a remote service (e.g. a flag service,
+ * Zookeeper, etc), this should be done asynchronously so that multiple flags can be registered
+ * in parallel, with the XXXFlag.get() calls blocking until the registration for that flag is
+ * complete. An application can also call [awaitRegistrationsComplete] to block until all
+ * pending registrations have completed, allowing e.g. the application to avoid processing
+ * incoming requests until all of its flags are available and watched
+ */
+interface DynamicFlagStore {
```
swankjesse: Are all flags strings behind-the-scenes? Or does registration bind the name to its type?

It’s weird that for booleans the store manages how those are mapped to persistence
But for JSONs that mapping happens in the caller

I think it might be simpler for this to be strings only?

misk/src/main/kotlin/misk/flags/Flags.kt
```diff
@@ -0,0 +1,207 @@
...
+
+import com.squareup.moshi.JsonAdapter
+import com.squareup.moshi.Moshi
+import kotlin.properties.ReadOnlyProperty
+import kotlin.reflect.KClass
+import kotlin.reflect.KProperty
+
+interface Flag {
+    val name: String
+    val description: String
```
swankjesse: (desc elsewhere vs. description here is my abbreviation annoyance. Don’t want to have to think about which to use)

misk/src/main/kotlin/misk/flags/Flags.kt
```diff
@@ -0,0 +1,207 @@
...
+        }
+
+        private class Property(val flag: misk.flags.IntFlag, val default: Int)
+            : ReadOnlyProperty<Flags, Int> {
+            override fun getValue(thisRef: Flags, property: KProperty<*>): Int =
+                    flag.get() ?: default
+        }
+    }
+
+    class LongFlag(
```
swankjesse: this seems dangerous

swankjesse: it truncates?

misk/src/main/kotlin/misk/flags/memory/InMemoryDynamicFlagStore.kt
```diff
@@ -0,0 +1,42 @@
...
+
+/** In memory implementation of a [DynamicFlagStore], suitable for use in testing */
+@Singleton
+class InMemoryDynamicFlagStore : DynamicFlagStore {
+    private val _booleanFlags = ConcurrentHashMap<String, InMemoryBooleanFlag>()
+    private val _stringFlags = ConcurrentHashMap<String, InMemoryStringFlag>()
+    private val _intFlags = ConcurrentHashMap<String, InMemoryIntFlag>()
+    private val _doubleFlags = ConcurrentHashMap<String, InMemoryDoubleFlag>()
+
+    val booleanFlags: Map<String, InMemoryBooleanFlag> get() = _booleanFlags
```
swankjesse: I think the duplication over types here is unnecessary. We wouldn’t support a string and an int flag with the same name; why are they in different maps?

swankjesse: (ie. I think you’re working for the type system, it’s not working for you)

misk/src/main/kotlin/misk/flags/memory/InMemoryDynamicFlagStore.kt
```diff
@@ -0,0 +1,42 @@
...
+
+import misk.flags.DynamicFlagStore
+import java.util.concurrent.ConcurrentHashMap
+import java.util.concurrent.TimeUnit
+import javax.inject.Singleton
+
+/** In memory implementation of a [DynamicFlagStore], suitable for use in testing */
+@Singleton
+class InMemoryDynamicFlagStore : DynamicFlagStore {
+    private val _booleanFlags = ConcurrentHashMap<String, InMemoryBooleanFlag>()
```
swankjesse: why _ prefix?

misk/src/test/kotlin/misk/flags/FlagPropertiesTest.kt
```diff
@@ -0,0 +1,105 @@
...
+import org.assertj.core.data.Offset
+import org.junit.jupiter.api.Test
+import javax.inject.Inject
+
+@ActionTest(startService = false)
+class FlagPropertiesTest {
+
+    data class JsonData(val message: String)
+
+    class MyFlags internal constructor(context: Flags.Context) : Flags("my-flags", context) {
```
swankjesse: this is super cool

misk/src/test/kotlin/misk/flags/FlagPropertiesTest.kt
```diff
@@ -0,0 +1,105 @@
...
+import misk.inject.KAbstractModule
+import misk.moshi.MoshiModule
+import misk.testing.ActionTest
+import misk.testing.ActionTestModule
+import org.assertj.core.api.Assertions.assertThat
+import org.assertj.core.data.Offset
+import org.junit.jupiter.api.Test
+import javax.inject.Inject
+
+@ActionTest(startService = false)
```
swankjesse: oh neat

misk/src/test/kotlin/misk/flags/FlagPropertiesTest.kt
```diff
@@ -0,0 +1,105 @@
...
+                override fun configureFlags() {
+                    bindFlags<MyFlags>()
+                    bindFlags<MyFlags>("unique-prefix")
+                }
+            })
+        }
+    }
+
+    @Inject
+    private
```
swankjesse: is this how the world formats these? feels really unfamiliar

misk/src/test/kotlin/misk/flags/FlagPropertiesTest.kt
```diff
@@ -0,0 +1,105 @@
...
+    @Named("unique-prefix")
+    var prefixedFlags: MyFlags
+
+    @Inject
+    private
+    lateinit var nakedFlags: MyFlags
+
+    @Inject
+    private
+    lateinit var flagStore: InMemoryDynamicFlagStore
```
swankjesse: nit: do we need `dynamic` in the name here? Will we ever have static flags? (I guess we call that configuration)

misk/src/test/kotlin/misk/flags/FlagsTest.kt
```diff
@@ -0,0 +1,157 @@
...
+
+    @Inject
+    private lateinit
+    @Named("my-int")
+    var int: IntFlag
+
+    @Inject
+    private lateinit
+    @Named("my-double")
+    var double: DoubleFlag
```
swankjesse: Should these be `Flag<Double>`, etc.?

misk/src/test/kotlin/misk/flags/FlagPropertiesTest.kt
```diff
@@ -0,0 +1,105 @@
...
+    }
+
+    @ActionTestModule
+    val testModule = object : KAbstractModule() {
+        override fun configure() {
+            install(MoshiModule())
+            install(InMemoryDynamicFlagStoreModule())
+            install(object : DynamicFlagModule() {
+                override fun configureFlags() {
+                    bindFlags<MyFlags>()
```
swankjesse: I think this is my favorite feature of the whole thing

misk/src/test/kotlin/misk/flags/FlagPropertiesTest.kt
```diff
@@ -0,0 +1,105 @@
...
+                }
+            })
+        }
+    }
+
+    @Inject
+    private
+    lateinit
+    @Named("unique-prefix")
+    var prefixedFlags: MyFlags
```
swankjesse: Eventually I’d like a way to programmatically get this without using the Injector. For example, for archiver we don’t know the prefix statically.

misk/src/main/kotlin/misk/flags/DynamicFlagModule.kt
mmihic: I think we want to allow naked (unannotated) flag classes for e.g. application-level flags

misk/src/main/kotlin/misk/flags/DynamicFlagModule.kt
mmihic: Will fix

misk/src/main/kotlin/misk/flags/DynamicFlagModule.kt
mmihic: Yes, as documented on the flag store, register should be an asynchronous, and only block until the value is retrieved. We can make description a required blocking operation as well, although I'm generally fine serving description out of local cache prior to whatever backend server needs to be reached has become available.

misk/src/main/kotlin/misk/flags/DynamicFlagStore.kt
mmihic: I think it's reasonable to allow the underlying backend to support optimized storage for common types (strings, numbers, booleans), and layer support for derived types (JSON, etc) on top of those base types.

misk/src/main/kotlin/misk/flags/memory/InMemoryDynamicFlagStore.kt
mmihic: Because there are public vals that expose the various maps as immutable maps, and it seems like using a prefix for private vals that are exposed as modified public vals seems like a Kotlin convention.

misk/src/test/kotlin/misk/flags/FlagPropertiesTest.kt
mmihic: Yes this was done with auto-formatting

PR #63, opened 2018-02-01
Prepare misk and exemplar for release 0.1.0

nbroyles:


gradle.properties
```diff
@@ -1,2 +1 @@
-GROUP=com.squareup.cloud-exemplar
-VERSION_NAME=0.1.0-SNAPSHOT
+GROUP=com.squareup
```
swankjesse: I think this really wants to be com.squareup.misk

swankjesse: "group" is closer to "project" in Maven terms . . . the Misk group should have all the Misk stuff, which will stay separate from our other open source projects

PR #66, opened 2018-02-02
Upgrade to Gradle 4.5 to support Java 9; add Java 9 to test matrix

maniksurtani:
Something changed in Java 9 that had the `LinkedHashMap` in `ActionScope` throwing `ConcurrentModificationException`s. @mmihic was there a specific reason you used a `LinkedHashMap`, did you need ordering guarantees?

### DIFF
--- a/misk/src/main/kotlin/misk/scope/ActionScope.kt
+++ b/misk/src/main/kotlin/misk/scope/ActionScope.kt
@@ -3,9 +3,6 @@ package misk.scope
 import com.google.inject.Key
 import com.google.inject.Provider
 import java.util.concurrent.Callable
-import java.util.concurrent.ConcurrentHashMap
-import java.util.concurrent.ConcurrentMap
-import java.util.concurrent.ConcurrentSkipListMap
 import javax.inject.Inject
 import javax.inject.Singleton
 import kotlin.reflect.KFunction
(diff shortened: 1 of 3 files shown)


nbroyles:
lgtm, though I'll defer to @mmihic to approve given the Concurrent/LinkedHashMap question

mmihic:
It shouldn't need to be a ``ConcurrentHashMap`` - we only ever access it from a single thread (it's always retrieved from thread local storage) and we never modify it during iteration (we only iterate over it in one place - to make a copy before handing off to another thread - and that iteration is entirely self contained). So the fact that we're getting concurrency exceptions means there is some sort of latent bug here which I need to investigate.

mmihic:
FYI I'm picking this up now

mmihic:
I fixed the ActionScoped problem; it was indeed a latent bug, and switching to `ConcurrentHashMap` wouldn't actually fix it. The behavior is technically legal under `LinkedHashMap` ([reference](https://docs.oracle.com/javase/8/docs/api/java/util/HashMap.html#computeIfAbsent-K-java.util.function.Function-)) (although apparently in practice unsupported). It is explicitly illegal under `ConcurrentHashMap` ([reference](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ConcurrentHashMap.html#computeIfAbsent-K-java.util.function.Function-)) (and apparently remains undetected ([reference](https://stackoverflow.com/questions/28840047/recursive-concurrenthashmap-computeifabsent-call-never-terminates-bug-or-fea/28845674#28845674))

mmihic:
lol that I can approve PRs which contain my own changes

misk/src/main/kotlin/misk/scope/ActionScope.kt
```diff
@@ -85,8 +85,21 @@ class ActionScope @Inject internal constructor(
     fun <T> get(key: Key<T>): T {
         check(tls.get() != null) { "not running within an ActionScope" }
 
+        // NB(mmihic): We don't use computeIfAbsent because computing the value of this
```
ryanhall07: whoaa, nasty bug. good catch!

PR #69, opened 2018-02-05
Extends misk testing framework

mmihic:
- Renames ``ActionTest`` to ``MiskTest`` - most of the time this annotation is used for testing things other than actions

- Allows registering ``BeforeEachCallback`` and ``AfterEachCallback`` extensions which get initialized via the Misk test injector, to support extensions that interact with objects in the Misk binding context (e.g. that perform seed data initialization before tests, or register background services, or setup preconditions in internal Misk services)

### DIFF
--- a/misk/testing/src/main/kotlin/misk/testing/AfterEachExtensionModule.kt
+++ b/misk/testing/src/main/kotlin/misk/testing/AfterEachExtensionModule.kt
@@ -7,7 +7,7 @@ import org.junit.jupiter.api.extension.AfterEachCallback
 import org.junit.jupiter.api.extension.BeforeEachCallback
 import kotlin.reflect.KClass
 
-/** Module registering a callback that fires before each test */
+/** Module registering a callback that fires after each test */
 class AfterEachExtensionModule<T : AfterEachCallback> constructor(
         private val kclass: KClass<T>
 ) : KAbstractModule() {
--- a/misk/testing/src/main/kotlin/misk/testing/AfterEachExtensionModule.kt
+++ b/misk/testing/src/main/kotlin/misk/testing/AfterEachExtensionModule.kt
@@ -1,0 +1,23 @@
+package misk.testing
+
+import com.google.inject.Module
+import misk.inject.KAbstractModule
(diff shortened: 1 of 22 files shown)


misk/testing/src/main/kotlin/misk/testing/AfterEachExtensionModule.kt
```diff
@@ -0,0 +1,23 @@
+package misk.testing
+
+import com.google.inject.Module
+import misk.inject.KAbstractModule
+import misk.inject.addMultibinderBinding
+import org.junit.jupiter.api.extension.AfterEachCallback
+import org.junit.jupiter.api.extension.BeforeEachCallback
+import kotlin.reflect.KClass
+
+/** Module registering a callback that fires before each test */
```
swankjesse: before each –> after each

misk/testing/src/main/kotlin/misk/testing/MiskTestExtension.kt
```diff
@@ -4,59 +4,92 @@ import com.google.common.util.concurrent.ServiceManager
...
+internal class MiskTestExtension : BeforeEachCallback, AfterEachCallback {
     override fun beforeEach(context: ExtensionContext) {
-        val modules = context.getActionTestModules()
+        val testModules = context.getActionTestModules().asSequence().toList().toTypedArray()
+        val callbackModules = if (context.startService()) serviceManagementModules else arrayOf()
+        val modules = Modules.combine(MiskTestingModule(), *callbackModules, *testModules)
 
         val injector = Guice.createInjector(modules)
         injector.injectMembers(context.requiredTestInstance)
         context.store("injector", injector)
```
swankjesse: cute

mmihic:
There's an argument to make that the extension should just create an injector and fire any before-each/after-each callbacks that have been installed by the test module, and that everything misk-related should just come through testing modules (e.g. there'd be no `startService` field on the annotation, instead tests would include a ``MiskTestingModule(startServices = true)`` in their list of modules. I didn't go that far in this change though.

PR #77, opened 2018-02-06
Allow customized exception mappings

mmihic:
Typically applications and component will want some level of control
over how specific exceptions get reported to clients; for example
IllegalArgumentExceptions and other client presented problems should be
returned with a 4xx response code rather than the usual 500. This change
adds the ability to register ExceptionMappers, which converts exceptions
thrown by the action into the appropriate Response format. It is
currently limited to only returning text/plain responses; a future
commit will allow for more encoding, for example allowing errors to be
returned in a specific JSON format.

Also introduces an ActionException base type, allowing actions to throw
exceptions that explicitly return a specific status code.

### DIFF
--- a/misk/src/main/kotlin/misk/exceptions/Exceptions.kt
+++ b/misk/src/main/kotlin/misk/exceptions/Exceptions.kt
@@ -1,0 +1,46 @@
+package misk.exceptions
+
+/** Common status codes for actions */
+enum class StatusCode(val code: Int) {
+    BAD_REQUEST(400),
+    NOT_FOUND(404),
+    UNAUTHENTICATED(401),
+    FORBIDDEN(403),
+    NOT_ACCEPTABLE(406),
+    ENHANCE_YOUR_CALM(420),
+    UNPROCESSABLE_ENTITY(429),
+    INTERNAL_SERVER_ERROR(500),
+    SERVICE_UNAVAILABLE(503);
+
+    val isClientError = code in (400..499)
+    val isServerError = code in (500..599)
+}
+
+/** Base class for exceptions thrown by actions, translated into responses */
+open class ActionException(
+        val statusCode: StatusCode,
+        message: String = statusCode.name,
+        cause: Throwable? = null
+) : Exception(message, cause)
+
+/** Base exception for when resources are not found */
+open class NotFoundException(message: String = "", cause: Throwable? = null) :
+        ActionException(StatusCode.NOT_FOUND, message, cause)
+
+/** Base exception for when authentication fails */
+open class UnauthenticatedException(message: String = "", cause: Throwable? = null) :
+        ActionException(StatusCode.UNAUTHENTICATED, message, cause)
+
+/** Base exception for when authenticated credentials lack access to a resource */
+open class UnauthorizedException(message: String = "", cause: Throwable? = null) :
+        ActionException(StatusCode.FORBIDDEN, message, cause)
+
+/** Base exception for when a resource is unavailable */
+open class ResourceUnavailableException(message: String = "", cause: Throwable? = null) :
+        ActionException(StatusCode.SERVICE_UNAVAILABLE, message, cause)
+
+/** Base exception for bad client requests */
+open class BadRequestException(message: String = "", cause: Throwable? = null) :
+        ActionException(StatusCode.BAD_REQUEST, message, cause)
+
+
--- a/misk/src/main/kotlin/misk/web/WebModule.kt
+++ b/misk/src/main/kotlin/misk/web/WebModule.kt
@@ -7,6 +7,9 @@ import misk.inject.addMultibinderBinding
 import misk.inject.addMultibinderBindingWithAnnotation
 import misk.inject.to
 import misk.scope.ActionScopedProviderModule
+import misk.web.exceptions.ActionExceptionMapper
+import misk.web.exceptions.ExceptionHandlingInterceptor
+import misk.web.exceptions.ExceptionMapperModule
 import misk.web.extractors.HeadersParameterExtractorFactory
 import misk.web.extractors.ParameterExtractor
 import misk.web.extractors.PathPatternParameterExtractorFactory
@@ -17,12 +20,12 @@ import misk.web.interceptors.MarshallerInterceptor
 import misk.web.interceptors.MetricsInterceptor
 import misk.web.interceptors.RequestLoggingInterceptor
 import misk.web.jetty.JettyModule
-import javax.servlet.http.HttpServletRequest
 import misk.web.marshal.JsonMarshaller
 import misk.web.marshal.JsonUnmarshaller
 import misk.web.marshal.MarshallerModule
 import misk.web.marshal.PlainTextMarshaller
 import misk.web.marshal.UnmarshallerModule
+import javax.servlet.http.HttpServletRequest
 
 class WebModule : KAbstractModule() {
     override fun configure() {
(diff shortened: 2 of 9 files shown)


misk/src/main/kotlin/misk/exceptions/Exceptions.kt
```diff
@@ -0,0 +1,46 @@
...
+    SERVICE_UNAVAILABLE(503);
+
+    val isClientError = code in (400..499)
+    val isServerError = code in (500..599)
+}
+
+/** Base class for exceptions thrown by actions, translated into responses */
+open class ActionException(
+        val statusCode: StatusCode,
+        message: String = statusCode.name,
```
swankjesse: nice defaults

misk/src/main/kotlin/misk/web/WebModule.kt
```diff
@@ -42,13 +47,35 @@ class WebModule : KAbstractModule() {
         // Create an empty set binder of interceptor factories that can be added to by users.
         newSetBinder<Interceptor.Factory>()
 
-        // Register built-in interceptors
+        // Register built-in interceptors. Interceptors run in the order in which they are
+        // installed, and the order of these interceptors is critical.
```
swankjesse: great comment

misk/src/main/kotlin/misk/web/WebModule.kt
```diff
@@ -42,13 +47,35 @@ class WebModule : KAbstractModule() {
...
         binder().addMultibinderBindingWithAnnotation<Interceptor.Factory, MiskDefault>().to<MetricsInterceptor.Factory>()
+
+        // Convert application level exceptions into their appropriate response format
+        binder().addMultibinderBindingWithAnnotation<Interceptor.Factory, MiskDefault>().to<ExceptionMappingInterceptor.Factory>()
+
+        // Log exceptions that come from the application at the appropriate level
+        binder().addMultibinderBindingWithAnnotation<Interceptor.Factory, MiskDefault>().to<ExceptionLoggingInterceptor.Factory>()
+
+        // Convert typed responses into a ResponseBody that can marshal the response according to
+        // the client's requested content-typ
```
swankjesse: typ

misk/src/main/kotlin/misk/web/WebModule.kt
```diff
@@ -42,13 +47,35 @@ class WebModule : KAbstractModule() {
...
         binder().addMultibinderBindingWithAnnotation<Interceptor.Factory, MiskDefault>().to<MarshallerInterceptor.Factory>()
+
+        // Wrap "raw" responses with a Response object
         binder().addMultibinderBindingWithAnnotation<Interceptor.Factory, MiskDefault>().to<BoxResponseInterceptorFactory>()
 
+        // Register build-in exception mappers
+        install(ExceptionMapperModule.create<ActionExceptionMapper>())
+        install(ExceptionMapperModule.create<IllegalArgumentExceptionMapper>())
```
swankjesse: I’m not sold on this one. I think if the application layer bubbles up an IllegalArgumentException, it’s just as likely to be the server’s fault as the caller’s

misk/src/main/kotlin/misk/web/exceptions/ActionExceptionMapper.kt
```diff
@@ -0,0 +1,31 @@
...
+        val message = if (th.statusCode.isClientError) th.message ?: th.statusCode.name
+        else th.statusCode.name
+        return Response(StringResponseBody(message), HEADERS, statusCode = th.statusCode.code)
+    }
+
+    override fun canHandle(th: Throwable): Boolean = th is ActionException
+
+    private companion object {
+        val HEADERS: Headers =
+                Headers.of(listOf("Content-Type" to MediaTypes.TEXT_PLAIN_UTF8).toMap())
```
swankjesse: I think returning errors as plaintext is nice.

misk/src/main/kotlin/misk/web/exceptions/ExceptionMapper.kt
```diff
@@ -0,0 +1,19 @@
...
+    /** @return true if the [ExceptionMapper] can handle the given exception */
+    fun canHandle(th: Throwable): Boolean
+
+    /** @return the [Response] corresponding to the exception. */
+    // TODO(mmihic): Allow control of marshalling based on content type. Ideally we could
+    // return a Response<*> and then content negotiation would control how that * gets mapped.
+    // Right now however the marshalling interceptor works off the static return type of the
+    // action method, so it blows up when we return a different type from here. This is a general
+    // issue; interceptors should be able to return responses that are of a different structure
+    // then the action method return value
```
swankjesse: neat

misk/src/main/kotlin/misk/web/exceptions/ExceptionMappingInterceptor.kt
```diff
@@ -0,0 +1,57 @@
...
+import javax.inject.Inject
+
+/**
+ * Converts application and component level dispatch exception into the appropriate response
+ * format. Allows application and component code to control how exceptions are represented to
+ * clients; for example by setting the status code appropriately, or by returning a specialized
+ * response format specific to the error. Components can control how exceptions are mapped
+ * by installing [ExceptionMapper] via the [ExceptionMapperModule]
+ */
+class ExceptionMappingInterceptor(private val mappers: Set<ExceptionMapper<*>>) : Interceptor {
```
swankjesse: Should exception mapping and logging be a single interceptor? Thinking being we can avoid accidentally swallowing or duplicating exceptions

misk/src/main/kotlin/misk/web/exceptions/IllegalArgumentExceptionMapper.kt
```diff
@@ -0,0 +1,23 @@
...
+
+import misk.exceptions.StatusCode
+import misk.web.Response
+import misk.web.ResponseBody
+import misk.web.marshal.StringResponseBody
+import misk.web.mediatype.MediaTypes
+import okhttp3.Headers
+
+/** Maps [IllegalArgumentException]s into BAD_REQUEST responses */
+internal class IllegalArgumentExceptionMapper : ExceptionMapper<IllegalArgumentException> {
```
swankjesse: Yeah, I think this is not something we should do by default.

We _may_ want a nice API to check an action’s arguments and throw a BadRequestException

misk/src/main/kotlin/misk/web/exceptions/ExceptionMappingInterceptor.kt
mmihic: Yeah, debated this while writing the code. Probably a good idea.

misk/src/main/kotlin/misk/web/exceptions/IllegalArgumentExceptionMapper.kt
mmihic: Hmm, ok, reason I sortof like this is that you can write library code that doesn't need to be aware of the fact that it is running in an action. But I can support removing this and adding the equivalent of ``check() {}`` to raise a ``BadRequestException`` instead of ``IllegalArgumentException``.

misk/src/main/kotlin/misk/web/exceptions/ActionExceptionMapper.kt
mmihic: For default errors yeah. I can see a case for returning structured errors for more complex scenarios (e.g. giving a high level status code + sub-code providing more details, or returning the specific list of problematic fields when throwing a validation exception). Could argue those cases should be handled by returning specific response types rather than exceptions, but often they are actually exceptional and forcing the path to go through response types results in loss of type information in the actions (e.g. need to return generic ``Response`` objects rather than the "expected" response type)

misk/src/main/kotlin/misk/exceptions/Exceptions.kt
```diff
@@ -0,0 +1,46 @@
+package misk.exceptions
+
+/** Common status codes for actions */
+enum class StatusCode(val code: Int) {
+    BAD_REQUEST(400),
+    NOT_FOUND(404),
+    UNAUTHENTICATED(401),
+    FORBIDDEN(403),
+    NOT_ACCEPTABLE(406),
+    ENHANCE_YOUR_CALM(420),
```
maniksurtani: lol

misk/src/main/kotlin/misk/exceptions/Exceptions.kt
```diff
@@ -0,0 +1,46 @@
+package misk.exceptions
+
+/** Common status codes for actions */
+enum class StatusCode(val code: Int) {
```
maniksurtani: Surely these must exist somewhere already? Guava?

mmihic: Guava doesn't have an HTTP stuff. They exist in Jersey, might existing in google's HTTP client.

PR #79, opened 2018-02-06
Add support for query strings in urls

eeSeeGee:
Still TODO: enum support

### DIFF
--- a/exemplar/src/main/kotlin/com/squareup/exemplar/HelloWebAction.kt
+++ b/exemplar/src/main/kotlin/com/squareup/exemplar/HelloWebAction.kt
@@ -2,6 +2,7 @@ package com.squareup.exemplar
 
 import misk.web.Get
 import misk.web.PathParam
+import misk.web.QueryParam
 import misk.web.RequestHeaders
 import misk.web.ResponseContentType
 import misk.web.actions.WebAction
@@ -13,8 +14,15 @@ import javax.inject.Singleton
 class HelloWebAction : WebAction {
     @Get("/hello/{name}")
     @ResponseContentType(MediaTypes.APPLICATION_JSON)
-    fun hello(@PathParam name: String, @RequestHeaders headers: Headers): HelloResponse {
-        return HelloResponse("YO", name.toUpperCase())
+    fun hello(
+            @PathParam name: String,
+            @RequestHeaders headers: Headers,
+            @QueryParam nickName: String?,
+            @QueryParam greetings: List<String>?
+    ): HelloResponse {
+        return HelloResponse(
+                greetings?.joinToString(separator = " ") ?: "YO",
+                nickName?.toUpperCase() ?: name.toUpperCase())
     }
 }
 
--- a/misk/src/main/kotlin/misk/web/extractors/QueryStringParameterExtractor.kt
+++ b/misk/src/main/kotlin/misk/web/extractors/QueryStringParameterExtractor.kt
@@ -1,0 +1,38 @@
+package misk.web.extractors
+
+import misk.web.QueryParam
+import misk.web.PathPattern
+import misk.web.Request
+import misk.web.actions.WebAction
+import java.util.regex.Matcher
+import kotlin.reflect.KFunction
+import kotlin.reflect.KParameter
+import kotlin.reflect.full.findAnnotation
+
+/**
+ * Creates a [ParameterExtractor] that extracts the Query String parameter with the same name as
+ * [parameter] and returns it as whatever type is specified in [parameter]. Returns null if the
+ * parameter doesn't occur in the request, or can't be parsed into the correct type.
+ */
+object QueryStringParameterExtractorFactory : ParameterExtractor.Factory {
+    override fun create(
+            function: KFunction<*>,
+            parameter: KParameter,
+            pathPattern: PathPattern
+    ): ParameterExtractor? {
+        val queryParamAnnotation = parameter.findAnnotation<QueryParam>() ?: return null
+        val parameterName =
+                if (queryParamAnnotation.value.isBlank()) parameter.name!!
+                else queryParamAnnotation.value
+        val queryParamProcessor = QueryStringParameterProcessor(parameter)
+
+        return object : ParameterExtractor {
+            override fun extract(webAction: WebAction, request: Request,
+                    pathMatcher: Matcher): Any? {
+                val parameterValues: List<String> = request.url.queryParameterValues(parameterName)
+                return queryParamProcessor.extractFunctionArgumentValue(parameterValues)
+            }
+        }
+    }
+
+}
--- a/misk/src/main/kotlin/misk/web/extractors/QueryStringParameterProcessor.kt
+++ b/misk/src/main/kotlin/misk/web/extractors/QueryStringParameterProcessor.kt
@@ -1,0 +1,66 @@
+package misk.web.extractors
+
+import java.math.BigDecimal
+import java.math.BigInteger
+import kotlin.reflect.KParameter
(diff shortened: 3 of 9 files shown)


exemplar/src/main/kotlin/com/squareup/exemplar/HelloWebAction.kt
```diff
@@ -13,8 +10,11 @@ import javax.inject.Singleton
 class HelloWebAction : WebAction {
     @Get("/hello/{name}")
     @ResponseContentType(MediaTypes.APPLICATION_JSON)
-    fun hello(@PathParam name: String, @RequestHeaders headers: Headers): HelloResponse {
-        return HelloResponse("YO", name.toUpperCase())
+    fun hello(@PathParam name: String, @RequestHeaders headers: Headers,
```
mmihic: Format when wrapping is to put each parameter on its own line, with the return type on a separate line. e.g.

````kotlin
fun hello(
   @PathParam name: String,
   @RequestHeaders headers: Headers,
   @QueryParam("nickname") nickname: String?
) : HelloResponse {
}


````

File: exemplar/src/main/kotlin/com/squareup/exemplar/HelloWebAction.kt [hunk 2]
```diff
@@ -13,8 +10,11 @@ import javax.inject.Singleton
...
+            @QueryParam("nickname") nickName: String?,
+            @QueryParam("greetings") greetings: List<String>?): HelloResponse {
```
mmihic: Technically `"greetings"` isn't required in this case, correct (it should deduce the query param name from the name of the parameter)? If so, plz remove the unnecessary qualifier (and do the same with `"nickname"` above). If not, plz make it so that we deduce the query name from the name of the parameter a la `@PathParam` :)

misk/src/main/kotlin/misk/web/extractors/QueryStringParameterExtractor.kt
```diff
@@ -0,0 +1,38 @@
...
+ * [parameter] and returns it as whatever type is specified in [parameter]. Returns null if the
+ * parameter doesn't occur in the request, or can't be parsed into the correct type.
+ */
+object QueryStringParameterExtractorFactory : ParameterExtractor.Factory {
+    override fun create(
+            function: KFunction<*>,
+            parameter: KParameter,
+            pathPattern: PathPattern
+    ): ParameterExtractor? {
+        val pathParamAnnotation = parameter.findAnnotation<QueryParam>() ?: return null
```
mmihic: `queryParamAnnotation` not `pathParamAnnotation`

misk/src/main/kotlin/misk/web/extractors/QueryStringParameterProcessor.kt
```diff
@@ -0,0 +1,85 @@
...
+                byteType -> strValue.toByte()
+                byteTypeNullable -> strValue.toByte()
+                byteArrayType -> strValue.toByteArray()
+                byteArrayTypeNullable -> strValue.toByteArray()
+                booleanType -> strValue.toBoolean()
+                booleanTypeNullable -> strValue.toBoolean()
+                else -> null
+            }
+        } catch(e: NumberFormatException) {
+            throw IllegalArgumentException(
```
mmihic: Use string interpolation for messages e.g

`"invalid parameter format for $name : $strValue"`

misk/src/main/kotlin/misk/web/extractors/QueryStringParameterProcessor.kt
```diff
@@ -0,0 +1,85 @@
...
+                    "Invalid parameter format for " + name + ": " + strValue, e)
+        }
+    }
+
+    companion object {
```
mmihic: yeah we should definitely pull this out and make a general (string) -> type conversion function that can be used in multiple places. We probably also want to support types that have a `valueOf(String)` static method defined on them (which would've fixed my enum case, not that I think about it)

misk/src/test/kotlin/misk/web/extractors/QueryStringRequestTest.kt
```diff
@@ -0,0 +1,119 @@
...
+import misk.web.actions.WebAction
+import misk.web.jetty.JettyService
+import misk.web.mediatype.MediaTypes
+import okhttp3.OkHttpClient
+import okhttp3.Request
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.jupiter.api.Test
+import javax.inject.Inject
+
+@MiskTest(startService = true)
```
mmihic: nice test

exemplar/src/main/kotlin/com/squareup/exemplar/HelloWebAction.kt
```diff
@@ -1,9 +1,6 @@
 package com.squareup.exemplar
 
-import misk.web.Get
-import misk.web.PathParam
-import misk.web.RequestHeaders
-import misk.web.ResponseContentType
+import misk.web.*
```
swankjesse: update your IntelliJ configs to disable wildcard imports
https://github.com/square/java-code-styles/

misk/src/main/kotlin/misk/web/extractors/QueryStringParameterProcessor.kt
```diff
@@ -0,0 +1,85 @@
+package misk.web.extractors
+
+import java.math.BigDecimal
+import java.math.BigInteger
+import kotlin.reflect.KParameter
+import kotlin.reflect.KType
+import kotlin.reflect.full.createType
+
+/**
+ * A query string parameter can be two things: a *primative* or a List<*primative*>
```
swankjesse: sp. primitive

misk/src/main/kotlin/misk/web/extractors/QueryStringParameterProcessor.kt
swankjesse: Agreed. I think we might want to be pretty opinionated on what types we support out of the box, and we should support ’em everywhere (path parameters, query parameters, config values, etc.)

From this list, I’d omit byte, byteArray, bigDecimal, bigInteger, and float. Maybe add `ByteString`, though we’d need to decide on hex or base64 as our thing

exemplar/src/main/kotlin/com/squareup/exemplar/HelloWebAction.kt
eeSeeGee: Gotcha, will do

exemplar/src/main/kotlin/com/squareup/exemplar/HelloWebAction.kt
eeSeeGee: Yup, optional, will dump

misk/src/main/kotlin/misk/web/extractors/QueryStringParameterExtractor.kt
eeSeeGee: I wrote all this code from scratch

misk/src/main/kotlin/misk/web/extractors/QueryStringParameterProcessor.kt
eeSeeGee: Yeah, I forgot Kotlin could do that until a little while ago

exemplar/src/main/kotlin/com/squareup/exemplar/HelloWebAction.kt
eeSeeGee: Yeah, that's set, I'm not sure why it's doing that.

PR #84, opened 2018-02-07
Google Cloud Datastore and Cloud Storage support

mmihic:
Adds modules for configuration access to Google Cloud Datastore and
Google Cloud Storage, as well as fakes / emulation support for writing
tests against those services.

Datastore testing is done against the local datastore emulator running
in memory-only mode. This ensures tests will experience as close as
possible real datastore semantics. Machines running these tests will
need to have the emulator installed (see `gcloud beta emulators`) but
the emulator itself will be automatically started and terminated by the
tests.

Cloud storage does not have an emulator, so tests will instead run
against a fake implementation of the storage RPC layer which keeps
everything in memory. As the cloud storage API and semantics are
substantially simpler than datastore, this is a reasonable compromise.

A future PR will contain an implementation of cloud storage that persists files on local disk, suitable for running development mode services.


misk/src/main/kotlin/misk/cloud/gcp/GoogleCloudModule.kt
```diff
@@ -0,0 +1,80 @@
...
+class GoogleCloudModule : KAbstractModule() {
+    override fun configure() {
+        binder().addMultibinderBinding<Service>().to<GoogleCloud>()
+    }
+
+    @Provides
+    @Singleton
+    fun provideServiceCredentials(env: Environment): Credentials =
+            if (env == Environment.DEVELOPMENT) NoCredentials.getInstance()
+            else ServiceAccountCredentials.getApplicationDefault()
```
nbroyles: This is configurable in some way right? The default may not be the appropriate service account (unless I'm misunderstanding how this works)

misk/testing/src/main/kotlin/misk/cloud/gcp/storage/InMemoryStorageRpc.kt
```diff
@@ -0,0 +1,394 @@
...
+import java.nio.file.FileAlreadyExistsException
+import java.util.concurrent.locks.ReentrantReadWriteLock
+import kotlin.concurrent.read
+import kotlin.concurrent.write
+
+internal class InMemoryStorageRpc : StorageRpc {
+    private val content = mutableMapOf<String, ByteArray>()
+    private val metadata = mutableMapOf<String, StorageObject>()
+    private val pendingContent = mutableMapOf<String, ByteArray>()
+    private val lock = ReentrantReadWriteLock()
```
nbroyles: Is this because we have one VM-wide datastore and tests could be running in parallel?

misk/src/main/kotlin/misk/cloud/gcp/GoogleCloudModule.kt
mmihic: We always use the default service account for permissions when running in the actual GCP environment.

misk/testing/src/main/kotlin/misk/cloud/gcp/storage/InMemoryStorageRpc.kt
mmihic: Usually you'll create a new instance per test so it's not an issue of multiple tests running in parallel. But you might use this for development mode services, in which case there could be multiple threads accessing this singleton concurrently.

nbroyles: Ah, got it. Makes sense.

maniksurtani:
What about the case when you have an app that isn't deployed to GCP, but
may still want to interact with GCD? Not an entirely uncommon setup, e.g.,
when building bridging apps.

misk/src/main/kotlin/misk/cloud/gcp/GoogleCloudModule.kt
maniksurtani: What about the case when you have an app that isn't deployed to GCP, but
may still want to interact with GCD? Not an entirely uncommon setup, e.g.,
when building bridging apps.

misk/testing/src/main/kotlin/misk/cloud/gcp/datastore/FakeDatastoreModule.kt
```diff
@@ -0,0 +1,52 @@
...
+
+        override fun startUp() {
+            // Reset on every restart / test run
+            datastoreHelper.reset()
+        }
+
+        override fun shutDown() {}
+    }
+
+    // NB(mmihic): We use a VM-wide singleton for the datastore because starting the datastore
```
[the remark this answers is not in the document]
mmihic: it's in a ``companion object`` which is basically the kotlin equivalent of static (sortof)

misk/testing/src/main/kotlin/misk/cloud/gcp/storage/InMemoryStorageRpc.kt
mmihic: because in multi-threading cases there will be way more readers than writers

misk/src/main/kotlin/misk/cloud/gcp/GoogleCloudModule.kt
mmihic: fair. can add more configurability down the line.

misk/src/main/kotlin/misk/cloud/gcp/GoogleCloudModule.kt
```diff
@@ -0,0 +1,80 @@
...
+/** Installs support for talking to real GCP services, either direct or via semulator */
+class GoogleCloudModule : KAbstractModule() {
+    override fun configure() {
+        binder().addMultibinderBinding<Service>().to<GoogleCloud>()
+    }
+
+    @Provides
+    @Singleton
+    fun provideServiceCredentials(env: Environment): Credentials =
+            if (env == Environment.DEVELOPMENT) NoCredentials.getInstance()
```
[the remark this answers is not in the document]
mmihic: yeah i'll file an issue for this and maybe do a PR for it before the enum branching gets too out of hand

misk/src/main/kotlin/misk/cloud/gcp/GoogleCloudModule.kt
```diff
@@ -0,0 +1,80 @@
...
+            else ServiceAccountCredentials.getApplicationDefault()
+
+    @Provides
+    @Singleton
+    fun provideCloudDatastore(credentials: Credentials, config: DatastoreConfig): Datastore =
+            DatastoreOptions.newBuilder()
+                    .setCredentials(credentials)
+                    .setHost(config.host)
+                    .setTransportOptions(HttpTransportOptions.newBuilder()
+                            .setConnectTimeout(config.connect_timeout_ms ?: -1)
```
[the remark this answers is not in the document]
mmihic: yeah i originally did it this way, not sure why i changed it.

PR #85, opened 2018-02-07
Expand upon routing rules for MediaRanges

tso:
Prior to this change if we had two routes with the same path
the one returned was arbitrary, even if one was the clear choice
in regards to content type. We now attempt to choose the route which
matches the desired content type.

We also attempt to route based, in some cases, on whether or not
the specified charset matches our route.

Along with this comes two small fixes:
  - Allow multiple media ranges to be specified in the accept header
  - Parse the URL only once per request

### DIFF
--- a/misk/src/main/kotlin/misk/web/BoundAction.kt
+++ b/misk/src/main/kotlin/misk/web/BoundAction.kt
@@ -54,9 +54,9 @@ internal class BoundAction<A : WebAction, out R>(
         return result
     }
 
-    fun match(jettyRequest: HttpServletRequest): RequestMatch? {
+    fun match(jettyRequest: HttpServletRequest, url: HttpUrl): RequestMatch? {
         // Confirm the path and method matches
-        val pathMatcher = pathMatcher(jettyRequest.httpUrl()) ?: return null
+        val pathMatcher = pathMatcher(url) ?: return null
         if (jettyRequest.method != httpMethod.name) return null
 
         // Confirm the request content type matches the types we accept, and pick the most specific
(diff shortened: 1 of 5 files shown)


misk/src/main/kotlin/misk/web/mediatype/MediaRange.kt
```diff
@@ -28,110 +28,103 @@ data class MediaRange(
...
+        return true
     }
 
     companion object {
         const val WILDCARD = "*"
         val ALL_MEDIA = MediaRange(WILDCARD, WILDCARD)
+    }
+}
 
-        fun parse(s: String): MediaRange {
```
mmihic: Why the rename? Would prefer that we keep the name as ``parse`` and possibly add a ``parseRanges`` (for parsing multiple ranges) to keep the naming consistent with its companion ``MediaType``

mmihic: Also strongly prefer this remain scoped vs being a naked package level functions.

tso:
Thanks for the context! I didn't see the point before, but now it makes a lot of sense. I've changed this to be less of a refactor and more of a fix + small refactor + some additional rules for routing.

I can separate into separate PRs if you'd like (fix, refactor for url parsing, and routing as separate PRs).

misk/src/main/kotlin/misk/web/BoundAction.kt
```diff
@@ -117,51 +117,27 @@ private fun HttpServletRequest.accepts(): List<MediaRange> {
...
-        val requestContentTypeMatch: MediaRange.Matcher?,
-        val responseContentTypeMatch: MediaRange.Matcher?
+        private val pathMatcher: Matcher,
+        private val responseContentTypeMatch: MediaRange.Matcher?
 ) : Comparable<RequestMatch> {
     override fun compareTo(other: RequestMatch): Int {
         val patternDiff = action.pathPattern.compareTo(other.action.pathPattern)
         if (patternDiff != 0) return patternDiff
+
+        if (responseContentTypeMatch != null && other.responseContentTypeMatch == null) return -1
```
mmihic: Oh, awesome! Ok thanks for doing this

PR #86, opened 2018-02-07
Use shadow plugin to generate exemplar fat-jar

nbroyles:
There previous technique caused really bizarre ZipException errors when attempting to include the exemplar JAR in our monorepo shim project. Using the Shadow plugin apparently sidesteps the issue as I was able to successfully build a project with the exemplar as a dependency.


exemplar/build.gradle
```diff
@@ -32,11 +32,21 @@ jar {
...
-  // and adds them to the JAR itself. One can extend this task, to skip certain
-  // files or particular types at will
-  from { configurations.compile.collect { it.isDirectory() ? it : zipTree(it) } }
+  classifier = 'unshaded'
 }
 
 apply from: rootProject.file('gradle/gradle-mvn-push.gradle')
+apply plugin: 'com.github.johnrengelman.shadow'
+
+task shadedJar(type: com.github.jengelman.gradle.plugins.shadow.tasks.ShadowJar) {
```
ryanhall07: nit: should the task be named shadowJar? not sure what the convention is for gradle. shaded might be a maven term

exemplar/build.gradle
```diff
@@ -32,11 +32,21 @@ jar {
...
+  classifier = null
+  manifest.attributes 'Main-Class': 'com.squareup.exemplar.ExemplarServiceKt'
+
+  from(project.convention.getPlugin(JavaPluginConvention).sourceSets.main.output)
+  configurations = [project.configurations.runtime]
+  exclude('META-INF/INDEX.LIST', 'META-INF/*.SF', 'META-INF/*.DSA', 'META-INF/*.RSA')
```
ryanhall07: are any of these default configs that don't need to specified? http://imperceptiblethoughts.com/shadow/

exemplar/build.gradle
nbroyles: Sure, I'll rename. I just made the assumption was a Java "thing" people knew.

exemplar/build.gradle
nbroyles: You've gotta carry this stuff along when you create a task that extends a ShadowJar or else the jar doesn't get build properly: https://github.com/johnrengelman/shadow/issues/108

exemplar/build.gradle
nbroyles: Actually, renaming this shadowJar would collide with the shadowJar task from the plugin. Just going to leave as-is. Since this is being published to Maven anyway, it's probably ok to reuse some terminology here.

ryanhall07: ok. i'd probably go with something like shadowJarTask

exemplar/build.gradle
ryanhall07: nice. thanks for the pointer

PR #91, opened 2018-02-10
Add support for commands

mmihic:
Allows applications to register additional commands that can be run
directly from the jar, typically used for utility commands. For example,
an application might register commands allowing for interaction with the
service from the command line, supporting scripting of administrative or
operational tasks.

If no command is specified on the command line, the service is started
and blocks until terminated (consistent with the prior behavior of
running the jar)

### DIFF
--- a/misk/src/main/kotlin/misk/MiskApplication.kt
+++ b/misk/src/main/kotlin/misk/MiskApplication.kt
@@ -1,24 +1,99 @@
 package misk
 
+import com.beust.jcommander.JCommander
+import com.beust.jcommander.ParameterException
+import com.google.common.annotations.VisibleForTesting
 import com.google.common.util.concurrent.ServiceManager
 import com.google.inject.Guice
 import com.google.inject.Module
+import misk.inject.KAbstractModule
 import misk.inject.getInstance
+import misk.logging.getLogger
 
-class MiskApplication(private vararg val modules: Module) {
-    fun startAndAwaitStopped() {
-        val injector = Guice.createInjector(*modules)
-        val serviceManager = injector.getInstance<ServiceManager>()
+/** The entry point for misk applications */
+class MiskApplication(private val modules: List<Module>, commands: List<MiskCommand> = listOf()) {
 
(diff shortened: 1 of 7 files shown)


mmihic:
Yeah the point of this is to make it easier to provide administrative commands that you'd either run interactively on a VM, or would put into an admin script / cronjob

misk/src/main/kotlin/misk/MiskApplication.kt
```diff
@@ -1,24 +1,99 @@
...
+
+    /**
+     * Runs the application, finding and executing the appropriate command based on the
+     * provided command line arguments
+     */
+    fun run(args: Array<String>) {
+        try {
+            doRun(args)
+        } catch (e: CliException) {
+            log.info(e.message)
```
ryanhall07: is info the right level?

mmihic:
Definitely not - this is intentionally designed to avoid that confusion. The problem in SC land is that the lifecycle of the application is controlled by the container, and the set of modules/services installed is global to the application rather than being specific to the command, so there was a) no way for a command to avoid starting services (the container did it automatically), b) no way for a command to control which services were installed (any dependencies declared on the application got installed, even if some weren't used by the command), and c) no way for a service to know whether it should let itself be started (a binary UTILITY mode is insufficient, since some commands might want some services and the service has no way of knowing).

The PR doesn't have any of that complexity - there are no modes, no enforced lifecycle, no automatic module/service installation, really no "container" per-se. There are just named commands that have parameters and a ``run`` method, with ``MiskApplication.run()``  being just a shell that parses the arguments to `main()` to determine which command to run and to populate that command's parameters with the values from the command line. Nothing is automatically installed by the container - if a command wants dependencies injected, it needs to explicitly specify the full set of modules/services to install. And nothing is automatically started - if a command wants to start services, it needs to explicitly do so in its ``run`` method. If no commands are installed or no arguments are passed, the application runs a default command that starts the server and waits for termination.

ryanhall07:
> if a command wants dependencies injected, it needs to explicitly specify the full set of modules/services to install.

nice! thanks for the thorough explanation!

PR #93, opened 2018-02-10
Update exemplar README.md due to shading

mmihic:
Now that we are using the shading plugin, it's no longer enough to run
gradle jar to produce an executable jar; you now must run gradle
assemble to produce the final executable shaded jar.


nbroyles:
Good catch!

PR #96, opened 2018-02-15
Add metrics backend for SignalFx

nbroyles:
Wire up SignalFx for a metric reporting option. Additionally, pull in the jackson-jsr310 dependency so that we can properly parse yaml configs for duration. Without this, loading duration from yaml will crash.


misk/src/main/kotlin/misk/metrics/backends/signalfx/SignalFxBackendModule.kt
```diff
@@ -0,0 +1,32 @@
...
+        binder().addMultibinderBinding<Service>().to<SignalFxReporterService>()
+    }
+
+    @Provides
+    @Singleton
+    fun signalFxReporter(
+            @AppName appName: String,
+            config: SignalFxBackendConfig,
+            metricRegistry: MetricRegistry
+            ): SignalFxReporter {
```
mmihic: nit: this indentation seems weird - are you using the standard codestyle from square/codestyles?

misk/src/test/kotlin/misk/config/ConfigTest.kt
```diff
@@ -33,6 +34,9 @@ class ConfigTest {
     @field:[Inject Named("consumer_b")]
     lateinit var consumerB: ConsumerConfig
 
+    @field:[Inject]
+    lateinit var durationConfig: DurationConfig
```
ryanhall07: why inject this?

misk/src/main/kotlin/misk/metrics/backends/signalfx/SignalFxBackendModule.kt
nbroyles: I am. Not sure what happened here, but cleaned it up.

misk/src/test/kotlin/misk/config/ConfigTest.kt
nbroyles: Good catch. Don't need it.

PR #98, opened 2018-02-16
Add utility classes for marshalling Wire Messages into HTTP requests

eeSeeGee:


misk/src/main/kotlin/misk/wire/ProtoMessageHttpClient.kt
```diff
@@ -0,0 +1,48 @@
...
+import okhttp3.OkHttpClient
+import okhttp3.Request
+import okhttp3.RequestBody
+import javax.inject.Inject
+
+/**
+ * This helper class is for making HTTP requests to an endpoint that handles protobuf Message claases.
+ */
+class ProtoMessageHttpClient {
+  @Inject
```
mmihic: nit: inject this as a ``val`` in the constructor

misk/src/main/kotlin/misk/wire/ProtoMessageHttpClient.kt
```diff
@@ -0,0 +1,48 @@
+package misk.wire
```
mmihic: nit: move to `misk.client`

misk/src/main/kotlin/misk/wire/AbstractClientCommand.kt
```diff
@@ -0,0 +1,24 @@
+package misk.wire
```
mmihic: I'm not sure this is sufficient value-add to add at the misk-level - the only thing it's really doing is defining a parameter for passing in the URL. I think that should just be left to the applications.

misk/src/main/kotlin/misk/wire/ProtoMessageHttpClient.kt
```diff
@@ -0,0 +1,48 @@
...
+        .addHeader("Accept", MediaTypes.APPLICATION_JSON)
+        .post(RequestBody.create(MediaTypes.APPLICATION_JSON_MEDIA_TYPE, requestJson))
+        .build()
+    val response = okHttp.newCall(request).execute()
+    if (response.code() != 200) {
+      throw RuntimeException("request failed (${response.code()} ${response.body()?.string()}"
+      )
+    }
+
+    return response.body()?.string()?.let { moshi.adapter(responseType).fromJson(it) }
```
mmihic: Please add tests.

misk/src/main/kotlin/misk/wire/ProtoMessageHttpClient.kt
eeSeeGee: Sure

misk/src/main/kotlin/misk/wire/ProtoMessageHttpClient.kt
eeSeeGee: Sure, makes sense.

misk/src/main/kotlin/misk/wire/ProtoMessageHttpClient.kt
eeSeeGee: Whoops, will do.

misk/src/main/kotlin/misk/client/ProtoMessageHttpClient.kt
```diff
@@ -0,0 +1,46 @@
...
+import javax.inject.Inject
+
+/**
+ * This helper class is for making HTTP requests to an endpoint that handles protobuf Message claases.
+ */
+class ProtoMessageHttpClient @Inject constructor(var moshi: Moshi) {
+
+  private val okHttp = OkHttpClient()
+
+  fun <O : Any> post(
```
ryanhall07: should this be private? would everybody use the friendlier version below?

ryanhall07:
A couple of nit fixes I noticed reviewing after merge

https://github.com/square/misk/pull/106

misk/src/main/kotlin/misk/client/ProtoMessageHttpClient.kt
mmihic: can't make this private if we intend to inline it, and you can only reify parameterized types in inline functions

PR #99, opened 2018-02-16
Add ErrorCodeLogLevelConfig to control error code log levels

ryanhall07:
This allows applications to use different log levels in different
environments.


misk/src/test/kotlin/misk/config/ConfigTest.kt
```diff
@@ -43,6 +45,9 @@ class ConfigTest {
         assertThat(testConfig.consumer_a).isEqualTo(ConsumerConfig(0, 1))
         assertThat(testConfig.consumer_b).isEqualTo(ConsumerConfig(1, 2))
         assertThat(testConfig.duration).isEqualTo(DurationConfig(Duration.parse("PT1S")))
+        assertThat((testConfig.error_code_log_level)).isEqualTo(
+            ErrorCodeLogLevelConfig(Level.INFO, Level.ERROR)
+        )
```
ryanhall07: this looks weird to me, but i'm using the SQ formatter

mmihic:
I think we want this to be a little more general purpose. ``ExceptionMapper``s can be used to map ay exception to a response, not just ``ActionException``s, so it should be possible to control whether to log other types of exceptions converted by ``ExceptionMapper``s

ryanhall07:
@mmihic yea this was just to allow config overrides for only the `ActionExceptionMapper`. maybe a better name for the config value? `action_exception_log_levels:` ?

the more generic exception handling is in https://github.com/square/misk/pull/108

ryanhall07:
@mmihic updated this to be a specific `ActionExceptionLogLevelConfig` for configuring the `ActionExceptionMapper`. PTAL

swankjesse:
Reformatted:
https://github.com/square/misk/pull/116

PR #101, opened 2018-02-17
Reformat all the code to match the latest SQ codestyle

ryanhall07:
Most of the noise is coming from the change to indent spaces 2 instead
of the Kotlin standard of 4.

https://github.com/square/java-code-styles/pull/51

This is causing a lot of local diff noise when developing, so figured
one giant change was better.

### DIFF
--- a/exemplar/src/main/kotlin/com/squareup/exemplar/HelloWebAction.kt
+++ b/exemplar/src/main/kotlin/com/squareup/exemplar/HelloWebAction.kt
@@ -12,18 +12,22 @@ import javax.inject.Singleton
 
 @Singleton
 class HelloWebAction : WebAction {
-    @Get("/hello/{name}")
-    @ResponseContentType(MediaTypes.APPLICATION_JSON)
-    fun hello(
-            @PathParam name: String,
-            @RequestHeaders headers: Headers,
-            @QueryParam nickName: String?,
-            @QueryParam greetings: List<String>?
-    ): HelloResponse {
-        return HelloResponse(
-                greetings?.joinToString(separator = " ") ?: "YO",
-                nickName?.toUpperCase() ?: name.toUpperCase())
-    }
+  @Get("/hello/{name}")
+  @ResponseContentType(MediaTypes.APPLICATION_JSON)
+  fun hello(
+      @PathParam name: String,
+      @RequestHeaders headers: Headers,
+      @QueryParam nickName: String?,
+      @QueryParam greetings: List<String>?
+  ): HelloResponse {
+    return HelloResponse(
+        greetings?.joinToString(separator = " ") ?: "YO",
+        nickName?.toUpperCase() ?: name.toUpperCase()
+    )
+  }
 }
 
-data class HelloResponse(val greeting: String, val name: String)
+data class HelloResponse(
+    val greeting: String,
+    val name: String
+)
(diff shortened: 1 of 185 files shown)


ryanhall07:
I only reformatted the KT and java files

```
~/Development/misk rhall-reformat git diff HEAD^1..HEAD --name-only | grep -v ".kt" | grep -v ".java"
~/Development/misk rhall-reformat
```

exemplar/src/main/kotlin/com/squareup/exemplar/HelloWebAction.kt
```diff
@@ -1,29 +1,29 @@
 package com.squareup.exemplar
 
-import misk.web.Get
-import misk.web.PathParam
-import misk.web.QueryParam
-import misk.web.RequestHeaders
-import misk.web.ResponseContentType
+import misk.web.*
```
tso: Probably want to not do start imports

ryanhall07: hmm that's not good! I'll look into my intellij did this. guess I should actually review this myself :)

ryanhall07: Fixed the issue with the sq formatter. Now I just need to get access rights to push my branch :)

ryanhall07: https://github.com/square/java-code-styles/pull/52

swankjesse:
I'd prefer to do this as a git history rewrite to avoid breaking `git blame`. It's more work but it avoids having a giant gap in the (short) history of the project.

ryanhall07:
Oh I've never done that before! Are you proposing we reformat each individual commit and amend it? I did a quick google search and couldn't find any tooling to support that

swankjesse:
Here’s a branch with the rewrites:
https://github.com/square/misk/commits/master_reformatted

swankjesse:
And here’s an example of a rewritten commit on [master](https://github.com/square/misk/commit/cdb284d3d3e80a35e16c2155aa5bdae9341fdbbb#diff-f812f946effe075065f618c590a74cf8R1) and [master_reformatted](https://github.com/square/misk/commit/3ae370707acc9d45fd095ee9ebbaec1e72ea8337#diff-f812f946effe075065f618c590a74cf8R1)

swankjesse:
... and the program that recursively “reformats” a directory of Kotlin files. This one just halves the leading spaces on `.kt` files and nothing else.

https://gist.github.com/swankjesse/7952d0c54efc2701fe6fc0bc184549cc

swankjesse:
@mmihic @nbroyles if y’all give me the green light, I’ll run this on `master`. I don’t think I can do it with a pull request though. It’ll just be a push to master with a new rewritten history.

The only hassle is that you’ll need to pull master and fixup in-flight PRs. Possibly by running the gist above.

ryanhall07:
@swankjesse sgtm!

stevecstian:
@swankjesse I'm not 100% sure, but I think you probably want to keep a tag or branch which points to the _`master-before-reformatting`_ to keep old commits reachable, especially for those commits mentioned in GitHub Issues.

PR #103, opened 2018-02-17
Fix wildcard based content routing

mmihic:
Fixed content based routing when the requests or actions have wildcards
in their accepted or emitted media ranges. The previous logic had
several issues:

1. We weren't considering the action's content type when ordering, so if
there were an action accepting text/* and another accepting text/plain,
and a request arrived with content-type text/plain, we'd randomly select
between those two actions. We now deterministically prefer the action
with the most specific matching content type (thus a request with text/plain
would match the action accepting text/plain, and a request text/yaml
would match the action accepting text/*)

2. We were ordering actions based on the Accepts header supplied by the
client rather than the media type defined on the action, so if there
were an action that could emit text/plain and another that could emit
*/* and a request arrived with Accepts: text/*, we'd randonly select
between those two actions. We now deteministically prefer the action
with the most specific matching response content type (thus a request
accepting text/* would prefer the action emitteing text/plain over the
action emitting */*).

As part of this, also cleaned up toString() methods on the various
matchers to make debugging easier.


misk/src/main/kotlin/misk/web/BoundAction.kt
```diff
@@ -114,25 +121,47 @@ private fun HttpServletRequest.accepts(): List<MediaRange> {
...
+internal open class RequestMatch(
+        private val pathPattern: PathPattern,
+        private val acceptedMediaRange: MediaRange,
+        private val requestCharsetMatch: Boolean,
+        private val responseContentType: MediaType
 ) : Comparable<RequestMatch> {
+
     override fun compareTo(other: RequestMatch): Int {
-        val patternDiff = action.pathPattern.compareTo(other.action.pathPattern)
+        val patternDiff = pathPattern.compareTo(other.pathPattern)
```
ryanhall07: What about using some kind of comparison chain? 

```
compareBy<RequestMatch>({ it.pathPattern}, {it.acceptedMediaRange}, ...).compare(this, other)
```

misk/src/main/kotlin/misk/web/BoundAction.kt
```diff
@@ -114,25 +121,47 @@ private fun HttpServletRequest.accepts(): List<MediaRange> {
...
+        // bother clarifying with the charset; we'd rather match a (text/*, text/plain)
+        // with no charset over a (text/*, text/*) with charset
+        if (requestCharsetMatch && !other.requestCharsetMatch) return -1
+        if (!requestCharsetMatch && other.requestCharsetMatch) return 1
 
         return 0
     }
 
+    override fun toString(): String =
+            "path: $pathPattern, accepts: $acceptedMediaRange, emits: $responseContentType"
```
ryanhall07: add the charset too?

misk/src/main/kotlin/misk/web/PathPattern.kt
```diff
@@ -8,14 +8,24 @@ import java.util.regex.Pattern
  * like this: `{username:[a-z]+}`. If no regex is specified the variable is a sequence of non-'/'
  * characters.
  */
-data class PathPattern(
-        val pattern: Pattern,
+class PathPattern(
```
ryanhall07: why remove data? because you want to control the equals/hashCode explicitly?

misk/src/test/kotlin/org/assertj/core/api/AssertExtensions.kt
```diff
@@ -19,4 +20,38 @@ fun <ACTUAL : CharSequence> AbstractCharSequenceAssert<*, ACTUAL>.isEqualToAsJso
...
+    }
+}
+
+fun <A : Comparable<A>> assertOrdering(vararg values: A) {
+    values.forEachIndexed { index, value ->
+        assertThat(value).isEqualByComparingTo(value)
+
+        val before = if (index == 0) listOf() else values.take(index - 1)
+        before.forEach {
+            assertThat(value).isGreaterThan(it)
```
ryanhall07: this assumes asc ordering right? maybe name the method `assertAscOrdering`

ryanhall07: also why not just use the built in sorted asserts? `assertThat(values).isSorted`

misk/src/main/kotlin/misk/web/PathPattern.kt
mmihic: correct

misk/src/test/kotlin/org/assertj/core/api/AssertExtensions.kt
mmihic: really hard to debug when the sorting goes wrong

misk/src/main/kotlin/misk/web/BoundAction.kt
mmihic: interesting, yeah, could do

misk/src/test/kotlin/org/assertj/core/api/AssertExtensions.kt
ryanhall07: yea that's fair. maybe just add a note, so a future contributor doesn't get smart and change it

misk/src/main/kotlin/misk/web/BoundAction.kt
mmihic: although tbh i sortof prefer to have request match be directly ``Comparable``

ryanhall07: sure, i'm just suggesting the impl of `BoundAction#compareTo` be a call to `compareBy`

mmihic: Ah, well, it gets to be tricker to combine ``compareBy`` with explicit comparison functions which is what's needed for the media type comparison and the charset comparisons (you need to declare anonymous types etc). I think having it as straight line code is easier to read

swankjesse:
Reformatted:
https://github.com/square/misk/pull/115

PR #104, opened 2018-02-17
Fix NotFoundAction handling

mmihic:
Previously POSTing to an unknown path would result in a 200 with an
empty response body, instead of the expected 404. Multiple problems:

1. The NotFoundAction was only bound for GET requests
2. The NotFoundAction only supported requests with Accept: text/plain

The NotFoundAction is now bound for both GET and POSTs, and supports
requests for any type of content (but always returns text/plain). Also
added a backstop in the WebActionsServlet to return a 404 if no suitable
action could be found; this handles cases where applications do not
install the NotFoundAction, or when a request arrives with an
unsupported HTTP method.

### DIFF
--- a/misk/src/main/kotlin/misk/web/actions/NotFoundAction.kt
+++ b/misk/src/main/kotlin/misk/web/actions/NotFoundAction.kt
@@ -2,18 +2,24 @@ package misk.web.actions
 
 import misk.web.Get
 import misk.web.PathParam
+import misk.web.Post
+import misk.web.RequestContentType
 import misk.web.Response
 import misk.web.ResponseContentType
 import misk.web.mediatype.MediaTypes
+import okhttp3.Headers
 import javax.inject.Singleton
 
 @Singleton
 class NotFoundAction : WebAction {
     @Get("/{path:.*}")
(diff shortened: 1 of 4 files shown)


misk/src/main/kotlin/misk/web/actions/NotFoundAction.kt
```diff
@@ -2,18 +2,24 @@ package misk.web.actions
...
+import okhttp3.Headers
 import javax.inject.Singleton
 
 @Singleton
 class NotFoundAction : WebAction {
     @Get("/{path:.*}")
-    @ResponseContentType(MediaTypes.TEXT_PLAIN_UTF8)
+    @Post("/{path:.*}")
+    @RequestContentType(MediaTypes.ALL)
+    @ResponseContentType(MediaTypes.ALL)
```
ryanhall07: why is the response type not plain text?

ryanhall07: I'm guessing this is required so it's not dependent on what the request accepts (e.g if the request accepts json, this should still work)

mmihic: correct

PR #108, opened 2018-02-21
Allow binding an ExceptionMapper by an Exception type

ryanhall07:
This adds deterministic ExceptionMapper resolution, instead of relying
on the binding order of ExceptionMappers.

If an Exception is thrown without an explicit bound ExceptionMapper, the
superclasses of the Exception are explored for possible bound
ExceptionMappers. The closest bound superclass wins.

e.g If an application adds an explicit binding for NotFoundException,
that explicit Mapper is preferred over the base ActionExceptionMapper.
However, if no binding for BadRequestException exists, the base
ActionExceptionMapper is used instead.

Since an Exception always resolves to the same ExceptionMapper for the
life of a VM, it is safe to cache the resolution result.

### DIFF
--- a/misk/src/main/kotlin/misk/web/WebModule.kt
+++ b/misk/src/main/kotlin/misk/web/WebModule.kt
@@ -2,6 +2,7 @@ package misk.web
 
 import misk.Interceptor
 import misk.MiskDefault
+import misk.exceptions.ActionException
 import misk.inject.KAbstractModule
 import misk.inject.addMultibinderBinding
 import misk.inject.addMultibinderBindingWithAnnotation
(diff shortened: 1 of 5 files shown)


misk/src/main/kotlin/misk/web/WebModule.kt
```diff
@@ -2,6 +2,8 @@ package misk.web
 
 import misk.Interceptor
 import misk.MiskDefault
+import misk.exceptions.ActionException
+import misk.exceptions.NotFoundException
```
mmihic: not used?

misk/src/main/kotlin/misk/web/exceptions/ExceptionMapperResolver.kt
```diff
@@ -0,0 +1,50 @@
+package misk.web.exceptions
+
+import java.util.LinkedList
+import java.util.concurrent.ConcurrentHashMap
+import java.util.concurrent.ConcurrentMap
+import javax.inject.Inject
+import javax.inject.Singleton
+import kotlin.reflect.KClass
+import kotlin.reflect.full.superclasses
+import kotlin.reflect.jvm.internal.impl.utils.DFS
```
mmihic: hmm

ryanhall07: fair enough

PR #109, opened 2018-02-22
Support multiple RequestContent types in one method

eeSeeGee:


misk/src/test/kotlin/misk/web/ContentBasedDispatchTest.kt
```diff
@@ -30,13 +37,21 @@ internal class ContentBasedDispatchTest {
 
     data class Packet(val message: String)
 
+    data class ComplexPacket(val message: String, val composedBy: String = "default") {
```
mmihic: What this test is covering is a bit hard to understand. I'd create separate tests for "use generic unmarshalling of types with valueOf methods" and "picks the appropriate unmarshaller based on incoming content type". The first could just be tested at the unmarshaller level (i.e. be in a simpler more isolated test class), the second is more appropriate to content based dispatch.

misk/src/main/kotlin/misk/web/extractors/RequestBodyParameterExtractor.kt
```diff
@@ -20,7 +20,8 @@ internal class RequestBodyParameterExtractor(
     override fun extract(webAction: WebAction, request: Request, pathMatcher: Matcher): Any? {
         val mediaType = request.headers["Content-Type"]?.let { MediaType.parse(it) }
         val unmarshaller = mediaType?.let { type ->
-            unmarshallerFactories.map { it.create(type, parameter.type) }.firstOrNull()
+            unmarshallerFactories.map { it.create(type, parameter.type) }
```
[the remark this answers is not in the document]
eeSeeGee: Nice, Kotlin sure has a lot of extensions

misk/src/main/kotlin/misk/web/marshal/Generic.kt
```diff
@@ -23,6 +26,13 @@ object GenericUnmarshallers {
             ByteString::class.java
     )
 
+    // This generic handler can potentially be called millions of times, each time returning what's
+    // effectively the same object over and over. Remember the results in this map to avoid
+    // polluting the GC, and also to avoid making the call at all when no conversion is possible.
+    // Rather than waste a bunch of time on sychronization for a very small map just keep a separate
+    // copy in each thread.
+    private val stringConverterMemo = ThreadLocal<LinkedHashMap<KType, ToStringConvertable?>>()
```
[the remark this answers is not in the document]
eeSeeGee: D'oh. I could do a concurrent hash map here, but maybe this object isn't really heavy enough to require the cost of locking.

misk/src/test/kotlin/misk/web/ContentBasedDispatchTest.kt
eeSeeGee: Yeah, I suppose the test outgrew the parameters of this task set. I'll split it out.

misk/src/main/kotlin/misk/web/marshal/Generic.kt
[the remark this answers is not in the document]
swankjesse: Probably just wants to be a Guava Cache?

misk/src/test/kotlin/misk/web/ContentBasedDispatchTest.kt
```diff
@@ -130,4 +195,19 @@ internal class ContentBasedDispatchTest {
...
+    class ComplexPacketUnmarshaller : Unmarshaller {
+        override fun unmarshal(source: BufferedSource): Any?
+            = ComplexPacket(source.readUtf8(), "unmarshaller")
+
+        class Factory: Unmarshaller.Factory {
+            override fun create(mediaType: MediaType, type: KType): Unmarshaller? {
+                if (mediaType.type() != complexMediaType.type() ||
+                    mediaType.subtype() != complexMediaType.subtype()) return null
+
+                if (GenericUnmarshallers.canHandle(type)) return null
```
swankjesse: this one makes me anxious. . . I think this precedence call should be somewhere else . . .if this marshaller is lower-precedence than the generic ones then it should be called later

eeSeeGee: This call doesn't really make sense now that I think of it; if GenericUnmarshallers can do a better job of converting that type why not just do it directly? I got this pattern from Json unmarshalling, seemed like a good idea when not thinking too much about it.

swankjesse:
Reformatted here:
young.20180222.support_multiple_request_content

swankjesse:
Reformatted here:
https://github.com/square/misk/pull/114

PR #110, opened 2018-02-22
Pass app config to interested modules instead of injecting

nbroyles:
This will allow us to install Modules based on configuration values without needing to do two-phase/bootstrapping injection.


misk/src/test/kotlin/misk/config/ConfigTest.kt
```diff
@@ -18,9 +15,10 @@ import javax.inject.Named
 
 @MiskTest
 class ConfigTest {
+    val config = MiskConfig.load<TestConfig>(TestConfig::class.java, "test_app", TESTING)
```
ryanhall07: it can't infer `TestConfig`?

ryanhall07:
you'll need update exemplar right?

misk/src/main/kotlin/misk/config/MiskConfig.kt
```diff
@@ -0,0 +1,76 @@
...
+import com.fasterxml.jackson.datatype.jsr310.JavaTimeModule
+import com.fasterxml.jackson.module.kotlin.KotlinModule
+import com.fasterxml.jackson.module.kotlin.MissingKotlinParameterException
+import misk.environment.Environment
+import java.io.BufferedReader
+import java.io.InputStreamReader
+import java.net.URL
+
+object MiskConfig {
+  fun <T : Config> load(
```
mmihic: provide an inline reified version of this and prefer that in samples. e.g.

````kotlin
inline fun <reified T: Config> load(appName: String, environment: Environment) = 
   load(T::class.java, appName, environment)
````

Use

````kotlin
val config = MiskConfig.load<TestConfig>("myapp", Environment.DEVELOPMENT)
````

File: misk/src/test/kotlin/misk/config/ConfigTest.kt (hunk 1 above)
Comment (mmihic) [reply misk/src/test/kotlin/misk/config/ConfigTest.kt]: It should. But also, see above comments on using an inlined reified version for this.

File: misk/src/main/kotlin/misk/config/MiskConfig.kt [hunk 3]
```diff
@@ -0,0 +1,76 @@
...
+      configClass: Class<out Config>,
+      appName: String,
+      environment: Environment
```
mmihic: Since we're now requiring that the environment be known before injection, we should move the lookup logic from ``EnvironmentModule.fromEnvironmentVariable`` to ``Environment.fromEnvironmentVariable`` so that you can get the ``Environment`` without creating an injector.

e.g.

````kotlin
enum class Environment {
   companion object {
       fun fromEnvironmentVariable() : Environment { /* existing lookup logic */   }
   }
}

class EnvironmentModule(val environment: Environment) : KAbstractModule() {
    override fun configure() {
      bind<Environment>().toInstance(environment)
    }
}
````

File: exemplar/src/main/java/com/squareup/exemplar/ExemplarJavaConfigModule.java [hunk 4]
```diff
@@ -4,7 +4,13 @@
...
 public class ExemplarJavaConfigModule extends AbstractModule {
+  private final ExemplarConfig config;
+
+  ExemplarJavaConfigModule(ExemplarConfig config) {
+    this.config = config;
+  }
+
   @Override protected void configure() {
-    install(new ConfigModule(ExemplarJavaConfig.class, "exemplar"));
+    install(new ConfigModule(ExemplarJavaConfig.class, "exemplar", config));
```
ryanhall07: this is cool

exemplar/src/main/java/com/squareup/exemplar/ExemplarJavaApp.java
```diff
@@ -2,23 +2,25 @@
...
+import misk.environment.Environment;
 import misk.environment.EnvironmentModule;
 import misk.hibernate.HibernateModule;
 import misk.moshi.MoshiModule;
 import misk.web.WebModule;
 
 public class ExemplarJavaApp {
   public static void main(String[] args) {
+    Environment environment = Environment.fromEnvironmentVariable();
+    ExemplarConfig config = MiskConfig.load(ExemplarConfig.class, "exemplar", environment);
```
ryanhall07: is this why you need the `@JvmStatic`?

nbroyles: Exactly. Otherwise, I'd have to do `MiskConfig.INSTANCE.load` which works, but is obviously more verbose.

exemplar/src/main/java/com/squareup/exemplar/ExemplarJavaConfigModule.java
```diff
@@ -4,7 +4,13 @@
 import misk.config.ConfigModule;
 
 public class ExemplarJavaConfigModule extends AbstractModule {
```
swankjesse: would it be simpler to just inline this entire module at the callsite?

nbroyles: I like that. Will fix up

PR #113, opened 2018-02-23
Wire up configuration to support distributed tracing in Misk

nbroyles:
Uses opentracing.io to allow us to have flexibility on which Tracer we choose. Initially wired up to support Uber's Jaeger out of the box. Follow up PRs will add support for automatically tracing certain actions/methods, and annotations to indicate you would like a desired method to be traced.


misk/src/main/kotlin/misk/tracing/backends/jaeger/JaegarSamplerConfig.kt
```diff
@@ -0,0 +1,9 @@
+package misk.tracing.backends.jaeger
+
+import misk.config.Config
+
+data class JaegerSamplerConfig (
+  val type: String?,
```
ryanhall07: what do null values mean for these configs? we should document them

misk/src/test/kotlin/misk/config/ConfigTest.kt
```diff
@@ -67,23 +67,30 @@ class ConfigTest {
 
     @Test
     fun friendlyErrorMessagesWhenFilesNotFound() {
-        assertThrows(IllegalStateException::class.java, {
+        val exception = assertThrows(IllegalStateException::class.java, {
```
ryanhall07: why the change?

build.gradle
```diff
@@ -35,6 +35,8 @@ ext {
           junitEngine          : 'org.junit.jupiter:junit-jupiter-engine:5.0.2',
           gcpCloudStorage      : 'com.google.cloud:google-cloud-storage:1.15.0',
           jCommander           : 'com.beust:jcommander:1.72',
+          openTracing          : 'io.opentracing:opentracing-api:0.31.0',
+          tracingJaeger        : 'com.uber.jaeger:jaeger-core:0.24.0',
```
ryanhall07: should the jaeger stuff be a separate artifact that clients can depend on if they choose to install jaeger? not sure where we draw line on what we package together. @mmihic @swankjesse any opinion?

misk/src/main/kotlin/misk/tracing/TracingModule.kt
```diff
@@ -0,0 +1,16 @@
+package misk.tracing
+
+import misk.inject.KAbstractModule
+import misk.tracing.backends.jaeger.JaegerBackendModule
+
+class TracingModule(val config: TracingConfig) : KAbstractModule() {
+  override fun configure() {
+    if (config.tracer.equals("jaeger", true)) {
```
mmihic: Maybe just use the presence of a non-null `jaegar` property in `TracingConfig` to decide this vs having a separate `type` property?

ryanhall07: I like that. Should then verify only a single property is set

mmihic:
This also begs for better validation on config (eg should be able to say
that one and only one of X properties must be set)

build.gradle
swankjesse: Yes, though I don’t think that’s necessary for this PR.

misk/src/main/kotlin/misk/tracing/backends/jaeger/JaegarSamplerConfig.kt
swankjesse: or give them defaults?

misk/src/main/kotlin/misk/tracing/backends/jaeger/JaegerBackendModule.kt
```diff
@@ -0,0 +1,36 @@
...
+class JaegerBackendModule(val config: JaegerBackendConfig?) : KAbstractModule() {
+  override fun configure() {
+    binder().addMultibinderBinding<Service>().to<TracingService>()
+  }
+
+  @Provides
+  @Singleton
+  fun jaegerTracer(@AppName appName: String): Tracer {
+    val reporter =
+        if (config?.reporter == null) Configuration.ReporterConfiguration()
```
swankjesse: could you default `reporter` to a new instance?

misk/src/main/kotlin/misk/tracing/backends/jaeger/JaegerBackendModule.kt
```diff
@@ -0,0 +1,36 @@
...
+        else Configuration.ReporterConfiguration(
+            config.reporter.log_spans, config.reporter.agent_host,
+            config.reporter.agent_port, config.reporter.flush_interval_ms,
+            config.reporter.max_queue_size)
+
+    val sampler =
+        if (config?.sampler == null) Configuration.SamplerConfiguration()
```
swankjesse: same

misk/src/test/kotlin/misk/config/ConfigTest.kt
```diff
@@ -67,23 +67,30 @@ class ConfigTest {
...
             MiskConfig.load<TestConfig>(TestConfig::class.java, "missing", defaultEnv)
-        }, "could not find configuration files - checked [missing-common.yaml, missing-testing.yaml]")
+        })
+
+        assertThat(exception.localizedMessage).contains("could not find configuration files -" +
+            " checked [missing-common.yaml, missing-testing.yaml]")
```
swankjesse: @ryanhall07 presumably for this assert?

misk/src/test/kotlin/misk/config/ConfigTest.kt
nbroyles: This method doesn't behave as expected. If you dig into the assertThrows impl, if the exception is the same as the argument passed, the string provided is ignored.

misk/src/main/kotlin/misk/tracing/backends/jaeger/JaegerBackendModule.kt
nbroyles: `config.reporter` is a different type from the object being instantiated here. Changing the type on the config would break loading from the yaml

build.gradle
nbroyles: How would this work when misk code has a references to classes in the jaeger artifact? Seems like you wouldn't really be able to separate the two, though there could be some Java/Kotlin sorcery I'm unaware of.

ryanhall07: You need to split the misk code and create a separate `misk-jaeger` artifact. if apps want jaeger, they include the `misk-jaeger` dep and then can install the `JaegerModule` in their app module.

PR #114, opened 2018-02-24
Support multiple RequestContent types in one method

swankjesse:


misk/src/main/kotlin/misk/web/marshal/Generic.kt
```diff
@@ -23,6 +27,21 @@ object GenericUnmarshallers {
...
   )
 
+  // This generic handler can potentially be called millions of times, each time returning what's
+  // effectively the same object over and over. Remember the results in this map to avoid
+  // polluting the GC, and also to avoid making the call at all when no conversion is possible.
+  private val stringConvertableCache = CacheBuilder.newBuilder()
+      .maximumSize(1000)
+      .build(
+          object: CacheLoader<KType, ToStringConvertable?>() {
+            override fun load(key: KType?): ToStringConvertable? {
```
mmihic: The ``key``parameter and the return value aren't nullable; this is Kotlin incorrectly guessing nullability for an API which hasn't been marked with ``@NonNull``. And in fact it's explicitly illegal to return null from a  [``CacheLoader``](https://google.github.io/guava/releases/23.0/api/docs/com/google/common/cache/CacheLoader.html#load-K-). Probably need to return a sentinel value instead to represent no converter available.

eeSeeGee: Updated, PTAL

PR #116, opened 2018-02-24
Add ActionExceptionLogLevelConfig to control log levels for ActionExceptions

swankjesse:
This allows applications to use different log levels in different
environments.


ryanhall07:
closes https://github.com/square/misk/issues/76

PR #118, opened 2018-02-24
Run IntelliJ's formatter on all of Misk.

swankjesse:
This fixes some places where my re-indenter cut 2-space indented files down to 1.
It also fixes some places where wrapped parameters were not indented +4.

And some other assorted reformattings to follow IntelliJ's rules. Probably the
most interesting is that if parameters are at all multiline, they are strictly
multiline.

### DIFF
--- a/misk/src/main/kotlin/misk/Action.kt
+++ b/misk/src/main/kotlin/misk/Action.kt
@@ -4,8 +4,8 @@ import kotlin.reflect.KFunction
 import kotlin.reflect.KType
 
 data class Action(
-  val name: String,
-  val function: KFunction<*>,
-  val parameterTypes: List<KType>,
-  val returnType: KType
+    val name: String,
+    val function: KFunction<*>,
+    val parameterTypes: List<KType>,
+    val returnType: KType
 )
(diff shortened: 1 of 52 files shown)


misk/src/main/kotlin/misk/Action.kt
```diff
@@ -4,8 +4,8 @@ import kotlin.reflect.KFunction
 import kotlin.reflect.KType
 
 data class Action(
-  val name: String,
-  val function: KFunction<*>,
-  val parameterTypes: List<KType>,
-  val returnType: KType
+    val name: String,
```
AlecStrong: might need to update either intellij or your code style, constructor properties should have 2 space indent (assuming 2 space indent is the norm for the project)

swankjesse: Yup. I think I got that wrong. Fixing.

misk/src/main/kotlin/misk/cloud/gcp/storage/BaseCustomStorageRpc.kt
```diff
@@ -126,23 +128,31 @@ abstract class BaseCustomStorageRpc : StorageRpc {
     throw UnsupportedOperationException()
   }
 
-  override fun deleteAcl(bucket: String?, entity: String?,
-      options: Map<StorageRpc.Option, *>?): Boolean {
+  override fun deleteAcl(
+      bucket: String?, entity: String?,
```
AlecStrong: looks like it missed some multilining in this file

misk/src/main/kotlin/misk/Action.kt
swankjesse: Actually hang on; I think the _Square Android_ style guide is the one that’s wrong. Should be double the normal indent for parameters lists.
https://android.github.io/kotlin-guides/style.html#functions

PR #119, opened 2018-02-24
Mechanically reformat the Misk codebase.

swankjesse:
This uses IntelliJ's formatter and Square's kotlin style rules.

The biggest difference with what is currently checked in is that
method parameters and constructor parameters are indented +2 instead
of their current +4.


swankjesse:
See also: https://github.com/square/java-code-styles/pull/54

PR #120, opened 2018-02-26
Simplify installing tracers

nbroyles:
Removes a superfluous tracer field and adds documentation to Jaeger configuration classes.

### DIFF
--- a/misk/src/main/kotlin/misk/tracing/TracingModule.kt
+++ b/misk/src/main/kotlin/misk/tracing/TracingModule.kt
@@ -5,12 +5,12 @@ import misk.tracing.backends.jaeger.JaegerBackendModule
 
 class TracingModule(val config: TracingConfig) : KAbstractModule() {
   override fun configure() {
-    if (config.tracer.equals("jaeger", true)) {
+    if (config.backends.jaeger != null) {
       install(JaegerBackendModule(config.backends.jaeger))
       return
     }
 
-    throw IllegalArgumentException("No backend for '${config.tracer}' tracer. Ensure that " +
-        "tracer name is spelled correctly and that a backend module exists for that tracer.")
+    throw IllegalStateException("No backends configured for tracing. Please update your yaml "
+      + "configuration.")
   }
 }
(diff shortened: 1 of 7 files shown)


misk/src/main/kotlin/misk/tracing/backends/jaeger/JaegarSamplerConfig.kt
```diff
@@ -2,6 +2,10 @@ package misk.tracing.backends.jaeger
 
 import misk.config.Config
 
+/**
+ * Configuration for Jaeger's sampler. If values are left null then Jaeger will provide defaults.
+ * See [com.uber.jaeger.Configuration.SamplerConfiguration] for default values.
```
ryanhall07: are the default values public? if so, might be nice to inline and remove the nullability

misk/src/main/kotlin/misk/tracing/TracingModule.kt
```diff
@@ -5,12 +5,12 @@ import misk.tracing.backends.jaeger.JaegerBackendModule
...
   override fun configure() {
-    if (config.tracer.equals("jaeger", true)) {
+    if (config.backends.jaeger != null) {
       install(JaegerBackendModule(config.backends.jaeger))
       return
     }
 
-    throw IllegalArgumentException("No backend for '${config.tracer}' tracer. Ensure that " +
-        "tracer name is spelled correctly and that a backend module exists for that tracer.")
+    throw IllegalStateException("No backends configured for tracing. Please update your yaml "
```
ryanhall07: basically the tab vs. whitespace debate of exception throwing. I see you've chosen your religion

misk/src/main/kotlin/misk/tracing/backends/jaeger/JaegarSamplerConfig.kt
nbroyles: Not all of them which is super annoying. That was my initial plan until I noticed difference levels of visibility. 

Also would've been nice if they had public builders for their configs too which would have made instantiating them a little less gross.

:-/

misk/src/main/kotlin/misk/tracing/TracingModule.kt
nbroyles: 😬

PR #122, opened 2018-02-26
Introduce websocket support in misk

tso:


exemplar/src/main/kotlin/com/squareup/exemplar/actions/ChatWebSocketAction.kt
```diff
@@ -0,0 +1,52 @@
...
+
+import misk.web.ConnectWebSocket
+import misk.web.PathParam
+import misk.web.actions.WebAction
+import misk.web.actions.WebSocket
+import misk.web.actions.WebSocketListener
+import javax.inject.Singleton
+
+@Singleton
+class ChatWebSocketAction : WebAction {
```
swankjesse: Wanna write a few lines of documentation explaining what this does?

exemplar/src/main/kotlin/com/squareup/exemplar/actions/ChatWebSocketAction.kt
```diff
@@ -0,0 +1,52 @@
...
+  private val rooms = mutableMapOf<String, Room>()
```
swankjesse: this needs _something_ for concurrency. I think a Guava cache is the easiest thing to do self-creating map thingies.

``` 
  LoadingCache<Key, Graph> graphs = CacheBuilder.newBuilder()
       .build(
           new CacheLoader<Key, Graph>() {
             public Graph load(Key key) throws AnyException {
               return createExpensiveGraph(key);
             }
           });
```

exemplar/src/main/kotlin/com/squareup/exemplar/actions/ChatWebSocketAction.kt
```diff
@@ -0,0 +1,52 @@
...
+  fun chat(@PathParam name: String, webSocket: WebSocket): WebSocketListener {
+    val room = rooms.getOrDefault(name, Room(name))
+    room.join(webSocket)
+    rooms[name] = room
+    return room
+  }
+}
+
+private class Room(private val name: String) : WebSocketListener() {
+  private val history = mutableListOf<String>()
```
swankjesse: CopyOnWriteArrayList ?

exemplar/src/main/kotlin/com/squareup/exemplar/actions/ChatWebSocketAction.kt
```diff
@@ -0,0 +1,52 @@
...
+  private val sockets = mutableListOf<WebSocket>()
```
swankjesse: same?

exemplar/src/main/kotlin/com/squareup/exemplar/actions/ClockWebSocketAction.kt
```diff
@@ -0,0 +1,37 @@
...
+import misk.web.actions.WebSocket
+import misk.web.actions.WebSocketListener
+import java.util.Date
+import java.util.concurrent.ScheduledExecutorService
+import java.util.concurrent.TimeUnit
+import javax.inject.Inject
+import javax.inject.Singleton
+
+@Singleton
+class ClockWebSocketAction : WebAction {
```
swankjesse: kill this? The chat example is strictly better

misk/src/main/kotlin/misk/web/BoundAction.kt
```diff
@@ -52,22 +53,26 @@ internal class BoundAction<A : WebAction, out R>(
     return result
   }
 
-  fun match(jettyRequest: HttpServletRequest, url: HttpUrl): RequestMatch? {
+  fun match(
+    requestMethod: String,
+    reqContentType: MediaType?,
```
swankjesse: supernit: req –> request

misk/src/main/kotlin/misk/web/BoundAction.kt
```diff
@@ -77,7 +82,7 @@ internal class BoundAction<A : WebAction, out R>(
     // Find values for all the parameters.
     val webAction = webActionProvider.get()
     val parameters = parameterExtractors.map {
-      it.extract(webAction, request, pathMather)
+      it.extract(webAction, request, null, pathMather)
```
swankjesse: should this parameter default to null?

misk/src/main/kotlin/misk/web/actions/WebSocket.kt
```diff
@@ -0,0 +1,18 @@
+package misk.web.actions
+
+import okio.ByteString
+
+interface WebSocket {
```
swankjesse: I think later we just steal all of the docs from OkHttp’s WebSocket

swankjesse: Should this have a field `Request` as the request that initialized it?

misk/src/main/kotlin/misk/web/interceptors/MarshallerInterceptor.kt
```diff
@@ -35,6 +36,9 @@ internal class MarshallerInterceptor constructor(private val marshaller: Marshal
     @JvmSuppressWildcards private val marshallerFactories: List<Marshaller.Factory>
   ) : Interceptor.Factory {
     override fun create(action: Action): Interceptor? {
+      if (action.returnType.classifier == WebSocketListener::class) {
+        return null
```
swankjesse: short comment?

misk/src/main/kotlin/misk/web/jetty/RealWebSocket.kt
```diff
@@ -0,0 +1,98 @@
...
+
+private const val MAX_QUEUE_SIZE = 16 * 1024 * 1024
+
+class RealWebSocket : WebSocket {
+
+  /** Total size of messages enqueued and not yet transmitted by Jetty. */
+  private var outgoingQueueSize = 0L
+
+  /** Messages to send when the Web Socket connects. */
+  private var queue = ArrayDeque<String>()
```
swankjesse: later we’ll need like a `Message` abstraction in here so it can be either a String or a ByteString

misk/src/main/kotlin/misk/web/jetty/WebActionsServlet.kt
```diff
@@ -38,19 +43,48 @@ internal class WebActionsServlet @Inject constructor(
         keyOf<Request>() to asRequest)
 
     scope.enter(seedData).use {
-      val candidateActions = boundActions.mapNotNull { it.match(request, asRequest.url) }
+      val candidateActions = boundActions.mapNotNull {
+        it.match(
+            request.method,
+            request.contentType?.let { MediaType.parse(it) },
```
swankjesse: nice

misk/src/main/kotlin/misk/web/jetty/WebActionsServlet.kt
```diff
@@ -38,19 +43,48 @@ internal class WebActionsServlet @Inject constructor(
...
   }
+
+  override fun configure(factory: WebSocketServletFactory?) {
+    factory!!.creator = WebSocketCreator { servletUpgradeRequest, servletUpgradeResponse ->
+      val request = servletUpgradeRequest.httpServletRequest
+      val asRequest = Request(
+          request.urlString(),
+          HttpMethod.valueOf(request.method),
+          Headers.of(),
+          Buffer()
```
swankjesse: comment: // Empty request body.

misk/src/main/kotlin/misk/web/jetty/WebActionsServlet.kt
```diff
@@ -38,19 +43,48 @@ internal class WebActionsServlet @Inject constructor(
...
     }
   }
+
+  override fun configure(factory: WebSocketServletFactory?) {
+    factory!!.creator = WebSocketCreator { servletUpgradeRequest, servletUpgradeResponse ->
+      val request = servletUpgradeRequest.httpServletRequest
+      val asRequest = Request(
+          request.urlString(),
+          HttpMethod.valueOf(request.method),
+          Headers.of(),
```
swankjesse: we should get the actual headers

misk/src/main/kotlin/misk/web/jetty/WebActionsServlet.kt
```diff
@@ -38,19 +43,48 @@ internal class WebActionsServlet @Inject constructor(
...
       val bestAction = candidateActions.sorted().firstOrNull()
       bestAction?.handle(asRequest, response)
       logger.debug { "Request handled by WebActionServlet" }
     }
   }
+
+  override fun configure(factory: WebSocketServletFactory?) {
+    factory!!.creator = WebSocketCreator { servletUpgradeRequest, servletUpgradeResponse ->
+      val request = servletUpgradeRequest.httpServletRequest
+      val asRequest = Request(
```
swankjesse: does request.asRequest() work?

misk/src/main/kotlin/misk/web/jetty/WebActionsServlet.kt
```diff
@@ -38,19 +43,48 @@ internal class WebActionsServlet @Inject constructor(
...
+            request.method,
+            null,
+            listOf(),
+            asRequest.url
+        )
+      }
+
+      val bestAction = candidateActions.sorted().firstOrNull() ?: return@WebSocketCreator null
+      val realWebSocket = RealWebSocket()
+      val webSocketListener = bestAction.handleWebSocket(asRequest, realWebSocket)
```
swankjesse: throw or explode if this is null? (is that possible?)

swankjesse: probably possible due to Java

misk/src/main/kotlin/misk/web/jetty/WebActionsServlet.kt
```diff
@@ -59,11 +93,31 @@ internal fun HttpServletRequest.asRequest(): Request {
...
-      HttpUrl.parse(urlString)!!,
+      urlString(),
       HttpMethod.valueOf(method),
       headersBuilder.build(),
       Okio.buffer(Okio.source(inputStream))
   )
 }
+
+private fun HttpServletRequest.urlString(): HttpUrl {
+  return HttpUrl.parse(
```
swankjesse: nit: you can make this a little tiny bit more efficient?
```
  return if (queryString != null)
      HttpUrl.parse(requestURL) else
      HttpUrl.parse(requestURL + "?" + queryString)

File: misk/src/test/kotlin/misk/web/WebSocketsTest.kt [hunk 17]
```diff
@@ -0,0 +1,74 @@
...
+  @Inject lateinit var jettyService: JettyService
+
+  @Test
+  fun basicWebSocket() {
+    val client = OkHttpClient.Builder()
+        .build()
+
+    val httpServerUrl = jettyService.httpServerUrl
+    val request = okhttp3.Request.Builder()
+        .url("ws://${httpServerUrl.host()}:${httpServerUrl.port()}/hello")
```
swankjesse: you can just pass a httpServerUrl and it should mostly do the right thing?

misk/src/test/kotlin/misk/web/WebSocketsTest.kt
```diff
@@ -0,0 +1,74 @@
...
+
+    var message: String? = null
+    client.newWebSocket(request, object : okhttp3.WebSocketListener() {
+      override fun onMessage(webSocket: okhttp3.WebSocket, text: String) {
+        message = text
+      }
+    })
+
+    var sleeps = 0
+    while (sleeps < TIMEOUT / SLEEP_DURATION) {
```
swankjesse: ya wanna use something like a LinkedBlockingDeque

misk/src/test/kotlin/misk/web/WebSocketsTest.kt
```diff
@@ -0,0 +1,74 @@
...
+      if (message != null) break
+      Thread.sleep(SLEEP_DURATION)
+      sleeps += 1
+    }
+
+    Assertions.assertEquals("Hello!", message)
```
swankjesse: nit: import so you don’t need the `Assertions` prefix

misk/src/test/kotlin/misk/web/WebSocketsTest.kt
```diff
@@ -0,0 +1,74 @@
...
+  }
+
+  @Singleton
+  class HelloWebSocket : WebAction {
+    @ConnectWebSocket("/hello")
+    fun hello(webSocket: WebSocket): WebSocketListener {
+      webSocket.send("Hello!")
+      return WebSocketListener()
```
swankjesse: ya need a test to confirm that the listener also gets stuff

misk/src/main/kotlin/misk/web/extractors/RequestBodyParameterExtractor.kt
```diff
@@ -17,7 +18,12 @@ internal class RequestBodyParameterExtractor(
   private val parameter: KParameter,
   private val unmarshallerFactories: List<Unmarshaller.Factory>
 ) : ParameterExtractor {
-  override fun extract(webAction: WebAction, request: Request, pathMatcher: Matcher): Any? {
+  override fun extract(
+    webAction: WebAction,
+    request: Request,
+    webSocket: WebSocket?,
```
ryanhall07: should a `WebSocket : WebAction`? `WebAction` appears to just be a marker interface

misk/src/main/kotlin/misk/web/actions/WebSocketListener.kt
```diff
@@ -0,0 +1,11 @@
+package misk.web.actions
+
+import okio.ByteString
+
+open class WebSocketListener {
```
mmihic: Why ``class`` and not ``interface``

exemplar/src/main/kotlin/com/squareup/exemplar/actions/ClockWebSocketAction.kt
```diff
@@ -0,0 +1,37 @@
...
+  @Inject private lateinit var executorService: ScheduledExecutorService
+
+  @ConnectWebSocket("/clock")
```
mmihic: Why does ``@ConnectWebSocket`` use annotation based dispatch but the rest of the API use a statically typed listener interface? Feels inconsistent and a little jarring to switch between the two modes of API. I'd propose either making all of websockets support be listener interface based, or use annotations for dispatching the other types of web socket events (Closing, Closed, OnMessage) in addition to the connect.

misk/src/main/kotlin/misk/web/extractors/RequestBodyParameterExtractor.kt
mmihic: I don't think that quite works - a ``WebAction`` is a target of dispatch, the ``WebSocket`` is a property of the request. But it does feel a little sad to expose web sockets to all parameter extractors, when the only one that ever will every use it is one that just returns the socket directly.  Might make more sense to make the ``WebSocket`` a property of the incoming request rather than a top-level parameter to ``extract``

misk/src/main/kotlin/misk/web/interceptors/MarshallerInterceptor.kt
mmihic: The fact that we have to hard-code implicit type-based conditionals to short-circuit the normal request routing logic when web sockets are involved makes me think we shouldn't be running web socket connects through the action routing system.

misk/src/main/kotlin/misk/web/jetty/WebActionsServlet.kt
mmihic: nit: change the signature to remove the nullability and get rid of the ``!!``

exemplar/src/main/kotlin/com/squareup/exemplar/actions/ClockWebSocketAction.kt
swankjesse: My model is that a websocket is a special interactive response. Instead of accepting a value as the request body you accept a handle to send messages. Instead of returning a value as the response body you return a handle to receive messages.

swankjesse:
I do like the idea of putting the Websocket onto the Request interface. That’s nice and makes it easier to get more interceptor features for WebSockets

PR #123, opened 2018-02-26
Add tracing interceptor for web actions

nbroyles:
Ensures that a span exists automatically for all Web Actions. Follow up PRs coming to extend functionality.

### DIFF
--- a/misk/src/main/kotlin/misk/web/interceptors/TracingInterceptor.kt
+++ b/misk/src/main/kotlin/misk/web/interceptors/TracingInterceptor.kt
@@ -1,0 +1,46 @@
+package misk.web.interceptors
+
+import com.google.inject.Inject
+import com.google.inject.Singleton
+import io.opentracing.Scope
+import io.opentracing.Tracer
+import io.opentracing.tag.Tags
+import misk.Action
+import misk.Chain
+import misk.Interceptor
+import misk.logging.getLogger
+
+private val logger = getLogger<TracingInterceptor>()
+
+/**
+ * Enables distributed tracing on all web actions, if a client has installed a tracer.
+ */
+internal class TracingInterceptor internal constructor(private val tracer: Tracer): Interceptor {
+  @Singleton
+  class Factory : Interceptor.Factory {
+    @Inject(optional=true) var tracer: Tracer? = null
+
+    override fun create(action: Action): Interceptor? {
+      // NOTE(nb): returning null ensures interceptor is filtered out when generating interceptors to
+      // apply for a specific action. See WebActionModule for implementation details
+      return if (tracer != null) TracingInterceptor(tracer!!) else null
+    }
+  }
+
+  override fun intercept(chain: Chain): Any? {
+    var result: Any? = null
+    var scope: Scope? = null
+
+    try {
+      scope = tracer.buildSpan("${chain.action}").startActive(true)
+      result = chain.proceed(chain.args)
+    } catch (exception: Exception) {
+      logger.warn("failed attempting to trace request", exception)
+      if (scope != null) Tags.ERROR.set(scope.span(), true)
+    } finally {
+      if (scope != null) scope.close()
+    }
+
+    return result
+  }
+}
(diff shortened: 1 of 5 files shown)


misk/src/main/kotlin/misk/web/interceptors/TracingInterceptor.kt
```diff
@@ -0,0 +1,46 @@
...
+ */
+internal class TracingInterceptor internal constructor(private val tracer: Tracer): Interceptor {
+  @Singleton
+  class Factory : Interceptor.Factory {
+    @Inject(optional=true) var tracer: Tracer? = null
+
+    override fun create(action: Action): Interceptor? {
+      // NOTE(nb): returning null ensures interceptor is filtered out when generating interceptors to
+      // apply for a specific action. See WebActionModule for implementation details
+      return if (tracer != null) TracingInterceptor(tracer!!) else null
```
ryanhall07: do you need the `!!` after checking null? I though KT was good about auto casting

misk/src/main/kotlin/misk/web/interceptors/TracingInterceptor.kt
```diff
@@ -0,0 +1,46 @@
...
+    var scope: Scope? = null
+
+    try {
+      scope = tracer.buildSpan("${chain.action}").startActive(true)
+      result = chain.proceed(chain.args)
+    } catch (exception: Exception) {
+      if (scope != null) Tags.ERROR.set(scope.span(), true)
+      logger.warn("failed attempting to trace request: $exception")
+    } finally {
+      if (scope != null) scope.close()
```
ryanhall07: can you auto close with a `use`? https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.io/use.html

misk/src/main/kotlin/misk/web/interceptors/TracingInterceptor.kt
```diff
@@ -0,0 +1,46 @@
...
+  override fun intercept(chain: Chain): Any? {
+    var result: Any? = null
+    var scope: Scope? = null
+
+    try {
+      scope = tracer.buildSpan("${chain.action}").startActive(true)
+      result = chain.proceed(chain.args)
+    } catch (exception: Exception) {
+      if (scope != null) Tags.ERROR.set(scope.span(), true)
+      logger.warn("failed attempting to trace request: $exception")
```
ryanhall07: use the flavor that accepts a `Throwable`

ryanhall07: I wonder if you should log this before the previous line, in case it also causes an exception

misk/src/test/kotlin/misk/web/interceptors/TracingInterceptorTest.kt
```diff
@@ -0,0 +1,65 @@
...
+    chain.proceed(chain.args) as Response<String>
+
+    val mockTracer = tracer as MockTracer
+    assertThat(mockTracer.finishedSpans().size).isEqualTo(1)
+  }
+
+  @Test
+  fun nothingIfNotInstalled() {
+    val module = NoopModule()
+    val injector = Guice.createInjector(module)
```
ryanhall07: this is a varag so you shouldn't have to pass anything. `Guice.createInjector()`

misk/src/main/kotlin/misk/web/interceptors/TracingInterceptor.kt
nbroyles: Smart cast doesn't work since tracer is a mutable property.

misk/src/main/kotlin/misk/web/interceptors/TracingInterceptor.kt
nbroyles: You could, but, unfortunately, you can't use `use` like `try-with-resources` where you can have catch block. So you wouldn't have access to the object created by `use` if you wanted to log anything from it in an exception that wraps it.

misk/src/test/kotlin/misk/web/interceptors/TracingInterceptorTest.kt
nbroyles: Ahh, clever.

PR #129, opened 2018-02-28
Add @Trace annotation to opt method into tracing

nbroyles:


misk/src/main/kotlin/misk/tracing/interceptors/TracingMethodInterceptor.kt
```diff
@@ -0,0 +1,34 @@
...
+  override fun invoke(invocation: MethodInvocation): Any? {
+    var result: Any? = null
+    var scope: Scope? = null
+    val tracer = tracerProvider.get()
+
+    try {
+      scope = tracer.buildSpan("${invocation.javaClass.simpleName}.${invocation.method.name}").startActive(true)
+      result = invocation.proceed()
+    } catch (exception: Exception) {
+      logger.warn("failed attempting to trace method", exception)
```
ryanhall07: do we really want this logged every time a traced method throws an exception? seems like a lot of noise and the log message reads like there is a problem with the tracing, not the thing being traced

misk/src/main/kotlin/misk/web/interceptors/TracingInterceptor.kt
```diff
@@ -18,12 +19,12 @@ private val logger = getLogger<TracingInterceptor>()
...
   @Singleton
   class Factory : Interceptor.Factory {
-    @Inject(optional=true) var tracer: Tracer? = null
+    @Inject lateinit var tracer: Tracer
 
     override fun create(action: Action): Interceptor? {
       // NOTE(nb): returning null ensures interceptor is filtered out when generating interceptors to
       // apply for a specific action. See WebActionModule for implementation details
-      return if (tracer != null) TracingInterceptor(tracer!!) else null
+      return if (tracer is NoopTracer) null else TracingInterceptor(tracer)
```
ryanhall07: why the special handling? shouldn't you be able to dispatch to the `NoopTracer` and it do nothing?

misk/src/test/kotlin/misk/tracing/interceptors/TracingMethodInterceptorTest.kt
```diff
@@ -0,0 +1,42 @@
...
+    assertThat(mockTracer.finishedSpans().size).isEqualTo(0)
+    testObject.testTraceMethod()
+    assertThat(mockTracer.finishedSpans().size).isEqualTo(1)
+  }
+}
+
+open class TestObject {
+  @Trace
+  open fun testTraceMethod() {
+    print("Trace me!")
```
ryanhall07: nit: remove?

misk/src/test/kotlin/misk/tracing/interceptors/TracingMethodInterceptorTest.kt
```diff
@@ -0,0 +1,42 @@
...
+
+    assertThat(mockTracer.finishedSpans().size).isEqualTo(0)
+    testObject.testTraceMethod()
+    assertThat(mockTracer.finishedSpans().size).isEqualTo(1)
+  }
+}
+
+open class TestObject {
+  @Trace
+  open fun testTraceMethod() {
```
ryanhall07: so what did you need to do to get this working?

misk/src/main/kotlin/misk/tracing/Trace.kt
```diff
@@ -0,0 +1,11 @@
+package misk.tracing
+
+/**
+ * Annotate a method with @Trace to enable tracing for the specified method.
+ *
+ * NOTE: This annotation is implemented using Guice's method interception capabilities.
+ * As such, methods that can be traced using this annotation are subject to the same limitations
+ * outlined in the <a href="https://github.com/google/guice/wiki/AOP">Guice wiki</a>.
```
ryanhall07: it sounds like there is some Kotlin issues though, like requiring `open`, since final is the default. it would be nice to document those too

misk/testing/src/main/kotlin/misk/testing/MockTracingBackendModule.kt
```diff
@@ -0,0 +1,11 @@
+package misk.testing
+
+import io.opentracing.Tracer
+import io.opentracing.mock.MockTracer
+import misk.inject.KAbstractModule
+
+class MockTracingBackendModule : KAbstractModule() {
+  override fun configure() {
+    bind(Tracer::class.java).to(MockTracer::class.java).asEagerSingleton()
```
ryanhall07: are we going with the term `Fake` or `Mock`? seems like `Fake` is used the most

misk/src/main/kotlin/misk/tracing/interceptors/TracingMethodInterceptor.kt
```diff
@@ -0,0 +1,34 @@
...
+import io.opentracing.tag.Tags
+import misk.logging.getLogger
+import org.aopalliance.intercept.MethodInterceptor
+import org.aopalliance.intercept.MethodInvocation
+import javax.inject.Inject
+import javax.inject.Provider
+
+private val logger = getLogger<TracingMethodInterceptor>()
+
+class TracingMethodInterceptor @Inject internal constructor(
```
ryanhall07: nit: should we have a single `TracingInterceptor` that implements both `MethodInterceptor` and `Interceptor`? might be able to DRY up some of the boilerplate, but not sure

misk/src/test/kotlin/misk/tracing/interceptors/TracingMethodInterceptorTest.kt
nbroyles: Actually mark the function open in addition to the class 🤦‍♂️

misk/src/main/kotlin/misk/tracing/Trace.kt
nbroyles: Good catch. Meant to document those, but forgot

misk/testing/src/main/kotlin/misk/testing/MockTracingBackendModule.kt
nbroyles: Meh, was just keeping the module name consistent with the `MockTracer` provided by the opentracing library. We can just rename the module in the future if it becomes confusing.

misk/src/main/kotlin/misk/tracing/interceptors/TracingMethodInterceptor.kt
nbroyles: Yeah, that's a good point. I think this is wrong in general. At a minimum we should be re-throwing the exception so that the tracer doesn't swallow it. Think I'm going to kill this catch altogether.

misk/src/main/kotlin/misk/tracing/interceptors/TracingMethodInterceptor.kt
nbroyles: Yeah, will give that a go and see how it looks.

misk/testing/src/main/kotlin/misk/testing/MockTracingBackendModule.kt
eeSeeGee: I believe conceptually a Fake is something that implements an interface in-memory and serves a variety of purposes, while a Mock implements some fraction of the API and has canned responses to specific inputs.

nbroyles: Agreed, and MockTracer functions very much like a Mock in that it does nothing except track calls to methods.

PR #130, opened 2018-02-28
Add MiskTracer to facilitate ad-hoc method tracing

nbroyles:


misk/src/main/kotlin/misk/tracing/MiskTracer.kt
```diff
@@ -0,0 +1,19 @@
...
+  private val tracer: Tracer
+) {
+  /**
+   * Adds span/sub-span to trace for method provided as a parameter. Use this method to do ad-hoc
+   * tracing on methods you're interested in.
+   *
+   * @param operationName name to be used for the span. Usually, should be a string representation
+   * of the method you plan to trace
+   */
+  fun trace(operationName: String, method: () -> Any?) {
```
mmihic: This should return the results from the method, and the signature should have a generic type bounds. e.g.

````kotlin
fun <R> trace(operationName: String, method: () -> R) : R = 
   tracer.buildSpan(operationName).startActive(true).use { method() }
````

File: misk/src/main/kotlin/misk/tracing/MiskTracer.kt [hunk 2]
```diff
@@ -0,0 +1,19 @@
...
+import javax.inject.Inject
+
+class MiskTracer @Inject internal constructor(
+  private val tracer: Tracer
+) {
+  /**
+   * Adds span/sub-span to trace for method provided as a parameter. Use this method to do ad-hoc
+   * tracing on methods you're interested in.
+   *
+   * @param operationName name to be used for the span. Usually, should be a string representation
```
mmihic: nit: call the parameter ``spanName`` to better reflect its meaning

misk/src/main/kotlin/misk/tracing/MiskTracer.kt
```diff
@@ -0,0 +1,19 @@
+package misk.tracing
+
+import io.opentracing.Tracer
+import javax.inject.Inject
+
+class MiskTracer @Inject internal constructor(
```
mmihic: nit: Don't preface classes with ``Misk`` unless absolutely necessary.

misk/src/main/kotlin/misk/tracing/MiskTracer.kt
```diff
@@ -0,0 +1,19 @@
...
+) {
+  /**
+   * Adds span/sub-span to trace for method provided as a parameter. Use this method to do ad-hoc
+   * tracing on methods you're interested in.
+   *
+   * @param operationName name to be used for the span. Usually, should be a string representation
+   * of the method you plan to trace
+   */
+  fun trace(operationName: String, method: () -> Any?) {
+    tracer.buildSpan(operationName).startActive(true).use { method() }
```
mmihic: Catch exceptions and set ERROR tag?

misk/src/main/kotlin/misk/tracing/MiskTracer.kt
nbroyles: Only doing it here because there are so many other overloaded uses of simply `Tracer`

PR #132, opened 2018-02-28
Inspect response status code to tag failed requests appropriately in trace

nbroyles:


misk/src/main/kotlin/misk/web/interceptors/TracingInterceptor.kt
```diff
@@ -28,9 +29,22 @@ internal class TracingInterceptor internal constructor(private val tracer: Trace
...
 
-    tracer.buildSpan("${chain.action}").startActive(true).use { result = chain.proceed(chain.args) }
+    try {
+      scope = tracer.buildSpan("${chain.action}").startActive(true)
+      result = chain.proceed(chain.args)
+      if (result is Response<*> && result.statusCode / 100 > 3) {
+        Tags.HTTP_STATUS.set(scope.span(), result.statusCode)
+        Tags.ERROR.set(scope.span(), true)
+      }
+    } catch (exception: Exception) {
```
ryanhall07: I think you want to make sure this is installed to run after the`ExceptionHandlingInterceptor`. that way exceptions are properly transformed into responses and you don't have to worry about catching and re-throwing exceptions. if an exception has leaked to here, it's bad and trace marking is probably not our main concern

ryanhall07: and by after, I really mean before in the chain so it runs after in response processing. closer to the transport layer

misk/src/test/kotlin/misk/web/interceptors/TracingInterceptorTest.kt
```diff
@@ -37,6 +38,22 @@ class TracingInterceptorTest {
...
+    chain.proceed(chain.args) as Response<String>
+
+    val mockTracer = tracer as MockTracer
+    assertThat(mockTracer.finishedSpans().size).isEqualTo(1)
+
+    val span = mockTracer.finishedSpans().first()
+    assertThat(span.tags().get(Tags.ERROR.key)).isEqualTo(true)
+    assertThat(span.tags().get(Tags.HTTP_STATUS.key)).isEqualTo(400)
+  }
+
```
mmihic: Add test for a custom exception that gets mapped to a non-successful response via the ``ExceptionMappingInterceptor``

misk/src/main/kotlin/misk/web/interceptors/TracingInterceptor.kt
mmihic: Correct

misk/src/main/kotlin/misk/web/interceptors/TracingInterceptor.kt
```diff
@@ -28,9 +29,22 @@ internal class TracingInterceptor internal constructor(private val tracer: Trace
...
-    var result: Any? = null
+    val result: Any?
+    var scope: Scope? = null
 
-    tracer.buildSpan("${chain.action}").startActive(true).use { result = chain.proceed(chain.args) }
+    try {
+      scope = tracer.buildSpan("${chain.action}").startActive(true)
+      result = chain.proceed(chain.args)
+      if (result is Response<*> && result.statusCode / 100 > 3) {
+        Tags.HTTP_STATUS.set(scope.span(), result.statusCode)
```
mmihic: set the ``HTTP_STATUS`` tag for non-errors as well (e.g. 2XX / 3XX) ?

misk/src/main/kotlin/misk/web/interceptors/TracingInterceptor.kt
```diff
@@ -28,9 +29,22 @@ internal class TracingInterceptor internal constructor(private val tracer: Trace
   }
 
   override fun intercept(chain: Chain): Any? {
-    var result: Any? = null
+    val result: Any?
+    var scope: Scope? = null
 
-    tracer.buildSpan("${chain.action}").startActive(true).use { result = chain.proceed(chain.args) }
+    try {
```
mmihic: should be able to condense this down to

````kotlin
val scope = tracer.buildSpan("${chain.action}").startActive(true)  
return scope.use { 
  try {
    val result = chain.proceed(chain.args)
    if (result is Response<*> && result.statusCode > 399) {
      Tags.HTTP_STATUS.set(scope.span(), result.statusCode)
      Tags.ERROR.set(scope.span(), true)
    }
    result
  } catch (e: Exception) {
    Tags.ERROR.set(scope.span(), true)
    throw e
  }
}
````

File: misk/src/main/kotlin/misk/web/interceptors/TracingInterceptor.kt (hunk 3 above)
Comment (nbroyles) [reply misk/src/main/kotlin/misk/web/interceptors/TracingInterceptor.kt]: No good reason not to, to be honest. I'll add it and if it gets annoying we can always just remove it.

File: misk/src/main/kotlin/misk/web/interceptors/TracingInterceptor.kt (hunk 1 above)
Comment (nbroyles) [reply misk/src/main/kotlin/misk/web/interceptors/TracingInterceptor.kt]: Ahh, that's my bad. I made that change in the `WebModule` but didn't add it to the PR. (@mmihic pointed out the same thing in the github issue).

File: misk/src/main/kotlin/misk/web/interceptors/TracingInterceptor.kt [hunk 5]
```diff
@@ -28,9 +29,22 @@ internal class TracingInterceptor internal constructor(private val tracer: Trace
...
+      scope = tracer.buildSpan("${chain.action}").startActive(true)
+      result = chain.proceed(chain.args)
+      if (result is Response<*> && result.statusCode / 100 > 3) {
```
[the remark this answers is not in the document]
nbroyles: Heh 🤦‍♂️

misk/src/main/kotlin/misk/web/WebModule.kt
```diff
@@ -69,6 +69,10 @@ class WebModule : KAbstractModule() {
     binder().addMultibinderBindingWithAnnotation<NetworkInterceptor.Factory, MiskDefault>()
         .to<MetricsInterceptor.Factory>()
 
+    // Traces requests as they work their way through the system.
```
ryanhall07: maybe comment what dependency this has in the binding ordering. e.g this must be bound before the `ExceptionHandlingInterceptor` because ...

misk/src/main/kotlin/misk/web/interceptors/TracingInterceptor.kt
```diff
@@ -28,9 +29,19 @@ internal class TracingInterceptor internal constructor(private val tracer: Trace
...
   override fun intercept(chain: NetworkChain): Response<*> {
-    var result: Response<*>? = null
-    tracer.buildSpan("${chain.action}")
-        .startActive(true).use { result = chain.proceed(chain.request) }
-    return result?: throw IllegalStateException("Null result in TracingInterceptor")
+    val scope = tracer.buildSpan("${chain.action}").startActive(true)
+    return scope.use {
+      try {
+        val result = chain.proceed(chain.request)
+        Tags.HTTP_STATUS.set(scope.span(), result.statusCode)
```
ryanhall07: nice!

misk/src/main/kotlin/misk/web/interceptors/TracingInterceptor.kt
```diff
@@ -28,9 +29,19 @@ internal class TracingInterceptor internal constructor(private val tracer: Trace
...
+        if (result.statusCode > 399) {
+          Tags.ERROR.set(scope.span(), true)
+        }
+        result
+      } catch (exception: Exception) {
+        Tags.ERROR.set(scope.span(), true)
+        throw exception
```
ryanhall07: this makes a lot more sense. thanks for iterating!

misk/src/test/kotlin/misk/web/interceptors/TracingInterceptorTest.kt
```diff
@@ -1,51 +1,91 @@
...
+  @Test
+  fun failedTraceWithException() {
+    get("/exception_trace")
 
     val mockTracer = tracer as MockTracer
     assertThat(mockTracer.finishedSpans().size).isEqualTo(1)
+
+    val span = mockTracer.finishedSpans().first()
+    assertThat(span.tags().get(Tags.ERROR.key)).isEqualTo(true)
+    assertThat(span.tags().get(Tags.HTTP_STATUS.key)).isEqualTo(420)
```
ryanhall07: lol

misk/src/test/kotlin/misk/web/interceptors/TracingInterceptorTest.kt
```diff
@@ -58,14 +98,45 @@ class TracingInterceptorTest {
...
   }
 
+  internal class FailedTracingTestAction : WebAction {
+    @Get("/failed_trace")
+    fun call(): Response<String> {
+      return Response("no good", statusCode = StatusCode.BAD_REQUEST.code)
+    }
+  }
+
+  internal class ExceptionThrowingTracingTestAction : WebAction {
```
ryanhall07: you still need a test case for a custom mapped exception

misk/src/test/kotlin/misk/web/interceptors/TracingInterceptorTest.kt
```diff
@@ -58,14 +98,45 @@ class TracingInterceptorTest {
...
+    @Get("/exception_trace")
+    fun call(): Response<String> {
+      throw ActionException(StatusCode.ENHANCE_YOUR_CALM, "Chill, man")
```
ryanhall07: in the voice of jim breuer

misk/src/test/kotlin/misk/web/interceptors/TracingInterceptorTest.kt
nbroyles: Wait, what do you mean by custom mapped exception?  Some exception I've created that extends `ActionException`?

Regardless of how it's defined, I'm not sure how it matters. Converting of any exception into a response w/ >= 400 status code is the domain of the `ExceptionHandlerInterceptor`. This test demonstrates that if the `ExceptionHandlerInterceptor` does it's job properly, the `TracingInterceptor` will act accordingly (i.e. tag the span with an error tag).

mmihic: Yeah this is good enough

ryanhall07: ah good point. yea the `ActionException` still invokes the `ExceptionInterceptor`

Issue #121, opened 2018-02-26
Separate network and application interceptors


swankjesse:
I wanna split the interceptors into two halves:

1. the “application interceptors” that get application-layer objects
2. the “network interceptors” that get network-layer objects

In the middle is our bridge. So users can decide whether they want a Response or a MyResponseObject.

swankjesse:
(stealing from [OkHttp’s interceptors](https://github.com/square/okhttp/wiki/Interceptors)).

eeSeeGee:
Looking at OkHttp, does it make sense to defer extracting parameters in `BoundAction.handle` and instead pass the `Request` object down the network interceptor chain until it terminates with a Misk-internal Interceptor, which does the parameter extraction and basically fulfills the existing chain with the application interceptors. On the way back out that Interceptor would convert a `T` into a `Response<T>` and the network interceptors would process a bunch of `Response` objects.

One thing that's interesting is what if an Application Interceptor decides to return a `Response`? Short circuit back to Network? Disallow?

swankjesse:
Yes, exactly.

I think `Response` is just a really low-level “application-layer” type.

mmihic:
I think that's fine as long as network layer interceptors have access to information about the target action. Otherwise interceptors like metrics and tracing won't be able to do their job.

eeSeeGee:
Yup, definitely, that will propagate.

eeSeeGee:
Implemented in #133

PR #133, opened 2018-03-01
Split Interceptor into NetworkInterceptor and ApplicationInterceptor

eeSeeGee:
This needs rebasing to account for the changing to Tracing interceptor today, but the meat is here.

Addresses #121


misk/src/main/kotlin/misk/web/BoundAction.kt
```diff
@@ -78,18 +81,20 @@ internal class BoundAction<A : WebAction, out R>(
...
+  internal fun handle(request: Request, pathMatcher: Matcher): Response<ResponseBody> {
     // Find values for all the parameters.
     val webAction = webActionProvider.get()
-    val parameters = parameterExtractors.map {
-      it.extract(webAction, request, pathMather)
-    }
 
-    val chain = webAction.asChain(function, parameters, *interceptors.toTypedArray())
+    // RequestBridgeInterceptor necessarily needs to be the last NetworkInterceptor run.
+    val interceptors = Lists.newArrayList(networkInterceptors)
```
mmihic: nit: prefer kotlin initializers (e.g. ``mutableListOf(*networkInterceptors)``

misk/src/main/kotlin/misk/web/RealNetworkChain.kt
```diff
@@ -0,0 +1,35 @@
...
+
+  override val request: Request
+    get() = _request
+
+  override val function: KFunction<*>
+    get() = _function
+
+  override fun proceed(request: Request): Response<*>? {
+    check(index < interceptors.size) { "final interceptor must be terminal" }
+    // NB(young): At this point it's possible the Request has been manipulated by Network
```
mmihic: I don't think we want to allow network interceptors to mutate ``Request`` properties

misk/src/main/kotlin/misk/web/actions/WebActions.kt
```diff
@@ -2,18 +2,29 @@ package misk.web.actions
...
+import misk.ApplicationInterceptor
+import misk.NetworkChain
+import misk.NetworkInterceptor
 import misk.web.RealChain
+import misk.web.RealNetworkChain
+import misk.web.Request
 import kotlin.reflect.KFunction
 import kotlin.reflect.KParameter
 
+fun WebAction.asNetworkChain(
```
mmihic: Is this used? In ``BoundActionMatch.handle`` we instantiate the ``RealNetworkChain`` directly

misk/src/main/kotlin/misk/NetworkInterceptor.kt
```diff
@@ -0,0 +1,11 @@
+package misk
+
+import misk.web.Response
+
+interface NetworkInterceptor {
+  fun intercept(chain: NetworkChain): Response<*>?
```
mmihic: Similar to above (but more important here) - we should require network interceptors to either return a valid response or throw an exception. I could sortof be talked into allowing null as a valid response from an action or application interceptor, but not from network interceptors.

misk/src/main/kotlin/misk/ApplicationInterceptor.kt
```diff
@@ -0,0 +1,9 @@
+package misk
+
+interface ApplicationInterceptor {
+  fun intercept(chain: Chain): Any?
```
mmihic: I think we want to require interceptors / actions to return a non-null response (which could be ``Nothing``) or throw an exception

misk/src/main/kotlin/misk/web/RealNetworkChain.kt
swankjesse: I’ve found it to be occasionally profitable. In particular, a network interceptor could gunzip an incoming request body.

misk/src/main/kotlin/misk/NetworkInterceptor.kt
swankjesse: +1

misk/src/main/kotlin/misk/ApplicationInterceptor.kt
swankjesse: +1

misk/src/main/kotlin/misk/web/BoundAction.kt
```diff
@@ -148,3 +153,28 @@ internal class RequestMatch(
 
 private fun MediaType.closestMediaRangeMatch(ranges: List<MediaRange>) =
     ranges.mapNotNull { it.matcher(this) }.sorted().firstOrNull()
+
+/** Acts as the bridge between network interceptors and application interceptors.
```
swankjesse: supernit: we like the `/**` to be on a line of its own, and only once space after the `*` on each line beneath

swankjesse: (or perhaps, _I_ like!)

misk/src/main/kotlin/misk/web/RealNetworkChain.kt
swankjesse: though I do think we should not re-route; that’s not good. If we were superparanoid we could check to see if the route is still valid and crash with an IllegalStateException if it doesn’t

misk/src/test/kotlin/misk/web/interceptors/MetricsInterceptorTest.kt
```diff
@@ -54,18 +63,28 @@ class MetricsInterceptorTest {
   }
 
   fun invoke(desiredStatusCode: Int): Response<String> {
+    val request = Request(
+        HttpUrl.parse("http://foo.bar/")!!,
```
swankjesse: Off topic: pretty much everyone in the world hates me because parse() can return null

misk/src/test/kotlin/misk/web/interceptors/TracingInterceptorTest.kt
```diff
@@ -40,25 +53,30 @@ class TracingInterceptorTest {
...
-  override fun configure() {}
-}
+  class MockTracingBackendModule : KAbstractModule() {
+    override fun configure() {
+      bind(Tracer::class.java).to(MockTracer::class.java).asEagerSingleton()
+    }
+  }
+
+  class NoopModule : KAbstractModule() {
+    override fun configure() {}
```
swankjesse: supernit: I think you want `= Unit`. I don’t like `{}` because it’s an extra special case in our style rules

misk/src/test/kotlin/misk/web/interceptors/UserInterceptorTest.kt
```diff
@@ -52,15 +54,50 @@ class UserInterceptorTest {
     assertThat(response.body()!!.source()!!.readUtf8()).isEqualTo("text")
   }
 
-  internal class UserCreatedInterceptor: Interceptor {
+  @Test
+  fun throwResponse() {
+    val response = get("/call/throw")
+    assertThat(response.code()).isEqualTo(500)
+    assertThat(response.body()!!.source()!!.readUtf8()).isEqualTo("internal server error")
```
swankjesse: The source shouldn’t be nullable. It might be, but it shouldn’t be

swankjesse: Also `response.body()!!.string()` is a shortcut for `response.body()!!.source()!!.readUtf8()`

misk/src/test/kotlin/misk/web/interceptors/UserInterceptorTest.kt
```diff
@@ -52,15 +54,50 @@ class UserInterceptorTest {
...
+  @Test
+  fun throwNetworkResponse() {
+    val response = get("/call/textResponse", "throw")
+    assertThat(response.code()).isEqualTo(500)
+    assertThat(response.body()!!.source()!!.readUtf8()).isEqualTo("internal server error")
+  }
+
+  internal class UserCreatedNetworkInterceptor: NetworkInterceptor {
+    override fun intercept(chain: NetworkChain): Response<*>? = when (chain.request.headers.get("mode")) {
+      "text" -> Response("net text response", UserInterceptorTest.TEXT_HEADERS, 410)
```
swankjesse: neat

misk/src/main/kotlin/misk/web/RealNetworkChain.kt
eeSeeGee: I was thinking it could potentially be useful if you're behind a proxy and need to move headers around, eg `X-Real-IP` or something.  I don't think we'll need to use it anytime soon anyway.

misk/src/main/kotlin/misk/web/actions/WebActions.kt
eeSeeGee: Whoops, good catch. This is the second half of a refactor.

misk/src/main/kotlin/misk/NetworkInterceptor.kt
eeSeeGee: Yeah, a stronger contract definitely makes sense here.

misk/src/test/kotlin/misk/web/interceptors/MetricsInterceptorTest.kt
eeSeeGee: Not really that big a deal in Kotlin, although if you don't know the language I guess you could think I REALLY want it parsed.

misk/src/test/kotlin/misk/web/interceptors/UserInterceptorTest.kt
mmihic: IMO cleaner to do the final ``!!``just a the end

``response.body()?.source()?.readUtf8()!!``

PR #134, opened 2018-03-01
Fix NotFoundAction handling

mmihic:
Previously POSTing to an unknown path would result in a 200 with an
empty response body, instead of the expected 404. Multiple problems:

1. The NotFoundAction was only bound for GET requests
2. The NotFoundAction only supported requests with Accept: text/plain

The NotFoundAction is now bound for both GET and POSTs, and supports
requests for any type of content (but always returns text/plain). Also
added a backstop in the WebActionsServlet to return a 404 if no suitable
action could be found; this handles cases where applications do not
install the NotFoundAction, or when a request arrives with an
unsupported HTTP method.

### DIFF
--- a/misk/src/main/kotlin/misk/web/WebActionModule.kt
+++ b/misk/src/main/kotlin/misk/web/WebActionModule.kt
@@ -20,24 +20,78 @@ import kotlin.reflect.KFunction
 import kotlin.reflect.full.findAnnotation
 
 class WebActionModule<A : WebAction> private constructor(
-  val webActionClass: KClass<A>,
-  val member: KFunction<*>,
-  val pathPattern: String,
-  val httpMethod: HttpMethod,
-  val acceptedContentTypes: List<MediaRange>,
-  val responseContentType: MediaType?
+  val webActionClass: KClass<A>
 ) : AbstractModule() {
   override fun configure() {
-    val provider = getProvider(webActionClass.java)
     @Suppress("UNCHECKED_CAST")
     val binder: Multibinder<BoundAction<A, *>> = Multibinder.newSetBinder(
         binder(),
         parameterizedType<BoundAction<*, *>>(subtypeOf<WebAction>(),
             subtypeOf<Any>()).typeLiteral()
     ) as Multibinder<BoundAction<A, *>>
-    binder.addBinding().toProvider(
-        BoundActionProvider(provider, member, pathPattern, httpMethod,
-            acceptedContentTypes, responseContentType))
+
+    // Find the function with Get or Post annotations. Only one such function is
+    // allow, but may have both Get and Post annotations if the action can handle
+    // both forms of HTTP method
+    val actionFunctions = webActionClass.members.mapNotNull {
+      if (it.findAnnotation<Get>() != null ||
+          it.findAnnotation<Post>() != null ||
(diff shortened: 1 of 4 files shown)


mmihic:
This was previously approved, but I think got lost in the reformatting

misk/src/main/kotlin/misk/web/WebActionModule.kt
```diff
@@ -20,24 +20,78 @@ import kotlin.reflect.KFunction
...
         binder(),
         parameterizedType<BoundAction<*, *>>(subtypeOf<WebAction>(),
             subtypeOf<Any>()).typeLiteral()
     ) as Multibinder<BoundAction<A, *>>
-    binder.addBinding().toProvider(
-        BoundActionProvider(provider, member, pathPattern, httpMethod,
-            acceptedContentTypes, responseContentType))
+
+    // Find the function with Get or Post annotations. Only one such function is
+    // allow, but may have both Get and Post annotations if the action can handle
```
swankjesse: nit: allowed ?

misk/src/main/kotlin/misk/web/WebActionModule.kt
```diff
@@ -20,24 +20,78 @@ import kotlin.reflect.KFunction
...
+
+    // Bind providers for each supported HTTP method
+    val actionFunction = actionFunctions.first()
+    val get = actionFunction.findAnnotation<Get>()
+    if (get != null) {
+      val getProvider = buildProvider(actionFunction, HttpMethod.GET, get.pathPattern)
+      binder.addBinding().toProvider(getProvider)
+    }
+
+    // Only one of ConnectWebSocket and Get may be specified
```
swankjesse: cool!

misk/src/main/kotlin/misk/web/WebActionModule.kt
```diff
@@ -20,24 +20,78 @@ import kotlin.reflect.KFunction
...
+    if (post != null) {
+      val postProvider = buildProvider(actionFunction, HttpMethod.POST, post.pathPattern)
+      binder.addBinding().toProvider(postProvider)
+    }
+  }
+
+  private fun buildProvider(function: KFunction<*>, httpMethod: HttpMethod, pathPattern: String):
+      BoundActionProvider<A, *> {
+    // NB(mmihic): The response media type may be ommitted; in this case only
+    // generic return types (String, ByteString, ResponseBody, etc) are supported
```
swankjesse: . . . or the marshaller must infer it?

misk/src/main/kotlin/misk/web/actions/NotFoundAction.kt
```diff
@@ -2,18 +2,24 @@ package misk.web.actions
...
 import misk.web.ResponseContentType
 import misk.web.mediatype.MediaTypes
+import okhttp3.Headers
 import javax.inject.Singleton
 
 @Singleton
 class NotFoundAction : WebAction {
   @Get("/{path:.*}")
-  @ResponseContentType(MediaTypes.TEXT_PLAIN_UTF8)
+  @Post("/{path:.*}")
```
swankjesse: nice

swankjesse:
And sorry about the reformatting slaughter.

PR #136, opened 2018-03-01
Fix Java path param dispatching

mmihic:
Java parameter types don't map directly onto created KTypes, so Java
applications attempting to use string conversion in e.g. PathParams or
FormParams would blow up unable to find an appropriate conversion
function. Fixed and added tests.


exemplar/src/main/java/com/squareup/exemplar/package-info.java
```diff
@@ -0,0 +1,4 @@
+@ParametersAreNonnullByDefault
+package com.squareup.exemplar;
+
+import javax.annotation.ParametersAreNonnullByDefault;
```
swankjesse: nice

misk/src/test/java/misk/web/extractors/JavaPathParamDispatchTest.java
```diff
@@ -0,0 +1,73 @@
...
+  }
+
+  public static final class GetObjectDetails implements WebAction {
+    @Get(pathPattern = "/objects/{objectType}/{name}/{version}")
+    @ResponseContentType(MediaTypes.TEXT_PLAIN_UTF8)
+    public String getObjectDetails(
+        @PathParam("objectType") ObjectType objectType,
+        @PathParam("name") String name,
+        @PathParam("version") long version
+    ) {
```
swankjesse: the Kotlin is working on you

misk/src/test/java/misk/web/extractors/JavaPathParamDispatchTest.java
```diff
@@ -0,0 +1,73 @@
...
+import okhttp3.Request;
+import org.junit.jupiter.api.Test;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+@MiskTest(startService = true)
+public class JavaPathParamDispatchTest {
+  @MiskTestModule final Module module = new TestModule();
+
+  public enum ObjectType {
```
swankjesse: ResourceType ?

ObjectType sounds like java.lang.Object which was threw me off for a sec

misk/src/main/kotlin/misk/web/extractors/StringConverter.kt
```diff
@@ -2,35 +2,29 @@ package misk.web.extractors
...
+internal fun converterFor(type: KType): StringConverter? {
+
+  return when {
+    type.isSubtypeOf(stringType) -> { it -> it }
+    type.isSubtypeOf(intType) -> numberConversionWrapper { it.toInt() }
+    type.isSubtypeOf(longType) -> numberConversionWrapper { it.toLong() }
+    type.isSubtypeOf(doubleType) -> numberConversionWrapper { it.toDouble() }
+    type.isSubtypeOf(booleanType) -> numberConversionWrapper { it.toBoolean() }
+    else -> createFromValueOf(type)
+  }
```
eeSeeGee: nice

PR #139, opened 2018-03-02
Auto-embed trace information into TypedHttpClient request headers

nbroyles:
Additionally, make TracingInterceptor aware of potential parent
span information in headers and use it to set parent context, if it's
available.

### DIFF
--- a/misk/src/main/kotlin/misk/client/ClientInvocationHandler.kt
+++ b/misk/src/main/kotlin/misk/client/ClientInvocationHandler.kt
@@ -1,6 +1,8 @@
 package misk.client
 
 import com.google.common.util.concurrent.SettableFuture
+import io.opentracing.Tracer
+import io.opentracing.contrib.okhttp3.TracingCallFactory
 import okhttp3.Interceptor
 import okhttp3.OkHttpClient
 import okhttp3.Request
@@ -28,7 +30,8 @@ internal class ClientInvocationHandler(
   retrofit: Retrofit,
   okHttpTemplate: OkHttpClient,
   networkInterceptorFactories: List<ClientNetworkInterceptor.Factory>,
-  applicationInterceptorFactories: List<ClientApplicationInterceptor.Factory>
+  applicationInterceptorFactories: List<ClientApplicationInterceptor.Factory>,
+  tracer: Tracer?
 ) : InvocationHandler {
 
   private val actionsByMethod = interfaceType.functions
@@ -56,8 +59,12 @@ internal class ClientInvocationHandler(
           clientBuilder.build()
         }
 
-    methodName to retrofit.newBuilder()
+    val retrofitBuilder = retrofit.newBuilder()
         .client(actionSpecificClient)
+
+    if (tracer != null) retrofitBuilder.callFactory(TracingCallFactory(actionSpecificClient, tracer))
+
+    methodName to retrofitBuilder
         .build()
         .create(interfaceType.java)
   }.toMap()
--- a/misk/src/main/kotlin/misk/client/TypedHttpClientModule.kt
+++ b/misk/src/main/kotlin/misk/client/TypedHttpClientModule.kt
@@ -1,15 +1,16 @@
 package misk.client
 
+import com.google.inject.Inject
 import com.google.inject.Key
 import com.google.inject.Provider
 import com.squareup.moshi.Moshi
+import io.opentracing.Tracer
 import misk.inject.KAbstractModule
 import misk.inject.newMultibinder
 import okhttp3.OkHttpClient
 import retrofit2.Retrofit
 import retrofit2.converter.moshi.MoshiConverterFactory
 import java.lang.reflect.Proxy
-import javax.inject.Inject
 import kotlin.reflect.KClass
 
 /** Creates a retrofit-backed typed client given an API interface and an HTTP configuration */
@@ -61,6 +62,9 @@ class TypedHttpClientModule<T : Any>(
     @Inject
     private lateinit var moshi: Moshi
 
+    @Inject(optional=true)
+    private val tracer: Tracer? = null
+
     override fun get(): T {
       val okhttp = httpClientProvider.get()
 
@@ -75,7 +79,8 @@ class TypedHttpClientModule<T : Any>(
           retrofit,
           okhttp,
           clientNetworkInterceptorFactories,
-          clientApplicationInterceptorFactories)
+          clientApplicationInterceptorFactories,
+          tracer)
 
       @Suppress("UNCHECKED_CAST")
       return Proxy.newProxyInstance(
--- a/misk/src/main/kotlin/misk/tracing/interceptors/TextMultimapExtractAdapter.kt
+++ b/misk/src/main/kotlin/misk/tracing/interceptors/TextMultimapExtractAdapter.kt
@@ -1,0 +1,31 @@
+package misk.tracing.interceptors
+
+import io.opentracing.propagation.TextMap
+
+/**
+ * Provides an interface for Tracer.extract() to read from a multimap datasource
+ * (i.e. HTTP headers provided via OkHttp)
+ */
+internal class TextMultimapExtractAdapter(
+  private val multimap: Map<String, List<String>>
+) : TextMap {
+  override fun put(key: String, value: String) {
+    throw UnsupportedOperationException("Cannot put into readonly data source")
+  }
+
+  override fun iterator(): MutableIterator<MutableMap.MutableEntry<String, String>> {
+    return multimap.flatMap {
+      it.value.map { mapValue -> ExtractEntry(it.key, mapValue) }
+    }.toMutableList().iterator()
+  }
+
+  private class ExtractEntry(
+    override val key: String,
+    override var value: String
+  ) : MutableMap.MutableEntry<String, String> {
+    override fun setValue(newValue: String): String {
+      value = newValue
+      return value
+    }
+  }
+}
--- a/misk/src/main/kotlin/misk/web/interceptors/TracingInterceptor.kt
+++ b/misk/src/main/kotlin/misk/web/interceptors/TracingInterceptor.kt
@@ -2,12 +2,15 @@ package misk.web.interceptors
 
 import com.google.inject.Inject
 import com.google.inject.Singleton
+import io.opentracing.SpanContext
 import io.opentracing.Tracer
+import io.opentracing.propagation.Format
 import io.opentracing.tag.Tags
 import misk.Action
 import misk.NetworkChain
 import misk.NetworkInterceptor
 import misk.logging.getLogger
+import misk.tracing.interceptors.TextMultimapExtractAdapter
 import misk.web.Response
 
 private val logger = getLogger<TracingInterceptor>()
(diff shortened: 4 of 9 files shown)


misk/src/main/kotlin/misk/tracing/interceptors/TextMultimapExtractAdapter.kt
```diff
@@ -0,0 +1,33 @@
...
+  private val multimap: Map<String, List<String>>
+) : TextMap {
+  override fun put(key: String?, value: String?) {
+    throw UnsupportedOperationException("Cannot put into readonly data source")
+  }
+
+  override fun iterator(): MutableIterator<MutableMap.MutableEntry<String, String>> {
+    return multimap.map {
+      val values = it.value
+      val mapEntries = values.map { mapValue -> ExtractEntry(it.key, mapValue) }
```
mmihic: nit, don't need the variable declaration

misk/src/main/kotlin/misk/tracing/interceptors/TextMultimapExtractAdapter.kt
```diff
@@ -0,0 +1,33 @@
...
+ */
+class TextMultimapExtractAdapter(
+  private val multimap: Map<String, List<String>>
+) : TextMap {
+  override fun put(key: String?, value: String?) {
+    throw UnsupportedOperationException("Cannot put into readonly data source")
+  }
+
+  override fun iterator(): MutableIterator<MutableMap.MutableEntry<String, String>> {
+    return multimap.map {
```
mmihic: nit: use `flatMap` rather than `flatten`

misk/src/test/kotlin/misk/client/TypedHttpClientTest.kt
```diff
@@ -42,22 +48,64 @@ internal class TypedHttpClientTest {
   @Inject
   private lateinit var jetty: JettyService
 
-  private lateinit var client: ReturnADinosaur
+  @Inject
+  private lateinit var tracer: Tracer
```
mmihic: Let's keep test cases focused - move trace testing to another dedicate test.

misk/src/main/kotlin/misk/web/interceptors/TracingInterceptor.kt
```diff
@@ -29,7 +32,21 @@ internal class TracingInterceptor internal constructor(private val tracer: Trace
   }
 
   override fun intercept(chain: NetworkChain): Response<*> {
-    val scope = tracer.buildSpan("${chain.action}").startActive(true)
+    var parentContext: SpanContext? = null
```
mmihic: can make this a val by doing:

````kotlin
val parentContext = try {
   tracer.extract(Format.Builtin.HTTP_HEADERS, TextMultimapExtractAdapter(chain.request.headers.toMultimap()))
} catch (e: Exception) {
   logger.info("blah blah")
   null
}

````

File: misk/src/main/kotlin/misk/web/interceptors/TracingInterceptor.kt [hunk 5]
```diff
@@ -29,7 +32,21 @@ internal class TracingInterceptor internal constructor(private val tracer: Trace
...
+    try {
+      parentContext = tracer.extract(Format.Builtin.HTTP_HEADERS,
+          TextMultimapExtractAdapter(chain.request.headers.toMultimap()))
+    } catch (exception: Exception) {
```
mmihic: nit: Use ``e`` for exception name

misk/src/main/kotlin/misk/web/interceptors/TracingInterceptor.kt
```diff
@@ -29,7 +32,21 @@ internal class TracingInterceptor internal constructor(private val tracer: Trace
...
+      logger.info("Failure attempting to extract span context. Existing context, if any," +
```
mmihic: What can cause this failure? Should this be at info level logging?

misk/src/main/kotlin/misk/tracing/interceptors/TextMultimapExtractAdapter.kt
```diff
@@ -0,0 +1,33 @@
...
+import io.opentracing.propagation.TextMap
+
+/**
+ * Provides an interface for Tracer.extract() to read from a multimap datasource
+ * (i.e. HTTP headers provided via OkHttp)
+ */
+class TextMultimapExtractAdapter(
+  private val multimap: Map<String, List<String>>
+) : TextMap {
+  override fun put(key: String?, value: String?) {
```
mmihic: nit: these are technically not nullable, are they?

misk/src/main/kotlin/misk/tracing/interceptors/TextMultimapExtractAdapter.kt
```diff
@@ -0,0 +1,33 @@
...
+
+  override fun iterator(): MutableIterator<MutableMap.MutableEntry<String, String>> {
+    return multimap.map {
+      val values = it.value
+      val mapEntries = values.map { mapValue -> ExtractEntry(it.key, mapValue) }
+      mapEntries
+    }.flatten().toMutableList().iterator()
+  }
+
+  internal class ExtractEntry(
```
mmihic: Should this be ``internal``, or ``private``?

misk/src/main/kotlin/misk/tracing/interceptors/TextMultimapExtractAdapter.kt
```diff
@@ -0,0 +1,33 @@
+package misk.tracing.interceptors
+
+import io.opentracing.propagation.TextMap
+
+/**
+ * Provides an interface for Tracer.extract() to read from a multimap datasource
+ * (i.e. HTTP headers provided via OkHttp)
+ */
+class TextMultimapExtractAdapter(
```
mmihic: Does this need to be ``public`` or can it be ``internal``?

misk/src/test/kotlin/misk/client/TypedHttpClientTest.kt
```diff
@@ -42,22 +48,64 @@ internal class TypedHttpClientTest {
...
-    val response = client.getDinosaur(Dinosaur.Builder().name("trex").build()).execute()
+    val client = clientInjector.getInstance(ReturnADinosaur::class.java)
+    val response = client.getDinosaur(dinosaurRequest).execute()
     assertThat(response.code()).isEqualTo(200)
     assertThat(response.body()).isNotNull()
     assertThat(response.body()?.name!!).isEqualTo("supertrex")
   }
 
+  @Test
+  fun embedsTrace() {
```
mmihic: We should also have a test that does multiple hops (from client to server to server again), to ensure that parent span ids propagate.

misk/src/main/kotlin/misk/web/interceptors/TracingInterceptor.kt
nbroyles: Corrupted trace data (e.g. mangled header) can cause a failure. Thinking about it, we probably want to bump this to warn. Should be rare enough and people probably want to know.

misk/src/main/kotlin/misk/tracing/interceptors/TextMultimapExtractAdapter.kt
nbroyles: Good catch. Not sure what happened here

misk/src/main/kotlin/misk/client/ClientInvocationHandler.kt
```diff
@@ -56,8 +59,12 @@ internal class ClientInvocationHandler(
           clientBuilder.build()
         }
 
-    methodName to retrofit.newBuilder()
+    val retrofitBuilder = retrofit.newBuilder()
         .client(actionSpecificClient)
+
+    if (tracer != null) retrofitBuilder.callFactory(TracingCallFactory(actionSpecificClient, tracer))
```
mmihic: Please add this to the retrofit object created in the `TypedHttpClientModule` rather than doing it here. All global retrofit properties should be established in the module; anything added here is specific to an individual action

misk/src/main/kotlin/misk/client/TypedHttpClientModule.kt
```diff
@@ -61,21 +62,24 @@ class TypedHttpClientModule<T : Any>(
...
+    @Inject(optional=true)
+    private val tracer: Tracer? = null
+
     override fun get(): T {
       val okhttp = httpClientProvider.get()
 
       val retrofit = Retrofit.Builder()
           .baseUrl(httpClientsConfig[name].url)
           .addConverterFactory(MoshiConverterFactory.create(moshi))
-          .build()
```
mmihic: Why this change?

misk/src/main/kotlin/misk/tracing/interceptors/TextMultimapExtractAdapter.kt
```diff
@@ -0,0 +1,32 @@
...
+internal class TextMultimapExtractAdapter(
+  private val multimap: Map<String, List<String>>
+) : TextMap {
+  override fun put(key: String, value: String) {
+    throw UnsupportedOperationException("Cannot put into readonly data source")
+  }
+
+  override fun iterator(): MutableIterator<MutableMap.MutableEntry<String, String>> {
+    return multimap.flatMap {
+      val values = it.value
```
mmihic: nit. just ``it.values.map { mapValues -> ExtractEntry(it.key, mapValue) }``

misk/src/main/kotlin/misk/web/interceptors/TracingInterceptor.kt
```diff
@@ -29,7 +32,21 @@ internal class TracingInterceptor internal constructor(private val tracer: Trace
   }
 
   override fun intercept(chain: NetworkChain): Response<*> {
-    val scope = tracer.buildSpan("${chain.action}").startActive(true)
+    val parentContext: SpanContext? = try {
+      tracer.extract(Format.Builtin.HTTP_HEADERS,
+          TextMultimapExtractAdapter(chain.request.headers.toMultimap()))
+    } catch (exception: Exception) {
```
mmihic: nit: use ``e`` not ``exception``

misk/src/main/kotlin/misk/client/TypedHttpClientModule.kt
nbroyles: Blerg, merge conflict resolution gone bad. Will fix.

misk/src/main/kotlin/misk/client/ClientInvocationHandler.kt
nbroyles: So, I originally had it in the `TypedHttpClientModule`, but that broke once the `ClientInvocationHandler` changes landed. The issue was that the `TracingCallFactory` wasn't  being given the correct `OkHttpClient` after the changes. The `OkHttpClient` in `TypedHttpClientModule` isn't the client performing the actual calls, rather a template to generate the `actionSpecificClient`. Therefore, to get tracing to work properly, it needs to be installed in the `ClientInvocationHandler` the the call factory can be applied to the correct `OkHttpClient`

PR #140, opened 2018-03-05
Adds client interceptor support

mmihic:
Supports both application level interceptors (which can intercept the
creation of the Retrofit call as well as the actual enqueue/execute) and
network level interceptors (which simply wrap the okhttp3 interceptor
chain). Both types of interceptors can be conditionally installed on a
per-client action (client + method).

### DIFF
--- a/misk/src/main/kotlin/misk/client/ClientNetworkInterceptor.kt
+++ b/misk/src/main/kotlin/misk/client/ClientNetworkInterceptor.kt
@@ -1,0 +1,15 @@
+package misk.client
+
+import okhttp3.Response
+
+/**
+ * Intercepts client side calls at the application level, able to view and modify the
+ * outgoing HTTP request and observe the returned HTTP response
+ */
+interface ClientNetworkInterceptor {
+  fun intercept(chain: ClientNetworkChain) : Response
+
+  interface Factory {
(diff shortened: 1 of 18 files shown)


misk/src/test/kotlin/misk/client/TypedHttpClientInterceptorTest.kt
```diff
@@ -0,0 +1,177 @@
...
+  private lateinit var client: ReturnADinosaur
+
+  @BeforeEach
+  fun createClient() {
+    val clientInjector = Guice.createInjector(MoshiModule(), ClientModule(jetty))
+    client = clientInjector.getInstance()
+  }
+
+  @Test
+  fun useTypedClient() {
```
eeSeeGee: Maybe a test for if an Interceptor throws? E.g. a ClientApplicationInterceptor that checks that you have a certain header in your request?

misk/src/main/kotlin/misk/client/ClientNetworkInterceptor.kt
```diff
@@ -0,0 +1,15 @@
+package misk.client
+
+import okhttp3.Response
+
+/**
+ * Intercepts client side calls at the application level, able to view and modify the
```
ryanhall07: application level? do you mean network level?

PR #141, opened 2018-03-06
Expose client certificates as action scoped vars

mmihic:
Allows interceptors and actions to make decisions based on the subject
of the authenticated client

### DIFF
--- a/misk/src/main/kotlin/misk/security/ssl/CertificatesModule.kt
+++ b/misk/src/main/kotlin/misk/security/ssl/CertificatesModule.kt
@@ -1,0 +1,61 @@
+package misk.security.ssl
+
+import com.google.inject.TypeLiteral
+import misk.scope.ActionScoped
+import misk.scope.ActionScopedProvider
+import misk.scope.ActionScopedProviderModule
+import misk.security.x509.X500Name
+import java.security.cert.X509Certificate
+import javax.inject.Inject
+import javax.servlet.http.HttpServletRequest
+
+/** Installs support for accessing client certificates */
+internal class CertificatesModule : ActionScopedProviderModule() {
+  override fun configureProviders() {
+    bindProvider(
+        type = certificateArrayType,
+        providerType = ClientCertProvider::class,
+        annotatedBy = ClientCertChain::class.java)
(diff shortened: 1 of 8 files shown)


misk/src/main/kotlin/misk/security/ssl/CertificatesModule.kt
```diff
@@ -0,0 +1,61 @@
...
+import misk.security.x509.X500Name
+import java.security.cert.X509Certificate
+import javax.inject.Inject
+import javax.servlet.http.HttpServletRequest
+
+/** Installs support for accessing client certificates */
+internal class CertificatesModule : ActionScopedProviderModule() {
+  override fun configureProviders() {
+    bindProvider(
+        type = certificateArrayType,
```
ryanhall07: do you have to specify the parameter names or are you just doing it for readability?

misk/src/main/kotlin/misk/security/ssl/CertificatesModule.kt
```diff
@@ -0,0 +1,61 @@
...
+    @Inject @JvmSuppressWildcards @ClientCertChain
+    lateinit var clientCert: ActionScoped<Array<X509Certificate>?>
+
+    override fun get(): X500Name? {
+      return clientCert.get()?.get(0)?.issuerX500Principal?.name?.let { X500Name.parse(it) }
+    }
+  }
+
+  private companion object {
+    val certificateArrayType = object : TypeLiteral<Array<X509Certificate>?>() {}
```
ryanhall07: this is nice

misk/src/main/kotlin/misk/security/x509/X500Name.kt
```diff
@@ -0,0 +1,97 @@
...
+  val organizationalUnit = get("OU")
+  val state = get("ST")
+  val locality = get("L")
+  val country = get("C")
+
+  operator fun get(componentName: String) = components[componentName.toUpperCase()]
+
+  fun asMap() = components
+
+  companion object {
```
ryanhall07: is there no java library that would do this parsing?

PR #143, opened 2018-03-07
Add backend for Zipkin tracer

nbroyles:
Wire up the configuration

### DIFF
--- a/misk/src/main/kotlin/misk/tracing/TracingModule.kt
+++ b/misk/src/main/kotlin/misk/tracing/TracingModule.kt
@@ -1,19 +1,33 @@
 package misk.tracing
 
+import com.google.common.util.concurrent.Service
 import misk.inject.KAbstractModule
+import misk.inject.addMultibinderBinding
+import misk.inject.to
 import misk.logging.getLogger
+import misk.tracing.backends.TracingBackendConfig
 import misk.tracing.backends.jaeger.JaegerBackendModule
 import misk.tracing.backends.noop.NoopTracerBackendModule
+import misk.tracing.backends.zipkin.ZipkinBackendModule
+import kotlin.reflect.full.memberProperties
 
 private val logger = getLogger<TracingModule>()
 
 class TracingModule(val config: TracingConfig) : KAbstractModule() {
   override fun configure() {
-    if (config.backends.jaeger != null) {
-      install(JaegerBackendModule(config.backends.jaeger))
-    } else {
-      install(NoopTracerBackendModule())
-      logger.info("No tracing backend configured. NoopTracer installed.")
+    binder().addMultibinderBinding<Service>().to<TracingService>()
+
+    val configuredTracers = TracingBackendConfig::class.memberProperties.mapNotNull { it.get(config.backends) }
+
+    check(configuredTracers.size == 1) { "More than one tracer has been configured. Please remove one." }
+
+    when {
+      config.backends.jaeger != null -> install(JaegerBackendModule(config.backends.jaeger))
+      config.backends.zipkin != null -> install(ZipkinBackendModule(config.backends.zipkin))
+      else -> {
+        install(NoopTracerBackendModule())
+        logger.info("No tracing backend configured. NoopTracer installed.")
+      }
     }
   }
 }
--- a/misk/src/main/kotlin/misk/tracing/backends/TracingBackendConfig.kt
+++ b/misk/src/main/kotlin/misk/tracing/backends/TracingBackendConfig.kt
@@ -2,7 +2,9 @@ package misk.tracing.backends
 
 import misk.config.Config
 import misk.tracing.backends.jaeger.JaegerBackendConfig
+import misk.tracing.backends.zipkin.ZipkinBackendConfig
 
 data class TracingBackendConfig(
-  val jaeger: JaegerBackendConfig?
+  val jaeger: JaegerBackendConfig?,
+  val zipkin: ZipkinBackendConfig?
 ) : Config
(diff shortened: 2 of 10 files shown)


misk/src/main/kotlin/misk/tracing/backends/zipkin/ZipkinBackendModule.kt
```diff
@@ -0,0 +1,40 @@
...
+import misk.inject.to
+import misk.tracing.TracingService
+import zipkin2.reporter.AsyncReporter
+import zipkin2.reporter.okhttp3.OkHttpSender
+import java.util.concurrent.TimeUnit
+import javax.inject.Singleton
+
+class ZipkinBackendModule(val config: ZipkinBackendConfig) : KAbstractModule() {
+  override fun configure() {
+    binder().addMultibinderBinding<Service>().to<TracingService>()
```
ryanhall07: why does each impl need to do this? shouldn't the `TracerModule` do this for you?

misk/src/main/kotlin/misk/tracing/backends/zipkin/ZipkinBackendModule.kt
```diff
@@ -0,0 +1,40 @@
...
+    val reporter = AsyncReporter.builder(OkHttpSender.create(config.collector_url))
+
+    if (config.reporter?.message_max_bytes != null)
+      reporter.messageMaxBytes(config.reporter.message_max_bytes)
+    if (config.reporter?.message_timeout_sec != null)
+      reporter.messageTimeout(config.reporter.message_timeout_sec, TimeUnit.SECONDS)
+    if (config.reporter?.queued_max_bytes != null)
+      reporter.queuedMaxBytes(config.reporter.queued_max_bytes)
+
+    return BraveTracer.create(Tracing.newBuilder()
```
ryanhall07: lol the Brave Little Tracer

misk/src/main/kotlin/misk/tracing/backends/zipkin/ZipkinReporterConfig.kt
```diff
@@ -0,0 +1,21 @@
+package misk.tracing.backends.zipkin
+
+import misk.config.Config
+
+data class ZipkinReporterConfig(
+  /**
+   * Maximum backlog of span bytes reported vs sent. Corresponds to ReporterMetrics.updateQueuedBytes.
+   * Default 1% of heap
+   */
+  val queued_max_bytes: Int?,
```
ryanhall07: any way to inline these defaults?

misk/src/main/kotlin/misk/tracing/backends/TracingBackendConfig.kt
```diff
@@ -2,7 +2,9 @@ package misk.tracing.backends
 
 import misk.config.Config
 import misk.tracing.backends.jaeger.JaegerBackendConfig
+import misk.tracing.backends.zipkin.ZipkinBackendConfig
 
 data class TracingBackendConfig(
-  val jaeger: JaegerBackendConfig?
+  val jaeger: JaegerBackendConfig?,
+  val zipkin: ZipkinBackendConfig?
```
ryanhall07: something needs to validate only one is set

misk/src/main/kotlin/misk/tracing/TracingModule.kt
```diff
@@ -4,13 +4,16 @@ import misk.inject.KAbstractModule
 import misk.logging.getLogger
 import misk.tracing.backends.jaeger.JaegerBackendModule
 import misk.tracing.backends.noop.NoopTracerBackendModule
+import misk.tracing.backends.zipkin.ZipkinBackendModule
 
 private val logger = getLogger<TracingModule>()
 
 class TracingModule(val config: TracingConfig) : KAbstractModule() {
   override fun configure() {
     if (config.backends.jaeger != null) {
```
mmihic: supernit: At this point might want to switch over the a ``when`` statement. e.g.

````kotlin
when {
  config.backends.jaeger != null -> install(JaegerBackendModule(config.backends.jaeger))
  config.backends.zipkin != null -> install(ZipkinBackendModule(config.backends.zipkin))
  else -> {
    install(NoopTracerBackendModule())
   logger.info("No tracing backend configured. NoopTracer installed")
  }
}
````

File: misk/src/main/kotlin/misk/tracing/backends/zipkin/ZipkinBackendConfig.kt [hunk 6]
```diff
@@ -0,0 +1,8 @@
+package misk.tracing.backends.zipkin
+
+import misk.config.Config
+
+data class ZipkinBackendConfig(
+  val collector_url: String,
+  val reporter: ZipkinReporterConfig?
```
mmihic: Is this really optional?

misk/src/main/kotlin/misk/tracing/backends/zipkin/ZipkinBackendModule.kt
nbroyles: Good catch. After the refactor to always install a tracer even there isn't one in the config (e.g. `NoopTracer`) this can be moved into the Tracing module. Will fix.

misk/src/main/kotlin/misk/tracing/backends/zipkin/ZipkinBackendConfig.kt
nbroyles: Yeah, Zipkin gives you sane defaults if not present.

misk/src/main/kotlin/misk/tracing/backends/zipkin/ZipkinReporterConfig.kt
nbroyles: So, we had the same debate with Jaeger. Both of these tracers don't expose their defaults. Is it worth just copying them over anyway?

ryanhall07: no. was just asking

PR #144, opened 2018-03-12
boilerplate create-react-app hello world

terrilldent:
This is the generic hello world setup for a "create-react-app"

To run locally you may need to run these commands:

```
npm install -g npx
yarn install
yarn start
```


exemplar/web/package.json
```diff
@@ -0,0 +1,16 @@
+{
+  "name": "exemplar",
+  "version": "0.1.0",
```
swankjesse: for me to follow up – are all of these required? Can we omit the version etc.?

exemplar/web/public/manifest.json
```diff
@@ -0,0 +1,15 @@
+{
```
swankjesse: neat

mmihic:
High level - do we want this in exemplar, or in another example app? I'm a little wary of exemplar becoming a grab bag of random examples of stuff making it confusing and not so useful as an example app. Think we should consider having a family of example apps each of which is targeted at specific subsets of misk (transactional RPC style app, webapps, background task app, app using various cloud services, etc).

exemplar/web/package.json
eeSeeGee: You can specifiy a minimum version but that's just asking for trouble.

swankjesse:
https://github.com/square/misk/pull/146

PR #150, opened 2018-03-13
Add a RELEASING.md to outline misk release process

nbroyles:


ryanhall07:
I don't have any experience cutting open source releases, but I'm surprised this isn't more automated with some kind maven plugin

PR #151, opened 2018-03-14
Add MoshiJsonAdapters and SocketEventJsonAdapter

tso:


misk/src/main/kotlin/misk/eventrouter/SocketEvent.kt
```diff
@@ -0,0 +1,40 @@
...
+
+sealed class SocketEvent {
+  data class Event(val topic: String, val message: String) : SocketEvent()
+  data class Subscribe(val topic: String) : SocketEvent()
+  data class Unsubscribe(val topic: String) : SocketEvent()
+  data class Ack(val topic: String) : SocketEvent()
+  class UnknownItem : SocketEvent()
+}
+
+object SocketEventJsonAdapter {
```
swankjesse: thumbs up

PR #152, opened 2018-03-14
Add basic frontend for exemplarchat

tso:
Looks like: image

### DIFF
--- a/exemplarchat/web/src/App.js
+++ b/exemplarchat/web/src/App.js
@@ -3,16 +3,58 @@ import logo from './logo.svg';
 import './App.css';
 
 class App extends Component {
+  constructor() {
+    super();
+
+    this.state = {
+      pendingMessage: "",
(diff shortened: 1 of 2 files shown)


exemplarchat/web/src/App.js
```diff
@@ -3,16 +3,58 @@ import logo from './logo.svg';
...
+
+    this.state = {
+      pendingMessage: "",
+      messages: [],
+    };
+  }
+
+  componentDidMount() {
+    const href = window.location.href;
+    const room = href.substr(href.lastIndexOf('/') + 1);
```
terrilldent: This is a bit hacky, but it'll work if you load the right path.

tso: I assume the non-hacky way is adding `react-router` but I was/am unsure what we are planning on doing for routing in general so I figured hacky is better than committing to something

PR #153, opened 2018-03-16
Add support for logging to StackDriver

nbroyles:


misk/src/main/kotlin/misk/tracing/TracingLoggingEnhancer.kt
```diff
@@ -0,0 +1,38 @@
...
+import com.google.cloud.logging.LoggingEnhancer
+import com.uber.jaeger.Span
+import io.opentracing.Tracer
+import io.opentracing.util.GlobalTracer
+
+/**
+ * Add this enhancer to your project's Logback configuration to add trace ids to logs generated by
+ * GCP's LoggingAppender
+ */
+class TracingLoggingEnhancer : LoggingEnhancer {
```
mmihic: Can you put this into the gcp package? Want to keep all cloud-specific stuff under the appropriate cloud package.

mmihic: Also how does this get installed? It'd be nice if there were just a ``GcpLoggingModule`` or some such that configure the right thing (doing it in a ``Guice`` module might be too late, we might need some out-of-band mechanism that selects your target cloud environment similar to what's done with config / ``Environment``)

nbroyles: Ah, sure, should have added more context. I'm actually trying to figure out exactly what you mentioned re: installing a module to enable right now. As it stands, this can currently be used by supplying a logback.xml config file in your projects resources folder. Because of the way Logback works, if that file is present, then it will attempt to configure the logger you specify in the config file. 

Planned on landing this PR since it's a complete unit of work and orthogonal to getting the module bit working.

nbroyles: @mmihic Updated with a static method we can use to enable from client apps

PR #157, opened 2018-03-20
Support loading keystores from combined private key and certificate chain PEM files

eeSeeGee:
Support for a key format easily pulled from kubernetes.


misk/src/main/kotlin/misk/web/PemComboFile.kt
```diff
@@ -0,0 +1,144 @@
+package misk.web
```
mmihic: plz move to the security package with the other ssl related things

misk/src/main/kotlin/misk/web/PemComboFile.kt
```diff
@@ -0,0 +1,144 @@
...
+ * client trusts) and also for TLS servers (which certificate chain a TLS server
+ * serves).
+ */
+data class PemComboFile(
+  private val certificates: List<ByteString>,
+  private val privateRsaKeys: List<ByteString>,
+  private val privateKeys: List<ByteString>,
+  private val passphrase: String
+) {
+  fun tlsClientTrustStore(): KeyStore {
```
mmihic: Not sure I understand why we have these two functions - why not just one that parses whatever is in the file and builds a keystore out of the contents?

misk/src/test/resources/ssl/client.pem
```diff
@@ -0,0 +1,28 @@
+-----BEGIN PRIVATE KEY-----
```
mmihic: Please update the README.md in this directory with the steps used to build these files, so that we know how to reconstruct them in case of expiration etc.

misk/src/main/kotlin/misk/web/PemComboFile.kt
mmihic: Specifically, the distinction between a "trust store" and a "key store" is an artificial one; it's just how the application code wants to use the contents of the file. It's a distinction that matters at the web server / client level, but shouldn't be something that permeates a basic file parser.

swankjesse: I recommended these, but you’re right that they’re wrong!

I still think we need something to help callers to differentiate our two main use cases:

 - 1 private key & _N_ certs that I use to prove my own identity
 - 0 private keys and _N_ certs that I trust when validating other services’ identity

One thing that makes this extra tricky is that we use the private key both for clients and servers. Typically with TLS it’s just the private key for servers.

misk/src/main/kotlin/misk/web/PemComboFile.kt
```diff
@@ -0,0 +1,144 @@
...
+    val keyStore = newEmptyKeyStore()
+    decodeCertificates().forEachIndexed { index, certificate ->
+      keyStore.setCertificateEntry(index.toString(), certificate)
+    }
+    return keyStore
+  }
+
+  fun tlsServerKeyStore(): KeyStore? {
+    if (certificates.isEmpty() || privateRsaKeys.size + privateKeys.size != 1) return null
```
swankjesse: document something like `/** Returns null if ... */`

misk/src/main/kotlin/misk/web/PemComboFile.kt
```diff
@@ -0,0 +1,144 @@
...
+
+    val result = newEmptyKeyStore()
+    val priKeySpec = if (privateKeys.isEmpty()) convertPKCS1toPKCS8(privateRsaKeys[0])
+      else PKCS8EncodedKeySpec(privateKeys[0].toByteArray())
+    val kf = KeyFactory.getInstance("RSA")
+    val priKey = kf.generatePrivate(priKeySpec)
```
swankjesse: priKey –> privateKey
kf –> keyFactory

etc.

misk/src/main/kotlin/misk/web/PemComboFile.kt
```diff
@@ -0,0 +1,144 @@
...
+
+      // skip version seq[0];
+      val modulus = seq[1].bigInteger
+      val publicExp = seq[2].bigInteger
+      val privateExp = seq[3].bigInteger
+      val prime1 = seq[4].bigInteger
+      val prime2 = seq[5].bigInteger
+      val exp1 = seq[6].bigInteger
+      val exp2 = seq[7].bigInteger
+      val crtCoef = seq[8].bigInteger
```
swankjesse: Use the names from the JDK?
```
RSAPrivateCrtKeySpec(BigInteger modulus, BigInteger publicExponent, BigInteger privateExponent, BigInteger primeP, BigInteger primeQ, BigInteger primeExponentP, BigInteger primeExponentQ, BigInteger crtCoefficient) 
```

misk/src/main/kotlin/misk/web/PemComboFile.kt
mmihic: I'd propose doing this with higher level APIs that wrap an underlying ``KeyStore`` and validate that they contain / expose only the expected contents e.g. have a ``CertStore`` that exposes the advertised certificate and private key and a ``TrustStore`` that exposes only a collection of trusted certificate chains. That way the PEM parser just builds a general ``KeyStore`` out of its contents (whatever they may be) and then the SSL layers enforce particular constraints on the contents of a ``KeyStore`` depending on the use case.

eeSeeGee: I like that concept, I'll add some higher level classes that serve as very basic wrappers for these two store types. We can iterate them into more general abstractions down the road.

PR #159, opened 2018-03-21
Fix event router tests

tso:
The code in `EventRouterTestingModule.kt` feels pretty bleh but I don't know of a better way to do it

### DIFF
--- a/misk/src/main/kotlin/misk/eventrouter/RealEventRouterModule.kt
+++ b/misk/src/main/kotlin/misk/eventrouter/RealEventRouterModule.kt
@@ -19,7 +19,13 @@ class RealEventRouterModule : KAbstractModule() {
   }
 
   @Provides @Singleton @ForEventRouterActions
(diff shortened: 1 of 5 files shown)


misk/src/main/kotlin/misk/eventrouter/RealEventRouterModule.kt
```diff
@@ -19,18 +19,33 @@ class RealEventRouterModule : KAbstractModule() {
   }
 
   @Provides @Singleton @ForEventRouterActions
-  fun provideExecutor(): ExecutorService =
+  fun actionExecutor(): ExecutorService =
+      ThreadPoolExecutor(1, 1, 1, TimeUnit.MINUTES, LinkedBlockingQueue())
+
+  // TODO(tso): make this single threaded _per subscriber_ rather than single threaded
```
swankjesse: nice

PR #162, opened 2018-03-27
Don't treat assembly as a release when running in CI

eeSeeGee:
This hinges on adding "CI=true" to Travis, which I've already done.

### DIFF
--- a/gradle/gradle-mvn-push.gradle
+++ b/gradle/gradle-mvn-push.gradle
@@ -21,7 +21,7 @@ version = VERSION_NAME
 group = GROUP
 
 def isReleaseBuild() {
-  return VERSION_NAME.contains("SNAPSHOT") == false
+  return !isCi && VERSION_NAME.contains("SNAPSHOT") == false
 }
 
 def getReleaseRepositoryUrl() {


gradle/gradle-mvn-push.gradle
```diff
@@ -21,7 +21,7 @@ version = VERSION_NAME
 group = GROUP
 
 def isReleaseBuild() {
-  return VERSION_NAME.contains("SNAPSHOT") == false
+  return !isCi && VERSION_NAME.contains("SNAPSHOT") == false
```
nbroyles: Ah, I never hit this because I never pushed a change to master with a VERSION_NAME that didn't contain SNAPSHOT. My approach was to simply bump the version for the purpose of the release on a branch, tag that commit and push the tags, and then submit a PR with new version plus '-SNAPSHOT' appended. I have no idea which approach is preferred, but either will work.

nbroyles: cc @swankjesse for his thoughts since I cribbed the release notes from SQLDelight.

eeSeeGee: Yeah, that's fine too, this is a nice safety check, I think.

PR #167, opened 2018-03-29
Fix tracing startup when none is configured

eeSeeGee:


misk/src/main/kotlin/misk/tracing/TracingModule.kt
```diff
@@ -19,7 +19,7 @@ class TracingModule(val config: TracingConfig) : KAbstractModule() {
 
     val configuredTracers = TracingBackendConfig::class.memberProperties.mapNotNull { it.get(config.backends) }
 
-    check(configuredTracers.size == 1) { "More than one tracer has been configured. Please remove one." }
+    check(configuredTracers.size < 2) { "More than one tracer has been configured. Please remove one." }
```
nbroyles: Heh, oops. Good catch.

PR #170, opened 2018-04-16
Adds a local cluster connector so that development functions

tso:


misk/src/main/kotlin/misk/eventrouter/RealEventRouterModule.kt
```diff
@@ -18,10 +19,15 @@ import javax.inject.Singleton
 
 class RealEventRouterModule : KAbstractModule() {
   override fun configure() {
+    val environment = Environment.fromEnvironmentVariable()
```
tso: This should be `@Inject lateinit var environment: Environment`

misk/src/main/kotlin/misk/eventrouter/LocalClusterConnector.kt
```diff
@@ -0,0 +1,19 @@
+package misk.eventrouter
+
+import misk.web.actions.WebSocket
+import misk.web.actions.WebSocketListener
+import javax.inject.Singleton
+
+@Singleton
+internal class LocalClusterConnector : ClusterConnector {
+  override fun joinCluster(topicPeer: TopicPeer) {
+    topicPeer.clusterChanged(ClusterSnapshot(listOf("local"), "local"))
```
tso: Should use the kubernetes config to get `my_pod_name`

exemplarchat/src/main/resources/chat-development.yaml
```diff
@@ -1,3 +1,3 @@
 kubernetes:
-  my_pod_namespace: none
-  my_pod_name: none
+  my_pod_namespace: default
+  my_pod_name: Localhost
```
swankjesse: should this be lowercase? I’ve never seen the string `Localhost` with a `L` and it looks so weird to my eyes

misk/src/main/kotlin/misk/eventrouter/LocalClusterConnector.kt
swankjesse: `local` or `Localhost` ? Are these synonyms?

misk/src/main/kotlin/misk/eventrouter/LocalClusterConnector.kt
```diff
@@ -0,0 +1,19 @@
...
+  }
+
+  override fun leaveCluster(topicPeer: TopicPeer) {
+  }
+
+  override fun connectSocket(hostname: String, listener: WebSocketListener): WebSocket {
+    throw IllegalStateException("In a single machine system there should be no network requests")
```
swankjesse: nice

misk/src/main/kotlin/misk/eventrouter/RealEventRouterModule.kt
swankjesse: ahh, you can’t ’cause this is itself a module. Pass the environment to the module constructor?

PR #171, opened 2018-04-16
Change CacheBuilder to be mapOf since no concurrency

tso:


misk/src/main/kotlin/misk/eventrouter/RealEventRouter.kt
```diff
@@ -245,6 +234,16 @@ internal class RealEventRouter : EventRouter {
     }
   }
 
+  private fun hostToSocket(hostname: String): WebSocket {
+    logger.debug { "connecting to $hostname" }
+    val ws = hostsToSockets[hostname]
+    if (ws == null) {
+      hostsToSockets = hostsToSockets.plus(
+          Pair(hostname, clusterConnector.connectSocket(hostname, webSocketListener)))
```
swankjesse: nit: probably better to use a MutableMap! Also weak to do three map lookups when two will do!

PR #176, opened 2018-04-29
Adds a DataSourceModule

mmihic:
Allows binding a qualified or unqualified DataSource based on
configuration. The bound DataSources uses HikariCP for the connection
pool, with a limited set of configuration options.

See https://github.com/square/misk/issues/175

### DIFF
--- a/misk/src/main/kotlin/misk/jdbc/DataSourceConfig.kt
+++ b/misk/src/main/kotlin/misk/jdbc/DataSourceConfig.kt
@@ -1,0 +1,39 @@
+package misk.jdbc
+
+import misk.config.Config
+import java.time.Duration
+
+/** Defines a type of datasource */
+enum class DataSourceType(
+  val driverClassName: String,
+  val buildJdbcUrl: (DataSourceConfig) -> String
+) {
+  MYSQL("com.mysql.jdbc.Driver", { config ->
+    val port = config.port ?: 3306
+    val host = config.host ?: "127.0.0.1"
+    val database = config.database ?: ""
+    "jdbc:mysql://${host}:$port/$database"
+
+  }),
+  HSQLDB("org.hsqldb.jdbcDriver", { config ->
+    "jdbc:hsqldb:mem:${config.database!!}"
+  })
+}
+
+/** Configuration element for an individual datasource */
+data class DataSourceConfig(
+  val type: DataSourceType,
+  val host: String? = null,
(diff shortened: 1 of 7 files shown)


misk/src/main/kotlin/misk/jdbc/DataSourceConfig.kt
```diff
@@ -0,0 +1,39 @@
...
+data class DataSourceConfig(
+  val type: DataSourceType,
+  val host: String? = null,
+  val port: Int? = null,
+  val database: String? = null,
+  val username: String? = null,
+  val password: String? = null,
+  val connection_properties: Map<String, String> = mapOf(),
+  val fixed_pool_size: Int = 10,
+  val connection_timeout: Duration = Duration.ofSeconds(30),
```
ryanhall07: this seems long for a default

misk/src/main/kotlin/misk/jdbc/DataSourceModule.kt
```diff
@@ -0,0 +1,60 @@
...
+        DataSourceModule(datasourceName, Key.get(DataSource::class.java, annotatedBy))
+
+    fun <A : Annotation> create(datasourceName: String, annotatedBy: Class<A>) =
+        DataSourceModule(datasourceName, Key.get(DataSource::class.java, annotatedBy))
+
+    fun <A : Annotation> create(datasourceName: String, annotatedBy: KClass<A>) =
+        DataSourceModule(datasourceName, Key.get(DataSource::class.java, annotatedBy.java))
+  }
+
+  private class DataSourceProvider(private val datasourceName: String) : Provider<DataSource> {
```
ryanhall07: should this be in a separate hikari package? do you think apps would want to provide their own Datasource?

mmihic: If so, they can implement their own module for it. The only thing this module does is setup the Hikari CP, so not much value in abstracting that out

misk/src/main/kotlin/misk/jdbc/DataSourceConfig.kt
mmihic: This is the default in Hikari CP; wanted the defaults here to mirror the defaults there as if you never set them

ryanhall07: ah. good idea

PR #180, opened 2018-05-01
Update DataSourceModuleTest to use yaml configuration to demonstrate …

jmuia:
…end-to-end DataSource binding

### DIFF
--- a/misk/src/test/resources/test_data_source_app-common.yaml
+++ b/misk/src/test/resources/test_data_source_app-common.yaml
@@ -1,0 +1,7 @@
+data_sources:
+  databases:
+    exemplar:
+      type: HSQLDB
+      username: snork
+      password: florp
+      database: lorfil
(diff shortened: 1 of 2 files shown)


misk/src/test/resources/test_data_source_app-common.yaml
```diff
@@ -0,0 +1,7 @@
+data_sources:
+  databases:
```
jmuia: it isn't great to have this additional level of nesting. I think we'd need to update the `DataSourcesConfig` class to implement the `Map` interface so Jackson populates it properly (?).

jmuia:
I wanted an example of pulling the data sources from configuration files. I didn't update the other configs that override properties to test failure cases, but I can if that makes sense.

misk/src/test/resources/test_data_source_app-common.yaml
mmihic: yeah totally agreed; can it just implement Map, or do we need a custom Yaml unmarshaller for that type?

jmuia: Not 100% sure, I'm going to investigate a bit and will follow up with another PR

PR #182, opened 2018-05-02
MiskCaller and authz support

mmihic:
Exposes an ActionScoped<MiskCaller> representing the service or user
calling into the current action. Provides authenticators which identify
the service via the OU of a client cert (if specified) or the user via
headers forwarded from trogdor. Provides an AccessInterceptor which
enforces annotation based authentication rules, only allowing calls into
actions from a designated set of allowed services or allowed roles.

### DIFF
--- a/misk/src/main/kotlin/misk/security/authz/AccessInterceptor.kt
+++ b/misk/src/main/kotlin/misk/security/authz/AccessInterceptor.kt
@@ -1,0 +1,56 @@
+package misk.security.authz
+
+import misk.Action
+import misk.ApplicationInterceptor
+import misk.Chain
+import misk.MiskCaller
+import misk.exceptions.UnauthenticatedException
+import misk.exceptions.UnauthorizedException
+import misk.scope.ActionScoped
+import javax.inject.Inject
+import kotlin.reflect.full.findAnnotation
+
+class AccessInterceptor private constructor(
+  private val allowedServices: Set<String>,
+  private val allowedRoles: Set<String>,
+  private val caller: ActionScoped<MiskCaller?>
+) : ApplicationInterceptor {
+
+  override fun intercept(chain: Chain): Any {
+    val caller = caller.get() ?: throw UnauthenticatedException()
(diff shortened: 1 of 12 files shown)


misk/src/main/kotlin/misk/security/authz/AccessInterceptor.kt
```diff
@@ -0,0 +1,56 @@
...
+import kotlin.reflect.full.findAnnotation
+
+class AccessInterceptor private constructor(
+  private val allowedServices: Set<String>,
+  private val allowedRoles: Set<String>,
+  private val caller: ActionScoped<MiskCaller?>
+) : ApplicationInterceptor {
+
+  override fun intercept(chain: Chain): Any {
+    val caller = caller.get() ?: throw UnauthenticatedException()
```
ryanhall07: nice! does it make sense to include an error message like `no caller found` or is giving away too much info?

misk/src/main/kotlin/misk/security/authz/TrogdorAuthenticator.kt
```diff
@@ -0,0 +1,26 @@
+package misk.security.authz
+
+import misk.MiskCaller
+import misk.scope.ActionScoped
+import misk.web.Request
+import javax.inject.Inject
+
+/** Maps the incoming request to a human user based on headers forwarded by trogdor */
```
ryanhall07: should we start putting sq specific stuff in a separate package?

ryanhall07:
we should also be careful about sq specific infrastructure in commit messages. ie. explain what trogdor is if you're going to reference it

misk/src/main/kotlin/misk/security/authz/TrogdorAuthenticator.kt
ryanhall07: I think ideally a separate repo, but that might be quite a pain with the state of our gradle build. related to the bug you filed recently about splitting everything up. so maybe just put in a separate package for now with the intention of moving out

PR #186, opened 2018-05-10
Add Backoff/ExponentialBackoff

mmihic:
Adds a stateful mechanism for computing backoffs, with a specific implementation for computing exponential backoffs with optional jitter.

### DIFF
--- a/misk/src/main/kotlin/misk/backoff/ExponentialBackoff.kt
+++ b/misk/src/main/kotlin/misk/backoff/ExponentialBackoff.kt
@@ -1,0 +1,71 @@
+package misk.backoff
+
+import java.time.Duration
+import java.util.concurrent.ThreadLocalRandom
+
+/**
+ * Performs exponential backoff with optional jitter. Durations are supplied as
+ * functions, so that they can change dynamically as the system is running (e.g.
+ * in response to changes in dynamic flags)
+ */
+class ExponentialBackoff(
+  private val baseDelay: () -> Duration,
+  private val maxDelay: () -> Duration,
+  private val jitter: () -> Duration // TODO(mmihic): Consider making jitter a %
+
+) : Backoff {
+  private var consecutiveRetryCount: Int = 0
+  private var maxRetryCount = Integer.MAX_VALUE
+
+  /**
+   * Creates a new unjittered [ExponentialBackoff] using a function for the base
+   * and max retry delays.
+   *
+   * @param baseDelay The [Supplier] for the base delay
+   * @param maxDelay The [Supplier] for maximum amount of time to wait between retries
+   */
+  constructor(baseDelay: () -> Duration, maxDelay: () -> Duration) :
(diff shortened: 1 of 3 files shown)


misk/src/main/kotlin/misk/backoff/ExponentialBackoff.kt
```diff
@@ -0,0 +1,71 @@
+package misk.backoff
+
+import java.time.Duration
+import java.util.concurrent.ThreadLocalRandom
+
+/**
+ * Performs exponential backoff with optional jitter. Durations are supplied as
```
ryanhall07: why an optional jitter? or at least make it the default

misk/src/main/kotlin/misk/backoff/ExponentialBackoff.kt
```diff
@@ -0,0 +1,71 @@
...
+  private val baseDelay: () -> Duration,
+  private val maxDelay: () -> Duration,
+  private val jitter: () -> Duration // TODO(mmihic): Consider making jitter a %
+
+) : Backoff {
+  private var consecutiveRetryCount: Int = 0
+  private var maxRetryCount = Integer.MAX_VALUE
+
+  /**
+   * Creates a new unjittered [ExponentialBackoff] using a function for the base
```
ryanhall07: nit: not a function, a supplier

misk/src/main/kotlin/misk/backoff/ExponentialBackoff.kt
```diff
@@ -0,0 +1,71 @@
...
+   */
+  constructor(baseDelay: Duration, maxDelay: Duration)
+      : this(baseDelay, maxDelay, Duration.ofMillis(0))
+
+  override fun reset() {
+    consecutiveRetryCount = 0
+  }
+
+  override fun nextRetry(): Duration {
+    // Cap the retry count to prevent long overflow. Since it's a function of the configurable base
```
ryanhall07: :)

misk/src/main/kotlin/misk/backoff/ExponentialBackoff.kt
mmihic: It's a function not a supplier

misk/src/main/kotlin/misk/backoff/ExponentialBackoff.kt
mmihic: What's a good default (given that we don't know any either the base or the max delay)?

ryanhall07: Maybe a default percentage?

mmihic: You ok if I deal with this in another PR? I have a higher priority branch that's using backoffs that I want to get in first.

ryanhall07: definitely. just a suggestion

PR #187, opened 2018-05-10
Add retry() helper

mmihic:
Allows retrying a function up to a certain number of times, with a backoff

### DIFF
--- a/misk/src/main/kotlin/misk/backoff/Retries.kt
+++ b/misk/src/main/kotlin/misk/backoff/Retries.kt
@@ -1,0 +1,34 @@
+package misk.backoff
+
+/**
+ * Retries the provided function up to a certain number of times, applying the given backoff
+ * between each retry. The retry function is provided with current retry count, in case this is
+ * relevant
+ */
+fun <A> retry(upTo: Int, withBackoff: Backoff, f: (retryCount: Int) -> A): A {
+  require(upTo > 0) { "must support at least one call" }
+
+  withBackoff.reset()
+
+  var lastException: Exception? = null
+
+  for (i in 0 until upTo) {
+    try {
+      val result = f(i)
+      withBackoff.reset()
+      return result
+    } catch (e: Exception) {
+      lastException = e
+
+      if (i + 1 < upTo) {
+        try {
+          Thread.sleep(withBackoff.nextRetry().toMillis())
+        } catch (e: InterruptedException) {
+          throw e
+        }
+      }
+    }
+  }
+
+  throw lastException!!
+}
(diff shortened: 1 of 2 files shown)


misk/src/main/kotlin/misk/backoff/Retries.kt
```diff
@@ -0,0 +1,34 @@
...
+      withBackoff.reset()
+      return result
+    } catch (e: Exception) {
+      lastException = e
+
+      if (i + 1 < upTo) {
+        try {
+          Thread.sleep(withBackoff.nextRetry().toMillis())
+        } catch (e: InterruptedException) {
+          throw e
```
ryanhall07: why not continue? I don't have a lot of knowledge of why this would happen

misk/src/main/kotlin/misk/backoff/Retries.kt
```diff
@@ -0,0 +1,34 @@
+package misk.backoff
+
+/**
+ * Retries the provided function up to a certain number of times, applying the given backoff
+ * between each retry. The retry function is provided with current retry count, in case this is
+ * relevant
+ */
+fun <A> retry(upTo: Int, withBackoff: Backoff, f: (Int) -> A): A {
```
ryanhall07: it's not immediately obvious that the Int parameter to the function is the current retry number. any way to improve that?

misk/src/test/kotlin/misk/backoff/RetryTest.kt
```diff
@@ -0,0 +1,67 @@
...
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.jupiter.api.Assertions.assertThrows
+import org.junit.jupiter.api.Test
+import java.time.Duration
+
+internal class RetryTest {
+  @Test fun retries() {
+    val backoff = ExponentialBackoff(Duration.ofMillis(10), Duration.ofMillis(100))
+
+    val result = retry(3, backoff) { retry ->
```
ryanhall07: i'm surprised the function is not passed as a parameter. definitely some KT thing I don't understand.

mmihic: If the final parameter to a function is itself a function, you can use this alternate syntax.

misk/src/main/kotlin/misk/backoff/Retries.kt
mmihic: ``InterruptedException`` is thrown when someone explicitly asks that the thread interrupt its work and not continue (e.g. shutdown now plz).  It's a signal that the thread should stop whatever it is doing.

misk/src/test/kotlin/misk/backoff/RetryTest.kt
ryanhall07: ok. that makes sense to exit

PR #189, opened 2018-05-10
Remove use of instance metadata endpoints

mmihic:
AWS and GCP instance metadata endpoints expose too much information to
non-root users. Deployed pods will need to restrict access to those
endpoints. We currently use those endpoints just to retrieve instance
name, region, and zone information for the purpose of tagging logs and
metrics; switch this to pull from environment variables instead,
expecting those env vars to be set in pod manifests by deployment tools.


misk/src/main/kotlin/misk/environment/EnvironmentModule.kt
```diff
@@ -2,8 +2,16 @@ package misk.environment
 
 import misk.inject.KAbstractModule
 
-class EnvironmentModule(val environment: Environment) : KAbstractModule() {
+/**
+ * Binds [Environment] and [InstanceMetadata] to make both available to services and actions
+ */
+class EnvironmentModule(
+  private val environment: Environment,
+  private val instanceMetadata: InstanceMetadata? = null
```
ryanhall07: when is this not null?

misk/src/main/kotlin/misk/environment/InstanceMetadata.kt
```diff
@@ -1,4 +1,33 @@
...
+  companion object {
+    private const val ENV_INSTANCE_NAME = "HOSTNAME"
+    private const val ENV_ZONE = "MY_ZONE"
+    private const val ENV_REGION = "MY_REGION"
+
+    private val logger = getLogger<InstanceMetadata>()
+
+    private fun getEnv(envVar: String, default: String): String {
+      return System.getenv(ENV_INSTANCE_NAME) ?: {
+        logger.warn { "no environment variable with key $envVar; defaulting to $default" }
```
ryanhall07: should this assert `DEVELOPMENT` env?

PR #190, opened 2018-05-11
Add support for protobuf over HTTP

nbroyles:
This commit adds the application/x-protobuf content type and adds a
protobuf marshaller and unmarshaller. Additionally, allow the ability to
specify how to convert request objects into request bodies on the
TypedHttpClient and add support for protobuf.


misk/src/main/kotlin/misk/web/extractors/RequestBodyParameterExtractor.kt
```diff
@@ -25,7 +25,7 @@ internal class RequestBodyParameterExtractor(
   ): Any? {
     val mediaType = request.headers["Content-Type"]?.let { MediaType.parse(it) }
     val unmarshaller = mediaType?.let { type ->
-      unmarshallerFactories.map { it.create(type, parameter.type) }.firstOrNull()
+      unmarshallerFactories.map { it.create(type, parameter.type) }.filterNotNull().firstOrNull()
```
nbroyles: There was a subtle bug here. `firstOrNull` returns null if the list is empty or the first element. If the list is [null, SupportedUnmarshaller.Factory], then this would return null which isn't what you want. Instead, filter the list of nulls first and then return `firstOrNull`

misk/src/test/kotlin/misk/client/TypedHttpClientTest.kt
```diff
@@ -42,16 +44,25 @@ internal class TypedHttpClientTest {
   @Inject
   private lateinit var jetty: JettyService
 
-  private lateinit var client: ReturnADinosaur
+  private lateinit var clientInjector: Injector
```
ryanhall07: why inject the injector instead of just injecting the fields you need?

misk/src/test/kotlin/misk/client/TypedHttpClientTest.kt
```diff
@@ -71,23 +82,39 @@ internal class TypedHttpClientTest {
...
       install(WebActionModule.create<ReturnADinosaurAction>())
+      install(WebActionModule.create<ReturnAProtoDinosaurAction>())
     }
   }
 
   class ClientModule(val jetty: JettyService) : KAbstractModule() {
     override fun configure() {
-      install(TypedHttpClientModule.create<ReturnADinosaur>("dinosaur"))
+      install(TypedHttpClientModule.create<ReturnADinosaur>("dinosaur", Names.named("dinosaur")))
+      install(TypedHttpClientModule.create<ReturnAProtoDinosaur>("protoDino", Names.named("protoDino")))
```
ryanhall07: really missing an `protosaurus` opportunity here

misk/src/main/kotlin/misk/client/TypedHttpClientModule.kt
```diff
@@ -89,5 +91,16 @@ class TypedHttpClientModule<T : Any>(
           invocationHandler
       ) as T
     }
+
+    private fun getConverter(endpointConfig: HttpClientEndpointConfig): Converter.Factory {
+      return when {
+        endpointConfig.converter == "json" -> MoshiConverterFactory.create(moshi)
```
ryanhall07: could we just inspect the retrofit interface and determine if the request/response types are wire messages?

ryanhall07: I didn't mean the implementations, but the actual request/response type arguments. e.g `ReturnADinosaur` extends `Message`

nbroyles: Ah, sure, I think you could do that, but that's kind of a tricky game. For example, would a string by plaintext or json? Retrofit allows you to actually specify the content-type headers and I think that's probably the right way to do it. Then you can just look at that and set the converter accordingly and default to something sane (e.g. json), if not present.

ryanhall07: a string is json, since it's not a Message.

but we're not using the content-type headers. we're statically defining it

nbroyles: You can statically define the content-type headers as well via an annotation.

Unless I'm missing something here, I still don't think you want to be inspecting arguments to determine how to encode and decode content. That's literally what the headers are for.

misk/src/test/kotlin/misk/client/TypedHttpClientTest.kt
nbroyles: Not injecting the injector. We're creating a custom one in the beforeEach that represents the client side injector. Injecting fields here would get them created with the wrong injector (and set of modules). This is the same as what was there before, just making it a member variable instead of a local one so that I can get the appropriate client.

misk/src/test/kotlin/misk/client/TypedHttpClientTest.kt
nbroyles: Mistakes truly were made.

build.gradle
```diff
@@ -48,6 +48,7 @@ ext {
           zipkinBrave          : 'io.zipkin.brave:brave:4.17.2',
           retrofit             : 'com.squareup.retrofit2:retrofit:2.3.0',
           retrofitMoshi        : 'com.squareup.retrofit2:converter-moshi:2.3.0',
+          retrofitWire         : 'com.squareup.retrofit2:converter-wire:2.4.0',
```
jrodbx: should this be 2.3.0 as the other retrofit deps?  or conversely, bump the others?

misk/src/main/kotlin/misk/client/ClientInvocationHandler.kt
```diff
@@ -62,13 +70,37 @@ internal class ClientInvocationHandler(
...
-    if (tracer != null) retrofitBuilder.callFactory(TracingCallFactory(actionSpecificClient, tracer))
+    if (tracer != null) retrofitBuilder.callFactory(
+        TracingCallFactory(actionSpecificClient, tracer))
+
+    val mediaTypes = getEndpointMediaTypes(methodName)
+    if (mediaTypes.contains(MediaTypes.APPLICATION_PROTOBUF)) {
+      retrofitBuilder.addConverterFactory(WireConverterFactory.create())
+    }
+    // Always add JSON as default. Ensure it's added last so that other converters get a chance since
+    // JSON converter will accept any object.
```
jrodbx: 👍

misk/src/main/kotlin/misk/client/ClientInvocationHandler.kt
```diff
@@ -62,13 +70,37 @@ internal class ClientInvocationHandler(
     val retrofitBuilder = retrofit.newBuilder()
         .client(actionSpecificClient)
 
-    if (tracer != null) retrofitBuilder.callFactory(TracingCallFactory(actionSpecificClient, tracer))
+    if (tracer != null) retrofitBuilder.callFactory(
+        TracingCallFactory(actionSpecificClient, tracer))
```
jrodbx: nit: same line?

misk/src/test/kotlin/misk/client/TypedHttpClientTest.kt
```diff
@@ -42,16 +45,25 @@ internal class TypedHttpClientTest {
...
+    clientInjector = Guice.createInjector(MoshiModule(), ClientModule(jetty))
   }
 
   @Test
   fun useTypedClient() {
+    val client: ReturnADinosaur = clientInjector.getInstance(Names.named("dinosaur"))
+    val response = client.getDinosaur(Dinosaur.Builder().name("trex").build()).execute()
+    assertThat(response.code()).isEqualTo(200)
+    assertThat(response.body()).isNotNull()
+    assertThat(response.body()?.name!!).isEqualTo("supertrex")
```
jrodbx: `body()!!.name!!`, since you control the test inputs?

misk/src/test/kotlin/misk/client/TypedHttpClientTest.kt
ryanhall07: got it. thanks

misk/src/main/kotlin/misk/client/ClientInvocationHandler.kt
```diff
@@ -64,11 +72,34 @@ internal class ClientInvocationHandler(
...
   }.toMap()
 
+  private fun getEndpointMediaTypes(methodName: String): List<String> {
+    val headers =
+        interfaceType.memberFunctions.find { it.name == methodName }?.findAnnotation<Headers>()
+            ?: return emptyList()
+
+    return headers.value.mapNotNull {
+      val (headerKey, headerValue) = it.split(":").map { it.trim() }
+      if (headerKey.equals("Accept", true) || headerKey.equals("Content-type", true)) {
```
ryanhall07: what happens if both header values are present and different? does retrofit even allow that?

misk/src/main/kotlin/misk/client/ClientInvocationHandler.kt
```diff
@@ -64,11 +72,34 @@ internal class ClientInvocationHandler(
...
+        headerValue.toLowerCase()
+      } else {
+        null
+      }
+    }.filterNot { it == MediaTypes.APPLICATION_JSON }
```
ryanhall07: why filter this out? because it's the default?

misk/src/main/kotlin/misk/client/ClientInvocationHandler.kt
```diff
@@ -64,11 +72,34 @@ internal class ClientInvocationHandler(
 
     if (tracer != null) retrofitBuilder.callFactory(TracingCallFactory(actionSpecificClient, tracer))
 
+    val mediaTypes = getEndpointMediaTypes(methodName)
+    if (mediaTypes.contains(MediaTypes.APPLICATION_PROTOBUF)) {
+      retrofitBuilder.addConverterFactory(WireConverterFactory.create())
+    }
+    // Always add JSON as default. Ensure it's added last so that other converters get a chance since
```
ryanhall07: what does it mean to allow multiple converters? shouldn't only a single converter be installed?

misk/src/main/kotlin/misk/client/ClientInvocationHandler.kt
nbroyles: It's atypical but allowed. An endpoint is free to send requests in format A and expect a response in format B. Retrofit supports that by allowing multiple converters, which it iterates through in order looking for one that supports the object it's attempting to convert.

misk/src/main/kotlin/misk/client/ClientInvocationHandler.kt
nbroyles: Correct

misk/src/main/kotlin/misk/client/ClientInvocationHandler.kt
nbroyles: Previous comment should clarify this.

PR #194, opened 2018-05-16
Cluster interface and DataSourceCluster bindings

jmuia:
The `Cluster` interface provides a consistent abstraction for interacting with various database-related resources based on their role in the underlying database topology.

Apply that abstraction to the `DataSourceModule`, binding `DataSource`s indirectly via `DataSourceCluster`s.


misk/src/main/kotlin/misk/jdbc/Cluster.kt
```diff
@@ -0,0 +1,17 @@
...
+ *
+ * Database clusters are often composed of a Read/Write master
+ * instance and a set of Read-Only replica instances.
+ *
+ * Examples include a Read-Only [DataSource] or a Read/Write
+ * [SessionFactory] in Hibernate.
+ */
+interface Cluster<T> {
+  fun writer() : T
+  fun reader() : T
```
jmuia: should this be `T?`?

mmihic:
IMO T - can default to the writer if no read specific endpoint available

misk/src/main/kotlin/misk/jdbc/Cluster.kt
```diff
@@ -0,0 +1,17 @@
+package misk.jdbc
+
+/**
+ * [Cluster] provides a consistent abstraction for
+ * interacting with various database-related resources [T]
+ * based on their role in the underlying database topology.
+ *
+ * Database clusters are often composed of a Read/Write master
+ * instance and a set of Read-Only replica instances.
```
ryanhall07: I would make it clear that even if the actual cluster has a set of replicas, Misk leaves it up to the underlying persistence layer to provide a single addressable endpoint that manages connections to the set

misk/src/main/kotlin/misk/jdbc/Cluster.kt
ryanhall07: we should make that clear in the javadoc. I would find that somewhat surprising behavior.

misk/src/main/kotlin/misk/jdbc/DataSourceModule.kt
```diff
@@ -4,47 +4,55 @@ import com.google.inject.Key
...
+        DataSourceModule(datasourceName, Key.get(DataSourceCluster::class.java, annotatedBy))
 
     fun <A : Annotation> create(datasourceName: String, annotatedBy: KClass<A>) =
-        DataSourceModule(datasourceName, Key.get(DataSource::class.java, annotatedBy.java))
+        DataSourceModule(datasourceName, Key.get(DataSourceCluster::class.java, annotatedBy.java))
   }
 
-  private class DataSourceProvider(private val datasourceName: String) : Provider<DataSource> {
-    @Inject lateinit var datasourcesConfig: DataSourcesConfig
+  private class DataSourceClusterProvider(
```
ryanhall07: do you still think we should make it a common practice to bind `T` to the `writer`? for example `@Inject DataSource theWriter`

jmuia: I was thinking about this too. I like this idea. In particular when you don't provide specific `reader` configuration and only care about a single rw database instance.

jmuia: I think I misread your comment yesterday so I want to confirm: are you suggesting the `DataSourceModule` should _also_ bind `T` to the `writer` so that application developers don't need to it themselves?

ryanhall07: yes. basically default to the writer

misk/src/main/kotlin/misk/jdbc/Cluster.kt
jmuia: One thing to consider: if you don't provide a specific `reader` then you'll likely use `@Inject @Exemplar DataSource` directly rather than `@Inject @Exemplar DataSourceCluster` anyways. This saves having to null-check the `reader`.

mmihic:
Do we potentially want different types for reader vs writer? (I’m thinking
ReadingTransacter vs Transacter)

misk/src/main/kotlin/misk/jdbc/Cluster.kt
```diff
@@ -0,0 +1,23 @@
...
+ * Misk defers management of these replicas to the underlying
+ * persistence layer and requires a single addressable endpoint.
+ *
+ * Examples include a Read-Only [DataSource] or a Read/Write
+ * [SessionFactory] in Hibernate.
+ *
+ * N.B. If a Read-Only resource is not provided Misk will
+ * default to the Read/Write resource.
+ */
+interface Cluster<T> {
```
tirsen: So a cluster is like a vitess cluster?

misk/src/main/kotlin/misk/jdbc/Cluster.kt
```diff
@@ -0,0 +1,23 @@
...
+  val writer: T
+  val reader: T
```
tirsen: With Vitess we would not need to have separate readers and writers. You can have a single connection and change whether you're reading from replicas or from the master with a `USE` statement. `USE @replica` makes your connection start doing replica reads, `USE @master` takes you back to the master.

I'm imagining we would instead have methods on the transacter that looks like this:
```
transacter.replicaRead().executeWithBackoff(s -> {
  s.createQuery(...)
  ...
})
```

(Sorry not valid Kotlin, I'm still a noob there.)

And this for master reads:
```
transacter.consistentRead().execute(...)
```

For slow OLAP style queries:
```
transacter.replicaRead().slow().execute(...)
```

And even in the future:
```
transacter.consistentReplicaRead().execute(...)
```
If Vitess implements this: http://www.proxysql.com/blog/proxysql-gtid-causal-reads

/cc @mmihic

mmihic:
A Cluster as defined here is specifically not a Vitess cluster - it is an
unsharded collection of database instances (master, sync/semi-sync standby,
zero or more local read replicas) pointing to the same actual dataset. It
is equivalent to an RDS cluster or ODS cluster.

PR #195, opened 2018-05-24
Add a JPAEntityModule for binding entities for a DataSource

jmuia:
JPA entities can be bound for a DataSource with the JPAEntityModule.
Multiple modules can contribute entities for a particular DataSource
using its qualifying annotation (or lack thereof).

Higher-level abstractions such as Hibernate can use the entities bound
for a particular DataSource as part of its relational mapping.

The annotation `@JvmName("-createKClass")` prevents declaration clashes
on the JVM due to type erasure. Declarations such as `fun create(entities: Set<KClass<*>>)`
and `fun create(entities: Set<Class<*>>)` will have the same signature.

The "-" prefix in the above annotation also "hides" the method from Java developers
as Java does not allow the hyphen even though the JVM does. Discussion here:
https://discuss.kotlinlang.org/t/more-characters-allowed-for-identifiers-than-grammar-specifies-what-is-supported/2359/2

### DIFF
--- a/misk/src/main/kotlin/misk/jdbc/JPAEntity.kt
+++ b/misk/src/main/kotlin/misk/jdbc/JPAEntity.kt
@@ -1,0 +1,7 @@
+package misk.jdbc
+
+/**
+ * [JPAEntity] is a wrapper class that allows for unqualified
+ * binding of JPA entity classes without collision.
+ */
+data class JPAEntity(val entity: Class<*>)
--- a/misk/src/test/kotlin/misk/jdbc/JPAEntityModuleTest.kt
+++ b/misk/src/test/kotlin/misk/jdbc/JPAEntityModuleTest.kt
@@ -1,0 +1,51 @@
+package misk.jdbc
+
+import com.google.inject.Guice
+import misk.inject.getSetOf
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.Test
+import javax.inject.Qualifier
+import kotlin.reflect.KClass
+
+internal class JPAEntityModuleTest {
+
+  @Test fun multipleDataSources() {
+    val dinosaurs = setOf(Triceratops::class, Stegosaurus::class)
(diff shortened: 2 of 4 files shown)


misk/src/test/kotlin/misk/jdbc/JPAEntityModuleTest.kt
```diff
@@ -0,0 +1,58 @@
...
+
+    val dinosaurs = setOf(Triceratops::class, Stegosaurus::class)
+    val shapes = setOf(Square::class, Circle::class)
+
+    assertEntities(injector.getSetOf(JPAEntity::class), dinosaurs)
+    assertEntities(injector.getSetOf(JPAEntity::class, Shapes::class), shapes)
+  }
+
+  @Suppress("UNCHECKED_CAST")
+  fun <T : Any> Injector.getSetOf(
```
ryanhall07: probably useful for others

misk/src/main/kotlin/misk/jdbc/JPAEntity.kt
```diff
@@ -0,0 +1,3 @@
+package misk.jdbc
+
+data class JPAEntity(val entity: Class<*>)
```
ryanhall07: might be worth documenting why this type exists

misk/src/test/kotlin/misk/jdbc/JPAEntityModuleTest.kt
```diff
@@ -0,0 +1,51 @@
...
+import misk.inject.getSetOf
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.Test
+import javax.inject.Qualifier
+import kotlin.reflect.KClass
+
+internal class JPAEntityModuleTest {
+
+  @Test fun multipleDataSources() {
+    val dinosaurs = setOf(Triceratops::class, Stegosaurus::class)
```
swankjesse: yes!

PR #198, opened 2018-05-27
Introduce FakeResourceLoader

swankjesse:
Always inject ResourceLoader, and use either the ResourceLoaderModule
or the FakeResourceLoaderModule.

This is so we don't need a bunch of resource files to test schema
migrations.


misk/src/test/kotlin/misk/jdbc/DataSourceModuleTest.kt
```diff
@@ -153,7 +154,7 @@ internal class DataSourceModuleTest {
         |""".trimMargin()
 
       val DROP_TABLE_PEOPLE = """
-        |DROP TABLE people
+        |DROP TABLE IF EXISTS people
```
tirsen: Ohhhh multi line strings. I LOVE YOU KOTLIN!!!

tirsen:
LGTM

misk/src/main/kotlin/misk/resources/ClasspathResourceLoader.kt
```diff
@@ -0,0 +1,29 @@
...
+import java.util.TreeMap
+
+/** Real implementation of [ResourceLoader] suitable for development and production. */
+internal object ClasspathResourceLoader : ResourceLoader() {
+  private val resourcesByPath: Map<String, ClassPath.ResourceInfo>
+
+  init {
+    val classLoader = ResourceLoader::class.java.classLoader
+    val classPath = ClassPath.from(classLoader)
+    resourcesByPath = TreeMap(classPath.resources.associateBy { r -> r.resourceName })
```
shellderp: use `it`?

misk/src/main/kotlin/misk/resources/ResourceLoaderModule.kt
```diff
@@ -4,6 +4,6 @@ import misk.inject.KAbstractModule
 
 class ResourceLoaderModule : KAbstractModule() {
   override fun configure() {
-    bind<ResourceLoader>().toInstance(ResourceLoader)
+    bind<ResourceLoader>().toInstance(ClasspathResourceLoader)
```
shellderp: this doesn't need `()`? :O

misk/src/main/kotlin/misk/resources/ClasspathResourceLoader.kt
swankjesse: Fix in follow-up!

PR #199, opened 2018-05-28
Implement SchemaMigrator.

swankjesse:
I'm deliberately avoiding adding a more capable dependency like Flyway. This
should be upwards-compatible if ever we need to support multiple databases.

### DIFF
--- a/misk/src/main/kotlin/misk/hibernate/SchemaMigrator.kt
+++ b/misk/src/main/kotlin/misk/hibernate/SchemaMigrator.kt
@@ -1,0 +1,149 @@
+package misk.hibernate
+
+import com.google.common.base.Stopwatch
+import misk.jdbc.DataSourceConfig
+import misk.logging.getLogger
+import misk.resources.ResourceLoader
+import org.hibernate.SessionFactory
+import org.hibernate.query.Query
+import java.sql.Connection
+import java.util.Collections
+import java.util.NavigableMap
+import java.util.NavigableSet
+import java.util.TreeMap
+import java.util.TreeSet
+import java.util.regex.Pattern
+import javax.persistence.PersistenceException
+
+private val logger = getLogger<SchemaMigrator>()
+
+/**
+ * Manages **available** and **applied** schema migrations.
+ *
+ * Available schema migrations are SQL files in the datasource's `migrations_path` directory. Each
+ * file should contain SQL statements terminated by a `;`. The files should be named like
+ * `v100__exemplar.sql` with a `v`, an integer version, two underscores, a description, and the
+ * `.sql` suffix. The integer identifier is the migration version. Versions do not need to be
+ * sequential. They are applied in increasing order.
+ *
+ * Applied schema migrations are tracked in the database in a `schema_version` table. Migrations may
+ * be applied either by [SchemaMigrator.applyAll] or manually. When you applying schema changes
+ * manually you must add a row to the `schema_version` table to record which version was applied.
+ */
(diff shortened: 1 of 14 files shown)


misk/src/main/kotlin/misk/hibernate/SchemaMigrator.kt
```diff
@@ -0,0 +1,147 @@
...
+import javax.persistence.PersistenceException
+
+private val logger = getLogger<SchemaMigrator>()
+
+/**
+ * Manages **available** and **applied** schema migrations.
+ *
+ * Available schema migrations are SQL files in the datasource's `migrations_path` directory. Each
+ * file should contain SQL statements terminated by a `;`. The files should be named like
+ * `v100__exemplar.sql` with a `v`, an integer version, two underscores, a description, and the
```
shellderp: is this a tradition we want to carry forward?

misk/src/main/kotlin/misk/hibernate/SchemaMigrator.kt
```diff
@@ -0,0 +1,147 @@
...
+  fun availableMigrations(): NavigableMap<Int, String> {
+    val resources = ResourceLoader.list(config.migrations_path)
+    return TreeMap(resources.associateBy { resourceVersion(it) })
+  }
+
+  /** Creates the `schema_version` table if it does not exist. Returns the applied migrations. */
+  fun initialize(): NavigableSet<Int> {
+    try {
+      val result = appliedMigrations()
+      logger.info {
```
shellderp: neat

misk/src/main/kotlin/misk/hibernate/SchemaMigrator.kt
```diff
@@ -0,0 +1,149 @@
...
+
+  /**
+   * Returns the versions of applied migrations. Throws a [PersistenceException] if the migrations
+   * table has not been initialized.
+   */
+  fun appliedMigrations(): NavigableSet<Int> {
+    sessionFactory.openSession().use { session ->
+      @Suppress("UNCHECKED_CAST") // createNativeQuery returns a raw Query.
+      val query = session.createNativeQuery("SELECT version FROM schema_version") as Query<String>
+      return TreeSet(query.list().map { it.toInt() })
```
shellderp: here you create a List then a TreeSet, but you could mapTo the TreeSet directly
```
return query.list().mapTo(TreeSet()) { it.toInt() }
```
I'm not picky though

misk/src/main/kotlin/misk/hibernate/SchemaMigrator.kt
swankjesse: I think it’s good enough!

misk/src/main/kotlin/misk/hibernate/SchemaMigrator.kt
swankjesse: I think it does the same thing underneath?

misk/src/main/kotlin/misk/hibernate/SchemaMigrator.kt
eeSeeGee: Any reason why version rather than timestamp?

tirsen: Timestamp in milliseconds would preserve some form of order while still providing for super low probability of conflicts. The only drawback I can see is that the file name becomes rather long!

PR #201, opened 2018-05-28
Drop support for unqualified datasources.

swankjesse:
We don't need 'em.


swankjesse:
Yep. I think if you really want you can do a binding to alias say, `@Movies Transacter` to `Transacter`

ryanhall07:
At first I was also against them, but came to the conclusion that every developer is probably going to create the linked binding, so you might as well provide it.

PR #202, opened 2018-05-28
Don't inject until after services are started.

swankjesse:
This way we can inject a SessionFactory instead of a Provider<SessionFactory>.

### DIFF
--- a/misk/src/test/kotlin/misk/hibernate/RawHibernateApiTest.kt
+++ b/misk/src/test/kotlin/misk/hibernate/RawHibernateApiTest.kt
@@ -18,7 +18,6 @@ import org.hibernate.SessionFactory
 import org.junit.jupiter.api.Test
 import java.util.Date
 import javax.inject.Inject
-import javax.inject.Provider
 
 /** Test that we can access Hibernate's SessionFactory directly. */
 @MiskTest(startService = true)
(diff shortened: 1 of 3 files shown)


misk/src/test/kotlin/misk/hibernate/RawHibernateApiTest.kt
```diff
@@ -38,21 +37,20 @@ class RawHibernateApiTest {
     }
   }
 
-  // TODO(jwilson): Inject the SessionFactory directly.
-  @Inject @Movies lateinit var sessionFactoryProvider: Provider<SessionFactory>
+  @Inject @Movies lateinit var sessionFactory: SessionFactory
```
tirsen: Does this mean we rebuild the SessionFactory before each test or do we reuse one? I think building a SessionFactory is pretty expensive!

misk/testing/src/main/kotlin/misk/testing/MiskTestExtension.kt
```diff
@@ -56,6 +61,19 @@ internal class MiskTestExtension : BeforeEachCallback, AfterEachCallback {
...
+  /** We inject after starting services and uninject after stopping services. */
+  @Singleton
+  class InjectUninject : BeforeEachCallback, AfterEachCallback {
+    override fun beforeEach(context: ExtensionContext) {
+      val injector = context.retrieve<Injector>("injector")
+      injector.injectMembers(context.requiredTestInstance)
+    }
+
+    override fun afterEach(context: ExtensionContext) {
+      uninject(context.requiredTestInstance)
```
tirsen: OOC why do we uninject?

misk/src/test/kotlin/misk/hibernate/RawHibernateApiTest.kt
swankjesse: The first one takes forever because I think it’s scanning the classpath. Subsequent ones are pretty decent. If necessary I’ll do the stuff to cache things between tests.

misk/testing/src/main/kotlin/misk/testing/MiskTestExtension.kt
swankjesse: Mostly because JUnit keeps instances of the test class in memory for the duration of the test run. If the test has nontrivial things injected, that ends up eating all the memory and slowing down a long test suite.

PR #204, opened 2018-05-29
Tidy up some test cases.

swankjesse:
Use assertThrows() where possible.
Fix some IntelliJ warnings around unsorted modifiers.
Prefer @MiskTest when that pattern applies.


misk/src/test/kotlin/misk/MiskApplicationTest.kt
```diff
@@ -39,9 +39,7 @@ internal class MiskApplicationTest {
   }
 
   class WithModule : MiskCommand("with-module", AppNameModule()) {
-    @Inject
-    lateinit
-    var injectedValue: MyInjectedValue
+    @Inject lateinit var injectedValue: MyInjectedValue
```
ryanhall07: can the formatter be updated to enforce this?

PR #205, opened 2018-05-29
Make all Kotlin warnings build errors

Syeberman:
Required fixing/supressing the current warnings from the build.


misk/src/main/kotlin/misk/metrics/backends/stackdriver/StackDriverBackendModule.kt
```diff
@@ -23,7 +23,11 @@ class StackDriverBackendModule : KAbstractModule() {
 
   @Provides
   @Singleton
-  fun monitoring(@AppName appName: String, config: StackDriverBackendConfig): Monitoring =
+  fun monitoring(
+    @AppName appName: String,
+    @Suppress("UNUSED_PARAMETER")
+    config: StackDriverBackendConfig
```
ryanhall07: just remove this?

PR #208, opened 2018-05-30
Fast fail on dependency cycles.

swankjesse:
I want to pass the interview.

### DIFF
--- a/misk/src/main/kotlin/misk/CoordinatedService.kt
+++ b/misk/src/main/kotlin/misk/CoordinatedService.kt
@@ -118,11 +118,46 @@ internal class CoordinatedService(val service: Service) : AbstractService() {
         }
       }
 
+      val validityMap = mutableMapOf<CoordinatedService, CycleValidity>()
+      for (service in coordinatedServices) {
+        val cycle = service.findCycle(validityMap)
+        if (cycle != null) {
+          errors.add("dependency cycle: ${cycle.joinToString(separator = " -> ")}")
+          break
+        }
+      }
+
       require(errors.isEmpty()) {
         "Service dependency graph has problems:\n  ${errors.joinToString(separator = "\n  ")}"
       }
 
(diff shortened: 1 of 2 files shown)


misk/src/main/kotlin/misk/CoordinatedService.kt
```diff
@@ -118,11 +118,46 @@ internal class CoordinatedService(val service: Service) : AbstractService() {
...
         "Service dependency graph has problems:\n  ${errors.joinToString(separator = "\n  ")}"
       }
 
       return ServiceManager(coordinatedServices)
     }
+
+    /** Returns the elements of a cycle, or null if there are no cycles originating at `node`. */
+    fun CoordinatedService.findCycle(
+      validityMap: MutableMap<CoordinatedService, CycleValidity>
+    ): MutableList<CoordinatedService>? {
```
ryanhall07: why nullable? I guess easier to bootstrap

misk/src/main/kotlin/misk/CoordinatedService.kt
```diff
@@ -118,11 +118,46 @@ internal class CoordinatedService(val service: Service) : AbstractService() {
         }
       }
 
+      val validityMap = mutableMapOf<CoordinatedService, CycleValidity>()
+      for (service in coordinatedServices) {
+        val cycle = service.findCycle(validityMap)
+        if (cycle != null) {
+          errors.add("dependency cycle: ${cycle.joinToString(separator = " -> ")}")
```
eeSeeGee: Maybe return some kind of Error class or even a string so that it's not the client's responsibility to traverse the cycle? Or so you see a use for the cycle list?

swankjesse: In this case this is a programmer-facing error message. You should be able to tell that your EggService can’t depend on your ChickenService.

PR #210, opened 2018-05-30
Replace stdlib-jre8 with stdlib-jdk8

drashkov:
jre8 was deprecated for kotlin 1.2 due to changes in java 9 support.


swankjesse:
Thanks! This bit me in IntelliJ today!

PR #211, opened 2018-05-31
Queries in the Misk Hibernate API.

swankjesse:
Currently implemented manually. Will implement with reflection.

### DIFF
--- a/misk/src/main/kotlin/misk/hibernate/Query.kt
+++ b/misk/src/main/kotlin/misk/hibernate/Query.kt
@@ -1,0 +1,24 @@
+package misk.hibernate
+
+import kotlin.reflect.KClass
+
+/** Base class for SQL queries. */
+interface Query<T> {
+  fun uniqueResult(session: Session): T?
+  fun <P : Projection> uniqueResultAs(session: Session, projection: KClass<P>): P?
+
+  fun list(session: Session): List<T>
+  fun <P : Projection> listAs(session: Session, projection: KClass<P>): List<P>
+
+  /** Creates instances of queries. */
+  interface Factory {
+    fun <T : Query<*>> newQuery(queryClass: KClass<T>): T
+  }
+}
+
+inline fun <reified P : Projection> Query<*>.listAs(session: Session) = listAs(session, P::class)
+
+inline fun <reified P : Projection> Query<*>.uniqueResultAs(session: Session) = uniqueResultAs(
+    session, P::class)
+
+inline fun <reified T : Query<*>> Query.Factory.newQuery(): T = newQuery(T::class)
--- a/misk/src/test/kotlin/misk/hibernate/ManualCharacterQuery.kt
+++ b/misk/src/test/kotlin/misk/hibernate/ManualCharacterQuery.kt
@@ -1,0 +1,165 @@
+package misk.hibernate
+
+import misk.hibernate.TransacterTest.ActorAndReleaseDate
+import misk.hibernate.TransacterTest.CharacterQuery
+import misk.hibernate.TransacterTest.MovieNameAndReleaseDate
+import java.time.LocalDate
+import javax.inject.Singleton
+import javax.persistence.criteria.CompoundSelection
+import javax.persistence.criteria.CriteriaBuilder
+import javax.persistence.criteria.Path
+import javax.persistence.criteria.Predicate
+import javax.persistence.criteria.Root
+import kotlin.reflect.KClass
+
+// TODO(jwilson): build this with metaprogramming (reflection/codegen) instead of code.
+class ManualCharacterQuery : CharacterQuery {
+  private val constraints = mutableListOf<ConstraintSpec>()
+
+  private fun addConstraint(path: List<String>, predicate: PredicateSpec): CharacterQuery {
+    constraints.add(ConstraintSpec(path, predicate))
+    return this
+  }
+
+  override fun name(name: String) = addConstraint(
+      path = listOf("name"),
+      predicate = PredicateSpec.Eq(name))
+
+  override fun actorName(name: String) = addConstraint(
+      path = listOf("actor", "name"),
+      predicate = PredicateSpec.Eq(name))
+
+  override fun movieName(name: String) = addConstraint(
+      path = listOf("movie", "name"),
+      predicate = PredicateSpec.Eq(name))
+
+  override fun movieReleaseDateBefore(upperBound: LocalDate) = addConstraint(
+      path = listOf("movie", "release_date"),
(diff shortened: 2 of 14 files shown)


misk/src/main/kotlin/misk/hibernate/Query.kt
```diff
@@ -0,0 +1,24 @@
+package misk.hibernate
+
+import kotlin.reflect.KClass
+
+/** Base class for SQL queries. */
+interface Query<T> {
+  fun uniqueResult(session: Session): T?
+  fun <P : Projection> uniqueResultAs(session: Session, projection: KClass<P>): P?
```
ryanhall07: should there be a method that does not return a nullable? I think a common pattern is just throwing a not found exception

misk/src/test/kotlin/misk/hibernate/ManualCharacterQuery.kt
```diff
@@ -0,0 +1,165 @@
...
+import java.time.LocalDate
+import javax.inject.Singleton
+import javax.persistence.criteria.CompoundSelection
+import javax.persistence.criteria.CriteriaBuilder
+import javax.persistence.criteria.Path
+import javax.persistence.criteria.Predicate
+import javax.persistence.criteria.Root
+import kotlin.reflect.KClass
+
+// TODO(jwilson): build this with metaprogramming (reflection/codegen) instead of code.
```
ryanhall07: we must go deeper :)

misk/src/test/kotlin/misk/hibernate/ManualCharacterQuery.kt
```diff
@@ -0,0 +1,165 @@
...
+      predicate = PredicateSpec.Lt(upperBound))
+
+  override fun uniqueResult(session: Session): DbCharacter? {
+    // TODO(jwilson): max results = 2
+    val list = list(session)
+    require(list.size <= 1) { "expected at most 1 result but was $list" }
+    return list.firstOrNull()
+  }
+
+  override fun <P : Projection> uniqueResultAs(session: Session, projection: KClass<P>): P? {
```
ryanhall07: can this just be a default impl on the interface? doesn't seem specific to the entity

misk/src/test/kotlin/misk/hibernate/ManualCharacterQuery.kt
ryanhall07: I wonder if we could do something cool and use the indexes of the schema to generate allowable where clauses.

misk/src/test/kotlin/misk/hibernate/TransacterTest.kt
```diff
@@ -13,23 +14,83 @@ class TransacterTest {
...
+  annotation class Constraint(
+    val path: String,
+    val operator: String = "="
+  )
+
+  annotation class Property(
+    val value: String
+  )
+
+  interface CharacterQuery : Query<DbCharacter> {
```
ryanhall07: should this live outside of the test?

misk/src/test/kotlin/misk/hibernate/TransacterTest.kt
```diff
@@ -13,23 +14,83 @@ class TransacterTest {
...
+    @Constraint("movie.release_date")
```
ryanhall07: I don't follow how these annotations are used

swankjesse: when I rebuild the implementation to use reflection they will be used

misk/src/test/kotlin/misk/hibernate/TransacterTest.kt
swankjesse: tough. The point of the test is to confirm that the query bindings work.

misk/src/test/kotlin/misk/hibernate/TransacterTest.kt
ryanhall07: ah. spent way too long trying to figure that one out :)

misk/src/test/kotlin/misk/hibernate/ManualCharacterQuery.kt
swankjesse: we can, but it’s an unbounded set

misk/src/main/kotlin/misk/hibernate/Query.kt
swankjesse: I’m trying to make it as easy as possible to migrate from Service Container, which returns null if there’s nothing found.

swankjesse: `!!` is good syntax to say “this is definitely here”

misk/src/test/kotlin/misk/hibernate/ManualCharacterQuery.kt
swankjesse: possibly? I’m reluctant to use default methods but that’s not well motivated

misk/src/test/kotlin/misk/hibernate/ManualCharacterQuery.kt
ryanhall07: i'm thinking we separate filter building into 2 steps, indexed and unindexed. you must pick an available index filter and then can add an additional unbounded set of unindexed filters.

we could even promote the usage of covered indices by combining indexed fields with a projection.

misk/src/test/kotlin/misk/hibernate/ManualCharacterQuery.kt
ryanhall07: probably better to start out without them until we see the same pattern repeating over and over again

misk/src/main/kotlin/misk/hibernate/Query.kt
tirsen: There's been plenty of bugs where we've just assumed we get something out of `uniqueResult` so I think having a reminder to think "what if this is not here?" is a good thing.

misk/src/main/kotlin/misk/hibernate/Query.kt
```diff
@@ -0,0 +1,24 @@
...
+
+  /** Creates instances of queries. */
+  interface Factory {
+    fun <T : Query<*>> newQuery(queryClass: KClass<T>): T
+  }
+}
+
+inline fun <reified P : Projection> Query<*>.listAs(session: Session) = listAs(session, P::class)
+
+inline fun <reified P : Projection> Query<*>.uniqueResultAs(session: Session) = uniqueResultAs(
```
tirsen: YEAH!

PR #212, opened 2018-06-01
Implement Query with dynamic proxies and reflection

swankjesse:


misk/src/main/kotlin/misk/hibernate/ReflectionQueryFactory.kt
```diff
@@ -0,0 +1,257 @@
...
+        // TODO(jwilson): validate & test validation more aggressively:
+        //  * the return type must be 'this'
+        //  * there should be the required number of parameters
+        //  * there should be a @Constraint annotation
+        //  * there should be a Java method
+        //  * the operator should be known
+
+        val path = constraint.path.split('.')
+        val handler = when {
+          constraint.operator == "=" -> object : QueryMethodHandler {
```
ryanhall07: neat

misk/src/test/kotlin/misk/hibernate/TransacterTest.kt
```diff
@@ -61,23 +61,14 @@ class TransacterTest {
...
-    val value: String
-  )
-
   interface CharacterQuery : Query<DbCharacter> {
-    @Constraint("movie.release_date")
+    @Constraint("name")
     fun name(name: String): CharacterQuery
 
-    @Constraint("movie.release_date")
+    @Constraint("actor.name")
```
ryanhall07: nice :)

misk/src/main/kotlin/misk/hibernate/Query.kt
```diff
@@ -22,3 +22,20 @@ inline fun <reified P : Projection> Query<*>.uniqueResultAs(session: Session) =
...
 
 inline fun <reified T : Query<*>> Query.Factory.newQuery(): T = newQuery(T::class)
+
+/**
+ * Annotations a function on a subinterface of [Query] to indicate which column (or path of columns)
+ * it constrains and using which operator.
+ */
+annotation class Constraint(
+  val path: String,
+  val operator: String = "="
```
ryanhall07: should probably document the supported operators

misk/src/main/kotlin/misk/hibernate/ReflectionQueryFactory.kt
```diff
@@ -0,0 +1,257 @@
...
+ */
+internal class ReflectionQuery<T : DbEntity<T>>(
+  private val factory: Factory,
+  private val rootEntityType: KClass<T>,
+  private val queryMethodHandlers: Map<Method, QueryMethodHandler>
+) : Query<T>, InvocationHandler {
+  private val constraints = mutableListOf<ConstraintSpec>()
+
+  override fun uniqueResult(session: Session): T? {
+    // TODO(jwilson): max results = 2
```
ryanhall07: this is to put a limit of 2 in the query for safety?

misk/src/main/kotlin/misk/hibernate/Query.kt
swankjesse: Yeah, Only supporting `=` and `<` is probably insufficient!

misk/src/main/kotlin/misk/hibernate/ReflectionQueryFactory.kt
swankjesse: yeah, so if the DB returns 1 result you’re good, and if it returns more than one you know your invariant was off

PR #213, opened 2018-06-01
Log the reasons why liveness/readiness checks fail

Syeberman:


misk/src/main/kotlin/misk/eventrouter/KubernetesClusterConnector.kt
```diff
@@ -35,13 +35,16 @@ internal class KubernetesClusterConnector : ClusterConnector {
...
+    }
     val sinceLastReceived = Duration.between(lastReceivedMessage, clock.instant()).seconds
     if (sinceLastReceived > config.kubernetes_read_timeout + config.kubernetes_connect_timeout) {
       return HealthStatus.unhealthy(
-          "I haven't received an update in $sinceLastReceived seconds.")
+          "k8s: I haven't received an update in $sinceLastReceived seconds.")
     }
-    return HealthStatus.healthy("I received a message $sinceLastReceived seconds ago.")
+    return HealthStatus.healthy(
+        "k8s: I received a message $sinceLastReceived seconds ago.")
```
ryanhall07: nit: why the `k8s` prefix? The logger name should help you search for these log message

misk/src/main/kotlin/misk/web/actions/LivenessCheckAction.kt
```diff
@@ -17,11 +21,17 @@ class LivenessCheckAction @Inject internal constructor(
...
+    }
+
+    if (failedServices.isEmpty()) {
+      return Response("", statusCode = 200)
     }
 
-    // TODO(jgulbronson) - Should return an empty body
-    return Response("", statusCode = if (isAlive) 200 else 503)
+    for (service in failedServices) {
+      logger.error("Service failed: $service")
```
swankjesse: this is just the service , not the reason?

misk/src/main/kotlin/misk/web/actions/ReadinessCheckAction.kt
```diff
@@ -18,13 +21,21 @@ class ReadinessCheckAction @Inject internal constructor(
...
-    // TODO(jgulbronson) - Should return an empty body
-    return Response(
-        "",
-        statusCode = if (isReady) 200 else 503
-    )
+    for (service in servicesNotRunning) {
+      logger.error("Service not running: $service")
+    }
+    for (healthCheck in failedHealthChecks) {
+      logger.error("Failed health check: ${healthCheck.messages}")
```
ryanhall07: it seems like you should get this for free when a `Action` fails. @mmihic thoughts?

samples/exemplarchat/src/main/kotlin/com/squareup/chat/healthchecks/ManualHealthCheck.kt
```diff
@@ -9,7 +9,11 @@ class ManualHealthCheck : HealthCheck {
   private var healthy = true
 
   override fun status(): HealthStatus {
-    return if (healthy) HealthStatus.healthy() else HealthStatus.unhealthy()
+    return if (healthy) {
```
mmihic: why the reformat?

misk/src/main/kotlin/misk/web/actions/ReadinessCheckAction.kt
mmihic: What should you get for free?

ryanhall07: if an action fails, misk logs the response for you. that would require adding the messages into the response, which seems reasonable?

misk/src/main/kotlin/misk/eventrouter/KubernetesClusterConnector.kt
Syeberman: The logger/class name isn't stored in `HealthStatus` or `HealthCheck`.  I could add that.

misk/src/main/kotlin/misk/web/actions/ReadinessCheckAction.kt
Syeberman: The current actions do not log on failure, they only return a `HealthStatus`.  We could update `HealthStatus.unhealthy` to automatically log...

samples/exemplarchat/src/main/kotlin/com/squareup/chat/healthchecks/ManualHealthCheck.kt
Syeberman: Line length.

misk/src/main/kotlin/misk/web/actions/LivenessCheckAction.kt
Syeberman: For some services, it's the service and state, for example `AbstractExecutionThreadService`:

```java
 public String toString() {
    return serviceName() + " [" + state() + "]";
  }
```

This isn't guaranteed by the `Service` interface, though.  Also, `serviceName()` isn't part of the interface either, so `toString()` is the only option available to get the name.

For a more detailed reason we could use `Service.failureCause`...but that throws if the service state isn't `FAILED`, so I'd need to write a helper method.

PR #216, opened 2018-06-04
URL shortener sample

swankjesse:
This is our first sample that uses Hibernate

### DIFF
--- a/samples/urlshortener/src/main/kotlin/com/squareup/urlshortener/UrlShortenerModule.kt
+++ b/samples/urlshortener/src/main/kotlin/com/squareup/urlshortener/UrlShortenerModule.kt
@@ -1,0 +1,28 @@
+package com.squareup.urlshortener
+
+import misk.MiskModule
+import misk.config.ConfigModule
+import misk.config.MiskConfig
+import misk.environment.Environment
+import misk.environment.EnvironmentModule
+import misk.hibernate.HibernateEntityModule
+import misk.hibernate.HibernateModule
+import misk.inject.KAbstractModule
+import misk.resources.ResourceLoaderModule
+
+/** Binds dependencies for all environments. */
+class UrlShortenerModule(val environment: Environment) : KAbstractModule() {
+  override fun configure() {
+    val config = MiskConfig.load<UrlShortenerConfig>("urlshortener", environment)
+    install(ConfigModule.create("urlshortener", config))
+
+    install(MiskModule())
+    install(ResourceLoaderModule())
+    install(EnvironmentModule(environment))
+
+    bind(UrlStore::class.java).to(RealUrlStore::class.java)
+
+    install(HibernateEntityModule(UrlShortener::class, listOf(DbShortenedUrl::class)))
+    install(HibernateModule(UrlShortener::class, config.data_source_cluster.writer))
+  }
+}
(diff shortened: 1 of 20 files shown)


samples/urlshortener/src/main/kotlin/com/squareup/urlshortener/UrlShortenerModule.kt
```diff
@@ -0,0 +1,28 @@
...
+import misk.resources.ResourceLoaderModule
+
+/** Binds dependencies for all environments. */
+class UrlShortenerModule(val environment: Environment) : KAbstractModule() {
+  override fun configure() {
+    val config = MiskConfig.load<UrlShortenerConfig>("urlshortener", environment)
+    install(ConfigModule.create("urlshortener", config))
+
+    install(MiskModule())
+    install(ResourceLoaderModule())
```
ryanhall07: I wonder if `ResourceLoaderModule` and `EnvironmentModule` should be in `MiskModule`?

samples/urlshortener/src/main/kotlin/com/squareup/urlshortener/UrlShortenerServiceModule.kt
```diff
@@ -0,0 +1,20 @@
+package com.squareup.urlshortener
+
+import com.google.inject.AbstractModule
+import misk.environment.Environment
+import misk.web.WebActionModule
+import misk.web.WebModule
+import misk.web.actions.DefaultActionsModule
+
+/** Binds all service dependencies including service-specific dependencies. */
+class UrlShortenerServiceModule : AbstractModule() {
```
ryanhall07: why this and `UrlShortenerModule`?

ryanhall07: my guess is to support testing? we probably want to give some guidance on how to structure modules

samples/urlshortener/src/main/kotlin/com/squareup/urlshortener/UrlShortenerModule.kt
swankjesse: This needs work. MiskModule is an awkward catch-all.

I think maybe we restructure modules to be task-focused first (ResourceLoaderModule, EnvironmentModule, ServiceManagerModule) and then offer convenient aggregates (MiskServiceModule, MiskTestModule) ? Something smarter than what we have now.

samples/urlshortener/src/main/kotlin/com/squareup/urlshortener/UrlShortenerServiceModule.kt
swankjesse: Definitely. And I think we need examples like this so that our advice is credible.

samples/urlshortener/src/main/kotlin/com/squareup/urlshortener/UrlShortenerModule.kt
ryanhall07: sure, I just thought `MiskModule` was the convenient aggregate.

swankjesse: Maybe? It might need a new name

PR #217, opened 2018-06-05
Support more operators in Query

swankjesse:


misk/src/test/kotlin/misk/hibernate/ReflectionQueryFactoryTest.kt
```diff
@@ -0,0 +1,178 @@
...
+      session.save(DbMovie(m1.name, m1.releaseDate, clock.instant()))
+      session.save(DbMovie(m2.name, m2.releaseDate, clock.instant()))
+      session.save(DbMovie(m3.name, m3.releaseDate, clock.instant()))
+      session.save(DbMovie(m4.name, m4.releaseDate, clock.instant()))
+      session.save(DbMovie(m5.name, m5.releaseDate, clock.instant()))
+      session.save(DbMovie(m98.name, m98.releaseDate, clock.instant()))
+      session.save(DbMovie(m99.name, m99.releaseDate, clock.instant()))
+
+      assertThat(queryFactory.newQuery<MovieQuery>()
+          .releaseDateLessThan(null)
```
ryanhall07: this seems buggy. why allow null values if the user is not explicitly looking for null?

swankjesse: Yeah, I’m torn between being consistent with what Hibernate does (nulls aren’t special) and avoiding surprising operations (`release_date = NULL` is different from `release_date IS NULL`).

swankjesse: I think we get good behavior “for free” by making calling code opt-into nullable types in their parameters. Most queries won’t declare a `?` on the parameter type:

```
    @Constraint(path = "release_date", operator = Operator.LE)
    fun releaseDateLessThanOrEqualTo(upperBound: LocalDate?): MovieQuery
```

PR #239, opened 2018-06-06
Switch tests to MySQL

drashkov:
first pass will fail, but it'll also test new TravisCI config.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,3 +16,5 @@ cache:
     - $HOME/.gradle/caches/
     - $HOME/.gradle/wrapper/
 
+services:
+  - mysql
--- a/build.gradle
+++ b/build.gradle
@@ -61,6 +61,7 @@ ext {
           slf4jApi             : 'org.slf4j:slf4j-api:1.7.25',
           mockitoCore          : 'org.mockito:mockito-core:2.8.9',
           jaxbApi              : 'javax.xml.bind:jaxb-api:2.3.0',
+          mysql                : 'mysql:mysql-connector-java:5.1.16',
   ]
 
   isCi = "true".equals(System.getenv('CI'))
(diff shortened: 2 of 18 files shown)


.travis.yml
```diff
@@ -16,3 +16,5 @@ cache:
     - $HOME/.gradle/caches/
     - $HOME/.gradle/wrapper/
 
+services:
+  - mysql
```
swankjesse: that’s amazing

build.gradle
```diff
@@ -61,6 +61,7 @@ ext {
           slf4jApi             : 'org.slf4j:slf4j-api:1.7.25',
           mockitoCore          : 'org.mockito:mockito-core:2.8.9',
           jaxbApi              : 'javax.xml.bind:jaxb-api:2.3.0',
+          mysql                : 'mysql:mysql-connector-java:5.1.16',
```
swankjesse: latest is 8.0.11 . . . any reason to not use the latest?

misk/src/main/kotlin/misk/jdbc/DataSourceConfig.kt
```diff
@@ -1,28 +1,31 @@
...
+        val testing = env == Environment.TESTING || env == Environment.DEVELOPMENT
+        val queryParams = if (testing) "?createDatabaseIfNotExist=true" else ""
+        "jdbc:mysql://$host:$port/$database$queryParams"
       }
   ),
   HSQLDB(
       driverClassName = "org.hsqldb.jdbcDriver",
       hibernateDialect = "org.hibernate.dialect.H2Dialect",
-      buildJdbcUrl = { config ->
+      buildJdbcUrl = { config, _ ->
```
swankjesse: neat

PR #241, opened 2018-06-07
Rollback transactions on exceptions

swankjesse:


misk/src/main/kotlin/misk/hibernate/RealTransacter.kt
```diff
@@ -7,13 +7,24 @@ internal class RealTransacter(
...
+      try {
+        result = lambda(realSession)
+        transaction.commit()
+        return result
+      } catch (e: Throwable) {
+        if (transaction.isActive) {
+          try {
+            transaction.rollback()
+          } catch (suppressed: Exception) {
+            e.addSuppressed(suppressed)
```
ryanhall07: TIL

misk/src/test/resources/misk/hibernate/migrations/v1__movies.sql
```diff
@@ -1,6 +1,7 @@
 CREATE TABLE movies(
   id bigint(20) NOT NULL PRIMARY KEY AUTO_INCREMENT,
-  name varchar(255) NOT NULL,
+  name varchar(191) NOT NULL,
```
ryanhall07: why?

swankjesse: https://dev.mysql.com/doc/refman/8.0/en/innodb-restrictions.html

> The index key prefix length limit is 767 bytes for InnoDB tables that use REDUNDANT or COMPACT row format. For example, you might hit this limit with a column prefix index of more than 191 characters on a TEXT or VARCHAR column, assuming a utf8mb4 character set and the maximum of 4 bytes for each character.

PR #243, opened 2018-06-07
Misk containers should not run as root

Syeberman:


samples/exemplar/Dockerfile
```diff
@@ -1,4 +1,9 @@
 FROM gcr.io/google-appengine/openjdk
 COPY ./build/libs /usr/local/exemplar
 WORKDIR /usr/local/exemplar
+
+RUN groupadd -g 999 appuser && \
+    useradd -r -u 999 -g appuser appuser
```
Syeberman: I'm open to other names or uids/gids.  I have no preference.

samples/exemplar/Dockerfile
```diff
@@ -1,4 +1,9 @@
...
+USER 999
```
Syeberman: k8s `RunAsNonRoot` requires this to be numeric so it can verify that it is not root.

Syeberman:
I believe they can create and `chown` an in-container tmp directory before the `USER` line in the Dockerfile.

mmihic:
👍 to this

PR #244, opened 2018-06-07
DbTimestampedEntity

swankjesse:
This builds on our own event listeners.

### DIFF
--- a/misk/src/main/kotlin/misk/hibernate/DbTimestampedEntity.kt
+++ b/misk/src/main/kotlin/misk/hibernate/DbTimestampedEntity.kt
@@ -1,0 +1,54 @@
+package misk.hibernate
+
+import java.time.Instant
+
+/**
+ * This complements [DbEntity] for entities that keep updated_at and created_at timestamps.
+ *
+ * Misk uses the application's clock time to set [created_at] and [updated_at]. We prefer the
+ * application's (ie. JVM's) clock over the database's clock because only the application's clock
+ * can be faked in tests.
+ *
+ * To use timestamped entities first add two columns in your `CREATE TABLE` statement. You may also
+ * add an optional index on `updated_at`:
+ *
+ * ```
+ * CREATE TABLE movies(
+ *   id bigint(20) NOT NULL PRIMARY KEY AUTO_INCREMENT,
+ *   created_at timestamp(3) NOT NULL DEFAULT NOW(3),
+ *   updated_at timestamp(3) NOT NULL DEFAULT NOW(3) ON UPDATE NOW(3),
+ *   ...
+ *   KEY `idx_updated_at` (`updated_at`),
+ * );
+ * ```
+ *
+ * The above SQL uses `timestamp(3)` and `NOW(3)` to get millisecond precision which interoperates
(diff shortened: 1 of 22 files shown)


misk/src/main/kotlin/misk/hibernate/HibernateModule.kt
```diff
@@ -77,5 +78,16 @@ class HibernateModule(
...
+
+    install(object : HibernateEntityModule(qualifier) {
+      override fun configureHibernate() {
+        bindListener(EventType.PRE_INSERT).to(TimestampListener::class.java)
+        bindListener(EventType.PRE_UPDATE).to(TimestampListener::class.java)
+      }
+    })
   }
+
+  override fun equals(other: Any?) = other is HibernateModule && other.qualifier == qualifier
```
ryanhall07: this is to allow multiple install calls? why does that happen?

misk/src/main/kotlin/misk/hibernate/DbTimestampedEntity.kt
```diff
@@ -0,0 +1,54 @@
...
+ *   override lateinit var created_at: Instant
+ *
+ *   ...
+ *
+ * }
+ * ```
+ *
+ * The timestamp listener will automatically populate these values on save and update.
+ */
+interface DbTimestampedEntity {
```
ryanhall07: do you think there is any value in splitting this interface up into mutable and immutable entities? I always found it unfortunate that an immutable entity was forced to have an `updated_at` field

misk/src/main/kotlin/misk/hibernate/HibernateModule.kt
swankjesse: I added it when I was trying to diagnose why I was getting Hibernate errors. But I totally don’t need it. I’ll remove.

misk/src/main/kotlin/misk/hibernate/DbTimestampedEntity.kt
swankjesse: Yeah, in this case I’m not quite ready to add that flexibility. In the spirit of being opinionated I want to have fewer types so our users have fewer decisions.

If you really need it, on immutable entities it’s very easy to set your own created_at date.

ryanhall07: yea good point!

misk/src/main/kotlin/misk/hibernate/DbTimestampedEntity.kt
```diff
@@ -0,0 +1,54 @@
+package misk.hibernate
+
+import java.time.Instant
+
+/**
+ * This complements [DbEntity] for entities that keep updated_at and created_at timestamps.
+ *
+ * Misk uses the application's clock time to set [created_at] and [updated_at]. We prefer the
+ * application's (ie. JVM's) clock over the database's clock because only the application's clock
+ * can be faked in tests.
```
tirsen: This isn't necessarily true... We could potentially do something here at the Vitess level...

swankjesse: yeah, but that’s also pretty awkward to hook into FakeClock, right?

PR #250, opened 2018-06-11
Support IN clauses in the query API

swankjesse:
Closes: https://github.com/square/misk/issues/246

### DIFF
--- a/misk/src/main/kotlin/misk/hibernate/ReflectionQueryFactory.kt
+++ b/misk/src/main/kotlin/misk/hibernate/ReflectionQueryFactory.kt
@@ -15,6 +15,7 @@ import javax.persistence.criteria.Predicate
 import javax.persistence.criteria.Root
 import kotlin.reflect.KClass
 import kotlin.reflect.KFunction
+import kotlin.reflect.KParameter
 import kotlin.reflect.full.allSupertypes
 import kotlin.reflect.full.declaredMemberFunctions
 import kotlin.reflect.full.findAnnotation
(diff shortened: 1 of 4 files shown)


misk/src/main/kotlin/misk/hibernate/ReflectionQueryFactory.kt
```diff
@@ -293,6 +294,37 @@ internal class ReflectionQuery<T : DbEntity<T>>(
...
+                return
+              }
+            }
+
+            object : QueryMethodHandler {
+              override fun invoke(query: ReflectionQuery<*>, args: Array<out Any>) {
+                query.addConstraint { root, builder ->
+                  val collection = argToCollection(args[0])
+                  if (collection.isEmpty()) {
+                    // The JDBC API forbids empty IN clauses so we use `WHERE 1 = 0` instead to
```
ryanhall07: anyway to not execute the query?

misk/src/test/kotlin/misk/hibernate/ReflectionQueryFactoryTest.kt
```diff
@@ -143,6 +143,69 @@ class ReflectionQueryFactoryTest {
...
+      session.save(DbMovie(m1.name, m1.releaseDate))
+      session.save(DbMovie(m2.name, m2.releaseDate))
+
+      assertThat(queryFactory.newQuery<OperatorsMovieQuery>()
+          .releaseDateInVararg(m1.releaseDate, null)
+          .listAs<NameAndReleaseDate>(session))
+          .containsExactly(m1)
+
+      assertThat(queryFactory.newQuery<OperatorsMovieQuery>()
+          .releaseDateInCollection(listOf(m1.releaseDate, null))
```
ryanhall07: so nulls are just filtered out? I don't see where that is happening

misk/src/main/kotlin/misk/hibernate/ReflectionQueryFactory.kt
swankjesse: Not really. This code will eventually compose and you might say `WHERE name IN () OR release_date < '2018-01-01'` and the IN clause reduces to false but not to an empty query.

misk/src/test/kotlin/misk/hibernate/ReflectionQueryFactoryTest.kt
swankjesse: I'm mostly just testing MySQL at this point. I think both with this and with the comparison operators it's worth writing documentation to warn how NULL will bite you in SQL.

I definitely don't want to change the semantics though; that makes mechanical porting extremely dangerous.

ryanhall07: ah got it.

misk/src/main/kotlin/misk/hibernate/ReflectionQueryFactory.kt
ryanhall07: true true

PR #251, opened 2018-06-12
Implement support for single-column projections.

swankjesse:
This changes the way that projections are accessed. Previously we had
generic listAs() and uniqueResultAs() functions that accepted a type
parameter. With this change each query class exposes its supported
projections.

There are two small benefits here:
 - we can reuse projections. Whether we get the movie NameAndReleaseDate
   from a movie query or a character query, both work because we can
   combine paths.
 - Projections are now linked to the query classes, which should make it
   possible to statically check types in an annotation processor.

Closes: https://github.com/square/misk/issues/247

### DIFF
--- a/misk/src/main/kotlin/misk/hibernate/ReflectionQueryFactory.kt
+++ b/misk/src/main/kotlin/misk/hibernate/ReflectionQueryFactory.kt
@@ -2,6 +2,7 @@ package misk.hibernate
 
 import com.google.common.cache.CacheBuilder
 import com.google.common.cache.CacheLoader
+import com.google.inject.TypeLiteral
 import misk.inject.typeLiteral
 import java.lang.reflect.InvocationHandler
 import java.lang.reflect.InvocationTargetException
@@ -16,6 +17,7 @@ import javax.persistence.criteria.Root
 import kotlin.reflect.KClass
 import kotlin.reflect.KFunction
 import kotlin.reflect.KParameter
+import kotlin.reflect.KType
 import kotlin.reflect.full.allSupertypes
 import kotlin.reflect.full.declaredMemberFunctions
 import kotlin.reflect.full.findAnnotation
@@ -27,7 +29,6 @@ import kotlin.reflect.jvm.javaMethod
  * reflection to call the primary constructor.
  */
 internal class ReflectionQuery<T : DbEntity<T>>(
-  private val factory: Factory,
   private val rootEntityType: KClass<T>,
   private val queryMethodHandlers: Map<Method, QueryMethodHandler>
 ) : Query<T>, InvocationHandler {
@@ -36,13 +37,6 @@ internal class ReflectionQuery<T : DbEntity<T>>(
   override fun uniqueResult(session: Session): T? {
     // TODO(jwilson): max results = 2
     val list = list(session)
-    require(list.size <= 1) { "expected at most 1 result but was $list" }
-    return list.firstOrNull()
-  }
-
-  override fun <P : Projection> uniqueResultAs(session: Session, projection: KClass<P>): P? {
-    // TODO(jwilson): max results = 2
-    val list = listAs(session, projection)
     require(list.size <= 1) { "expected at most 1 result but was $list" }
     return list.firstOrNull()
   }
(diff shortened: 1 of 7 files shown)


misk/src/test/kotlin/misk/hibernate/ReflectionQueryFactoryTest.kt
```diff
@@ -236,6 +272,15 @@ class ReflectionQueryFactoryTest {
 
     @Constraint(path = "release_date", operator = Operator.IS_NULL)
     fun releaseDateIsNull(): OperatorsMovieQuery
+
+    @Select
+    fun listAsNameAndReleaseDate(session: Session): List<NameAndReleaseDate>
+
+    @Select("name")
+    fun uniqueName(session: Session): String?
```
jmuia: Thoughts on validating single-column projections as nullable?

If the return type were `String` you could write Kotlin code like `queryFactory.newQuery<OperatorsMovieQuery>().uniqueName(session).capitalize()` and it would compile fine even though the name could be `null`.

misk/src/main/kotlin/misk/hibernate/ReflectionQueryFactory.kt
```diff
@@ -59,27 +53,13 @@ internal class ReflectionQuery<T : DbEntity<T>>(
...
   override fun invoke(proxy: Any, method: Method, args: Array<out Any>?): Any? {
-    val constraintHandler = queryMethodHandlers[method]
+    val handler = queryMethodHandlers[method]
 
     return when {
-      constraintHandler != null -> {
-        constraintHandler.invoke(this, args ?: arrayOf())
-        proxy
+      handler != null -> {
+        val result = handler.invoke(this, args ?: arrayOf())
```
jmuia: in practice, the return type here is either a Query or the final type (eg. List<Something>, SomeProjection, etc.), right?

misk/src/main/kotlin/misk/hibernate/ReflectionQueryFactory.kt
```diff
@@ -144,79 +120,75 @@ internal class ReflectionQuery<T : DbEntity<T>>(
...
-              "\n  ${errors.joinToString(separator = "\n  ")}"
-        }
+    private fun select(
+      criteriaBuilder: CriteriaBuilder,
+      queryRoot: Root<*>
+    ) = criteriaBuilder.array(*properties.map { queryRoot.traverse<Any?>(it) }.toTypedArray())
 
-        return ProjectionHandler(constructor!!, properties)
+    private fun toValue(row: Any): Any? {
+      return when {
```
jmuia: for custom types (eg. `IdType`) will the row here already be the correct type due to Hibernate?

misk/src/main/kotlin/misk/hibernate/ReflectionQueryFactory.kt
```diff
@@ -144,79 +120,75 @@ internal class ReflectionQuery<T : DbEntity<T>>(
...
           return
         }
 
+        val select = function.findAnnotation<Select>()
+        if (select != null) {
+          createSelect(errors, function, result, select)
+          return
+        }
+
+        errors.add("${function.name}() must be annotated @Constraint or @Select")
```
jmuia: maybe check for multiple annotations?

misk/src/main/kotlin/misk/hibernate/ReflectionQueryFactory.kt
```diff
@@ -343,11 +315,81 @@ internal class ReflectionQuery<T : DbEntity<T>>(
...
+        val selectMethodHandler: SelectMethodHandler
+        if (isProjection) {
+          @Suppress("UNCHECKED_CAST") // The line above confirms that this cast is safe.
+          val projectionClass = (elementType.rawType as Class<out Projection>).kotlin
+          val constructor = projectionClass.primaryConstructor
+          val parameters = if (constructor != null) {
+            constructor.parameters
+          } else {
+            errors.add("${projectionClass.java.name} has no primary constructor")
+            listOf()
```
jmuia: should this just return early?

misk/src/test/kotlin/misk/hibernate/ReflectionQueryFactoryValidationTest.kt
```diff
@@ -68,48 +68,63 @@ class ReflectionQueryFactoryValidationTest {
...
-      transacter.transaction { session ->
-        queryFactory.newQuery<EmptyQuery>().listAs<ParameterAnnotationRequiredOnProjection>(session)
+      transacter.transaction {
+        queryFactory.newQuery<ParameterAnnotationRequiredOnProjectionQuery>()
       }
     }.cause).hasMessage("""
-        |Projection class ${ParameterAnnotationRequiredOnProjection::class.java.name} has problems:
-        |  parameter 0 is missing a @Property annotation""".trimMargin())
+        |Query class ${ParameterAnnotationRequiredOnProjectionQuery::class.java.name} has problems:
+        |  ${ParameterAnnotationRequiredOnProjection::class.java.name} parameter 0 is missing a @Property annotation""".trimMargin())
```
jmuia: nit: this line is long

misk/src/test/kotlin/misk/hibernate/ReflectionQueryFactoryTest.kt
swankjesse: ooh, good idea. As these are general-purpose they’ll almost always need to be nullable. Lets enforce that.

misk/src/main/kotlin/misk/hibernate/ReflectionQueryFactory.kt
swankjesse: exactly

misk/src/main/kotlin/misk/hibernate/ReflectionQueryFactory.kt
swankjesse: yep! Yay!

misk/src/main/kotlin/misk/hibernate/ReflectionQueryFactory.kt
swankjesse: Fixed.

misk/src/test/kotlin/misk/hibernate/ReflectionQueryFactoryValidationTest.kt
swankjesse: yeah I don’t have a nice fix here

misk/src/main/kotlin/misk/hibernate/ReflectionQueryFactory.kt
swankjesse: fixed

misk/src/test/kotlin/misk/hibernate/ReflectionQueryFactoryTest.kt
swankjesse: Fixed.

ryanhall07:
this is cool!

PR #254, opened 2018-06-14
LogCollector for testing

swankjesse:
This requires Logback as the SLF4J backend. That's what we're using
elsewhere (ie. CloudwatchLogAppender) and having multiple is bad news.

### DIFF
--- a/misk/testing/src/main/kotlin/misk/logging/RealLogCollector.kt
+++ b/misk/testing/src/main/kotlin/misk/logging/RealLogCollector.kt
@@ -1,0 +1,62 @@
+package misk.logging
+
+import ch.qos.logback.classic.Level
+import ch.qos.logback.classic.Logger
+import ch.qos.logback.classic.spi.ILoggingEvent
+import ch.qos.logback.core.UnsynchronizedAppenderBase
+import com.google.common.util.concurrent.AbstractIdleService
+import com.google.common.util.concurrent.Service.State.NEW
+import org.slf4j.LoggerFactory
+import javax.inject.Singleton
+import kotlin.reflect.KClass
+
+@Singleton
+internal class RealLogCollector : AbstractIdleService(), LogCollector {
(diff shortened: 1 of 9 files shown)


misk/testing/src/main/kotlin/misk/logging/RealLogCollector.kt
```diff
@@ -0,0 +1,62 @@
...
+import com.google.common.util.concurrent.Service.State.NEW
+import org.slf4j.LoggerFactory
+import javax.inject.Singleton
+import kotlin.reflect.KClass
+
+@Singleton
+internal class RealLogCollector : AbstractIdleService(), LogCollector {
+  private val events = mutableListOf<ILoggingEvent>()
+
+  private val appender = object : UnsynchronizedAppenderBase<ILoggingEvent>() {
```
ryanhall07: why an unsynchronized appender, just to synchronize below?

misk/testing/src/test/kotlin/misk/logging/LogCollectorTest.kt
```diff
@@ -0,0 +1,74 @@
...
+  @MiskTestModule
+  val module = Modules.combine(
+      MiskModule(),
+      LogCollectorModule()
+  )
+
+  @Inject lateinit var logCollector: LogCollector
+
+  @Test
+  fun test() {
```
ryanhall07: nit: better name

misk/testing/src/main/kotlin/misk/logging/RealLogCollector.kt
swankjesse: I’ve got to synchronize with both producers and consumers

misk/testing/src/test/kotlin/misk/logging/LogCollectorTest.kt
swankjesse: I like that it’s cute. Fixed.

PR #255, opened 2018-06-14
Upgrade to the latest Kubernetes API client

swankjesse:
It doesn't have a transitive dependency on Log4J 1.x.


swankjesse:
Updated PR includes a fix for a 9-argument method that became a 12-argument method!

PR #257, opened 2018-06-14
Implement row counts.

swankjesse:
We're opinionated here. If you don't specify a row count you get warnings at 2000
rows and errors at 3000 rows.

Whether or not you specify a row count you get hard errors at 10,000 rows.

### DIFF
--- a/misk/src/main/kotlin/misk/hibernate/ReflectionQueryFactory.kt
+++ b/misk/src/main/kotlin/misk/hibernate/ReflectionQueryFactory.kt
@@ -4,6 +4,7 @@ import com.google.common.cache.CacheBuilder
 import com.google.common.cache.CacheLoader
 import com.google.inject.TypeLiteral
 import misk.inject.typeLiteral
+import misk.logging.getLogger
 import java.lang.reflect.InvocationHandler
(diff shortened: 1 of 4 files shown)


misk/src/main/kotlin/misk/hibernate/ReflectionQueryFactory.kt
```diff
@@ -32,16 +38,24 @@ internal class ReflectionQuery<T : DbEntity<T>>(
   private val rootEntityType: KClass<T>,
   private val queryMethodHandlers: Map<Method, QueryMethodHandler>
 ) : Query<T>, InvocationHandler {
+  override var maxRows = -1
```
ryanhall07: why the default of -1 instead of just MAX_MAX_ROW?

swankjesse: If the user provides a limit, we happily truncate. If we add the limit ourselves and need to truncate, that’s an error.

swankjesse: lemme add one more test for that.

PR #258, opened 2018-06-14
Add health checks to the HibernateModule for database connectivity an…

### LINKED ISSUE #227: Hibernate health checks
Health checks should do two things:
 * confirm connectivity
 * defend against extreme clock skew

jmuia:
…d clock skew.

Health checking queries are performed on-demand rather than periodically in the background and reported when asked.

Closes #227

### DIFF
--- a/misk/src/main/kotlin/misk/hibernate/HibernateHealthCheck.kt
+++ b/misk/src/main/kotlin/misk/hibernate/HibernateHealthCheck.kt
@@ -1,0 +1,53 @@
+package misk.hibernate
+
+import misk.healthchecks.HealthCheck
+import misk.healthchecks.HealthStatus
+import misk.jdbc.DataSourceConfig
+import misk.logging.getLogger
+import org.hibernate.SessionFactory
+import java.sql.Timestamp
+import java.time.Clock
+import java.time.Duration
+
+private val logger = getLogger<HibernateHealthCheck>()
+
+/**
+ * HealthCheck to confirm database connectivity and defend against clock skew.
+ */
+class HibernateHealthCheck(
+  private val sessionFactory: SessionFactory,
+  private val config: DataSourceConfig,
+  private val clock: Clock
+) : HealthCheck {
+
+  override fun status(): HealthStatus {
+    val databaseInstant = try {
+      sessionFactory.openSession().use { session ->
+        session.createNativeQuery("SELECT NOW()").uniqueResult() as Timestamp
+      }.toInstant()
+    } catch (e: Exception) {
+      return HealthStatus.unhealthy("Hibernate: failed to query ${config.database} database")
+    }
+
+    val delta = Duration.between(clock.instant(), databaseInstant).abs()
+    val driftMessage = "Hibernate: host and ${config.database} database " +
+        "clocks have drifted ${delta.seconds}s apart"
+
+    return when {
+      delta > CLOCK_SKEW_UNHEALTHY_THRESHOLD -> {
+        HealthStatus.unhealthy(driftMessage)
+      }
+      delta > CLOCK_SKEW_WARN_THRESHOLD -> {
+        logger.warn { driftMessage }
+        HealthStatus.healthy(driftMessage)
+      }
+      else ->
+        HealthStatus.healthy("Hibernate: ${config.database} database")
+    }
+  }
+
+  companion object {
+    val CLOCK_SKEW_WARN_THRESHOLD = Duration.ofSeconds(10)
+    val CLOCK_SKEW_UNHEALTHY_THRESHOLD = Duration.ofSeconds(30)
+  }
+}
--- a/misk/src/main/kotlin/misk/hibernate/HibernateHealthCheckModule.kt
+++ b/misk/src/main/kotlin/misk/hibernate/HibernateHealthCheckModule.kt
@@ -1,0 +1,30 @@
+package misk.hibernate
+
+import misk.healthchecks.HealthCheck
+import misk.inject.KAbstractModule
+import misk.inject.addMultibinderBinding
+import misk.inject.asSingleton
+import misk.jdbc.DataSourceConfig
+import org.hibernate.SessionFactory
+import java.time.Clock
+import javax.inject.Inject
+import javax.inject.Provider
+
+/**
+ * Binds a HealthCheck for Hibernate.
+ */
+class HibernateHealthCheckModule(
+  private val sessionFactoryProvider: Provider<SessionFactory>,
+  private val config: DataSourceConfig
+) : KAbstractModule() {
+
+  override fun configure() {
+    binder().addMultibinderBinding<HealthCheck>()
+      .toProvider(object : Provider<HibernateHealthCheck> {
+        @Inject lateinit var clock: Clock
+
+        override fun get(): HibernateHealthCheck =
+          HibernateHealthCheck(sessionFactoryProvider.get(), config, clock)
+      }).asSingleton()
+  }
+}
--- a/misk/src/main/kotlin/misk/time/ClockModule.kt
+++ b/misk/src/main/kotlin/misk/time/ClockModule.kt
@@ -2,9 +2,12 @@ package misk.time
 
 import misk.inject.KAbstractModule
 import java.time.Clock
+import java.time.ZoneOffset
+import java.util.TimeZone
 
 class ClockModule : KAbstractModule() {
   override fun configure() {
     bind<Clock>().toInstance(Clock.systemUTC())
+    TimeZone.setDefault(TimeZone.getTimeZone(ZoneOffset.UTC))
   }
 }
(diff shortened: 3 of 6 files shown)


misk/src/main/kotlin/misk/jdbc/DataSourceConfig.kt
```diff
@@ -43,7 +43,9 @@ data class DataSourceConfig(
   val fixed_pool_size: Int = 10,
   val connection_timeout: Duration = Duration.ofSeconds(30),
   val connection_max_lifetime: Duration = Duration.ofMinutes(30),
-  val migrations_path: String
+  val migrations_path: String,
+  // Threshold beyond which clock skew is considered unacceptable and the host is reported unhealthy.
+  val clock_skew_threshold: Duration = Duration.ofMinutes(1)
```
jmuia: I'm not sure what a good default is.

jmuia:
Need to add tests.

misk/src/main/kotlin/misk/hibernate/HibernateHealthCheck.kt
```diff
@@ -0,0 +1,56 @@
...
+      return when (e) {
+        is NoResultException,
+        is NonUniqueResultException,
+        is IllegalStateException,
+        is QueryTimeoutException,
+        is TransactionRequiredException,
+        is PessimisticLockException,
+        is LockTimeoutException,
+        is PersistenceException ->
+          HealthStatus.unhealthy("Hibernate: failed to query ${config.database}")
```
ryanhall07: why not report all exceptions as unhealthy?

misk/src/main/kotlin/misk/hibernate/HibernateHealthCheckModule.kt
```diff
@@ -0,0 +1,30 @@
...
+ * Binds a HealthCheck for Hibernate.
+ */
+class HibernateHealthCheckModule(
+  private val sessionFactoryProvider: Provider<SessionFactory>,
+  private val config: DataSourceConfig
+) : KAbstractModule() {
+
+  override fun configure() {
+    binder().addMultibinderBinding<HealthCheck>()
+      .toProvider(object : Provider<HibernateHealthCheck> {
```
ryanhall07: why the manual provider?

misk/src/main/kotlin/misk/hibernate/HibernateHealthCheck.kt
```diff
@@ -0,0 +1,56 @@
...
+import javax.persistence.LockTimeoutException
+import javax.persistence.NoResultException
+import javax.persistence.NonUniqueResultException
+import javax.persistence.PersistenceException
+import javax.transaction.TransactionRequiredException
+
+/**
+ * HealthCheck to confirm database connectivity and defend against clock skew.
+ */
+class HibernateHealthCheck(
```
ryanhall07: test?

misk/src/main/kotlin/misk/hibernate/HibernateHealthCheck.kt
jmuia: It felt bad catching `Exception`. These are the specific exceptions thrown by an API that I'm no longer calling. woops. Will remove.

misk/src/main/kotlin/misk/hibernate/HibernateHealthCheck.kt
jmuia: yep. Thoughts on how to do this? I'm thinking about how to do the clock skew since we use a FakeClock for tests but the time returned is from the database (if that's the test MySQL it will be the hosts time).

misk/src/main/kotlin/misk/hibernate/HibernateHealthCheckModule.kt
jmuia: I wanted to get the SessionFactory lazily, since it's provided by the SessionFactoryService (which needs to be started)

misk/src/main/kotlin/misk/hibernate/HibernateHealthCheck.kt
ryanhall07: you don't have to test the clock skew if it's difficult or flaky. but it seems like you could set the FakeClock to the real current time and it would be close enough.

misk/src/main/kotlin/misk/jdbc/DataSourceConfig.kt
swankjesse: I don’t think this needs to be configurable. If we pick a reasonable enough default I think that should be good for everyone.

My preference for the default is to warn at > 10s and become unhealthy at > 30s. This shouldn’t happen in practice.

misk/src/main/kotlin/misk/hibernate/HibernateHealthCheck.kt
```diff
@@ -0,0 +1,37 @@
...
+        session.createNativeQuery("SELECT NOW()").uniqueResult() as Timestamp
+      }.toInstant()
+    } catch (e: Exception) {
+      return HealthStatus.unhealthy("Hibernate: failed to query ${config.database}")
+    }
+
+    val delta = Duration.between(clock.instant(), databaseInstant).abs()
+
+    return if (delta > config.clock_skew_threshold) {
+      HealthStatus.unhealthy("Hibernate: host clock has drifted ${delta.seconds}s from ${config.database}")
```
swankjesse: we can’t _really_ blame the host; all we know is that the host and one of its DBs disagree.

misk/src/main/kotlin/misk/hibernate/HibernateHealthCheck.kt
swankjesse: yep!

misk/src/main/kotlin/misk/hibernate/HibernateHealthCheckModule.kt
ryanhall07: can't you just inject a `Provider<SessionFactory>` into the health check?

jmuia: I don't think so, since the `SessionFactory` could be qualified.

ryanhall07: ah right. thanks!

misk/src/main/kotlin/misk/hibernate/HibernateHealthCheck.kt
jmuia: Fixed language to make that more clear.

misk/src/main/kotlin/misk/jdbc/DataSourceConfig.kt
jmuia: Fixed.

misk/src/main/kotlin/misk/hibernate/HibernateHealthCheck.kt
jmuia: Added tests.

misk/src/main/kotlin/misk/time/ClockModule.kt
```diff
@@ -2,9 +2,12 @@ package misk.time
 
 import misk.inject.KAbstractModule
 import java.time.Clock
+import java.time.ZoneOffset
+import java.util.*
```
swankjesse: nit: configure IntelliJ to forbid wildcards

misk/src/main/kotlin/misk/time/ClockModule.kt
```diff
@@ -2,9 +2,12 @@ package misk.time
...
 
 class ClockModule : KAbstractModule() {
   override fun configure() {
     bind<Clock>().toInstance(Clock.systemUTC())
+    TimeZone.setDefault(TimeZone.getTimeZone(ZoneOffset.UTC))
```
swankjesse: I think we gotta move this somewhere else. Not urgent, but this is not how Guice wants to work
https://github.com/google/guice/wiki/ModulesShouldBeFastAndSideEffectFree

misk/src/main/kotlin/misk/hibernate/HibernateHealthCheck.kt
```diff
@@ -0,0 +1,54 @@
...
+      return HealthStatus.unhealthy("Hibernate: failed to query ${config.database} database")
+    }
+
+    val delta = Duration.between(clock.instant(), databaseInstant).abs()
+    val driftMessage = "Hibernate: host and ${config.database} database " +
+        "clocks have drifted ${delta.seconds}s apart"
+
+    return when {
+      delta > CLOCK_SKEW_UNHEALTHY_THRESHOLD -> {
+        logger.warn { driftMessage }
```
swankjesse: This is repetitive. We should assume whoever is calling the health check will log for us.

misk/src/main/kotlin/misk/time/ClockModule.kt
jmuia: Ah, whoops. I will fix in a follow-up.

PR #259, opened 2018-06-14
Support running tests against Vitess

tirsen:
Assumes the Vitess cluster is already started using:
cd misk/src/test/resources/misk/hibernate
docker-compose up


tirsen:
There's a bit of code dupe with the migrations. Not sure how to fix that...

misk/src/main/kotlin/misk/hibernate/ForceUTCTimeZoneService.kt
```diff
@@ -0,0 +1,19 @@
...
+import com.google.common.util.concurrent.AbstractIdleService
+import java.time.ZoneOffset
+import java.util.TimeZone
+
+/**
+ * This works around a Hibernate issue when loading LocalDate instances from the database when
+ * running on a machine which doesn't have UTC as the system timezone. All production servers have
+ * UTC as the system timezone so this is essentially a no op in those case.
+ */
+class ForceUTCTimeZoneService : AbstractIdleService() {
```
swankjesse: awesome. I think this probably belongs somewhere other than misk.hibernate (not urgent). Also we do `ForceUtcTimeZoneService` not `ForceUTCTimeZoneService` per https://google.github.io/styleguide/javaguide.html#s5.3-camel-case

misk/src/main/kotlin/misk/hibernate/HibernateModule.kt
```diff
@@ -77,6 +77,9 @@ class HibernateModule(
           qualifier, environment, schemaMigratorProvider)
     }).asSingleton()
 
+    binder().addMultibinderBinding<Service>().toProvider(
```
swankjesse: or this:
```
binder().addMultibinderBinding<Service>().to(ForceUTCTimeZoneService::class.java)
```

misk/src/main/kotlin/misk/hibernate/ForceUTCTimeZoneService.kt
swankjesse: add `@Singleton`

misk/src/main/kotlin/misk/hibernate/RealTransacter.kt
```diff
@@ -41,5 +47,62 @@ internal class RealTransacter(
...
+        return SINGLE_SHARD_LIST
+      }
+    }
+
+    private fun parseShard(string: String): Shard {
+      val (keyspace, shard) = string.split('/', limit = 2)
+      return Shard(Keyspace(keyspace), shard)
+    }
+
+    override fun <T> target(shard: Shard, function: () -> T): T {
```
swankjesse: yeah!

misk/src/main/kotlin/misk/hibernate/RealTransacter.kt
```diff
@@ -41,5 +47,62 @@ internal class RealTransacter(
...
+      if (config.type == DataSourceType.VITESS) {
+        val previousTarget = doWork {
```
swankjesse: what’s the scope of a USE statement? I’m anxious Hikari will give us 3 different physical connections for these 3 calls

misk/src/main/kotlin/misk/hibernate/SchemaMigrator.kt
```diff
@@ -37,6 +37,9 @@ internal class SchemaMigrator(
 ) {
   /** Returns a map from version to path. */
   fun availableMigrations(): NavigableMap<Int, String> {
+    if (config.migrations_path == null) {
```
swankjesse: hmmmm . . . is this something to support?

misk/src/main/kotlin/misk/hibernate/Session.kt
```diff
@@ -1,11 +1,20 @@
 package misk.hibernate
 
+import java.sql.Connection
 import kotlin.reflect.KClass
 
 interface Session {
   val hibernateSession: org.hibernate.Session
   fun <T : DbEntity<T>> save(entity: T): Id<T>
   fun <T : DbEntity<T>> load(id: Id<T>, type: KClass<T>): T
+  fun shards(): List<Shard>
```
swankjesse: Set<Shard> for stronger semantics? (No duplicates!)

misk/src/main/kotlin/misk/hibernate/Session.kt
```diff
@@ -1,11 +1,20 @@
...
+  fun <T> target(shard: Shard, function: () -> T): T
+  fun <T> doWork(work: (Connection) -> T): T
```
swankjesse: should we give this a nicer name? This is what Hibernate uses but it never made much sense to me

misk/src/main/kotlin/misk/hibernate/Session.kt
```diff
@@ -1,11 +1,20 @@
...
+}
+
+data class Keyspace(val name: String)
```
swankjesse: assert well-formed? Would be lousy to accidentally let whitespace or delimiters in

misk/src/main/kotlin/misk/hibernate/Session.kt
```diff
@@ -1,11 +1,20 @@
...
+data class Shard(val keyspace: Keyspace, val name: String) {
+  override fun toString() = "${keyspace.name}/$name"
```
swankjesse: yes!

misk/src/main/kotlin/misk/jdbc/DataSourceConfig.kt
```diff
@@ -28,7 +28,18 @@ enum class DataSourceType(
       buildJdbcUrl = { config, _ ->
         "jdbc:hsqldb:mem:${config.database!!};sql.syntax_mys=true"
       }
-  )
+  ),
+  VITESS(
```
swankjesse: yeah!

misk/src/test/kotlin/misk/hibernate/TransacterTest.kt
```diff
@@ -49,14 +49,14 @@ class TransacterTest {
       val lauraDernMovies = queryFactory.newQuery<CharacterQuery>()
           .actorName("Laura Dern")
           .listAsMovieNameAndReleaseDate(session)
-      assertThat(lauraDernMovies).containsExactly(
+      assertThat(lauraDernMovies).containsExactlyInAnyOrder(
```
swankjesse: With Vitess do we lose the implicit `ORDER BY ID` ? or is it just that on insert the IDs aren’t predictable?
Too bad, but I can’t think of a good alternative

misk/src/test/resources/misk/hibernate/docker-compose.yaml
```diff
@@ -0,0 +1,11 @@
+version: '3'
+services:
+  vitess:
+    image: "vitess/base:latest"
+    command: "../../../bin/vttestserver -alsologtostderr -port 27000 -web_dir=web/vtctld/app -web_dir2=web/vtctld2/app -mysql_bind_host=0.0.0.0 -schema_dir=schema  -keyspaces=movies,actors --num_shards=2,1"
```
eeSeeGee: You can make this a little more readable by splitting to multiple lines if you want. You can use an array of tokens
```
command: [
  "../../../bin/vttestserver",
  "-alsologtostderr",
  "-port", "27000",
   ...
]
```
or I believe the bar operator works here too:
```
command: |
  ../../../bin/vttestserver
  -alsologtostderr
  -port 27000
  ...
```

swankjesse: the implied working directory here doesn’t feel right. No guidance on what to do instead

misk/src/main/kotlin/misk/hibernate/ForceUTCTimeZoneService.kt
jmuia: I needed this change for testing Hibernate health checks (#258) and it seemed simple enough to just add the relevant line `TimeZone.setDefault(TimeZone.getTimeZone(ZoneOffset.UTC))` to the `ClockModule`

tirsen: It might belong in misk.time but currently this is only installed in the HibernateModule and strictly only necessary there (as Hibernate doesn't handle if mysqld is in a different timezone to the server).

misk/src/main/kotlin/misk/hibernate/HibernateModule.kt
tirsen: Aha!

misk/src/test/resources/misk/hibernate/docker-compose.yaml
tirsen: Nice! I couldn't get the pipe thing to work though but the other one is nice enough. :-)

misk/src/main/kotlin/misk/hibernate/RealTransacter.kt
```diff
@@ -41,5 +47,62 @@ internal class RealTransacter(
     override fun <T : DbEntity<T>> load(id: Id<T>, type: KClass<T>): T {
       return session.get(type.java, id)
     }
+
+    override fun shards(): List<Shard> {
+      if (config.type == DataSourceType.VITESS) {
+        return doWork { connection ->
+          val shards = ArrayList<Shard>()
+          connection.createStatement().use { statement ->
+            val resultSet = statement.executeQuery("SHOW VITESS_SHARDS")
```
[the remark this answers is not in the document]
tirsen: Yeah this is the new stuff that I want us to eventually switch to in Franklin. Notice I use the vanilla mysql JDBC driver!

misk/src/main/kotlin/misk/hibernate/RealTransacter.kt
```diff
@@ -41,5 +47,62 @@ internal class RealTransacter(
     override fun <T : DbEntity<T>> load(id: Id<T>, type: KClass<T>): T {
       return session.get(type.java, id)
     }
+
+    override fun shards(): List<Shard> {
+      if (config.type == DataSourceType.VITESS) {
+        return doWork { connection ->
+          val shards = ArrayList<Shard>()
```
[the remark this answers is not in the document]
tirsen: I made this even nicer using a ResultSet extension function. I love Kotlin!

misk/src/main/kotlin/misk/hibernate/RealTransacter.kt
tirsen: I don't think it will but I agree that calling the function inside `doWork` is clearer.

misk/src/main/kotlin/misk/hibernate/SchemaMigrator.kt
tirsen: vttestserver (the Vitess test cluster) supports migrations out of the box. You have to restart it to re-apply new migrations but it's really pretty damn fast to restart (seconds). We could potentially restart it before each test...

misk/src/main/kotlin/misk/hibernate/Session.kt
tirsen: done

misk/src/test/kotlin/misk/hibernate/TransacterTest.kt
tirsen: Yeah for a cross shard query like this one you can't efficiently support the implicit order by primary key because you would essentially have to load each shards result into memory and sort them. That's fine if you explicitly require a sort order but silently always paying this price doesn't seem like a good idea.

misk/src/test/resources/misk/hibernate/docker-compose.yaml
tirsen: Yeah that's not entirely pretty but it's not our docker image and there's strength in being able to just use the one that's provided by the open source project. Maybe an absolute path is a bit better.

I'm sure we're going to iterate on this for a bit until we get it just right.

tirsen: An absolute path is a lot better

misk/src/main/kotlin/misk/hibernate/Session.kt
tirsen: I changed it to `useConnection` which I'd argue is one step up from `doWork` but far from perfect. The main issue is that the word `use` seems to imply that the connection is closed afterwards?

PR #260, opened 2018-06-15
Add a test to confirm that unknown properties trigger errors.

swankjesse:


misk/src/test/resources/unknownproperty-common.yaml
```diff
@@ -0,0 +1,17 @@
...
+  port: 1234
+  idle_timeout: 30000
+
+consumer_a:
+  max_items: 1
+
+consumer_b:
+  min_items: 1
+  max_items: 2
+  blue_items: 3
```
ryanhall07: super nit: I would either comment or change the name to make it clear that this property is the unknown one

swankjesse: good call. Fixed!

Issue #224, opened 2018-06-05
Transacter.inTransaction


swankjesse:
Make it easy to check whether a transaction is in flight. This is useful for defensive programming:

```
  require(!transacter.inTransaction)
```

PR #267, opened 2018-06-19
Transacter.inTransaction()

jmuia:
Provide an API to check whether a transaction is in flight (#224).
This is useful for defensive programming such as `require(!transacter.inTransaction())`.

As distinct units of work run in their own threads, the thread-local variable holding on to the Session provides sufficient isolation on the Transacter singleton.

### DIFF
--- a/misk/src/main/kotlin/misk/hibernate/RealTransacter.kt
+++ b/misk/src/main/kotlin/misk/hibernate/RealTransacter.kt
@@ -6,15 +6,20 @@ import kotlin.reflect.KClass
 internal class RealTransacter(
   private val sessionFactory: SessionFactory
 ) : Transacter {
+
+  private val TLS = ThreadLocal<Session>()
+
+  override val inTransaction: Boolean
+    get() = TLS.get() != null
+
   override fun <T> transaction(lambda: (session: Session) -> T): T {
-    sessionFactory.openSession().use { session ->
-      val transaction = session.beginTransaction()!!
-      val realSession = RealSession(session)
+    return withSession { session ->
+      val transaction = session.hibernateSession.beginTransaction()!!
       val result: T
       try {
-        result = lambda(realSession)
+        result = lambda(session)
         transaction.commit()
-        return result
+        return@withSession result
       } catch (e: Throwable) {
         if (transaction.isActive) {
           try {
@@ -25,6 +30,27 @@ internal class RealTransacter(
         }
         throw e
       }
+    }
+  }
+
+  private fun <T> withSession(lambda: (session: Session) -> T): T {
+    check(TLS.get() == null) { "Attempted to start a nested session" }
+
+    val realSession = RealSession(sessionFactory.openSession())
+    TLS.set(realSession)
+
+    try {
+      return lambda(realSession)
+    } finally {
+      closeSession()
+    }
+  }
+
+  private fun closeSession() {
+    try {
+      TLS.get().hibernateSession.close()
+    } finally {
+      TLS.remove()
     }
   }
 
--- a/misk/src/main/kotlin/misk/hibernate/Transacter.kt
+++ b/misk/src/main/kotlin/misk/hibernate/Transacter.kt
@@ -1,5 +1,23 @@
 package misk.hibernate
 
+/**
+ * Provides explicit block-based transaction demarcation.
+ */
 interface Transacter {
+  /**
+   * @returns [true] if the calling thread is currently within a transaction block.
+   */
+  val inTransaction: Boolean
+
+  /**
+   * Executes [lambda] in a transaction. The transaction will be committed once [lambda] completes.
+   * If [lambda] raises an exception the transaction will be rolled back.
+   *
+   * @param lambda a function to execute in a transaction.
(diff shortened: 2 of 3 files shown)


misk/src/main/kotlin/misk/hibernate/RealTransacter.kt
```diff
@@ -28,6 +30,29 @@ internal class RealTransacter(
     }
   }
 
+  override fun inTransaction(): Boolean = TLS.get() != null
```
mmihic: nit: IMO make this a read-only property

misk/src/main/kotlin/misk/hibernate/Transacter.kt
```diff
@@ -2,4 +2,5 @@ package misk.hibernate
 
 interface Transacter {
   fun <T> transaction(lambda: (session: Session) -> T): T
-}
\ No newline at end of file
+  fun inTransaction(): Boolean
```
ryanhall07: maybe update the doc of this interface to make it clear `transaction` throws an exception `inTransaction` returns true

misk/src/test/kotlin/misk/hibernate/TransacterTest.kt
```diff
@@ -92,6 +92,27 @@ class TransacterTest {
...
+
+    transacter.transaction {
+      assertThat(transacter.inTransaction()).isTrue()
+    }
+
+    assertThat(transacter.inTransaction()).isFalse()
+  }
+
+  @Test
+  fun nestedTransactionUnsupported() {
```
ryanhall07: YES!

misk/src/main/kotlin/misk/hibernate/RealTransacter.kt
jmuia: Fixed.

misk/src/main/kotlin/misk/hibernate/Transacter.kt
jmuia: Updated doc.

misk/src/test/kotlin/misk/hibernate/TransacterTest.kt
swankjesse: Double yes!

misk/src/main/kotlin/misk/hibernate/RealTransacter.kt
```diff
@@ -6,15 +6,20 @@ import kotlin.reflect.KClass
 internal class RealTransacter(
   private val sessionFactory: SessionFactory
 ) : Transacter {
+
+  private val TLS = ThreadLocal<Session>()
```
swankjesse: nit: this field name is not right. `threadLocalSession` ?

misk/src/main/kotlin/misk/hibernate/Transacter.kt
```diff
@@ -1,5 +1,23 @@
 package misk.hibernate
 
+/**
+ * Provides explicit block-based transaction demarcation.
```
swankjesse: nice

misk/src/main/kotlin/misk/hibernate/Transacter.kt
```diff
@@ -1,5 +1,23 @@
...
+ */
 interface Transacter {
+  /**
+   * @returns [true] if the calling thread is currently within a transaction block.
```
swankjesse: nit: `@return`, or even better `Returns`

`@return` is a useful tag when there’s a lot of primary documentation and you want to go in detail on how to interpret the return value. But here the entire point of the method is to return something, so there’s no reason to hide the documentation under the `@return` tag.

```
/** 
 * Returns true if the calling thread is currently within a transaction block.  
 */
```

swankjesse: Also, we only wanna use `code blocks` and `[braces]` when they disambiguate. For words like true, null, etc., we want to use plain English as much as possible. (And if the reader wants the precision of code, they have the source right there)

misk/src/main/kotlin/misk/hibernate/Transacter.kt
```diff
@@ -1,5 +1,23 @@
...
+   */
+  val inTransaction: Boolean
+
+  /**
+   * Executes [lambda] in a transaction. The transaction will be committed once [lambda] completes.
+   * If [lambda] raises an exception the transaction will be rolled back.
+   *
+   * @param lambda a function to execute in a transaction.
```
swankjesse: nit: this doc doesn’t add anything.
https://publicobject.com/2007/08/05/stop_writing_redundant_javadoc/

misk/src/main/kotlin/misk/hibernate/Transacter.kt
```diff
@@ -1,5 +1,23 @@
...
+   * @return the result of [lambda] function.
+   *
+   * @throws IllegalStateException if attempting a nested transaction, which is unsupported.
+   * @throws Throwable for various other runtime exceptions.
```
swankjesse: especially this is not useful. It’s scary!

PR #268, opened 2018-06-19
Replace class literals with reified methods

swankjesse:
To help, introduce our own assertThrows() method that works better with Kotlin

### DIFF
--- a/misk/testing/src/main/kotlin/misk/testing/JUnit.kt
+++ b/misk/testing/src/main/kotlin/misk/testing/JUnit.kt
@@ -1,0 +1,18 @@
+package misk.testing
+
+/** Confirms that `block` throws `T`, and returns what was thrown. */
+inline fun <reified T : Throwable> assertThrows(block: () -> Unit): T {
+  var throwable: Throwable? = null
+  try {
+    block()
+  } catch (expected: Throwable) {
+    throwable = expected
+  }
+
+  when (throwable) {
+    is T -> return throwable
+    null -> throw AssertionError("expected ${T::class.simpleName} was not thrown")
+    else -> throw AssertionError(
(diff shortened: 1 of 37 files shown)


samples/urlshortener/src/test/kotlin/com/squareup/urlshortener/ShortenUrlWebActionTest.kt
```diff
@@ -39,14 +39,14 @@ class ShortenUrlWebActionTest {
   @Test
   fun malformedLongUrl() {
     val longUrl = ""
-    assertThrows(BadRequestException::class.java) {
+    assertThrows<BadRequestException> {
```
ryanhall07: nice

misk/testing/src/main/kotlin/misk/testing/JUnit.kt
```diff
@@ -0,0 +1,18 @@
+package misk.testing
+
+/** Confirms that `block` throws `T`, and returns what was thrown. */
+inline fun <reified T : Throwable> assertThrows(block: () -> Unit): T {
```
ryanhall07: where do we draw the line on the number of assertion helpers we add to misk? don't want it turning into something like assertj (which you could just use)

swankjesse: We should be comfortable adding small APIs to make Kotlin nicer. All such APIs should be candidates for upstreaming to their respective projects.

swankjesse: (I'd like to submit PRs to create a guice-kotlin module with our Guice.kt stuff, for example.)

swankjesse:
Test failures appear to be a long-fixed compiler bug. Will upgrade to Kotlin 1.2.50 and rebase on that.
https://github.com/square/misk/pull/269

PR #270, opened 2018-06-20
Use a mock SessionFactory to test database failures during health check.

### LINKED ISSUE #264: HealthCheckTest#databaseConnectivityFailure causes Service shutdown failures
Test performs correctly, but closing the SessionFactory means that any code that runs after the test and relies on the SessionFactory will fail (eg. TruncateTablesService shutdown statements). It would be good to use another method to fake a database failure.

jmuia:
Closes #264.


swankjesse:
Yeah this is much better.

PR #273, opened 2018-06-20
Tidy up some Guice APIs

swankjesse:
Be more consistent with multibindings, and avoid needing to deference
the binder in the common case that we're in a KAbstractModule.

### DIFF
--- a/misk/src/main/kotlin/misk/MiskApplication.kt
+++ b/misk/src/main/kotlin/misk/MiskApplication.kt
@@ -16,7 +16,7 @@ class MiskApplication(private val modules: List<Module>, commands: List<MiskComm
   constructor(vararg modules: Module) : this(modules.toList())
   constructor(vararg commands: MiskCommand) : this(listOf(), commands.toList())
 
-  private val commands = commands.map { it.name to it }.toMap()
+  private val commands = commands.associateBy { it.name }
   private val jc: JCommander
 
   init {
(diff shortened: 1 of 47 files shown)


misk/src/main/kotlin/misk/inject/KAbstractModule.kt
```diff
@@ -34,13 +38,33 @@ abstract class KAbstractModule : AbstractModule() {
...
-  protected inline fun <reified T : Any> AnnotatedBindingBuilder<in T>.to(): ScopedBindingBuilder {
+  protected inline fun <reified T : Any> LinkedBindingBuilder<in T>.to(): ScopedBindingBuilder {
     return to(T::class.java)
   }
 
-  protected inline fun <reified T : Any> newSetBinder(): Multibinder<T> {
-    return Multibinder.newSetBinder(binder(), T::class.java)
-  }
-}
+  protected inline fun <reified T : Any> addMultibinderBinding(
```
ryanhall07: nit: `addMultiBinding`?

misk/src/main/kotlin/misk/MiskApplication.kt
```diff
@@ -16,7 +16,7 @@ class MiskApplication(private val modules: List<Module>, commands: List<MiskComm
   constructor(vararg modules: Module) : this(modules.toList())
   constructor(vararg commands: MiskCommand) : this(listOf(), commands.toList())
 
-  private val commands = commands.map { it.name to it }.toMap()
+  private val commands = commands.associateBy { it.name }
```
tirsen: Ohhh nice!

misk/src/main/kotlin/misk/concurrent/WrappingListeningExecutorService.kt
```diff
@@ -20,7 +20,7 @@ abstract class WrappingListeningExecutorService : ForwardingListeningExecutorSer
   }
 
   override fun <T> submit(runnable: Runnable, result: T): ListenableFuture<T> {
-    return submit<T>({ runnable.run(); result })
+    return submit<T> { runnable.run(); result }
```
adrw: Kotlin-esque 👍

misk/src/main/kotlin/misk/inject/KAbstractModule.kt
adrw: @swankjesse thoughts?

swankjesse: Yeah, that’s better

swankjesse: Renamed to `multibind()`.

PR #275, opened 2018-06-22
 Basic transacter retries and opt-out API

jmuia:
Adds retries to transactions by default with an API to opt out.
Opt out API is intended to be extensible for various types of transacter configuration.

~~Depends on #274 for retry exception matching (temporarily cherry-picked for dev't).~~

~~Jitter will be added in a follow-up.~~

### DIFF
--- a/misk-hibernate/src/main/kotlin/misk/hibernate/RealTransacter.kt
+++ b/misk-hibernate/src/main/kotlin/misk/hibernate/RealTransacter.kt
@@ -1,22 +1,68 @@
 package misk.hibernate
 
 import com.google.common.collect.ImmutableSet
+import misk.backoff.ExponentialBackoff
+import misk.logging.getLogger
 import org.hibernate.SessionFactory
+import org.hibernate.StaleObjectStateException
+import org.hibernate.exception.LockAcquisitionException
 import java.sql.Connection
 import java.sql.ResultSet
+import java.time.Duration
+import javax.persistence.OptimisticLockException
 import kotlin.reflect.KClass
 
-internal class RealTransacter(
+private val logger = getLogger<RealTransacter>()
+
+internal class RealTransacter private constructor(
   private val sessionFactory: SessionFactory,
-  private val config: DataSourceConfig
+  private val config: DataSourceConfig,
+  private val threadLocalSession: ThreadLocal<Session>,
+  private val options: TransacterOptions
 ) : Transacter {
 
-  private val threadLocalSession = ThreadLocal<Session>()
+  constructor(sessionFactory: SessionFactory, config: DataSourceConfig) :
+      this(sessionFactory, config, ThreadLocal(), TransacterOptions())
 
   override val inTransaction: Boolean
     get() = threadLocalSession.get() != null
 
   override fun <T> transaction(lambda: (session: Session) -> T): T {
+    assert(options.maxAttempts > 0)
+
+    val backoff = ExponentialBackoff(
+      Duration.ofMillis(options.minRetryDelayMillis),
+      Duration.ofMillis(options.maxRetryDelayMillis),
+      Duration.ofMillis(options.retryJitterMillis)
+    )
+    var attempt = 0
+
+    while (true) {
+      try {
+        attempt++
+        return transactionInternal(lambda)
+      } catch (e: Exception) {
+        if (!isRetryable(e)) throw e
+
+        if (attempt >= options.maxAttempts) {
+          logger.info(e) { "${config.database} recoverable transaction exception (attempt: $attempt), max attempts exceeded" }
+          throw e
+        }
+
+        logger.info(e) { "${config.database} recoverable transaction exception (attempt: $attempt), retrying" }
+        val sleepDuration = backoff.nextRetry()
+        if (!sleepDuration.isZero) {
+          try {
+            Thread.sleep(sleepDuration.toMillis())
+          } catch (e: InterruptedException) {
+            throw e
+          }
+        }
+      }
+    }
+  }
+
+  private fun <T> transactionInternal(lambda: (session: Session) -> T): T {
     return withSession { session ->
       val transaction = session.hibernateSession.beginTransaction()!!
       val result: T
@@ -36,6 +82,13 @@ internal class RealTransacter(
       }
     }
   }
+
+  override fun retries(): Transacter = withOptions(options.copy(maxAttempts = 2))
+
+  override fun noRetries(): Transacter = withOptions(options.copy(maxAttempts = 1))
+
+  private fun withOptions(options: TransacterOptions): Transacter =
+    RealTransacter(sessionFactory, config, threadLocalSession, options)
 
   private fun <T> withSession(lambda: (session: Session) -> T): T {
     check(threadLocalSession.get() == null) { "Attempted to start a nested session" }
@@ -57,6 +110,24 @@ internal class RealTransacter(
       threadLocalSession.remove()
     }
   }
+
+  private fun isRetryable(e: Exception): Boolean {
+    return when (e) {
+      is RetryTransactionException,
+      is StaleObjectStateException,
+      is LockAcquisitionException,
+      is OptimisticLockException -> true
+      else -> false
+    }
+  }
+
+  // NB: all options should be immutable types as copy() is shallow.
+  internal data class TransacterOptions(
+    val maxAttempts: Int = 2,
+    val minRetryDelayMillis: Long = 100,
+    val maxRetryDelayMillis: Long = 100,
+    val retryJitterMillis: Long = 400
+  )
 
   internal class RealSession(
     val session: org.hibernate.Session,
--- a/misk-hibernate/src/main/kotlin/misk/hibernate/Transacter.kt
+++ b/misk-hibernate/src/main/kotlin/misk/hibernate/Transacter.kt
@@ -13,11 +13,26 @@ interface Transacter {
    * Starts a transaction on the current thread, executes lambda, and commits the transaction.
    * If lambda raises an exception the transaction will be rolled back instead of committed.
    *
+   * If retries are permitted (the default), a failed but recoverable transaction will be
+   * reattempted after rolling back.
+   *
    * It is an error to start a transaction if another transaction is already in progress.
    */
   fun <T> transaction(lambda: (session: Session) -> T): T
+
+  fun retries(): Transacter
+
+  fun noRetries(): Transacter
 }
 
 fun Transacter.shards() = transaction { it.shards() }
 fun <T> Transacter.transaction(shard: Shard, lambda: (session: Session) -> T) =
     transaction { it.target(shard) { lambda(it) } }
+
+/**
+ * Thrown to explicitly trigger a retry, subject to retry limits and config such as noRetries().
+ */
+class RetryTransactionException(
+  message: String? = null,
+  cause: Throwable? = null
+) : Exception(message, cause)
--- a/misk-hibernate/src/test/kotlin/misk/hibernate/TransacterTest.kt
+++ b/misk-hibernate/src/test/kotlin/misk/hibernate/TransacterTest.kt
@@ -8,6 +8,7 @@ import org.assertj.core.api.Assertions.assertThat
 import org.hibernate.exception.ConstraintViolationException
 import org.junit.jupiter.api.Test
 import java.time.LocalDate
+import java.util.concurrent.atomic.AtomicInteger
 import javax.inject.Inject
 
 @MiskTest(startService = true)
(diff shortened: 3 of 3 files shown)


misk/src/main/kotlin/misk/backoff/Retries.kt
```diff
@@ -3,9 +3,17 @@ package misk.backoff
 /**
  * Retries the provided function up to a certain number of times, applying the given backoff
  * between each retry. The retry function is provided with current retry count, in case this is
- * relevant
+ * relevant.
+ *
+ * retryException returns true if the caught exception should be retried, subject to the limit.
```
swankjesse: `@param` ?

misk/src/main/kotlin/misk/backoff/Retries.kt
```diff
@@ -3,9 +3,17 @@ package misk.backoff
...
+ * By default all exceptions are retried.
  */
-fun <A> retry(upTo: Int, withBackoff: Backoff, f: (retryCount: Int) -> A): A {
+fun <A> retry(
+  upTo: Int,
+  withBackoff: Backoff,
+  retryException: ((e: Exception) -> Boolean)? = null,
```
swankjesse: rather than making the function optional, default it to a lambda that returns true?

misk/src/main/kotlin/misk/backoff/Retries.kt
```diff
@@ -3,9 +3,17 @@ package misk.backoff
 /**
  * Retries the provided function up to a certain number of times, applying the given backoff
  * between each retry. The retry function is provided with current retry count, in case this is
- * relevant
+ * relevant.
+ *
+ * retryException returns true if the caught exception should be retried, subject to the limit.
+ * By default all exceptions are retried.
```
ryanhall07: I'm not convinced we want to retry exceptions by default. While it is a nice programming model (abstracts away failures), it has a real impact on the performance of the production system (especially when things aren't going well).

I'd like to spend some time thinking through other options and discussing the tradeoffs.

An alternative design is to never retry transactions and kick the exception up to the caller to retry.

* I think the caller needs to handle failures anyways, since many things can fail, not just database queries. Ideally we just make actions idempotent and they are safe to retry.
* Gives us natural jitter, since retries would probably come over the network
* It's easier to reason about retries in the entire system, since we haven't layered nested retries deep down in the stack.

This is just an alternative idea, but we should spend some time thinking through the trade offs.

misk/src/main/kotlin/misk/backoff/Retries.kt
swankjesse: Actually I think this method is too complex and should be replaced with a class. A method that accepts multiple lambdas is a bad method.

I’m tempted to leave the basic retry thing here as-is, but to write a custom retry loop in Transacter because it has special requirements. It might be the only thing in our entire codebase that needs these bells and whistles, and it keeps the entire machine simpler if we do them directly.

misk/src/main/kotlin/misk/hibernate/RealTransacter.kt
```diff
@@ -1,24 +1,44 @@
...
   private val sessionFactory: SessionFactory,
-  private val config: DataSourceConfig
+  private val config: DataSourceConfig,
+  private val threadLocalSession: ThreadLocal<Session>,
+  private val options: TransacterOptions
 ) : Transacter {
 
-  private val threadLocalSession = ThreadLocal<Session>()
+  constructor(sessionFactory: SessionFactory, config: DataSourceConfig) :
+      this(sessionFactory, config, ThreadLocal(), TransacterOptions())
```
swankjesse: just make ThreadLocal() and TransacterOptions() default values?

misk/src/main/kotlin/misk/hibernate/RealTransacter.kt
```diff
@@ -1,24 +1,44 @@
...
 
   override fun <T> transaction(lambda: (session: Session) -> T): T {
+    return when {
+      options.maxAttempts > 1 -> {
+        val backoff = ExponentialBackoff(Duration.ZERO, Duration.ZERO)
+        retry(options.maxAttempts, backoff, ::isRetryable) { transactionInternal(lambda) }
+      }
+      options.maxAttempts == 1 -> transactionInternal(lambda)
+      else -> throw IllegalArgumentException()
+    }
```
swankjesse: I find this really tough to follow

swankjesse: can we make it so the retry function handles `maxAttempts == 1` by not retrying?

misk/src/main/kotlin/misk/hibernate/RealTransacter.kt
```diff
@@ -1,24 +1,44 @@
...
     get() = threadLocalSession.get() != null
 
   override fun <T> transaction(lambda: (session: Session) -> T): T {
+    return when {
+      options.maxAttempts > 1 -> {
+        val backoff = ExponentialBackoff(Duration.ZERO, Duration.ZERO)
+        retry(options.maxAttempts, backoff, ::isRetryable) { transactionInternal(lambda) }
+      }
+      options.maxAttempts == 1 -> transactionInternal(lambda)
+      else -> throw IllegalArgumentException()
```
swankjesse: does options enforce this invariant? If so this should throw an AssertionError()

AssertionError: “this should never happen, and I will not be able to write a test case for this”
IllegalArgumentException: “a caller who doesn’t follow the rules can cause this to happen and I will write a test case for this”

misk/src/main/kotlin/misk/hibernate/RealTransacter.kt
```diff
@@ -1,24 +1,44 @@
...
+    }
+  }
+
+  private fun <T> transactionInternal(lambda: (session: Session) -> T): T {
```
swankjesse: off topic: Is there a way to structure this so we don’t need outer methods and inner methods? I find this style does the trick but it’s awkward to debug

misk/src/main/kotlin/misk/hibernate/RealTransacter.kt
```diff
@@ -39,6 +59,16 @@ internal class RealTransacter(
     }
   }
 
+  override fun retries(numRetries: Int): Transacter {
```
swankjesse: can we always specify things in number of attempts rather than in number of retries? If the code says retries(3) I’m going to implicitly think 3 attempts, not 4 attempts.

swankjesse: also – should we impose some simplicity on the caller here? Perhaps “retries” and “no retries” should be our only options. I don’t like the idea of somebody specifying 2 attempts.

misk/src/main/kotlin/misk/hibernate/RealTransacter.kt
```diff
@@ -60,6 +90,22 @@ internal class RealTransacter(
     }
   }
 
+  private fun isRetryable(e: Exception): Boolean {
+    // TODO(jmuia): check the causal chain.
```
swankjesse: yeah, and a test case that confirms that we retry the right ones!

misk/src/main/kotlin/misk/hibernate/Transacter.kt
```diff
@@ -16,8 +16,21 @@ interface Transacter {
    * It is an error to start a transaction if another transaction is already in progress.
    */
   fun <T> transaction(lambda: (session: Session) -> T): T
+
+  fun retries(numRetries: Int): Transacter
+  fun noRetries(): Transacter = retries(0)
```
swankjesse: nice

misk/src/main/kotlin/misk/hibernate/Transacter.kt
```diff
@@ -16,8 +16,21 @@ interface Transacter {
...
 }
 
 fun Transacter.shards() = transaction { it.shards() }
 fun <T> Transacter.transaction(shard: Shard, lambda: (session: Session) -> T) =
     transaction { it.target(shard) { lambda(it) } }
+
+/**
+ * Thrown to explicitly trigger a retry, subject to retry limits and config such as noRetries().
+ */
+class RetryTransactionException : Exception {
```
swankjesse: nice!

misk/src/main/kotlin/misk/hibernate/Transacter.kt
```diff
@@ -16,8 +16,21 @@ interface Transacter {
...
+  constructor() : super()
+  constructor(message: String) : super(message)
+  constructor(message: String, cause: Throwable) : super(message, cause)
```
swankjesse: does the API let you specify defaults for these so you only need one overload?

misk/src/test/kotlin/misk/backoff/RetryTest.kt
```diff
@@ -23,6 +23,22 @@ internal class RetryTest {
...
+    }
+  }
+
+  @Test fun failsImmediatelyIfNotRetryable() {
+    val backoff = ExponentialBackoff(Duration.ofMillis(10), Duration.ofMillis(100))
+
+    val exception = assertThrows<IllegalStateException> {
+      retry(3, backoff, { it is RetryableException }) { retry ->
+        if (retry < 1) throw RetryableException()
+        throw IllegalStateException("failed on retry $retry")
```
swankjesse: nit: use an AssertionError to force the test to fail

misk/src/main/kotlin/misk/backoff/Retries.kt
swankjesse: Definitely worth discussion. Lets talk on Wednesday’s sync.

misk/src/main/kotlin/misk/backoff/Retries.kt
jmuia: Ah, I used a default lambda originally. The code is cleaner that way, but I wondered about performance compared to a null-check.

This retry function is currently used for sending logs to CloudWatch, but doesn't have special requirements. I will rewrite the retry logic in Transacter. I think you're right it will be simpler to do this directly, particularly if there are more bells and whistles to add (such as "overriding" the backoff based on specific db failures). If more use cases show up that have similar requirements, we can consider DRYing it up.

misk/src/main/kotlin/misk/hibernate/Transacter.kt
jmuia: it wasn't obvious at first, but I think I can do:
```
class RetryTransactionException(
  message: String? = null,
  cause: Throwable? = null
) : Exception(message, cause)
```
and it doesn't matter much because both values can be null anyways

misk/src/main/kotlin/misk/backoff/Retries.kt
swankjesse: 👍🏻

misk/src/main/kotlin/misk/hibernate/RealTransacter.kt
jmuia: I like that it adds a bit of friction, since it's important to pass the current thread local (and easy to forget with a default value).

I've also added a helper withOptions to make this less error-prone.

misk/src/main/kotlin/misk/hibernate/RealTransacter.kt
jmuia: fixed.

misk/src/main/kotlin/misk/hibernate/RealTransacter.kt
jmuia: fixed

misk/src/main/kotlin/misk/hibernate/RealTransacter.kt
jmuia: I haven't thought of an approach that doesn't make the code complex and deeply nested

misk-hibernate/src/main/kotlin/misk/hibernate/RealTransacter.kt
```diff
@@ -1,22 +1,68 @@
...
+  constructor(sessionFactory: SessionFactory, config: DataSourceConfig) :
+      this(sessionFactory, config, ThreadLocal(), TransacterOptions())
 
   override val inTransaction: Boolean
     get() = threadLocalSession.get() != null
 
   override fun <T> transaction(lambda: (session: Session) -> T): T {
+    assert(options.maxAttempts > 0)
+
+    val backoff = ExponentialBackoff(
```
jmuia: I'd like to make a different Backoff implementation, but this will do for now

misk-hibernate/src/main/kotlin/misk/hibernate/RealTransacter.kt
```diff
@@ -58,6 +111,25 @@ internal class RealTransacter(
...
+      else -> false
+    }
+  }
+
+  // NB: all options should be immutable types as copy() is shallow.
+  internal data class TransacterOptions(
+    val maxAttempts: Int = 2,
+    val minRetryDelayMillis: Long = 100,
+    val maxRetryDelayMillis: Long = 100,
+    val retryJitterMillis: Long = 400
```
jmuia: the ExponentialBackoff adds jitter to the backoff. So this will be [100,500)

misk-hibernate/src/main/kotlin/misk/hibernate/RealTransacter.kt
```diff
@@ -37,6 +83,13 @@ internal class RealTransacter(
     }
   }
 
+  override fun retries(): Transacter = withOptions(options.copy(maxAttempts = 2))
```
jmuia: I don't like that this duplicates the default value in a different place. May add a constant.

misk/src/main/kotlin/misk/hibernate/RealTransacter.kt
jmuia: I think I'll do this in a follow up

misk-hibernate/src/main/kotlin/misk/hibernate/RealTransacter.kt
```diff
@@ -58,6 +111,25 @@ internal class RealTransacter(
     }
   }
 
+  private fun isRetryable(e: Exception): Boolean {
+    // TODO(jmuia): check the causal chain.
```
swankjesse: I wonder if this is actually necessary

misk-hibernate/src/main/kotlin/misk/hibernate/Transacter.kt
```diff
@@ -13,11 +13,26 @@ interface Transacter {
...
    * If lambda raises an exception the transaction will be rolled back instead of committed.
    *
+   * If retries are permitted (the default), a failed but recoverable transaction will be
+   * reattempted after rolling back.
+   *
    * It is an error to start a transaction if another transaction is already in progress.
    */
   fun <T> transaction(lambda: (session: Session) -> T): T
+
+  fun retries(): Transacter
```
swankjesse: why would I call this?

jmuia: It provides an option to turn retries back on. Eg. you pass around a configured transacter (rather than injecting a "default" one) and want to ensure that retries are enabled.

misk-hibernate/src/main/kotlin/misk/hibernate/RealTransacter.kt
jmuia: I can remove the TODO and if, in practice, we see these exceptions wrapped up we can revisit implementing this check.

PR #277, opened 2018-06-22
UpstreamResourceInterceptor 

adrw:
- Intercepts requests and forwards them by `okhttp3` request to another serve such as the Webpack-Dev-Server
- Avoids need for `CORS` policies and headers since Misk server handles all requests


docs/UpstreamResourceInterceptor Spec.md
```diff
@@ -0,0 +1,98 @@
+# Admin Web App Resource Mapping Spec
```
swankjesse: When I was writing this in I was thinking of it as more of a whiteboard than a permanent piece of documentation. I think we should either clean it up to be a very short document explaining our routes, or delete it.

docs/UpstreamResourceInterceptor Spec.md
```diff
@@ -0,0 +1,98 @@
...
+
+
+| PATH                  | DEVELOPMENT                               | PRODUCTION                        |
+| --------------------- | ------------------------------------------| --------------------------------- |
+| /_admin/js/core.js    | http://localhost:3000/_admin/js/core.js   | misk.jar web/_admin/js/core.js    |
+| /_admin/config        | http://localhost:3000/_admin/             | misk.jar web/_admin/index.html    |
+| /_admin/guice         | http://localhost:3000/_admin/             | misk.jar web/_admin/index.html    |
+| /                     | http://localhost:3001/                    | urlshortener.jar web/index.html   |
+
+`<script src="/js/core.js">`
```
swankjesse: (as is this document doesn’t make sense!)

misk/src/main/kotlin/misk/config/AppName.kt
```diff
@@ -2,10 +2,11 @@ package misk.config
 
 import javax.inject.Qualifier
 
+//@TODO(jwilson swankjesse) make sure same targets everywhere
```
swankjesse: please let me have only one name

misk/src/main/kotlin/misk/config/ConfigAdminAction.kt
```diff
@@ -0,0 +1,27 @@
...
+import misk.web.Get
+import misk.web.RequestContentType
+import misk.web.ResponseContentType
+import misk.web.actions.WebAction
+import misk.web.mediatype.MediaTypes
+import javax.inject.Inject
+import javax.inject.Singleton
+
+@Singleton
+class ConfigAdminAction : WebAction {
```
swankjesse: rebase on the other PR once that’s merged?

misk/src/main/kotlin/misk/web/Request.kt
```diff
@@ -3,13 +3,53 @@ package misk.web
...
+  val method: HttpMethod = HttpMethod.GET,
   val headers: Headers = Headers.of(),
   val body: BufferedSource,
   val websocket: WebSocket? = null
 )
+
+fun Request.toOkHttp3(newUrl: HttpUrl?): Response {
+  val client = OkHttpClient()
+
+  val okUrl = if (newUrl != null) newUrl else this.url
```
swankjesse: is this ever non-null?

swankjesse: Maybe we should have two methods, one to convert a Request to an okhttp3.Request, and a separate method to change that request’s URL? As-is this method does too many things

misk/src/main/kotlin/misk/web/Request.kt
```diff
@@ -3,13 +3,53 @@ package misk.web
...
+  val okRequestBody = if (this.method == HttpMethod.GET) null else object : okhttp3.RequestBody() {
```
swankjesse: expand this if ? This statement is too awkward for a single line

swankjesse: add a TODO to fix the requests to not have a request body if they shouldn’t have a request body? GET is a good approximation here, but in practice it’s much more complicated than that. (HEAD has none, DELETE sometimes has a body, etc.)

misk/src/main/kotlin/misk/web/Request.kt
```diff
@@ -3,13 +3,53 @@ package misk.web
...
+    override fun contentType(): MediaType? = null
+    override fun writeTo(sink: BufferedSink) {
+      sink.writeAll(body)
+    }
+  }
+
+  val okRequest = okhttp3.Request.Builder()
+      .url(okUrl)
+      .method(this.method.asString(), okRequestBody)
```
swankjesse: off topic: APIs that use `asString()` instead of `toString()` are terrible

misk/src/main/kotlin/misk/web/Request.kt
```diff
@@ -3,13 +3,53 @@ package misk.web
...
+      .headers(this.headers)
+      .build()
```
swankjesse: hang on, there should be 3 methods here.

At this point we have our OkHttp3 request and this method should be done. There’s _another_ method here that executes that request, and a third method that converts the OkHttp response into a Misk response.

misk/src/main/kotlin/misk/web/Request.kt
```diff
@@ -3,13 +3,53 @@ package misk.web
...
+
+  try {
+    return client.newCall(okRequest).execute()
+  } catch (e: IOException) {
+    return Response.Builder()
```
swankjesse: would life be simpler if we never create an okhttp3.Response in this case? Ie. if we went directly to the Misk Response? Right now we’re creating an okhttp3.Response but we’re always going to immediately convert it to a Misk one anyway

misk/src/main/kotlin/misk/web/actions/UpstreamResourceInterceptor.kt
```diff
@@ -0,0 +1,69 @@
...
+class UpstreamResourceInterceptor(
+  private val mappings: MutableList<out Mapping>
+) : NetworkInterceptor {
+  @Suppress("UNUSED_PARAMETER")
+  override fun intercept(chain: NetworkChain): Response<*> {
+    val requestSegments = chain.request.url.pathSegments()
+    var matchedMapping: Mapping? = null
+
+    for (mapping in this.mappings) {
+      if (!pathSegmentsMatch(mapping.localPathPrefix.drop(1).dropLast(1).split('/'), requestSegments)) continue
```
swankjesse: I suspect the drops are a signal that something is wrong here. Can you just do

chain.request.url.encodedPath().startsWith(mapping.localPathPrefix)

swankjesse: ie. match on the strings rather than the list of path segments

misk/src/main/kotlin/misk/web/actions/UpstreamResourceInterceptor.kt
```diff
@@ -0,0 +1,69 @@
...
+      if (matchedMapping == null || mapping.localPathPrefix.count { ch -> ch == '/' } > matchedMapping.localPathPrefix.count { ch -> ch == '/' }) matchedMapping = mapping
+    }
+
+    if (matchedMapping == null) return chain.proceed(chain.request)
+
+    val upstreamUrl = matchedMapping.upstreamBaseUrl
+    val upstreamSegments = upstreamUrl.pathSegments()
+    val upstreamPlusRequestSegments = upstreamSegments.subList(0, upstreamSegments.size - 1) + requestSegments.subList(
```
swankjesse: I think this will also be easier as a string replacement. Just take the suffix of the request after the localPathPrefix and append that to the upstreamUrl

misk/src/test/kotlin/misk/web/actions/UpstreamResourceInterceptorTest.kt
```diff
@@ -0,0 +1,116 @@
...
+import org.junit.jupiter.api.Test
+import java.net.HttpURLConnection
+import kotlin.reflect.KFunction
+
+internal class UpstreamResourceInterceptorTest {
+  val upstreamServer = MockWebServer()
+  lateinit var mapping: UpstreamResourceInterceptor.Mapping
+  lateinit var interceptor: UpstreamResourceInterceptor
+
+  // TODO: probably discourage path rewriting
```
swankjesse: what action do we take on this?

docs/UpstreamResourceInterceptor Spec.md
adrw: Fair point, I think I'll delete it and bake the routing conversation into the eventual New Module docs.

misk/src/main/kotlin/misk/web/Request.kt
adrw: Good catch, added TODO and issue #279

misk/src/main/kotlin/misk/web/Request.kt
adrw: would `!!.toString()` be preferable which I think is what `asString` wraps into a single function?

misk/src/main/kotlin/misk/web/Request.kt
adrw: tradeoff between having single function do all the work vs more generic function that others could use. Though, is there a foreseeable use case where someone would convert a Misk request -> okHttp but then *not* return a Misk response from the okHttp response?

misk/src/main/kotlin/misk/web/Request.kt
adrw: Refactored into a few single purpose methods 👍

misk/src/main/kotlin/misk/web/Request.kt
adrw: Turns out they also still have `toString()` implemented so switched to that.

misk/src/main/kotlin/misk/web/Request.kt
adrw: Refactor now returns Misk response directly.

misk/src/main/kotlin/misk/web/actions/UpstreamResourceInterceptor.kt
adrw: Yeah, that makes a lot more sense 👍

misk/src/main/kotlin/misk/web/actions/UpstreamResourceInterceptor.kt
adrw: This too 👍

misk/src/test/kotlin/misk/web/actions/UpstreamResourceInterceptorTest.kt
adrw: Still thinking about this one, will post a full reply before PR is closed.

adrw: The current implementation works as follows
`localhost:8080/_admin/config/js/config.js` matches on `/_admin/config` and forwards to `localhost:3001/{upstreamPrefix}/_admin/config/js/config.js`.

Since the chain URL suffix is simply added to the upstreamPrefix there is no possibility of complex rewriting of the URL suffix (which I think was the goal of the stated todo).

If we wanted to ensure that the upstreamPrefix is always empty and thus `localhost:8080/{match}/` maps directly to `localhost:3001/{match}/` then we could set a requirement that the upstream URL always has 0 path segments. Though I don't think this would likely impact the Webpack-Dev-Server use of the class, I'm not sure whether the additional requirement would cause issues in later use if we wanted to map some links to `localhost:3001/build/{match}` for example. 

For now, should we could add in the additional requirement and remove it if it becomes a blocker?

misk/src/main/kotlin/misk/web/Request.kt
swankjesse: yay

misk/src/test/kotlin/misk/web/actions/UpstreamResourceInterceptorTest.kt
swankjesse: Not sure I follow.

misk/src/main/kotlin/misk/web/Request.kt
```diff
@@ -3,13 +3,69 @@ package misk.web
...
   val url: HttpUrl,
-  val method: HttpMethod,
+  val method: HttpMethod = HttpMethod.GET,
   val headers: Headers = Headers.of(),
   val body: BufferedSource,
   val websocket: WebSocket? = null
 )
+
+fun okhttp3.Request.exec(): okhttp3.Response {
+  val client = OkHttpClient()
```
swankjesse: for follow up – we absolutely need to inject this. Creating a new instance is super expensive.

misk/src/main/kotlin/misk/web/Request.kt
```diff
@@ -3,13 +3,69 @@ package misk.web
...
+        .addHeader("Content-Type", "text/plain; charset=utf-8")
+        .body(ResponseBody.create(MediaType.parse("Failed to fetch upstream URL ${this.url()}"), "Failed to fetch upstream URL ${this.url()}"))
+        .message("Failed to fetch upstream URL ${this.url()}")
+        .code(HttpURLConnection.HTTP_UNAVAILABLE)
+        .build()
+  }
+}
+
+fun Request.forwardRequestTo(proxyUrl: HttpUrl): misk.web.Response<*> {
+  return this.toOkHttp3().setUrl(proxyUrl).exec().toMisk()
```
swankjesse: nit: we prefer to omit `this.` prefixes unless they’re required. 

Shipley’s law!
https://twitter.com/jessewilson/status/417478030806622208?lang=en

misk/src/main/kotlin/misk/web/Request.kt
```diff
@@ -3,13 +3,69 @@ package misk.web
...
+  return okhttp3.Request.Builder()
+      .url(newUrl)
+      .method(this.method(), this.body())
+      .headers(this.headers())
+      .build()
+}
+
+fun Request.toOkHttp3(): okhttp3.Request {
+  // @TODO(adrw) https://github.com/square/misk/issues/279
+  val okRequestBody = if (this.method == HttpMethod.GET) null
```
swankjesse: style nit: I think both `if` and `else` should have braces or neither

misk/src/main/kotlin/misk/web/Response.kt
```diff
@@ -31,3 +32,12 @@ fun Response<ResponseBody>.writeToJettyResponse(jettyResponse: HttpServletRespon
 }
 
 private fun HttpServletResponse.bufferedSink() = Okio.buffer(Okio.sink(outputStream))
+
+fun Response.toMisk() : misk.web.Response<*> {
```
swankjesse: nice

misk/src/main/kotlin/misk/web/actions/UpstreamResourceInterceptor.kt
```diff
@@ -0,0 +1,49 @@
...
+import misk.web.Response
+import misk.web.forwardRequestTo
+import okhttp3.HttpUrl
+
+class UpstreamResourceInterceptor(
+  private val mappings: MutableList<out Mapping>
+) : NetworkInterceptor {
+  @Suppress("UNUSED_PARAMETER")
+  override fun intercept(chain: NetworkChain): Response<*> {
+    var matchedMapping = findMappingMatch(chain) ?: return chain.proceed(chain.request)
```
swankjesse: cute

misk/src/main/kotlin/misk/web/actions/UpstreamResourceInterceptor.kt
```diff
@@ -0,0 +1,49 @@
...
+import misk.web.NetworkChain
+import misk.web.NetworkInterceptor
+import misk.web.Response
+import misk.web.forwardRequestTo
+import okhttp3.HttpUrl
+
+class UpstreamResourceInterceptor(
+  private val mappings: MutableList<out Mapping>
+) : NetworkInterceptor {
+  @Suppress("UNUSED_PARAMETER")
```
swankjesse: is this still required?

misk/src/main/kotlin/misk/web/actions/UpstreamResourceInterceptor.kt
```diff
@@ -0,0 +1,49 @@
...
+  override fun intercept(chain: NetworkChain): Response<*> {
+    var matchedMapping = findMappingMatch(chain) ?: return chain.proceed(chain.request)
+    val proxyUrl = HttpUrl.parse(
+        matchedMapping.upstreamBaseUrl.toString() + chain.request.url.encodedPath().removePrefix(
```
swankjesse: this is awesome

misk/src/main/kotlin/misk/web/actions/UpstreamResourceInterceptor.kt
```diff
@@ -0,0 +1,49 @@
...
+            matchedMapping.localPathPrefix))!!
+    return chain.request.forwardRequestTo(proxyUrl)
+  }
+
+  private fun findMappingMatch(chain: NetworkChain): Mapping? {
+    var matchedMapping: Mapping? = null
+    for (mapping in mappings) {
+      if (!chain.request.url.encodedPath().startsWith(mapping.localPathPrefix)) continue
+      if (matchedMapping == null || mapping.localPathPrefix.count { ch -> ch == '/' } > matchedMapping.localPathPrefix.count { ch -> ch == '/' }) matchedMapping =
```
swankjesse: it looks like you’re trying to handle the case where multiple mappings match. I think we should forbid this and just return the first path that matches

Maybe we should confirm that none of the mappings’ localPathPrefix is a prefix of any others

misk/src/main/kotlin/misk/web/actions/UpstreamResourceInterceptor.kt
```diff
@@ -0,0 +1,49 @@
...
+          mapping
+    }
+    return matchedMapping
+  }
+
+  /**
+   * Imagine that we had the following Mapping:
```
swankjesse: Please rephrase this to match the readers’ perspective.

“Maps URLs requested against this server to URLs of servers to delegate to”

misk/src/test/kotlin/misk/web/actions/UpstreamResourceInterceptorTest.kt
adrw: My bad, take 2:

If we wanted to eliminate all URL rewriting we can just force that the upstreamBaseURL is always the `host` + `localPathPrefix`. This way, the incoming links match on the `localPathPrefix` and are simply forwarded to the right `host`, with no URL rewriting.

misk/src/main/kotlin/misk/web/actions/UpstreamResourceInterceptor.kt
adrw: I'd be okay forbidding outright but maximal matching is likely necessary for the following case.

`/_admin/*`  -> `localhost:3001`: generic mapping for all _admin actions
`/_admin/config/` -> `localhost:3001`: the index page for the `config` module
`/_admin/config/all` -> `localhost:8080`: all `config` domain API calls should be handled by Misk, not forwarded to webpack

Options
1) Leave rules as maximal matching, though would need to be well documented and potentially add in `*` matching so all API calls don't need to be explicitly specified as a matching.
2) Enforce non-overlapping first prefix matching and re-attempt all failed forwarding URLs on Misk server. Potential problem is how to handle redirecting `/_admin/`which contains the main `index.html` and chrome/sidebar UI but would be a prefix for all modules.

misk/src/test/kotlin/misk/web/actions/UpstreamResourceInterceptorTest.kt
swankjesse: Gotcha. I don’t feel strongly!

misk/src/main/kotlin/misk/web/actions/UpstreamResourceInterceptor.kt
swankjesse: I think we want to minimize how much configuration is required. If possible we make it so the web and Misk agree on paths so there’s less rewriting.

adrw: Agreed 👍

PR #280, opened 2018-06-24
Switch to the standard shadowJar task

swankjesse:


samples/exemplar/build.gradle
```diff
@@ -37,18 +39,11 @@ jar {
   classifier = 'unshaded'
 }
 
-apply from: rootProject.file('gradle/gradle-mvn-push.gradle')
-apply plugin: 'com.github.johnrengelman.shadow'
-
-task shadedJar(type: com.github.jengelman.gradle.plugins.shadow.tasks.ShadowJar) {
-  classifier = null
```
jmuia: this will now be `all` by default

swankjesse: yeah, I think that’s probably fine.

PR #282, opened 2018-06-25
Run Vitess tests in Travis

tirsen:


tirsen:
WIP: Just testing it out running on Travis

misk-hibernate/src/test/kotlin/misk/hibernate/ReflectionQueryFactoryTest.kt
```diff
@@ -42,12 +42,12 @@ class ReflectionQueryFactoryTest {
       assertThat(queryFactory.newQuery<OperatorsMovieQuery>()
           .releaseDateLessThan(m3.releaseDate)
           .listAsNameAndReleaseDate(session))
-          .containsExactly(m1, m2)
+          .containsExactlyInAnyOrder(m1, m2)
```
tirsen: There's no "default order" for cross shard queries in Vitess.

build.gradle
```diff
@@ -1,5 +1,5 @@
 ext {
-  kotlinVersion = '1.2.50'
+  kotlinVersion = '1.2.51'
```
tirsen: My Kotlin plugin complained about this but if you want i can take it out of the review.

build.gradle
```diff
@@ -69,6 +69,7 @@ ext {
           mockitoCore          : 'org.mockito:mockito-core:2.8.9',
           jaxbApi              : 'javax.xml.bind:jaxb-api:2.3.0',
           mysql                : 'mysql:mysql-connector-java:5.1.16',
+          vitess               : 'io.vitess:vitess-jdbc:2.2.0'
```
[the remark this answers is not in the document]
swankjesse: Tracking bug?

misk-hibernate-testing/src/main/kotlin/misk/hibernate/HibernateTestingModule.kt
```diff
@@ -2,6 +2,7 @@ package misk.hibernate
 
 import com.google.common.util.concurrent.Service
 import misk.inject.KAbstractModule
+import misk.inject.asSingleton
```
swankjesse: ?

misk-hibernate-testing/src/main/kotlin/misk/hibernate/TruncateTablesService.kt
```diff
@@ -35,8 +38,11 @@ internal class TruncateTablesService(
   override val producedKeys = setOf<Key<*>>()
 
   override fun startUp() {
-    truncateUserTables()
-    executeStatements(startUpStatements, "startup")
+    // This is generally the first thing that happens in a test, if this fails it may be because Vitess isn't ready yet so we retry it a few times.
+    retry(10, ExponentialBackoff(Duration.ofMillis(20), Duration.ofMillis(1000))) {
```
swankjesse: This is the wrong place. The service that sets up the Vitess connection should do this before it self-reports as healthy.

misk-hibernate/src/main/kotlin/misk/hibernate/JdbcExtensions.kt
```diff
@@ -0,0 +1,15 @@
+package misk.hibernate
+
+import com.google.common.collect.Iterables.getOnlyElement
+import java.sql.ResultSet
+
+fun <T> ResultSet.map(function: (ResultSet) -> T): List<T> {
```
swankjesse: cute

swankjesse: Should this be `internal` ? I’m reluctant to publish APIs on JDBC types

misk-hibernate/src/main/kotlin/misk/hibernate/JdbcExtensions.kt
```diff
@@ -0,0 +1,15 @@
...
+  val result = mutableListOf<T>()
+  while (this.next()) {
+    result.add(function(this))
+  }
+  return result
+}
+
+fun <T> ResultSet.single(function: (ResultSet) -> T): T = getOnlyElement(map(function))
```
swankjesse: same

swankjesse: Also – elsewhere we have a method uniqueResult() that returns a `T?` . . . I think this should either be `internal` or be symmetric with that method

misk-hibernate/src/main/kotlin/misk/hibernate/RealTransacter.kt
```diff
@@ -166,12 +166,14 @@ internal class RealTransacter private constructor(
...
       if (config.type == DataSourceType.VITESS) {
         return useConnection { connection ->
-          val previousTarget = connection.createStatement().use { statement ->
-            val rs = statement.executeQuery("SELECT database()")
-            check(rs.next())
-            val previousTarget = rs.getString(1)
+          // This broke on master Vitess so we temporarily disable it.
+          val previousTarget = ""
+          // TODO re-enable this when it's working again: https://github.com/vitessio/vitess/pull/3962
+//          val previousTarget = connection.createStatement().use { statement ->
```
swankjesse: nit: delete code or use `if (false)` rather than commenting it out. It’s very scary to uncomment commented-out code because it doesn’t participate in refactorings.

misk-hibernate/src/test/kotlin/misk/hibernate/MoviesTestModule.kt
```diff
@@ -17,7 +17,7 @@ class MoviesTestModule : KAbstractModule() {
     install(Modules.override(MiskServiceModule()).with(FakeClockModule()))
     install(EnvironmentModule(Environment.TESTING))
 
-    val config = MiskConfig.load<MoviesConfig>("moviestestmodule", Environment.TESTING)
+    val config = MiskConfig.load<MoviesConfig>("moviestestmodule_vitess", Environment.TESTING)
```
swankjesse: this seems off

Either Vitess should be the default and the file should be called moviestestmodule, or Vitess shouldn’t be the default and we should have a separate thing called `MoviesVitessTestModule`

misk-hibernate/src/test/kotlin/misk/hibernate/ReflectionQueryFactoryTest.kt
```diff
@@ -324,7 +324,7 @@ class ReflectionQueryFactoryTest {
       assertThat(queryFactory.newQuery<OperatorsMovieQuery>()
           .apply { maxRows = 4 }
           .listAsNames(session))
-          .containsExactly("Rocky 1", "Rocky 2", "Rocky 3")
+          .hasSize(3)
```
swankjesse: For me for follow up: 
https://github.com/square/misk/issues/346

misk-hibernate/src/test/kotlin/misk/hibernate/TransacterTest.kt
```diff
@@ -78,6 +80,7 @@ class TransacterTest {
   }
 
   @Test
+  @Disabled("Uniqueness constraints aren't reliably enforced on Vitess")
```
swankjesse: Yep. We have to create a new test that is reliable under Vitess.

swankjesse:
# COOL

misk-hibernate/src/test/kotlin/misk/hibernate/MoviesTestModule.kt
tirsen: Yeah you're right. Shall I just make Vitess the default?

We could run some form of Travis matrix build too and run tests against both Vitess and MySQL if we want to.

tirsen: Note to self: we can use the Travis `env` configuration option to permute on a "database" option for our build matrix. https://docs.travis-ci.com/user/customizing-the-build/#Build-Matrix

Long term I think we should only run tests on Vitess though but maybe intermediately we should support both Vitess and MySQL?

misk-hibernate-testing/src/main/kotlin/misk/hibernate/TruncateTablesService.kt
tirsen: Done. I added a PingDatabaseService that the SessionFactoryService depends on.

misk-hibernate/src/main/kotlin/misk/hibernate/JdbcExtensions.kt
tirsen: Can't be internal alas :-( because it's also used in the misk-hibernate-testing module (which has the same package but different module so can't access internal :-( )

misk-hibernate/src/main/kotlin/misk/hibernate/JdbcExtensions.kt
tirsen: Thanks! I was trying to find a precedent to be symmetrical with. :-)

misk-hibernate/src/main/kotlin/misk/hibernate/RealTransacter.kt
tirsen: Nice! DOne!

misk-hibernate/src/test/kotlin/misk/hibernate/MoviesTestModule.kt
tirsen: Made Vitess default for now. Later on I can add a MySQL build matrix dimension.

misk-hibernate/src/test/kotlin/misk/hibernate/TransacterTest.kt
tirsen: Yeah not sure how to do that... :-) Uniqueness constraints isn't something that scales out well. The best alternative is cross shard leases. Maybe we should have something built-in that supports that pattern.

misk-hibernate/src/main/kotlin/misk/hibernate/JdbcExtensions.kt
swankjesse: Gotcha.

misk-hibernate/src/test/kotlin/misk/hibernate/MoviesTestModule.kt
swankjesse: Perfect.

misk-hibernate/src/test/kotlin/misk/hibernate/TransacterTest.kt
swankjesse: Yeah I think the goal of the test wasn't uniqueness, just something that'd fail in SQL. Perhaps we could force an error by duplicating a primary key.

tirsen: Gotcha. I'll figure something out...

PR #292, opened 2018-07-03
Implement support for boxed tokens.

swankjesse:
This was a lot of work mostly because Hibernate initialization is a bear.

Closes: https://github.com/square/misk/issues/234

### DIFF
--- a/misk-hibernate/src/main/kotlin/misk/hibernate/BoxedStringType.kt
+++ b/misk-hibernate/src/main/kotlin/misk/hibernate/BoxedStringType.kt
@@ -1,0 +1,105 @@
+package misk.hibernate
+
+import org.hibernate.engine.spi.SharedSessionContractImplementor
+import org.hibernate.usertype.UserType
+import java.io.Serializable
+import java.sql.PreparedStatement
+import java.sql.ResultSet
+import java.sql.Types
+import kotlin.reflect.KClass
+import kotlin.reflect.KFunction
+import kotlin.reflect.KProperty1
+import kotlin.reflect.full.memberProperties
+import kotlin.reflect.full.primaryConstructor
+
+/**
+ * Binds any data class that wraps a String to a varchar in MySQL. This is most useful to enforce
+ * additional type-safety on identifier columns.
+ *
+ * To use, create a data class with a single non-null `String` property. The name of this property
+ * is not significant.
+ *
+ * ```
+ * data class MovieToken(val token: String)
+ * ```
+ *
+ * To use with comparison operators (ie. WHERE clauses with `<` or `>`) you must also implement
+ * `Comparable`:
+ *
+ * ```
+ * data class MovieToken(val token: String) : Comparable<MovieToken> {
+ *   override fun compareTo(other: MovieToken) = token.compareTo(other.token)
+ * }
+ * ```
+ */
+internal class BoxedStringType<T : Any> private constructor(
+  private val property: KProperty1<T, String>,
+  private val constructor: KFunction<T>
+) : UserType {
+  override fun hashCode(x: Any) = x.hashCode()
+
+  override fun deepCopy(value: Any) = value
+
+  override fun replace(original: Any, target: Any, owner: Any?) = original
+
+  override fun equals(x: Any, y: Any) = x == y
+
+  override fun returnedClass() = constructor.returnType.javaClass
+
+  override fun assemble(cached: Serializable, owner: Any?) = constructor.call(cached)
+
+  @Suppress("UNCHECKED_CAST") // Hibernate promises to call us only with the types we support.
+  override fun disassemble(value: Any?) = property.get(value as T)
+
+  override fun nullSafeSet(
+    st: PreparedStatement,
+    value: Any?,
+    index: Int,
+    session: SharedSessionContractImplementor?
+  ) {
+    if (value == null) {
+      st.setNull(index, Types.VARCHAR)
+    } else {
+      st.setString(index, disassemble(value))
+    }
+  }
+
+  override fun nullSafeGet(
+    rs: ResultSet,
+    names: Array<out String>,
+    session: SharedSessionContractImplementor?,
+    owner: Any?
+  ): Any? {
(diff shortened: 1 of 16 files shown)


misk/src/main/kotlin/misk/hibernate/TokenType.kt
```diff
@@ -0,0 +1,105 @@
...
+ * To use with comparison operators (ie. WHERE clauses with `<` or `>`) you must also implement
+ * `Comparable`:
+ *
+ * ```
+ * data class MovieToken(val token: String) : Comparable<MovieToken> {
+ *   override fun compareTo(other: MovieToken) = token.compareTo(other.token)
+ * }
+ * ```
+ */
+internal class TokenType<T : Any> private constructor(
```
mmihic: Oh I love this. So much cleaner compared to the approach in Java

misk-hibernate/src/main/kotlin/misk/hibernate/TokenType.kt
```diff
@@ -0,0 +1,105 @@
...
+import kotlin.reflect.KFunction
+import kotlin.reflect.KProperty1
+import kotlin.reflect.full.memberProperties
+import kotlin.reflect.full.primaryConstructor
+
+/**
+ * Binds any data class that wraps a String to a varchar in MySQL. This is most useful to enforce
+ * additional type-safety on identifier columns.
+ *
+ * To use, create a data class with a single non-null `String` property. The name of this property
```
jmuia: Is non-null actually a requirement? What happens if the data type is non-null but the database value is null?

misk-hibernate/src/main/kotlin/misk/hibernate/TokenType.kt
```diff
@@ -0,0 +1,105 @@
...
+ * is not significant.
+ *
+ * ```
+ * data class MovieToken(val token: String)
+ * ```
+ *
+ * To use with comparison operators (ie. WHERE clauses with `<` or `>`) you must also implement
+ * `Comparable`:
```
jmuia: Given the requirement that the class boxes a string, should we just do this? Or at least enforce it (check in TokenType.create()).

misk-hibernate/src/main/kotlin/misk/hibernate/TokenType.kt
```diff
@@ -0,0 +1,105 @@
...
+ *
+ * ```
+ * data class MovieToken(val token: String) : Comparable<MovieToken> {
+ *   override fun compareTo(other: MovieToken) = token.compareTo(other.token)
+ * }
+ * ```
+ */
+internal class TokenType<T : Any> private constructor(
```
jmuia: Since this is general purpose (rather than specific to tokens), maybe call it as it is: `BoxedStringType` or something.

misk-hibernate/src/main/kotlin/misk/hibernate/SessionFactoryService.kt
```diff
@@ -104,19 +111,73 @@ internal class SessionFactoryService(
...
     val metadataSources = MetadataSources(registry)
     for (entityClass in entityClasses) {
       metadataSources.addAnnotatedClass(entityClass.entity.java)
     }
-    val metadataBuilder = metadataSources.metadataBuilder
-    metadataBuilder.applyBasicType(IdType, Id::class.qualifiedName)
-    metadataBuilder.applyBasicType(ByteStringType, ByteString::class.qualifiedName)
-    val metadata = metadataBuilder.build()
+
+    val metadataBuilder: MetadataBuilder = metadataSources.metadataBuilder
```
jmuia: nit: the metadataBuilder variable is no longer used (aside from calling build() immediately)

misk-hibernate/src/main/kotlin/misk/hibernate/TokenType.kt
swankjesse: In that case we want to get a null token, not a non-null token with a null string inside.

Ie. you get a `val GoodLuckToken?` which has a non-null string field, not a `val GoodLuckToken` that has a nullable string field.

misk-hibernate/src/main/kotlin/misk/hibernate/TokenType.kt
swankjesse: I think it’s awkward for us to do this automatically.

swankjesse: (Hibernate itself wants to cast these things to Comparable)

misk-hibernate/src/main/kotlin/misk/hibernate/TokenType.kt
swankjesse: Fixed.

misk-hibernate/src/main/kotlin/misk/hibernate/SessionFactoryService.kt
swankjesse: Added a comment to note how scary Hibernate is!

misk-hibernate/src/main/kotlin/misk/hibernate/BoxedStringType.kt
```diff
@@ -0,0 +1,105 @@
...
+      val memberPropertiesList = propertyType.memberProperties.toList()
+      if (memberPropertiesList.size != 1) return null
+
+      val onlyProperty: KProperty1<T, *> = memberPropertiesList[0]
+      if (onlyProperty.returnType.classifier != String::class) return null
+      @Suppress("UNCHECKED_CAST") // Guarded by a runtime check above.
+      val stringProperty = onlyProperty as KProperty1<T, String>
+
+      val constructor: KFunction<T>? = propertyType.primaryConstructor
+      if (constructor == null) return null // No primary constructor.
```
jmuia: nit: could be kotlin-y and use a `return when`

swankjesse: Will try in follow up.

PR #294, opened 2018-07-03
Extract Hibernate and database related code to separate package

jmuia:


misk/hibernate/src/main/kotlin/misk/hibernate/SchemaMigrator.kt
```diff
@@ -37,10 +36,7 @@ internal class SchemaMigrator(
 ) {
   /** Returns a map from version to path. */
   fun availableMigrations(): NavigableMap<Int, String> {
-    if (config.migrations_path == null) {
-      return TreeMap()
-    }
-    val resources = resourceLoader.list(config.migrations_path)
```
jmuia: I've refactored this as smart casts do not work across packages (an alternative would be `val resources = resourceLoader.list(config.migrations_path!!`)

settings.gradle
```diff
@@ -1,4 +1,6 @@
 include ':misk/testing'
```
jmuia: I'd like to standardize this to colons over slashes.

jmuia:
I tried my best to avoid redundant/unnecessary dependencies when moving them around. I'll find and apply a tool to do this automatically in a separate PR.

misk/hibernate/gradle.properties
```diff
@@ -0,0 +1,4 @@
+POM_ARTIFACT_ID=misk-hibernate
+POM_NAME=Misk
+POM_DESCRIPTION=An open source application container from Square
```
swankjesse: mind renaming to `misk-hibernate` ?

PR #298, opened 2018-07-04
Always use ':' in project names

swankjesse:


swankjesse:
Turns out that Gradle imposes a relationship between a project and its enclosing project, and this gets in the way. I’m going to flatten the directory structure so we now have these:

```
misk
misk-hibernate
misk-hibernate-testing
misk-testing
```

The small upside is there’s fewer places to remember whether to use a `:` or a `-` in an identifier.

jmuia:
👍

PR #300, opened 2018-07-04
Use TokenGenerator in the UrlShortener

swankjesse:


samples/urlshortener/src/main/resources/urlshortener-testing.yaml
```diff
@@ -0,0 +1,5 @@
+data_source_cluster:
+  writer:
+    database: urls_testing
+
+
```
jmuia: super nit: too many empty lines

PR #302, opened 2018-07-05
FrintJS abandoned, Wiring Webpack Part 1

adrw:
- Nuked most existing code because dead ends were hit with FrintJS being unable to handle hot reloads within dynamically loaded tabs
- First principles plumbing tests started and passed to ensure that hot reloads and forwarding through Misk work from the beginning
  - Forwarding all requests through Misk :8080 now works under testing
  - Loading modules with hot reloads while editing with all requests redirecting through Misk :8080 now work
- Linting added with tslint to ensure better JS code standards
- Dashboard "chrome / UI" is no longer a special tab/module but can be added to any group of tabs arbitrarily
- `modules` are now known as `tabs`. Tabs are a single full dashboard set of views that contain the necessary _admin information for a specific module. Example tabs are config, logging, threads... similar paradigm to the current _admin
- Next PRs will have dashboard and config tabs fully fleshed out and reimplemented to ensure hot reloading while editing and showing of live data from the Misk server
- Added nomenclature around Misk _admin endpoints that they are denoted by `_admin/api/{tab}/endpoint` for easier forwarding rules so that all `_admin/api/*` are not forwarded but are handled by Misk
- Fix Todo re: discouraging path rewriting. 
  - UpstreamResourceInterceptor tests rewritten to reflect that the upstream prefix and local prefix (local is what incoming request URL is matched on) are identical always. 
  - Thus every request forwarded Webpack is received by Webpack with the same url structure as is navigated. All tab developers then will be working on the same url domain as they've been given.
  - Ex. intercepted request `_admin/config/` will be forwarded to config webpack which will receive a request for `_admin/config`. Prior to this, it could receive a request for any arbitrary prefix set in the configured mapping in UpstreamResourceInterceptor
- Enforce that all mappings must end with trailing slash. To access config, url must be `_admin/config/`, not `_admin/config` to prevent prefix matching issues


misk/src/main/kotlin/misk/web/actions/UpstreamResourceInterceptor.kt
```diff
@@ -72,4 +78,14 @@ class UpstreamResourceInterceptor(
           upstreamBaseUrl.encodedPath().endsWith("/"))
     }
   }
+
+  class Factory @Inject internal constructor() : NetworkInterceptor.Factory {
+    override fun create(action: Action): NetworkInterceptor? {
+      return UpstreamResourceInterceptor(okhttp3.OkHttpClient(), mutableListOf<Mapping>(
+          UpstreamResourceInterceptor.Mapping("/_admin/test/", HttpUrl.parse("http://localhost:8000/")!!),
```
mmihic: Does this mean that we need to have production services listening on both `http` and `https`? What's port 3100?

misk/src/main/kotlin/misk/web/actions/UpstreamResourceInterceptor.kt
```diff
@@ -72,4 +78,14 @@ class UpstreamResourceInterceptor(
...
+          UpstreamResourceInterceptor.Mapping("/_admin/dashboard/", HttpUrl.parse("http://localhost:3100/")!!),
```
mmihic: Since this list is in misk internal code, how do applications add their own dashboards here?

adrw: This hasn't been built yet but mappings will be dynamically built when Misk evaluates all the injected modules, which ones have web `_admin` components, and then based on likely a simple YAML file adds the mappings. The YAML file will contain keys for the resource location when in development (url with port to the webpack dev server) and in production (pulled from within the Misk JAR file) and which mode it currently is in.

misk/src/main/kotlin/misk/web/actions/UpstreamResourceInterceptor.kt
adrw: Misk will only ever listen on whatever standard port we determine (currently `:8080`). These ports are the upstream ones. `:3[0-9][0-9][0-9]` ports are live webpack dev servers for dashboard tabs in development.

mmihic: Gotcha. So we'll probably need to have this look at the ``WebConfig`` to figure out what port/protocol we're actually running. FWI in staging/production misk is misk is going to be listening only on port `8443`.

I don't know anything about webpack - are these dev servers embedded into the misk process?

adrw: Ah okay, thanks for the heads up on `8443`.

Webpack will be run independently of Misk on developer local machines and are distinct NodeJS servers. Still trying to determine how much automation we bake in to Misk for starting these servers if the YAML says that a tab is in development mode vs. developers are responsible for starting and handling the webpack servers themselves. Leaning towards putting it in developers hands so any webpack specific issues or development is easier to debug or search on Stack Overflow because it doesn't have a Misk wrapper or composer manipulating it.

misk/src/main/kotlin/misk/web/actions/UpstreamResourceInterceptor.kt
```diff
@@ -11,17 +12,32 @@ import okhttp3.HttpUrl
...
+ * Rules
+ * - No overlapping mapping prefixes
+ *    "_admin/config/" and "_admin/config/subtab/" will not resolve consistently
+ *
+ * - All upstreamBaseUrl ends with "/"
+ * - All local prefix mappings end with "/"
+ *
+ *
+ */
+
```
swankjesse: nit: we don’t like unnecessary blank lines

misk/src/main/kotlin/misk/web/actions/UpstreamResourceInterceptor.kt
swankjesse: FWIW, this thing is only for development.

misk/src/main/kotlin/misk/web/actions/UpstreamResourceInterceptor.kt
```diff
@@ -11,17 +12,32 @@ import okhttp3.HttpUrl
...
   private val mappings: List<Mapping>
 ) : NetworkInterceptor {
-//  TODO(adrw) enforce no prefix overlapping
+//  TODO(adrw) enforce no prefix overlapping https://github.com/square/misk/issues/303
   override fun intercept(chain: NetworkChain): Response<*> {
     val matchedMapping = findMappingMatch(chain) ?: return chain.proceed(chain.request)
     val proxyUrl = HttpUrl.parse(
-        matchedMapping.upstreamBaseUrl.toString() + chain.request.url.encodedPath().removePrefix(
-            matchedMapping.localPathPrefix))!!
+        matchedMapping.upstreamBaseUrl.toString().dropLast(1)
```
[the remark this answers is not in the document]
adrw: Yes, I like that 👍

misk/src/test/kotlin/misk/web/actions/UpstreamResourceInterceptorTest.kt
```diff
@@ -43,7 +42,74 @@ internal class UpstreamResourceInterceptorTest {
   }
 
   @Test
-  internal fun requestForwardedToUpstreamServerIfPathMatches() {
+  internal fun mappingFailsIfLocalWithoutLeadingSlash() {
+    var failed = false
+
+    try {
+      mappingTestMapping = UpstreamResourceInterceptor.Mapping(
```
swankjesse: use assertThrows ? some examples in the project

misk/src/test/kotlin/misk/web/actions/UpstreamResourceInterceptorTest.kt
```diff
@@ -43,7 +42,74 @@ internal class UpstreamResourceInterceptorTest {
...
+      failed = true
+    }
+
+    assertThat(failed).isTrue()
+  }
+
+  @Test
+  internal fun requestForwardedToUpstreamServerIfPathMatchesWithTrailingSlash() {
+    upstreamServer.enqueue(MockResponse()
+        .setResponseCode(404)
```
swankjesse: nit: 404 is an awkward choice here, because I think that’s also the code we’ll use if it isn’t intercepted?

misk/web/tabs/config/tslint.json
```diff
@@ -0,0 +1,25 @@
...
+    "tslint-config-prettier", 
+    "tslint-consistent-codestyle",
+    "tslint-eslint-rules", 
+    "tslint-immutable",
+    "tslint-react",
+    "tslint-sonarts"
+  ],
+  "jsRules": {},
+  "rules": {
+    // "interface-name": false,
```
swankjesse: nit: comment why these are commented? Or nuke ’em?

misk/web/tabs/config/webpack.config.js
```diff
@@ -0,0 +1,70 @@
...
+  output: {
+    filename: 'tab_config.js',
+    path: path.join(__dirname, 'dist/_admin/config'),
+    publicPath: "/_admin/config/"
+  },
+  devServer: {
+    // host: 'localhost',
+    port: '3200',
+    hot: true,
+    // headers: {
```
swankjesse: comment why these are commented? Or nuke ’em?

misk/src/test/kotlin/misk/web/actions/UpstreamResourceInterceptorTest.kt
adrw: Yeah we'll change that, already had issues where tests had false positives/negatives because of that.

misk/web/tabs/config/tslint.json
adrw: Still ironing out what a good, practical linting config is going forward but will remove those soon

misk/web/tabs/config/webpack.config.js
adrw: Yeah I'll nuke them

PR #306, opened 2018-07-05
Be more consistent about how we specify dependencies

swankjesse:


gradle.properties
```diff
@@ -1,6 +1,6 @@
 GROUP=com.squareup.misk
 
-VERSION_NAME=0.4.0-SNAPSHOT
+VERSION_NAME=0.9.0-SNAPSHOT
```
adrw: This PR is mostly just version bump and whitespace fixes?

swankjesse: you got it

PR #307, opened 2018-07-05
Split remaining service dependencies into their own modules

swankjesse:
This is AWS, Graphite, Jaeger, SignalFx, and Zipkin.

The Zipkin/Jaeger change is unsatisfying because we used to support choosing a tracer
at runtime depending on what was in the config. This was nice behavior but it meant
that every application had the union of all tracing dependencies in their binary.
I'm going to try to find a nicer way in follow-up.


misk/src/main/kotlin/misk/tracing/TracingModule.kt
```diff
@@ -1,31 +0,0 @@
...
-import misk.logging.getLogger
-import misk.tracing.backends.TracingBackendConfig
-import misk.tracing.backends.jaeger.JaegerBackendModule
-import misk.tracing.backends.noop.NoopTracerBackendModule
-import misk.tracing.backends.zipkin.ZipkinBackendModule
-import kotlin.reflect.full.memberProperties
-
-private val logger = getLogger<TracingModule>()
-
-class TracingModule(val config: TracingConfig) : KAbstractModule() {
```
swankjesse: Removed without replacement. I’m going to try to circle back to this.
https://github.com/square/misk/issues/308

misk/src/main/kotlin/misk/tracing/backends/noop/NoopTracingModule.kt
```diff
@@ -6,7 +6,7 @@ import io.opentracing.Tracer
 import io.opentracing.noop.NoopTracerFactory
 import misk.inject.KAbstractModule
 
-class NoopTracerBackendModule : KAbstractModule() {
+class NoopTracingModule : KAbstractModule() {
```
nbroyles: Is this still needed now that all the tracing stuff is split out?

swankjesse: Yeah. Though I need to find a nicer way.

PR #316, opened 2018-07-10
Be consistent about how we register multibindings

swankjesse:
Having convenience modules is sort of handy, but also potentially awkward because the
module needs to choose whether its a KClass or an instance, and what the scope is.
Instead just make multibind() great and use it everywhere.

Also attempt to document which types are eligible for multibind().


misk/src/main/kotlin/misk/client/ClientApplicationInterceptor.kt
```diff
@@ -16,6 +16,14 @@ interface ClientApplicationInterceptor {
    */
   fun intercept(chain: ClientChain)
 
+  /**
```
adrw: Documentation for multibinding how-to 👍

misk/src/main/kotlin/misk/web/marshal/Json.kt
```diff
@@ -21,6 +22,7 @@ class JsonMarshaller<T>(val adapter: JsonAdapter<T>) : Marshaller<T> {
     }
   }
 
+  @Singleton
```
adrw: should `@Singleton` on factory classes be an enforced / documented idiom?

swankjesse: probably documented. We wanna say “everything that can be singleton should be singleton”

PR #318, opened 2018-07-11
Dynamically Built List of Installed Tabs Using Multibinding

adrw:
- New multibinding for 
  - AdminTab: used in React for building menus, loading scripts for those tabs...etc
  - UpstreamResourceInterceptor.Mapping: includes the mapping from a url path prefix to either a server base url or a Jar path determinant on the upstreamMode (server or jar)
- All tab related endpoints should start with `/api/_admin/{tabname}`
- Multibinding for AdminTab should go with the module that contains the related endpoints. If there are no endpoints, it can go in `AdminTabModule`
- Tests rewritten using new HttpClient
- Plans start soon for refactoring all ProxyInterceptor and ResourceInterceptor code into actions...


misk/src/main/kotlin/misk/config/ConfigAdminAction.kt
```diff
@@ -14,7 +14,7 @@ class ConfigAdminAction : WebAction {
   @Inject @AppName lateinit var appName: String
   @Inject lateinit var environment: Environment
 
-  @Get("/_admin/api/config/all")
+  @Get("/api/_admin/config/all")
```
swankjesse: 👍🏻

BTW, the _ in _admin _might_ be unnecessary. We generally use `_` as a URL prefix for internal-facing URLs and don’t send those through our HTTP gateway. But for Misk we aren’t doing that, the `_` is just a naming convention to hint to the reader that it’s internal only.

I’d be tempted to rename this to either `/_api/admin/config/call` (keeping the hint) or perhaps `/api/admin/config/call` (because it’s not a user-facing URL anyway)

misk/src/main/kotlin/misk/config/ConfigWebModule.kt
```diff
@@ -2,9 +2,23 @@ package misk.config
...
 
 class ConfigWebModule : KAbstractModule() {
   override fun configure() {
     install(WebActionModule.create<ConfigAdminAction>())
+    multibind<AdminTab>().toInstance(AdminTab(
+        "Config",
+        "config",
+        "/_admin/config/"
+    ))
+    multibind<UpstreamResourceInterceptor.Mapping>().toInstance(UpstreamResourceInterceptor.Mapping(
```
swankjesse: TODO: we only wanna do this in development

misk/src/main/kotlin/misk/web/AdminTabModule.kt
```diff
@@ -0,0 +1,32 @@
+package misk.web
+
+import misk.inject.KAbstractModule
+import misk.web.actions.AdminTab
+
+class AdminTabModule : KAbstractModule() {
+  override fun configure() {
+    // only bind tabs here that do not consume any endpoint
```
swankjesse: wanna promote this to a class-level user-facing comment? Like “Binds the admin UI framework. Individual tabs should be bound with their other code.”

misk/src/main/kotlin/misk/web/actions/AdminTabAction.kt
```diff
@@ -0,0 +1,41 @@
...
+}
+
+data class AdminTab(
+  val name: String,
+  val slug: String,
+  val url_path_prefix: String,
+  val icon: String = "widget-button"
+) {
+  init {
+    require(name.first().isUpperCase() &&
```
swankjesse: for things like isUpperCase I prefer plain documentation vs. runtime enforcement. This kind of thing tends to become too strict in the weeds. (for example, our “iOS Apps” tab).

misk/src/main/kotlin/misk/web/actions/UpstreamResourceInterceptor.kt
```diff
@@ -13,41 +14,50 @@ import okhttp3.OkHttpClient
...
+import javax.inject.Singleton
 
 /**
  * UpstreamResourceInterceptor
  *
  * Rules
  * - No overlapping mapping prefixes
  *    "/_admin/config/" and "/_admin/config/subtab/" will not resolve consistently
- * - All upstreamBaseUrl ends with "/"
+ * - All server_url ends with "/"
```
swankjesse: this doesn’t read right (it didn’t before either)

misk/src/main/kotlin/misk/web/actions/UpstreamResourceInterceptor.kt
```diff
@@ -13,41 +14,50 @@ import okhttp3.OkHttpClient
...
   override fun intercept(chain: NetworkChain): Response<*> {
     val matchedMapping = findMappingMatch(chain) ?: return chain.proceed(chain.request)
     val proxyUrl = matchedMapping.upstreamBaseUrl.newBuilder()
         .encodedPath(chain.request.url.encodedPath())
         .query(chain.request.url.query())
         .build()
-    return forwardRequestTo(chain.request, proxyUrl)
+    return when (matchedMapping.upstreamMode) {
+      Mode.SERVER -> forwardRequestTo(chain.request, proxyUrl)
+      // TODO(adrw) build out JAR resource getting and returning in a response
```
swankjesse: yeah yay

misk/src/main/kotlin/misk/web/actions/UpstreamResourceInterceptor.kt
```diff
@@ -58,14 +68,18 @@ class UpstreamResourceInterceptor(
...
-      Response(
-          "Failed to fetch upstream URL ${clientRequest.url()}".toResponseBody(),
-          Headers.of("Content-Type", "text/plain; charset=utf-8"),
-          HttpURLConnection.HTTP_UNAVAILABLE
-      )
+      FetchFailResponse(clientRequest)
     }
   }
 
+  private fun FetchFailResponse(clientRequest: okhttp3.Request): Response<ResponseBody> {
```
swankjesse: nit: fun should start with a lowercase `F`, we’re not C# programmers

misk/src/main/kotlin/misk/web/actions/UpstreamResourceInterceptor.kt
```diff
@@ -76,31 +90,39 @@ class UpstreamResourceInterceptor(
   /**
    * Maps URLs requested against this server to URLs of servers to delegate to
    *
-   * localPathPrefix: `/_admin/`
-   * upstreamBaseUrl: `http://localhost:3000/`
+   * url_path_prefix: `/_admin/`
+   * server_url: `http://localhost:3000/`
```
swankjesse: nice

misk/src/main/kotlin/misk/web/actions/UpstreamResourceInterceptor.kt
```diff
@@ -76,31 +90,39 @@ class UpstreamResourceInterceptor(
...
    *
    * An incoming request then for `/_admin/config.js` would route to `http://localhost:3000/_admin/config.js`.
    */
+  // Make register tab a subclass of mapping
```
swankjesse: what’s this comment mean?

misk/src/main/kotlin/misk/web/actions/UpstreamResourceInterceptor.kt
```diff
@@ -76,31 +90,39 @@ class UpstreamResourceInterceptor(
...
-          upstreamBaseUrl.pathSegments().size == 1)
+          upstreamBaseUrl.pathSegments().size == 1 &&
+          upstreamBaseJar.startsWith("/") &&
+          upstreamBaseJar.endsWith("/"))
     }
   }
 
+  enum class Mode {
+    SERVER,
+    JAR
```
swankjesse: I have a different idea on how I’d like this to work.

I think what I want is that you either use a `UpstreamResourceInterceptor.Mapping` to get a resource from a URL, or you use something like a `ResourceInterceptor.Mapping` to get a resource from the JAR, and the caller always picks which to install.

Ie. this class doesn’t care whether you’re a SERVER or a JAR, it just does SERVER things. There’s another class that just does JAR things, and in the module you have to decide which you want.

misk/src/main/kotlin/misk/web/actions/UpstreamResourceInterceptor.kt
```diff
@@ -76,31 +90,39 @@ class UpstreamResourceInterceptor(
...
   class Factory @Inject internal constructor() : NetworkInterceptor.Factory {
+    @Inject private lateinit var mappings: List<Mapping>
     override fun create(action: Action): NetworkInterceptor? {
-      return UpstreamResourceInterceptor(okhttp3.OkHttpClient(), mutableListOf<Mapping>(
-          UpstreamResourceInterceptor.Mapping("/@misk/", HttpUrl.parse("http://localhost:9100/")!!),
-          UpstreamResourceInterceptor.Mapping("/_admin/test/", HttpUrl.parse("http://localhost:8000/")!!),
-          UpstreamResourceInterceptor.Mapping("/_admin/dashboard/", HttpUrl.parse("http://localhost:3100/")!!),
-          UpstreamResourceInterceptor.Mapping("/_admin/config/", HttpUrl.parse("http://localhost:3200/")!!)
-      ))
+      return UpstreamResourceInterceptor(okhttp3.OkHttpClient(), mappings)
```
swankjesse: TODO for me: we still need to make it so OkHttpClient is injected

misk/src/test/kotlin/misk/web/actions/AdminTabActionTest.kt
```diff
@@ -0,0 +1,117 @@
...
+internal class AdminTabActionTest {
+  @MiskTestModule
+  val module = TestModule()
+
+  @Inject lateinit var adminTabAction: AdminTabAction
+  @Inject lateinit var adminTabs: List<AdminTab>
+
+  private val logger = getLogger<AdminTabActionTest>()
+
+  class TestModule : KAbstractModule() {
```
swankjesse: does this need to itself install the AdminTabsModule ?

misk/src/test/kotlin/misk/web/actions/UpstreamResourceInterceptorTest.kt
```diff
@@ -16,55 +21,102 @@ import org.junit.jupiter.api.BeforeEach
...
   @BeforeEach
   internal fun setUp() {
     upstreamServer.start()
 
-    mapping = UpstreamResourceInterceptor.Mapping(
-        "/local/prefix/",
-        upstreamServer.url("/"))
-
-    interceptor = UpstreamResourceInterceptor(OkHttpClient(), mutableListOf(mapping))
+    interceptor = UpstreamResourceInterceptor(OkHttpClient(), mutableListOf(
```
swankjesse: can you change this so it’s injected? That would be neater

misk/src/test/kotlin/misk/web/actions/UpstreamResourceInterceptorTest.kt
```diff
@@ -16,55 +21,102 @@ import org.junit.jupiter.api.BeforeEach
...
+          "/local/prefix/",
+          HttpUrl.parse("http://localhost:418")!!, "/", UpstreamResourceInterceptor.Mode.SERVER
+      ))
+    }
+  }
+
+  @Test
+  internal fun testBindings() {
+
+    assert(mappingBindings.size == 1)
```
swankjesse: nit: we always use `assertThat()` . . . it has nicer failure messaging

misk/web/@misk/components/tslint.json
```diff
@@ -19,6 +19,7 @@
     "ordered-imports": [true],
     "no-console": [false],
     // "no-var-requires": true,
+    "no-implicit-dependencies": false,
```
swankjesse: this is curious!

misk/src/main/kotlin/misk/config/ConfigWebModule.kt
adrw: Good point, would an environment variable check on production/dev mode be the way to do that?

misk/src/main/kotlin/misk/web/actions/AdminTabAction.kt
adrw: Fair point and good edge case with the `iOS Apps`, I'll remove the syntax enforcement requirements and add documentation.

misk/src/main/kotlin/misk/web/actions/UpstreamResourceInterceptor.kt
adrw: oh okay, yeah that makes more sense. I'll split out that functionality into a new `ResourceInterceptor` class then and add separate multibindings as required.

misk/web/@misk/components/tslint.json
adrw: Bit of a design decision to loosen the linting requirements to allow for a simpler `package.json`.

With it set to true, any `import x from 'libraryX'` must have `libraryX` declared explicitly as a dependency in the `package.json`.

One of the goals with `@misk/common` was to allow developers to import only that single dependency and then get access to all common libraries (React, BlueprintJS...) that are used across the tabs. A single dependency would also give better control over consistency of version numbers which would prevent cases where a tab developer bumps their React version that includes some breaking changes and then doesn't work interoperably with the rest of the app.

With a single dependency, two courses of action for imports are available to us. 

1) `@misk/common` imports and then re-exports all methods from all of the common libraries. Import statements then are `import React from '@misk/common'`. This would eliminate the "no-implicit-dependencies" error in linting.

2) `@misk/common` only has the common libraries set as dependencies so they will always be available to tab developers but won't be explicitly stated in `package.json` but their import statements will look identical to all online documentation for those libraries: `import React from 'react'`.

The current approach was to go with 2), use `@misk/common` as a single dependency to maintain version consistency across the tabs and reduce the need for tab developers to worry about managing their `package.json` versions.

Thoughts?

misk/src/main/kotlin/misk/config/ConfigAdminAction.kt
adrw: I guess the initial reasoning for maintaining the `_admin` in the endpoints was to maintain consistency between the url domain given to a tab `_admin/config` and their related endpoints. That didn't come to fully pass given the complexities of matching routes and only passing web ones onto related servers/resources and not losing the endpoint calls within that, thus `/api/_admin/config`. 

1) If we know for sure that there will not be any internal api calls to a future dashboard for `/admin/config/` then I think it'd be okay to go with the suggested `/_api/admin/config`. 

2) Or we abandon the initial goal of shared url namespaces with the related front end view tabs and say that the scoping is module based regardless of whether it goes to a view at `/admin/config` or `/_admin/config`. Then all api endpoints within the config module would live at `/api/config`.

Leaning towards 2, since the shared url domain between view and API seems to be already broken enough that it might make sense to go with a simpler url scheme for endpoints. Thoughts?

misk/web/@misk/components/tslint.json
swankjesse: Got it. Yeah, lets make the dependency manager do the dependency manager’s job.

misk/src/main/kotlin/misk/web/actions/UpstreamResourceInterceptor.kt
adrw: Also, should this be called `JarResourceInterceptor` for better transparency of where it will be getting these resources. Or in Java ecosystem maybe that's a redundant prefix.

misk/src/main/kotlin/misk/web/actions/UpstreamResourceInterceptor.kt
adrw: Was taking notes of what I thought you meant but turns out I misunderstood.

misk/src/main/kotlin/misk/config/ConfigAdminAction.kt
swankjesse: Yeah I think 2. I’d like the user-facing URLs to have `_` in ’em for social reasons. APIs don’t need ’em.

 * `/_admin/config/` I visit this in my browser
 * `/api/config/` the above URL makes its API calls here

misk/src/main/kotlin/misk/config/ConfigWebModule.kt
swankjesse: That’s probably the most capable way to do it. Lets leave as-is for now and follow-up.

misk/src/main/kotlin/misk/web/actions/UpstreamResourceInterceptor.kt
swankjesse: How about this:

 * WebProxyInterceptor - what we’re currently calling UpstreamResourceInterceptor
 * ResourceInterceptor - serves from the .jar file or filesystem

adrw: 💯

adrw:
Todo
- [ ] Inject okhttp.client
- [ ] Build tests to handing bindings
- [ ] Inject WebProxyInterceptor only in dev environment

adrw:
Added issue #323

PR #319, opened 2018-07-12
AccessAnnotation: a symbolic way to lock down APIs

swankjesse:


misk/src/main/kotlin/misk/security/authz/AccessAnnotation.kt
```diff
@@ -0,0 +1,45 @@
...
+ *   @PaleontologistAccess
+ *   fun dig()
+ * }
+ * ```
+ *
+ * Finally we use multibindings to specify which services and roles are permitted:
+ *
+ * ```
+ * multibind<AccessAnnotation>().toInstance(AccessAnnotation(
+ *     DigUpDinosaurAction::class, roles = listOf("paleontologist", "intern")))
```
mmihic: Should this be ``PaleontologistAccess``?

misk/src/main/kotlin/misk/security/authz/AccessAnnotation.kt
```diff
@@ -0,0 +1,45 @@
...
+ * ```
+ */
+data class AccessAnnotation(
+  val annotation: KClass<out Annotation>,
+  val services: List<String> = listOf(),
```
mmihic: I think ``AccessAnnotation`` is fine for roles since the mapping from concrete entity (a username) to the code facing entity (a role) is maintained externally. But I don't think we should be using code to describe the the inbound RBAC rules for peer services; we should be moving these to an externally managed system as well.

misk/src/main/kotlin/misk/security/authz/AccessAnnotation.kt
swankjesse: Yes!

misk/src/main/kotlin/misk/security/authz/AccessAnnotation.kt
swankjesse: Sure! Though I’m not quite sure what that looks like. One option is to bind the AccessAnnotation to something computed rather than something static. The Misk API stays the same mostly.

PR #322, opened 2018-07-13
Reduce the number of modules in Misk's API.

swankjesse:
It's too difficult to create a working application with many small modules.
It's worth the slightly-larger injector to always include features like
a clock and default web actions.


misk/src/main/kotlin/misk/web/WebModule.kt
```diff
@@ -111,5 +120,18 @@ class WebModule : KAbstractModule() {
     // Bind _admin static resources to web
     multibind<StaticResourceMapper.Entry>()
         .toInstance(StaticResourceMapper.Entry("/_admin/", "web/_admin", "misk/web/_admin/build"))
+
+    // Bind build-in actions.
+    install(WebActionModule.create<AdminTabAction>())
```
adrw: Definitely like these default actions being baked into WebModule by default 👍

PR #326, opened 2018-07-17
Enable Conscrypt and HTTP/2

swankjesse:
Closes: https://github.com/square/misk/issues/253 (http/2)
Closes: https://github.com/square/misk/issues/164 (boringssl via conscrypt)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,6 +1,7 @@
 ext {
   kotlinVersion = '1.2.50'
   awsVersion = '1.11.143'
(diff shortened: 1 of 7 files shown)


build.gradle
```diff
@@ -1,6 +1,7 @@
 ext {
   kotlinVersion = '1.2.50'
   awsVersion = '1.11.143'
+  jettyVersion = '9.4.11.v20180605'
```
ryanhall07: nice pattern

PR #338, opened 2018-07-18
Webpack Externals & Modular Builds Work

adrw:
All of the weeks of Webpack research have finally paid off: common `vendors.js` and `styles.js` and super thin builds for each tab.

- `@misk/common` now properly exports as a library exposing `externals` array that individual tabs' Webpack uses to deduplicate their exported code for thinner builds (ie. don't include `React` library with every tab if it will be loaded once in the wrapper in the `vendors.js`)
- `@misk/common` also ships with compiled `styles.js` and `vendors.js` which when loaded in the browser, contain all common styles and vendor libraries that are accessible within any of the tabs loaded at the time
- All tabs are bumped to the latest `webpack.config.js` common internal template
- `@misk/common` 0.0.10 published on NPM and updated to BlueprintJS 3.0+
- `@misk/dev` 0.0.6 published on NPM 
- All tabs now produce _very, very thin_ builds compared to when not using Webpack externals (~5kb vs >850kb)

Next Steps
---
- Flesh out loader tab to be the only tab that has script tags for all of the tabs, and the core `styles.js` and `vendors.js`
- Loader will also have the main router that shows/hides tabs as they are clicked


misk/web/README.md
```diff
@@ -2,84 +2,43 @@ Misk Admin
...
+  - `@misk/common`: common dependencies
+  - `@misk/components`: reusable React components (not working yet)
+  - `@misk/dev`: common devDependencies
+- `tabs`: all modular parts of the dashboard
+  - `config`: config tab
+  - `dashboard`: UI wrapper that adds dashboard navbar, menu...etc
+  - `loader`: thin wrapper that has the main router, all script tags, handles hide/show each tab when it's clicked
+  - `healthcheck`: to be built
+  - `guice`: to be built
+  - `hibernate`: to be built
```
swankjesse: this working doc is great

misk/web/README.md
```diff
@@ -2,84 +2,43 @@ Misk Admin
...
-- [ ] Fix `yarn dist` not working because of bug with `react-dom.production.min.js`
-- [ ] Add reasonable linting and syntax enforcement with `eslint` or `tslint`
-- [ ] Get repo working with tsconfig.json `strict: true`
-- [ ] Exemplar module UI
-- [ ] Find out how to access .env files in JS
-- [ ] Add CSP to prevent malicious takeover of admin dashboard [Terril CSP Cash.me](https://medium.com/square-corner-blog/content-security-policy-for-single-page-web-apps-78f2b2cf1757)
+
+## Creating a new Tab
+
+- Install IntelliJ Ultimate and [Setup Typescript Support](https://www.jetbrains.com/help/idea/2017.1/typescript-support.html). Typescript errors will now appear as you type.
```
swankjesse: ooooh cool

misk/web/tabs/loader/.gitignore
```diff
@@ -0,0 +1,15 @@
+# Logs
```
swankjesse: should we just put a .gitignore at the root of the project? These are going to be repetitive

misk/web/tabs/loader/.gitignore
```diff
@@ -0,0 +1,15 @@
...
+logs
```
swankjesse: though I guess names like this are tricker

misk/web/tabs/loader/src/containers/TabContainer.tsx
```diff
@@ -0,0 +1,47 @@
...
+  pathname: string,
+  search: string
+}
+
+const Container = styled.div`
+  margin-left: 180px;
+  margin-top: 20px;
+`
+
+class TabContainer extends React.Component<ITabProps, {children : any}> {
```
swankjesse: Cool!

misk/web/tabs/loader/.gitignore
adrw: yup, I'll clean these up

misk/web/tabs/loader/.gitignore
adrw: I'll find a way to make it work, I don't have many logs outputting them right now.

PR #343, opened 2018-07-19
Use URL-like schemes in resource addresses

swankjesse:
This replaces the backend-type prefix /resources or /memory with a scheme
like resources: or memory:.

Closes: https://github.com/square/misk/issues/342

### DIFF
--- a/misk/src/main/kotlin/misk/resources/ResourceLoader.kt
+++ b/misk/src/main/kotlin/misk/resources/ResourceLoader.kt
@@ -11,22 +11,22 @@ import javax.inject.Singleton
  * ResourceLoader is a testable API for loading resources from the classpath, from the filesystem,
  * from memory, or from another [Backend] source.
  *
- * Resource paths look like UNIX filesystem paths: `/resources/migrations/v1.sql`. Paths always
- * start with a backend name like `/resources` or `/memory` and are followed by a path that is
- * backend-specific. This path may have any number of segments.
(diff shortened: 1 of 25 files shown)


misk/src/main/kotlin/misk/resources/ResourceLoader.kt
```diff
@@ -11,22 +11,22 @@ import javax.inject.Singleton
  * ResourceLoader is a testable API for loading resources from the classpath, from the filesystem,
  * from memory, or from another [Backend] source.
  *
- * Resource paths look like UNIX filesystem paths: `/resources/migrations/v1.sql`. Paths always
- * start with a backend name like `/resources` or `/memory` and are followed by a path that is
- * backend-specific. This path may have any number of segments.
+ * Resource addresses have a scheme name, a colon, and an absolute filesystem-like path:
+ * `resources:/migrations/v1.sql`. Schemes identify backends `resources:` or `memory:`. Paths start
```
eeSeeGee: Is it strange that the colon isn't followed by a pair of slashes? Am I just a slave to http conventions?

swankjesse: URLs have a bunch of parts each of which has its own delimiter rules. Everything but the scheme is delimited on the left.
scheme:
//authority
/path
?query
#fragment

We don’t have an authority so we don’t need double slashes.

PR #347, opened 2018-07-25
Start Vitess automatically when we run tests

tirsen:


misk-hibernate/src/main/kotlin/misk/hibernate/PingDatabaseService.kt
```diff
@@ -31,10 +31,14 @@ class PingDatabaseService @Inject constructor(
...
-            c.createStatement().executeQuery("SELECT 1 FROM dual").uniqueResult { it.getInt(1) }
-        check(result == 1)
+      try {
+        dataSource.connection.use { c ->
+          val result =
+              c.createStatement().executeQuery("SELECT 1 FROM dual").uniqueResult { it.getInt(1) }
+          check(result == 1)
+        }
+      } catch (e: Exception) {
+        e.printStackTrace()
```
eeSeeGee: You don't want to throw here?

.travis.yml
```diff
@@ -4,12 +4,6 @@ jdk:
   - oraclejdk8
   - oraclejdk9
 
-before_script:
-  - docker pull vitess/base
-  - pushd misk-hibernate/src/test/resources/misk/hibernate
-  - docker-compose up -d
```
drashkov: You can also do docker-compose -f <file> so you don't need to change dirs.

misk-hibernate-testing/src/main/kotlin/misk/hibernate/StartVitessService.kt
```diff
@@ -0,0 +1,128 @@
...
+    docker.startContainerCmd(containerId!!).exec()
+    StartVitessService.logger.info("Started Vitess with container id $containerId")
+  }
+
+  fun stop() {
+    if (!isRunning()) {
+      return
+    }
+
+    docker.killContainerCmd(containerId!!).exec()
```
eeSeeGee: Does the container also need to be `rm`'d?

misk-hibernate/src/main/kotlin/misk/hibernate/PingDatabaseService.kt
tirsen: Oops me debugging

tirsen: reverting

misk-hibernate-testing/src/main/kotlin/misk/hibernate/StartVitessService.kt
tirsen: Hmm I thought `kill` deleted it too. The container isn't there when I check with `docker container ls`.

eeSeeGee: Ah, yeah I think that's right.

PR #348, opened 2018-07-26
Don't wait for services to stop

tirsen:
This makes the tests somewhat flaky and is not necessary.


tirsen:
This just takes it back to what it was before. The wait was not there earlier, I added it it in my previous PR because I needed it for a little while but I don't need it now and I forgot to revert that bit. Unfortunately it made the build flaky so I'd like to revert it.

I agree we should probably put it there eventually but there does seem to be services that intermittently don't shut down cleanly so now master is flaky.

PR #349, opened 2018-07-26
Add uploader

drashkov:


misk-aws/build.gradle
```diff
@@ -37,3 +37,7 @@ dependencies {
   testCompile dep.junitParams
   testCompile project(':misk-testing')
 }
+
+if (rootProject.file("../upload.gradle").exists()) {
+  apply from: rootProject.file("../upload.gradle")
```
tirsen: This will work obviously but it feels kinda weird to load an internal file from an external open source project. It would be nicer if `polyrepo` could "push" changes to the sub-projects build instead. Maybe using some kind of local hook.

Here's a suggestion:
```
if (rootProject.file("hooks.gradle").exists()) {
  apply from: rootProject.file("hooks.gradle")
}
```
Then `polyrepo` writes `hooks.gradle` when checking out the project which would for example load `upload.gradle`. `hooks.gradle` is in `.gitignore`.

PR #351, opened 2018-07-27
Re-enable SHOW VITESS_TARGET for shard targeting

tirsen:


tirsen:
This is now fixed in Vitess master.

misk-hibernate/src/main/kotlin/misk/hibernate/RealTransacter.kt
```diff
@@ -168,12 +168,9 @@ internal class RealTransacter private constructor(
         return useConnection { connection ->
           // This broke on master Vitess so we temporarily disable it.
```
eeSeeGee: Remove this too?

PR #355, opened 2018-07-30
Remove the Vitess container

tirsen:
Just killing a container does not remove it.

### DIFF
--- a/misk-hibernate-testing/src/main/kotlin/misk/hibernate/StartVitessService.kt
+++ b/misk-hibernate-testing/src/main/kotlin/misk/hibernate/StartVitessService.kt
@@ -82,7 +82,7 @@ internal class DockerVitessCluster(
       return
     }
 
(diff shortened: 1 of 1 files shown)


tirsen:
I got confused because `docker container ls` only lists running containers, you have to run `docker container ls -a` to see all containers.

misk-hibernate-testing/src/main/kotlin/misk/hibernate/StartVitessService.kt
```diff
@@ -82,7 +82,7 @@ internal class DockerVitessCluster(
       return
     }
 
-    docker.killContainerCmd(containerId!!).exec()
+    docker.removeContainerCmd(containerId!!).exec()
```
eeSeeGee: With docker command line if you try to remove a running container it'll fail, but I'll assume this works.

PR #357, opened 2018-07-30
Use prometheus metrics client

mmihic:


mmihic:
Sorry, brain splat - meant ClusterIP not NodePort. You create a service
dedicated to exposing metrics for the app, the scraper just discovers the
endpoints for that service via a service level label selector.

eeSeeGee:
Neat, I dig it.

PR #358, opened 2018-07-31
Gotta remove the container with force

tirsen:


tirsen:
Yeah you were right, you gotta remove it with force. I don't understand why the build didn't break.

PR #361, opened 2018-08-01
Fail fast if any services are not Singleton

mmihic:
readiness checks will fail with a difficult to debug error if any
services are not Singletons (since the readiness check will get a new
instance, which will be in the NEW state). Check for this explicitly
and fail fast with an obvious error. Note this doesn't handle cases
where services are registered asEagerSingleton or bound toInstance.

### DIFF
--- a/misk/src/main/kotlin/misk/MiskServiceModule.kt
+++ b/misk/src/main/kotlin/misk/MiskServiceModule.kt
@@ -30,6 +30,18 @@ class MiskServiceModule : KAbstractModule() {
   @Provides
   @Singleton
   fun provideServiceManager(services: List<Service>): ServiceManager {
+    // Confirm all services have been registered as singleton. If they aren't singletons,
+    // _readiness checks will fail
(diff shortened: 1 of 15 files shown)


misk/src/main/kotlin/misk/MiskServiceModule.kt
```diff
@@ -30,6 +30,18 @@ class MiskServiceModule : KAbstractModule() {
...
+    val invalidServices = services.map { it.javaClass }.filter { !isSingleton(it) }.map { it.name }
+    check(invalidServices.isEmpty()) {
+      "the following services are not marked as @Singleton: ${invalidServices.sorted().joinToString(
+          ", ")}"
+    }
+
     return CoordinatedService.coordinate(services)
   }
+
+  private fun isSingleton(clazz: Class<*>) =
```
ryanhall07: I wonder what would happen if you injected the `List<Service>` twice and compared they are the same instances to check for singleton scoping

mmihic: Yeah debated that approach. Will consider

PR #362, opened 2018-08-03
Add caller identity as a label in request metrics

mmihic:


misk/src/main/kotlin/misk/web/interceptors/MetricsInterceptor.kt
```diff
@@ -12,26 +14,31 @@ import javax.inject.Singleton
...
+  private val caller: ActionScoped<MiskCaller?>
 ) : NetworkInterceptor {
   override fun intercept(chain: NetworkChain): Response<*> {
     val (elapsedTime, result) = timed { chain.proceed(chain.request) }
 
     val elapsedTimeMillis = elapsedTime.toMillis().toDouble()
-    requestDuration.labels(actionName, "all").observe(elapsedTimeMillis)
-    requestDuration.labels(actionName, "${result.statusCode / 100}xx").observe(elapsedTimeMillis)
-    requestDuration.labels(actionName, "${result.statusCode}").observe(elapsedTimeMillis)
+    val callingPrincipal = caller.get()?.principal ?: "unknown"
```
eeSeeGee: Is this a cardinality explosion?

mmihic: It's the same cardinality we have today (and it's kindof critical to identifying poorly behaved clients). We can aggregate all clients into a single timeseries via recording rules on the Prometheus side.

PR #363, opened 2018-08-07
Completed WebProxyInterceptor -> WebProxyAction

adrw:
- Interceptor and related tests has been rewritten as an Action
- New binding nomenclature for WebActions with WebActionEntry in KAbstractModules. Can include optional prefix parameter which is a string URL prefix on the bound action.
```Kotlin
multibind<WebActionEntry>().toInstance(WebActionEntry<NotFoundAction>())
multibind<WebActionEntry>().toInstance(WebActionEntry<WebProxyAction>("/_admin/dashboard"))
```
- New binding nomenclature for WebProxyEntry in KAbstractModules
```Kotlin
multibind<WebProxyEntry>().toInstance(
        WebProxyEntry("/_admin/dashboard", "http://localhost:3110/"))
```


misk/src/main/kotlin/misk/config/ConfigModule.kt
```diff
@@ -24,6 +25,9 @@ class ConfigModule<T : Config>(
     bind<String>().annotatedWith<AppName>().toInstance(appName)
     bind(configClass).toInstance(config)
     bindConfigClassRecursively(configClass)
+
+    multibind<WebActionEntry>().toInstance(
```
mmihic: Is this necessary given that it is also bound in `ConfigWebModule`? Think we should leave the low level `ConfigModule` independent of web concerns.

misk/src/main/kotlin/misk/healthchecks/ClusterWideHealthModule.kt
```diff
@@ -12,8 +12,10 @@ import javax.inject.Singleton
 
 class ClusterWideHealthModule : KAbstractModule() {
   override fun configure() {
-    multibind<WebActionEntry>().toInstance(WebActionEntry(ClusterWideHealthPageAction::class))
-    multibind<WebActionEntry>().toInstance(WebActionEntry(ClusterWideHealthService::class))
+    multibind<WebActionEntry>().toInstance(
```
mmihic: love this new syntax

misk/src/main/kotlin/misk/eventrouter/RealEventRouterModule.kt
```diff
@@ -30,7 +30,8 @@ class RealEventRouterModule(val environment: Environment) : KAbstractModule() {
     }
     bind<ClusterMapper>().to<ConsistentHashing>()
     install(MoshiAdapterModule(SocketEventJsonAdapter))
-    multibind<WebActionEntry>().toInstance(WebActionEntry(EventRouterConnectionAction::class))
+    multibind<WebActionEntry>().toInstance(
```
mmihic: nit, but the line length here seems a little aggressively short.

misk/src/main/kotlin/misk/web/Request.kt
```diff
@@ -17,11 +18,14 @@ data class Request(
 ) {
   fun toOkHttp3(): okhttp3.Request {
     // TODO(adrw) https://github.com/square/misk/issues/279
-    val okRequestBody = if (this.method == HttpMethod.GET) {
-      null
-    } else {
-      object : RequestBody() {
-        override fun contentType() = null
+    val okRequestBody = when {
```
mmihic: nit maybe do 
```kotlin:
when method {
   HttpMethod.CONNECT, HttpMethod.GET, HttpMethod.HEAD, HttpMethod.OPTIONS, HttpMethod.TRACE -> null
   else -> object : RequestBody() { /* etc */ } 
}
```

misk/src/main/kotlin/misk/web/Response.kt
```diff
@@ -48,7 +48,7 @@ fun Response<ResponseBody>.writeToJettyResponse(jettyResponse: HttpServletRespon
 
 private fun HttpServletResponse.bufferedSink() = Okio.buffer(Okio.sink(outputStream))
 
-fun okhttp3.Response.toMisk() : Response<*> {
+fun okhttp3.Response.toMisk() : Response<ResponseBody> {
```
mmihic: I don't think this is correct - further upstream we might not have converted to a `ResponseBody` yet

misk/src/main/kotlin/misk/web/actions/WebActionEntry.kt
```diff
@@ -18,8 +17,14 @@ data class WebActionEntry(
   val pathPrefix: String = ""
 ) {
   init {
-    require(pathPrefix.matches(Regex("(/[^/]+)*"))) {
+    require(pathPrefix.matches(Regex("(/[^/]+)*")) && !pathPrefix.startsWith("/api")) {
```
mmihic: Why the special call-out for `/api` ? What implication does this have for the URI space available for application installed actions?

misk/src/main/kotlin/misk/web/actions/WebEntryCommon.kt
```diff
@@ -0,0 +1,27 @@
+package misk.web.actions
+
+import okhttp3.HttpUrl
+
+object WebEntryCommon {
+  /**
```
mmihic: nit: use proper KDoc comments

```kotlin
/** @return entry whose url_path_prefix must closely matches the given url */
```

misk/src/main/kotlin/misk/web/actions/WebProxyAction.kt
```diff
@@ -0,0 +1,108 @@
...
+import okhttp3.Headers
+import okhttp3.HttpUrl
+import okhttp3.OkHttpClient
+import java.io.IOException
+import java.net.HttpURLConnection
+import javax.inject.Inject
+import javax.inject.Named
+import javax.inject.Singleton
+
+/**
```
mmihic: Can we bundle all of the proxy stuff up into a dedicated package? It's difficult to differentiate now what is required for systems using proxies support, and what isn't. Given that proxies imply required configuration (http clients and such) it would be good to have it all separately located vs intermingled with API action support

misk/build.gradle
```diff
@@ -1,6 +1,13 @@
 import org.junit.platform.console.options.Details
 
 apply plugin: 'kotlin'
+
+kotlin {
```
mmihic: Does this need to be part of the misk build proper, or just the testing area? Makes me a little nervous to add to misk proper; I don't think we've yet vetted that coroutines are acceptable in server-side production code

misk/src/main/kotlin/misk/eventrouter/RealEventRouterModule.kt
adrw: Reformatted so they fit on single line

misk/src/main/kotlin/misk/web/Request.kt
adrw: Fixed

misk/src/main/kotlin/misk/web/actions/WebActionEntry.kt
adrw: Good catch, only meant that to be a requirement on WebProxyEntry. Purpose was to encourage an idiom that web facing endpoints use prefix "/api". This then prevents having a URI prefix that matches on a WebProxyEntry which would forward the entry to a build server instead of Misk handling it.

misk/build.gradle
adrw: Only used in `WebProxyActionTest` so I switched it to testCompile and the test still passes.

misk/src/main/kotlin/misk/web/Response.kt
adrw: Added better type safety so that if okhttp3.Response has null ResponseBody and empty body is returned instead of null.

Issue #364, opened 2018-08-14
Fix trailing slash requirement remaining on `/_admin/`


adrw:
Trailing slashes not necessary in routing URLs except `/_admin/`. Find out why, and fix it so it isn't necessary. Issue is likely within `/_admin` being the prefix bound to the WebProxyAction.

adrw:
Will now enforce all prefixes start & end with "/" to prevent this ambiguity

Issue #335, opened 2018-07-18
Make loader tab a thin wrapper that holds the main router and all script tags for the tabs


adrw:
- Likely also make each tab a library that loads in as a window level module which can then be easily accessed between modules and with the router be turned off/on (ie. have a show/hide function that is exported and available in the window.module)

adrw:
Closed in #393

PR #365, opened 2018-08-14
Loader Tab

adrw:
- Add back open CORS interceptor for development, necessary for requests coming from webpack dev servers
- Cleanup unnecessary AdminTab and WebProxyEntry multibindings
- Add `dayjs`, `immutable`, and `redux-saga` to `@misk/common` library for *redux-saga unidirectional React data flow* pattern
- Move dashboard `div` mount within DashboardContainer component in Config tab so it is self contained and div isn't in the Config index HTML template
- Fix public paths and WebProxyEntry rules to properly forward links without trailing slash. Ie. now can browse to `/_admin/dashboard` instead of needing trailing slash `/_admin/dashboard/`. Remaining issue: #364
- #335 : Implement redux-saga flow in Loader. Successfully loads `adminTabs` list from Misk server, maps tabs to `ScriptComponent` which currently renders the meta data for testing. Initial work had been to have a script tag within this component, this worked to load components but lead to infinite script include loop when this was done in React lifecycle function `compononentDidMount()`. Better way moving forward is to export tabs as libraries, bind them to window variables, and access them as external libraries.
- Moved loader webpack config back to having `/` public path so that React Router would work again. Nesting routes for each tab now work in the `LoaderContainer`.

Next Steps
---
- Test theory of exporting tabs as libraries and using as external components by fixing `@misk/components` to work as library
- Add redux-saga flow to other tabs
- Fix webpack for other tabs so their React Routers work


misk/src/main/kotlin/misk/web/actions/AdminTabAction.kt
```diff
@@ -44,11 +44,15 @@ data class AdminTab(
   val icon: String = "widget-button"
 ) {
   init {
-    require(slug.filter { char -> !char.isUpperCase()
-            && !char.isWhitespace() }.length.equals(slug.length) &&
-        icon.filter { char -> !char.isUpperCase()
-            && !char.isWhitespace() }.length.equals(icon.length) &&
-        url_path_prefix.startsWith("/") &&
-        url_path_prefix.endsWith("/"))
+    require(slug.filter { char ->
```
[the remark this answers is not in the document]
adrw: Documentation with the guidelines/rules currently lives at the top of that file though I'm not sure that shows up in the PR unfortunately.

```
/**
 * Admin Tab Action
 *
 * Returns all Admin Tabs for primary use in dynamically building the menu bar of the /_admin dashboard
 *
 * Guidelines
 * - name should start with a capital letter unless it is a proper noun (ie. iOS)
 * - slug must be valid slug (lowercase and no white space)
 * - icon must be valid slug (lowercase and no white space).
 *      - Not enforced in Misk but icon must also be a valid slug in BlueprintJS/Icons
 * - url_path_prefix must start and end with "/"
 *
 */
```

misk/web/tabs/dashboard/src/index.html
```diff
@@ -9,9 +9,9 @@
 
 <body>
     <div id="dashboard"></div>
-    <div id="config"></div>
 
     <!-- Misk Common Libraries -->
+    <div id="thisTabIsDashboard"></div>
```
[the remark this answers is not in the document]
adrw: was using for testing to know which `index.html` template was being rendered, will remove it.

misk/web/@misk/common/src/vendors.js
```diff
@@ -15,4 +17,5 @@ window.ReactRouter = require('react-router');
 window.ReactRouterDom = require('react-router-dom');
 window.ReactRouterRedux = require('react-router-redux');
 window.Redux = require('redux');
+window.ReduxSaga = require('redux-saga');
```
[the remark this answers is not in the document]
eventualbuddha: Oof, yeah, I agree with @terrilldent. These should not be globals if it's at all possible to avoid it. Happy to join in on that conversation.

misk/web/tabs/loader/src/containers/LoaderContainer.tsx
```diff
@@ -0,0 +1,79 @@
...
+}
+
+export interface IAdminTab {
+  icon: IconName
+  name: string
+  slug: string
+  url_path_prefix: string
+}
+
+const Container = styled.div``
```
terrilldent: Looks like this could be replaced with a `<div>`

misk/web/tabs/loader/src/actions/types.ts
```diff
@@ -0,0 +1,22 @@
...
+  "DELETE",
+  "SUCCESS",
+  "FAILURE"
+])
+
+export const ADMINTABS = createActionTypes("ADMINTABS", [
+  "FAILURE",
+  "GET_ALL",
+  "SUCCESS"
+])
```
eventualbuddha: Are these different from `enum`s? A basic reading of this leads me to believe you won't get any type-checking guarantees with this setup.

misk/web/tabs/loader/src/containers/LoaderContainer.tsx
eventualbuddha: Accidental template string instead of semicolon? This kind of thing would be caught with [tslint](https://github.com/palantir/tslint) and/or [prettier](https://www.github.com/prettier/prettier).

misk/web/tabs/loader/src/containers/LoaderContainer.tsx
```diff
@@ -0,0 +1,79 @@
...
+export interface IAdminTabs {
+  [key:string]: IAdminTab
+  toJS: any
+}
+
+export interface IAdminTab {
+  icon: IconName
+  name: string
+  slug: string
+  url_path_prefix: string
```
eventualbuddha: Is there a good reason to use an underscored name instead of camel case?

misk/web/tabs/loader/src/containers/index.ts
```diff
@@ -1,3 +1,2 @@
-import { NavContainer } from "./NavContainer"
-import TabContainer from "./TabContainer"
-export { NavContainer, TabContainer }
\ No newline at end of file
+import LoaderContainer from "./LoaderContainer"
+export { LoaderContainer }
```
eventualbuddha: Could be `export { default as LoaderContainer } from "./LoaderContainer";`.

misk/web/tabs/loader/src/sagas/itemSaga.ts
```diff
@@ -0,0 +1,86 @@
...
+// yield put(push('/next-page'))
+
+import axios from "axios"
+import { all, call, put, takeLatest } from "redux-saga/effects"
+
+import {
+  IAction, IActionType, ITEM, dispatchItem
+} from "../actions"
+
+function * handleGet () {
```
eventualbuddha: Are all these supposed to be generators in the typical sense? It kind of looks like it's using generators in place of async/await.

misk/web/tabs/loader/src/containers/LoaderContainer.tsx
eventualbuddha: Looks like it might be intentional, but a little confusing. I still like the idea of using tslint & prettier.

adrw: tslint is used with extensive rules though I agree that the empty styles is a confusing syntax so I'll be only using styled divs when the div has styles to be declared.

misk/web/tabs/loader/src/containers/LoaderContainer.tsx
adrw: This was a decision made to match variables against the request code in the Misk Kotlin server so that the variables stand out as ones that have come from a request.

misk/web/tabs/loader/src/containers/index.ts
adrw: thanks! Will fix this syntax across the code base

misk/web/tabs/loader/src/sagas/itemSaga.ts
adrw: tldr; yes, but generators provide additional features useful in writing tests. We're using the redux-saga flow initially to try and provide a bit of structure for requests, actions, and updating the state given that contributors to this project will likely not be primarily front end or React fluent engineers.

This provides a better description of the use of ES6 generators in the redux-saga flow: https://github.com/redux-saga/redux-saga

misk/web/@misk/common/src/vendors.js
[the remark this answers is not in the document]
adrw: Would appreciate your input, I'll ping you on Slack.

misk/web/tabs/loader/src/actions/types.ts
adrw: Your reading is correct, the types are not type checked currently. I've created an issue to come back to that includes some initial research on how best to do this as there doesn't seem to be an immediately default or intuitive approach. #367

adrw:
@eventualbuddha on the contrary, I appreciate the external feedback especially from others that are very fluent in current trends in front end / Javascript. Feel free to chime in on any upcoming PRs!

PR #370, opened 2018-08-15
@misk/common externals issue & Loader Tab Links

adrw:
Why
---
- Console error where Blueprint library couldn't be found anymore
- Want a way in Loader tab to test links between tabs in React Router


adrw:
@terrilldent No testing yet but I agree it would be greatly helpful. Should also be part of the Travis CI.

Issue #337, opened 2018-07-18
Dashboard


adrw:
Dashboard uses `/api/admin/admintabs/all?` endpoint to get all tabs and render menu items for them with links

adrw:
Export simple rendering components for NavContainer (topbar and side menu) that render by consuming inputted AdminTabs[]

adrw:
#400 removes Dashboard tab as core components now live in `@misk/components`

PR #373, opened 2018-08-15
[#337] Add interfaces to @misk/common, prepare Dashboard for public path changes

adrw:
Why
---
- Preparing Dashboard tab to change public path to "/" similar to Loader
  - Since Dashboard will be loaded like a simple rendering component, core interfaces for AdminTabs should be moved to `@misk/common` so they are shared between Loader and Dashboard tabs
  - `@misk` packages should be removed from `@misk/common` externals to prevent circular dependencies
  - `makeExternals` in `@misk/common` is difficult to parse nested ternary functions, should be reformatted for better readability
- By the above 3 items, release new version 0.0.13 of `@misk/common`
- [#337] Update Dashboard issue for new pure component approach

### DIFF
--- a/misk/web/@misk/common/src/externals.ts
+++ b/misk/web/@misk/common/src/externals.ts
@@ -2,7 +2,7 @@ export interface IExternal {
   [key: string]: string|string[]
 }
 
-export default makeExternals({
+export const externals = makeExternals({
   "@blueprintjs/core": ["Blueprint", "Core"],
   "@blueprintjs/icons": ["Blueprint", "Icons"],
   "axios": "Axios",
--- a/misk/web/@misk/common/src/externals.ts
+++ b/misk/web/@misk/common/src/externals.ts
@@ -23,6 +23,16 @@ export const externals = makeExternals({
   "styled-components": "StyledComponents"
 })
 
+/**
+ * 
+ * @param inExternals : IExternal
+ * 
+ * Create Webpack compatible externals object with following modifications
+ * - Concatenate scoped package arrays into single strings
+ * 
+ * Todo
+ * - Provide distinct package strings for amd, commonjs, commonjs2, root if necessary
+ */
 function makeExternals(inExternals: IExternal) : IExternal {
   const outExternals: IExternal = {}
   Object.keys(inExternals).forEach((name, index) => {
--- a/misk/web/@misk/common/src/externals.ts
+++ b/misk/web/@misk/common/src/externals.ts
@@ -5,8 +5,6 @@ export interface IExternal {
 export default makeExternals({
   "@blueprintjs/core": ["Blueprint", "Core"],
   "@blueprintjs/icons": ["Blueprint", "Icons"],
-  "@misk/common": ["Misk", "Common"],
-  "@misk/components": ["Misk", "Components"],
   "axios": "Axios",
   "connected-react-router": "ConnectedReactRouter",
   "dayjs": "Dayjs",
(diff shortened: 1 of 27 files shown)


misk/web/@misk/common/src/externals.ts
```diff
@@ -28,7 +26,10 @@ export default makeExternals({
 function makeExternals(inExternals: IExternal) : IExternal {
   const outExternals: IExternal = {}
   Object.keys(inExternals).forEach((name, index) => {
-    outExternals[name] = inExternals.hasOwnProperty(name) ? (Array.isArray(inExternals[name]) ? (inExternals[name] as string[]).join("") : inExternals[name]) : name
-  })
+    outExternals[name] = inExternals.hasOwnProperty(name) ? 
+      (Array.isArray(inExternals[name]) ? 
+        (inExternals[name] as string[]).join("") : inExternals[name]
```
terrilldent: Serializing all the externals... this is an interesting piece of code, maybe we can add a comment about why this is happening.

misk/web/@misk/common/src/vendors.js
```diff
@@ -1,5 +1,5 @@
-window.BlueprintCore = require('@blueprintjs/core')
-window.BlueprintIcons = require('@blueprintjs/icons')
+window.BlueprintCore = require('@blueprintjs/core');
+window.BlueprintIcons = require('@blueprintjs/icons');
```
terrilldent: Linter didn't catch this earlier?

misk/web/tabs/dashboard/src/components/SidebarComponent.tsx
```diff
@@ -1,35 +1,24 @@
...
 }
 
 const Sidebar = styled.div`
   position: absolute;
 `
 
-export class SidebarComponent extends React.Component<ISidebarProps, {}> {
-  private renderedMenuItems: any
+const buildMenuItems = (adminTabs: IMiskAdminTabs) => (
+  Object.entries(adminTabs).map((tab) => <MenuItem key={tab[0]} href={tab[1].url_path_prefix} className="" icon={tab[1].icon} text={tab[1].name}/>)
```
terrilldent: Can we use array destructuring to name `tab[0]` and `tab[1]` here? It could help some know what those objects are supposed to be:

```
  Object.entries(adminTabs).map(([tabName, tabConfig]) => {
    <MenuItem key={tabName} href={tabConfig.url_path_prefix} className="" icon={tabConfig.icon} text={tabConfig.name}/>)
  }
```

misk/web/@misk/common/src/externals.ts
terrilldent: Why is this happening?

misk/web/@misk/common/src/vendors.js
adrw: Unfortunately not. Linting currently only runs on Typescript files since every file in the repo except this one is Typescript.

misk/web/@misk/common/src/externals.ts
adrw: Webpack requires externals in a specific object format, this function produces an object to that specification. Currently that involves concatenating arrays of scoped packages (ie. `[Blueprint, Core] -> "BlueprintCore"). Later that may involve declaring specific strings for the different module frameworks (amd, commonjs, commonjs2, root).

misk/web/tabs/dashboard/src/components/SidebarComponent.tsx
adrw: Really like this, fixed it in Loader tab as well!

misk/web/@misk/common/src/externals.ts
terrilldent: This belongs in a comment "Webpack requires externals in a specific object format, this function produces an object to that specification. Currently that involves concatenating arrays of scoped packages (ie. `[Blueprint, Core] -> "BlueprintCore")."

adrw: Documentation was added to the function:

```
/**
 * 
 * @param inExternals : IExternal
 * 
 * Create Webpack compatible externals object with following modifications
 * - Concatenate scoped package arrays into single strings
 * 
 * Todo
 * - Provide distinct package strings for amd, commonjs, commonjs2, root if necessary
 */
```

PR #379, opened 2018-08-16
Loader Saga in Loader

adrw:
Why?: Needed to get by web request all adminTabs and their respective compiled JS components, add them to state, and make those state items accessible to `LoaderContainer`

Issue #383 created in response to @terrilldent concerns about try/catch use in Sagas handling generator functions


misk/web/tabs/dashboard/webpack.config.js
```diff
@@ -29,7 +29,10 @@ module.exports = {
   output: {
     filename: '_admin/dashboard/tab_dashboard.js',
     path: path.join(__dirname, 'dist'),
-    publicPath: "/"
+    publicPath: "/",
+    library: ['Tabs', 'Dashboard'],
+    libraryTarget: 'umd',
+    globalObject: 'typeof self !== \'undefined\' ? self : this'
```
terrilldent: This is crazy. What's happening here?

misk/web/tabs/loader/src/containers/LoaderContainer.tsx
```diff
@@ -34,13 +38,18 @@ class LoaderContainer extends React.Component<ITabProps> {
 
   render() {
     const { adminTabs } = this.props.adminTabs
+    const { adminTabComponents } = this.props.adminTabComponents
     if (adminTabs) {
       const tabRouteComponents = Object.entries(adminTabs).map(([key,tab]) => this.buildTabRouteComponent(tab))
-      const tabLinks = Object.entries(adminTabs).map(([key,tab]) => <Link key={key} to={`/_admin/test/${tab.slug}`}>{tab.name}</Link>)
+      const tabLinks = Object.entries(adminTabs).map(([key,tab]) => <Link key={key} to={`/_admin/test/${tab.slug}`}>{tab.name}<br/></Link>)
+      console.log(adminTabs, adminTabComponents)
```
terrilldent: intentional console.log?

misk/web/tabs/loader/src/sagas/loaderSaga.ts
```diff
@@ -0,0 +1,102 @@
...
+
+    for (const key in adminTabs) {
+      if (adminTabs.hasOwnProperty(key)) {
+        const tab = adminTabs[key]
+        const url = `${tab.url_path_prefix}/tab_${tab.slug}.js`
+        const { data } = yield call(axios.get, url)
+        yield put(dispatchLoader.success({ adminTabComponents: { [tab.slug]: data } }))
+      }
+    }
+  } catch (e) {
```
terrilldent: This paradigm feels weird to me. There's a lot of code in the try/catch. I'd like to do a gut check that this is a normal paradigm for saga functions. Is there a control flow issue where throwing an exception ever would be bad? Let's discuss.

misk/web/tabs/loader/src/sagas/loaderSaga.ts
```diff
@@ -0,0 +1,102 @@
...
+  try {
+    let adminTabs: any = {}
+    try {
+      const { data } = yield call(axios.get, "http://0.0.0.0:8080/api/admintab/all")
+      adminTabs = data.adminTabs
+      yield put(dispatchLoader.success({ adminTabs }))
+      if (Object.entries(adminTabs).length === 0) {
+        yield put(dispatchLoader.success({ adminTabComponents: { } }))
+      }
+    } catch (e) {
```
terrilldent: Note: Nested try/catch that are doing the same thing are a clear sign there's something to be re-thought. As discussed in person I think the try/catch should be very targeted. If these functions were required to never throw errors then the function that calls them would likely protect against it.

misk/web/tabs/loader/src/sagas/loaderSaga.ts
terrilldent: I would narrow it to something like this: 

```
  try {
    const { data } = yield call(axios.get, url)
    yield put(dispatchLoader.success({ adminTabComponents: { [tab.slug]: data } }))
  } catch (e) {
    yield put(dispatchLoader.failure({ error: { ...e } }))
  }
```

misk/web/tabs/loader/src/sagas/loaderSaga.ts
```diff
@@ -0,0 +1,74 @@
...
+        yield put(dispatchLoader.failure({ error: { ...e } }))
+      }
+    }
+  }
+}
+
+function * handleGetOneComponent (action: IAction<IActionType, { tab: IMiskAdminTab }>) {
+  try {
+    const { tab } = action.payload
+    const url = `${tab.url_path_prefix}/tab_${tab.slug}.js`
```
terrilldent: This line and the one above belong outside the `try`

misk/web/tabs/loader/src/containers/LoaderContainer.tsx
adrw: This file is still going through a lot of work, won't be present in final PR.

misk/web/tabs/loader/src/sagas/loaderSaga.ts
adrw: +1, refactored with the principles we discussed offline

misk/web/tabs/dashboard/webpack.config.js
adrw: - publicPath: url root that the package can be accessed at from webpack dev server and within the build structure if compiled with production flags
- library: library name, this one being in array means that it is a scoped package. Thus this would be accessible in the window by `Tabs.Dashboard`. The array names allow all tabs to be in a central location and prevent namespace issues if tabs end up being named the same as existing NPM packages that might be included in `@Misk/Common` and are bound as globals in window
- libraryTarget: type of module export `umd` is universal module description, options also include var if it's being bound to a variable and others
- globalObject: by default it tries to bind to window but this check prevents errors if it's ever included and run from Node or a non-browser environment. Without this, webpack complains during compilation.

terrilldent: It's worth adding a note about the `globalObject` in particular.

adrw: added to the next PR

PR #380, opened 2018-08-16
Make Tabs Loadable

adrw:
Why?: existing tabs need to have modified webpack configs and LoaderContainer needs work to properly load from script tags and mount the tabs.


misk/web/tabs/loader/src/components/ScriptComponent.tsx
```diff
@@ -1,21 +1,20 @@
...
   return (
     <div>
-      <div id={props.tab.slug}>
-        <p>{props.tab.name}: {props.tab.slug} {props.tab.url_path_prefix} <Icon icon={props.tab.icon}/></p>
-      </div>
       <Helmet>
-        {/* <script async src={`${props.tab.url_path_prefix}/tab_${props.tab.slug}.js`}/> */}
-      </Helmet>     
+        <title>Test Title</title>
+        <script async={true} src={`${props.tab.url_path_prefix}/tab_${props.tab.slug}.js`}/>
```
terrilldent: I'm letting the <script> tags fly for now, but I'm not forgetting about it. Hoping we can find an alternative eventually.

adrw: For sure! Not ideal but we have the ground work in place with the previous PR (each component's code now loads successfully into the state) to spend some more time later finding the best way to do that. Initial research and testing hasn't yielded much but there likely is a better way. #385

PR #386, opened 2018-08-17
Config Saga

adrw:
Why?: Config was initially hacked together with requests happening in componentDidMount() and other lifecycle methods. Moving to redux-saga will provide better structure for the tab.


misk/web/tabs/config/src/actions/types.ts
```diff
@@ -0,0 +1,17 @@
...
+// export const CONFIG = createActionTypes("CONFIG", [
+//   "FAILURE",
+//   "GET_All",
+//   "SUCCESS"
+// ])
+
+export enum CONFIG {
+  FAILURE = "CONFIG_FAILURE",
+  GET_ALL = "CONFIG_GET_ALL",
+  SUCCESS = "CONFIG_SUCCESS"
```
terrilldent: I love when variable name lengths all match up

adrw: There is still beauty in this crazy world.

PR #392, opened 2018-08-20
Adds support for clob properties in Hibernate mapping

davidgarth:
Adds a mapping to support String properties annotated with @Lob. Note sure if this is the best way to add `@Lob` support though.

### DIFF
--- a/misk-hibernate/src/main/kotlin/misk/hibernate/SessionFactoryService.kt
+++ b/misk-hibernate/src/main/kotlin/misk/hibernate/SessionFactoryService.kt
@@ -192,6 +192,7 @@ internal class SessionFactoryService(
       "char" -> Char::class
       "float" -> Float::class
       "double" -> Double::class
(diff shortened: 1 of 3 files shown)


ryanhall07:
can you add a test?

misk-hibernate/src/main/kotlin/misk/hibernate/SessionFactoryService.kt
```diff
@@ -192,6 +192,7 @@ internal class SessionFactoryService(
       "char" -> Char::class
       "float" -> Float::class
       "double" -> Double::class
+      "materialized_clob" -> String::class
```
ryanhall07: what if the lob does not have a character type?

davidgarth: Good point, this can't be done here.

davidgarth:
Need to find a better way to do this.

PR #394, opened 2018-08-20
Adds support for nullable Ids

davidgarth:
Nullable ids or entity references fail on dirty check or deep copy.


misk-hibernate/src/test/kotlin/misk/hibernate/DbCharacter.kt
```diff
@@ -34,16 +34,16 @@ class DbCharacter() : DbEntity<DbCharacter>, DbTimestampedEntity {
 
   @ManyToOne(fetch = FetchType.LAZY)
   @JoinColumn(name = "actor_id", updatable = false, insertable = false)
-  lateinit var actor: DbActor
+  var actor: DbActor? = null
 
   @Column
-  lateinit var actor_id: Id<DbActor>
+  var actor_id: Id<DbActor>? = null
```
davidgarth: The easiest way to test this was just to make the actor optional for a character.

PR #396, opened 2018-08-21
Expose jaeger spans in MDC

mmihic:


misk-jaeger/src/main/kotlin/misk/tracing/backends/jaeger/MDCScopeManager.kt
```diff
@@ -0,0 +1,63 @@
...
+      MDC.remove("parent-id")
+      threadLocalStorage.remove()
+    } else {
+      putToMDC(parentScope.span())
+      threadLocalStorage.set(parentScope)
+    }
+  }
+
+  private fun putToMDC(span: Span) {
+    (span as? com.uber.jaeger.Span)?.let {
```
ryanhall07: when wouldn't this be a jaeger trace?

misk-jaeger/src/main/kotlin/misk/tracing/backends/jaeger/MDCScopeManager.kt
```diff
@@ -0,0 +1,63 @@
...
+
+import io.opentracing.Scope
+import io.opentracing.ScopeManager
+import io.opentracing.Span
+import org.slf4j.MDC
+
+/**
+ * [MDCScopeManager] is a [ScopeManager] that records the current trace-id, span-id, parent-id
+ * in the current logging [MDC]. It is jaeger specific because getting access to these values
+ * is dependent on the specific tracing vendor implementation.
```
ryanhall07: in the future could probably just have some interface that provides the MDC fields, since the majority of this manager has nothing to do with jaegar

misk-jaeger/src/main/kotlin/misk/tracing/backends/jaeger/MDCScopeManager.kt
mmihic: Just being overly cautious; someone could use `MDCScopeManager` outside of jaeger

misk-jaeger/src/main/kotlin/misk/tracing/backends/jaeger/MDCScopeManager.kt
mmihic: agreed

mmihic: well, sortof - when trace-ids/span-ids/parent-ids are available is actually highly dependent on the implementation. some implementations don't assign them until a span is finished and sent to the reporter

ryanhall07: lol standards

PR #409, opened 2018-08-23
Add session-level tracing to RealTransacter

wesleyk:
Notes:
- We record spans on a per-txn basis, so an invocation to Transacter#transaction could actually result in many spans, in the case of retries.
- We leverage the MockTracer for testing. Installing the MockTracer is a bit hairy because of how the GlobalTracer leverages global state.
- Removed unused NoopTracingModule. Folks should prefer the Mock one.

r @ryanhall07 @swankjesse @mmihic @nbroyles


misk-hibernate/src/test/kotlin/misk/hibernate/TransacterTest.kt
```diff
@@ -55,8 +60,8 @@ class TransacterTest {
           .actorName("Laura Dern")
           .listAsMovieNameAndReleaseDate(session)
       assertThat(lauraDernMovies).containsExactlyInAnyOrder(
-          NameAndReleaseDate("Star Wars", LocalDate.of(1977, 5, 25)),
-          NameAndReleaseDate("Jurassic Park", LocalDate.of(1993, 6, 9)))
+          ReflectionQueryFactoryTest.NameAndReleaseDate("Star Wars", LocalDate.of(1977, 5, 25)),
```
wesleyk: I can revert these

wesleyk: actually intellij seems to want this here

misk-hibernate/src/main/kotlin/misk/hibernate/RealTransacter.kt
```diff
@@ -64,13 +66,21 @@ internal class RealTransacter private constructor(
 
   private fun <T> transactionInternal(lambda: (session: Session) -> T): T {
     return withSession { session ->
+      // Begin span
+      val scope = GlobalTracer.get().buildSpan(SESSION_SPAN_NAME)
```
mmihic: you can also use 
```kotlin
val result = Tracer.traceWithSpan(SESSION_SPAN_NAME) { span -> 
   Tags.COMPONENT_SET(span, SESSION_SPAN_TAG)
    
    
    // rest of code
}
```

misk-testing/src/main/kotlin/misk/tracing/MockTracingModule.kt
```diff
@@ -0,0 +1,33 @@
...
+import io.opentracing.Tracer
+import io.opentracing.mock.MockTracer
+import io.opentracing.util.GlobalTracer
+import misk.inject.KAbstractModule
+import javax.inject.Singleton
+
+
+class MockTracingModule : KAbstractModule() {
+  companion object {
+    val tracer : MockTracer = MockTracer()
```
mmihic: having a global seems dangerous for testing

wesleyk: Yeah the challenge is that the underlying GlobalTracer tracer is already a global. I can at least keep this as an instance

wesleyk: eh we kind of need to keep it global because of the GlobalTracer; otherwise different test cases end up getting different versions of the tracer

misk-hibernate/src/main/kotlin/misk/hibernate/RealTransacter.kt
```diff
@@ -129,6 +143,12 @@ internal class RealTransacter private constructor(
     val retryJitterMillis: Long = 400
   )
 
+  companion object {
+    val APPLICATION_TRANSACTION_SPAN_NAME = "app-db-transaction"
+    val DB_TRANSACTION_SPAN_NAME = "db-session"
+    val TRANSACTER_SPAN_TAG = "hibernate-transacter"
```
ryanhall07: just for my knowledge, what does this do for you?

misk-testing/src/main/kotlin/misk/tracing/MockTracingModule.kt
ryanhall07: that seems like what you want? a new tracer for every test case

mmihic: imo `@Inject` the `Tracer` vs relying on `GlobalTracer`

misk-hibernate/src/main/kotlin/misk/hibernate/RealTransacter.kt
wesleyk: just another tag, referenced as "component". A bit redundant, we can remove it down the line if we don't find it useful. From the docs:

"COMPONENT is a low-cardinality identifier of the module, library, or package that is instrumented."

misk-testing/src/main/kotlin/misk/tracing/MockTracingModule.kt
wesleyk: Yeah, but you can't override the GlobalTracer's tracer once it's set

wesleyk: Hmm good idea re: injecting, I'll look into that

wesleyk:
BTW to circle back on this, we removed the global tracer usage here: https://github.com/square/misk/pull/411

PR #410, opened 2018-08-23
Add request attributes to logging context

mmihic:


misk/src/main/kotlin/misk/web/interceptors/RequestLogContextInterceptor.kt
```diff
@@ -0,0 +1,62 @@
...
+) : NetworkInterceptor {
+
+  override fun intercept(chain: NetworkChain): Response<*> {
+    val request = currentRequest.get()
+    return try {
+      MDC.put(MDC_ACTION, action.name)
+      MDC.put(MDC_CALLING_PRINCIPAL, currentCaller.get()?.principal ?: "unknown")
+      MDC.put(MDC_REMOTE_IP, request.remoteAddr)
+      MDC.put(MDC_REMOTE_PORT, request.remotePort.toString())
+      MDC.put(MDC_REQUEST_URI, request.requestURI)
```
ryanhall07: I could see somebody adding a new field here and forgetting to update the `allContextNames` below. Can we some kind of test that verifies that MDC is cleared of all keys?

mmihic: Yeah, this ends up being super tricky. Need to think about how to test it.

ryanhall07: maybe in a test just clear the MDC before, run the interceptor and verify nothing is in the MDC after?

mmihic: yes but that's tricky because there needs to be something on the other side of the interceptor, plus a bunch of fakes for the action scoped, etc. it'll take a bunch of code

PR #411, opened 2018-08-23
Prefer injecting tracer in RealTransacter for better testability

wesleyk:
r @rhall @mmihic


misk-testing/src/main/kotlin/misk/testing/MockTracingBackendModule.kt
```diff
@@ -6,6 +6,7 @@ import misk.inject.KAbstractModule
 
 class MockTracingBackendModule : KAbstractModule() {
   override fun configure() {
-    bind<Tracer>().to<MockTracer>().asEagerSingleton()
+    bind<MockTracer>().asEagerSingleton()
+    bind<Tracer>().to<MockTracer>()
```
wesleyk: needed to split this out to ensure I could inject a MockTracer and get the same underlying instance

misk-hibernate/src/main/kotlin/misk/hibernate/HibernateModule.kt
```diff
@@ -72,8 +73,10 @@ class HibernateModule(
           sessionFactoryProvider.get(), config)
     }).asSingleton()
 
-    bind(transacterKey).toProvider(Provider<Transacter> {
-      RealTransacter(qualifier, sessionFactoryProvider.get(), config)
+    bind(transacterKey).toProvider(object : Provider<Transacter> {
+      @com.google.inject.Inject(optional=true) val tracer: Tracer? = null
```
ryanhall07: do we want to force misk apps to use guice?

ryanhall07: so does guice blow up if you try injecting a null value for a nullable type in kt?

wesleyk: guice is already all over misk, my impression was that the two are coupled

wesleyk: > you try injecting a null value for a nullable type in kt

Yea, tried locally and that's what I get

ryanhall07: well is guice impl all over or just javax.Inject? if we decide we want to be opinionated about guice, then we should just use the guice annotations everywhere IMO

wesleyk: it does seem like it's opinionated. There's a class called Guice.

PR #412, opened 2018-08-24
Specify dependencies in dependencies.gradle in Misk

swankjesse:


dependencies.gradle
```diff
@@ -0,0 +1,68 @@
+ext.dep = [
+  "kotlinStdLib": "org.jetbrains.kotlin:kotlin-stdlib-jdk8:1.2.51",
```
ryanhall07: it feels like our polyrepo impl is bleeding into misk here by requiring a json format for the gradle file

swankjesse: A little bit. I think the upside of easier dependency management is worth the cost.

PR #417, opened 2018-08-27
All arbitrary tags to be passed in logs

mmihic:
Any passed tags are convered into MDC entries, which in turn get emitted
in JSON logs and indexed for searching

### DIFF
--- a/misk/src/test/kotlin/misk/logging/LoggingTest.kt
+++ b/misk/src/test/kotlin/misk/logging/LoggingTest.kt
@@ -1,13 +1,109 @@
 package misk.logging
 
+import ch.qos.logback.classic.Level
+import ch.qos.logback.classic.LoggerContext
+import com.google.inject.util.Modules
+import misk.MiskServiceModule
+import misk.testing.MiskTest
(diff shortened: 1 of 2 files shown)


misk/src/test/kotlin/misk/logging/LoggingTest.kt
```diff
@@ -1,13 +1,109 @@
...
+    }
+    assertThat(untaggedEvents[1]).satisfies {
+      assertThat(it.level).isEqualTo(Level.INFO)
+      assertThat(it.message).isEqualTo("untagged info")
+      assertThat(it.throwableProxy).isNull()
+      assertThat(it.mdcPropertyMap).isEmpty()
+    }
+
+    // Log with tags
+    logger.info("user-id" to "blerb", "alias-id" to "d6F1EF53") { "tagged info" }
```
wesleyk: TIL https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/to.html, interesting syntax

ryanhall07: nice! this is super easy to use

PR #424, opened 2018-08-28
Add Common TsConfig to @misk/common

adrw:
Ensure common Typescript build environment across all Misk front end development with all `tsconfig.json` Typescript compiler options extending the same base in `@misk/common/src/base`. Existing tabs updated to extend base options.

Side Effect: many fixes to code in Config/Loader tabs since the common `tsconfig.json` rules are stricter than the existing ones that were in each tab.

Depends on PR #423 merging

### DIFF
--- a/misk/web/@misk/common/tsconfig.json
+++ b/misk/web/@misk/common/tsconfig.json
@@ -1,58 +1,6 @@
 {
-    "compileOnSave": true,
+    "extends": "./src/base/tsconfig.base",
     "compilerOptions": {
-        "allowSyntheticDefaultImports": true,
-        "declaration": true,
-        "esModuleInterop": true,
-        "experimentalDecorators": true,
-        "importHelpers": true,
-        "jsx": "react",
-        "lib": [
-            "dom",
-            "dom.iterable",
-            "es5",
-            "es2015.collection",
-            "es2015.iterable"
-        ],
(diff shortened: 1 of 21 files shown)


misk/web/@misk/common/tsconfig.json
```diff
@@ -1,27 +1,6 @@
 {
-    "compileOnSave": true,
+    "extends": "./src/base/tsconfig.base",
```
swankjesse: nice!

misk/web/tabs/config/src/components/ConfigComponent.tsx
```diff
@@ -1,5 +1,5 @@
 // tslint:disable-next-line:no-var-requires
-const YAML = require("json2yaml")
+// const YAML = require("json2yaml")
```
swankjesse: delete?

misk/web/tabs/loader/src/index.tsx
```diff
@@ -1,3 +1,4 @@
+///<reference types="webpack-env" />
```
swankjesse: delete?

adrw: This is a funky Typescript thing that isn't a comment but is consumed by the Typescript compiler and points it to the correct reference typings.

PR #430, opened 2018-08-29
Upgrade Misk to Okio 2

swankjesse:


misk-gcp/src/main/kotlin/misk/cloud/gcp/storage/LocalStorageRpc.kt
```diff
@@ -109,7 +111,7 @@ class LocalStorageRpc(root: Path, moshi: Moshi = Moshi.Builder().build()) : Base
     try {
       val upload = beginUpload(obj, options)
       var destOffset: Long = 0
-      val source = Okio.buffer(Okio.source(content))
+      val source = content.source().buffer()
```
adrw: very cool new syntax 👍

PR #434, opened 2018-09-04
Common Webpack

adrw:
- Working common Webpack config that every tab builds off of. 
- Multiple fixes to bugs in recent "Common *" PRs as division of `@misk/` package purposes is being finalized (ie. `@misk/typescript` has been deprecated as it makes more sense to ship this in `@misk/dev`).
- All Webpack configs building off the common base config now require a standard set of metadata in to be added to `package.json`. This allows tab developers to never have to change the code in their tab Webpack.
```JSON
"dependencies: {},
"miskTabWebpack": {
    "name": "Loader",
    "slug": "loader",
    "relative_path_prefix": "_admin/",
    "port": "3100"
  }
```
- Ideally, there is some way to force update the child webpack config and tab tsconfig, tslint in each tab. Upgrade scripts may be the only way possible given that each of these configs has to live as a real file in the tab repo. 
  - Wait............ try sym links? Nope
  - Find some way to have a webpackConfig creator factory that takes input (`__dirname, MiskTabPackageJSONMetadata`)
- [ ] Finish transitioning `@misk/webpack` to a Typescript repo that exports a builder function


adrw:
Likely to be replaced by #435

adrw:
Replaced by #435

PR #441, opened 2018-09-06
sq-market Font

adrw:
Added sq-market font to `@misk/common`  for easy inclusion in all tabs.
miskCachedUrls framework added to @misk/common to ensure refreshed remote included assets like sq-market font


misk/web/@misk/common/README.md
```diff
@@ -101,6 +101,24 @@ Contributing
 - Add window to javascript require include in `src/vendors.js`
 - Update `README.md` with a copy of the updated `package.json` list of packages
 
+#Adding a Cached Remote Asset
+- There are sometimes cases where remote assets should be cached in order to be included in the single line imports of `vendors.js` or `styles.js`
+- Add a new task to `package.json` under the metadata key   `"miskCachedUrls"` with the format as follows
+  ```JSON
+  "miskCachedUrls": {
+    "taskname": {
```
swankjesse: taskname?

swankjesse: maybe rename to myfont?

misk/web/@misk/common/package.json
```diff
@@ -52,5 +53,11 @@
   "devDependencies": {
     "@misk/dev": "^0.0.25",
     "@misk/tslint": "^0.0.3"
+  },
+  "miskCachedUrls": {
+    "sq-market-font": {
+      "filepath": "sq-market.css",
+      "url": "https://d1g145x70srn7h.cloudfront.net/fonts/sqmarket/sq-market.css"
```
swankjesse: we’ll probably need a better strategy for this.

swankjesse: like, one problem with this is that if the resource does change, we need a mechanism to force it to get updated.

It might be simpler to check these files into Misk and make the update process completely manual.

misk/web/@misk/common/README.md
adrw: This is the docs for how to add a static asset download task, taskname is just a filler.

misk/web/@misk/common/package.json
adrw: Agreed, the forced way would be to eventually store all of these `@misk/common` generated static assets (library code, fonts, css) on a service specific CDN with a static URL that never changes and just have the URL point at the updated assets. Then the tab code never has to change or wait for the maintainer to eventually pull the newest release from NPM.

adrw: #444

PR #450, opened 2018-09-10
Basic gRPC connectivity in Misk.

swankjesse:
This is still early, but it works. It adds an @Grpc action annotation
that works like @Post, but that enables some gRPC specific features:

 - We send the gRPC trailer
 - We enforce gRPC content types (and require no content-type is specified)
 - We add gRPC-specific response headers

There's lots of work for follow-up PRs:

 - grpc client support (via Retrofit, ideally)
 - a misk-grpc-testing module that tests connectivity between the Google
   Netty client and our server. This should test message encodings!
 - streamed request and response messages
 - gradle support for generating protos


misk/src/main/kotlin/misk/web/DispatchMechanism.kt
```diff
@@ -0,0 +1,36 @@
...
+   * This can be misleading: web sockets don't behave like normal GETs (they are upgraded), and GRPC
+   * calls don't behave like normal POSTS (they are duplex).
+   */
+  val method: HttpMethod
+    get() {
+      return when (this) {
+        DispatchMechanism.GET -> HttpMethod.GET
+
+        // gRPC layers over POST. https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md
+        DispatchMechanism.GRPC -> HttpMethod.POST
```
adrw: Good wrapping of HttpMethod -> DispatchMechanism 👍

misk/src/main/kotlin/misk/web/actions/WebActionFactory.kt
```diff
@@ -40,20 +39,21 @@ internal class WebActionFactory {
...
+          it.findAnnotation<ConnectWebSocket>() != null ||
+          it.findAnnotation<Grpc>() != null) {
         it as? KFunction<*>
             ?: throw IllegalArgumentException("expected $it to be a function")
       } else null
     }
 
     require(actionFunctions.isNotEmpty()) {
-      "no Get or Post annotations on ${webActionClass.simpleName}"
+      "no @Get, @Post, @ConnectWebSocket, or @Grpc annotations on ${webActionClass.simpleName}"
```
adrw: Updated messages 👍

misk/src/main/kotlin/misk/grpc/GrpcEncoding.kt
```diff
@@ -0,0 +1,29 @@
+package misk.grpc
+
+import okio.BufferedSource
+import okio.GzipSource
+import okio.Source
+import java.net.ProtocolException
+
+internal abstract class GrpcEncoding(val name: String) {
```
ryanhall07: why an abstract class? seems like an interface

misk/src/main/kotlin/misk/grpc/GrpcEncoding.kt
```diff
@@ -0,0 +1,29 @@
...
+
+internal object IdentityGrpcEncoding : GrpcEncoding("identity") {
+  override fun decode(source: BufferedSource) = source
+}
+
+internal object GzipGrpcEncoding : GrpcEncoding("gzip") {
+  override fun decode(source: BufferedSource) = GzipSource(source)
+}
+
+internal fun String.toGrpcEncoding(): GrpcEncoding {
```
ryanhall07: interesting. is this going to add `toGrpcEndoing` as a method to all strings?

misk/src/main/kotlin/misk/web/jetty/WebActionsServlet.kt
```diff
@@ -49,32 +51,33 @@ internal class WebActionsServlet @Inject constructor(
...
+        }
+
+        val bestAction = candidateActions.sorted().firstOrNull()
+        if (bestAction != null) {
+          bestAction.handle(asRequest, response)
+          return
+        }
       }
+    } catch (e: ProtocolException) {
+      // Probably an unexpected HTTP method. Send a 404 below.
```
ryanhall07: can we log the exception server side in case our assumption is wrong?

PR #454, opened 2018-09-11
Test gRPC with a protoc client and a Misk server

swankjesse:
This PR is a little bigger than I'd prefer because it promotes some of our
testing TLS code from misk/src/test to misk-testing/src/main. This is
because now both of our gRPC tests use TLS.


swankjesse:
Gah. https://github.com/junit-team/junit5/issues/1582

misk-grpc-tests/src/main/kotlin/misk/grpc/miskserver/GetFeatureGrpcAction.kt
```diff
@@ -0,0 +1,17 @@
...
+
+class GetFeatureGrpcAction : WebAction {
+  @Grpc("/routeguide.RouteGuide/GetFeature")
+  fun sayHello(@RequestBody point: Point): Feature {
+    return Feature.Builder()
+        .name("maple tree")
+        .location(point)
+        .build()
+  }
+}
```
oldergod: supernit: new line

PR #458, opened 2018-09-12
Vitess doesn't have a username/password it uses mTLS for auth

tirsen:


misk-hibernate/src/main/kotlin/misk/hibernate/SessionFactoryService.kt
```diff
@@ -81,8 +81,12 @@ internal class SessionFactoryService(
       applySetting(AvailableSettings.DRIVER, config.type.driverClassName)
       applySetting("hibernate.hikari.driverClassName", config.type.driverClassName)
       applySetting(AvailableSettings.URL, config.type.buildJdbcUrl(config, environment))
-      applySetting(AvailableSettings.USER, config.username)
-      applySetting(AvailableSettings.PASS, config.password)
+      if (config.username != null) {
```
ryanhall07: should we add some validation that you need to either set username/password or provide the mTLS creds?

PR #466, opened 2018-09-17
Basic entity group framework

tirsen:
Just the basics to code against. No runtime checks.


misk-hibernate/src/main/kotlin/misk/hibernate/Cid.kt
```diff
@@ -0,0 +1,12 @@
+package misk.hibernate
+
+import java.io.Serializable
+import javax.persistence.Embeddable
+
+/** Entity group child id, maps to two long columns one for the root id and one for the row */
+// experiment: this is also available for the root entity where the rootId and memberId is the
```
swankjesse: I like it

misk-hibernate/src/main/kotlin/misk/hibernate/Cid.kt
```diff
@@ -0,0 +1,12 @@
...
+// same, let's see how that plays out
+@Embeddable
+data class Cid<R : DbRoot<R>, T : DbMember<R, T>>(val rootId: Id<R>, val id: Id<T>) : Serializable {
```
swankjesse: R/T is nice

misk-hibernate/src/main/kotlin/misk/hibernate/DbChild.kt
```diff
@@ -0,0 +1,10 @@
+package misk.hibernate
+
+/**
+ * Marker interface for persistent entities. Ensures that only persistent entities can be passed
+ * into [Session] methods.
+ */
+interface DbChild<R : DbRoot<R>, T : DbChild<R, T>> : DbMember<R, T> {
```
swankjesse: should we rename to DbEntity ?

misk-hibernate/src/main/kotlin/misk/hibernate/DbMember.kt
```diff
@@ -0,0 +1,10 @@
+package misk.hibernate
+
+/**
+ * Marker interface for persistent entities. Ensures that only persistent entities can be passed
+ * into [Session] methods.
+ */
+interface DbMember<R : DbRoot<R>, T : DbMember<R, T>> : DbEntity<T> {
+  val rootId : Id<R>
```
swankjesse: this can be inferred from cid, right?

misk-hibernate/src/main/kotlin/misk/hibernate/SessionFactoryService.kt
```diff
@@ -170,14 +172,34 @@ internal class SessionFactoryService(
...
 
+  fun maybeAddSimpleValueProperty(
+    value: Value?,
+    property: Property,
+    entityBinding: PersistentClass,
+    result: MutableSet<KClass<*>>
+  ) {
+    if (value is SimpleValue) {
+      checkNotNull(value.typeName)
+      { "property ${property.name} in class ${entityBinding.className} has null type" }
```
swankjesse: nit: the `{` goes on the preceding line, like any other control flow

misk-hibernate/src/test/kotlin/misk/hibernate/CidGenerator.kt
```diff
@@ -0,0 +1,165 @@
+@file:JvmName("CidGenerator")
```
swankjesse: nit: unnecessary?

misk-hibernate/src/test/kotlin/misk/hibernate/CidGenerator.kt
```diff
@@ -0,0 +1,165 @@
...
+import java.sql.SQLException
+import java.util.Properties
+
+@Suppress("RedundantVisibilityModifier", "unused")
+public class CidGenerator : AbstractPostInsertGenerator(), Configurable {
+  private lateinit var parentColumn: String
+
+  @Throws(MappingException::class)
+  override fun configure(type: Type, params: Properties, serviceRegistry: ServiceRegistry) {
+    this.parentColumn = params.getProperty("parentColumn")
```
swankjesse: are parent and root the same thing? Should we call it rootColumn?

misk-hibernate/src/test/kotlin/misk/hibernate/CidGenerator.kt
```diff
@@ -0,0 +1,165 @@
...
+    checkNotNull(parentColumn)
+  }
+
+  fun generate(s: SessionImplementor, obj: Any): Serializable {
```
swankjesse: curious – why override?

misk-hibernate/src/test/kotlin/misk/hibernate/CidGenerator.kt
```diff
@@ -0,0 +1,165 @@
...
+    return super.generate(s, obj)
+  }
+
+  @Throws(HibernateException::class)
+  override fun getInsertGeneratedIdentifierDelegate(
+    persister: PostInsertIdentityPersister, dialect: Dialect, isGetGeneratedKeysEnabled: Boolean
```
swankjesse: style nit: typically when we wrap params we wrap em all the way

swankjesse: (which is very different from Java)

misk-hibernate/src/test/kotlin/misk/hibernate/CidGenerator.kt
```diff
@@ -0,0 +1,165 @@
...
+    }
+
+    override fun performInsert(
+      insertSQL: String,
+      session: SharedSessionContractImplementor,
+      binder: Binder
+    ): Serializable {
+      @Suppress("UNCHECKED_CAST")
+      val entity = binder.entity as DbChild<Nothing, Nothing>
+      // Square service container supports pre-assigning the cid by the app in which case we don't
```
swankjesse: nice comment!

misk-hibernate/src/test/kotlin/misk/hibernate/CidGenerator.kt
```diff
@@ -0,0 +1,165 @@
...
+          insert.setLong(insert.parameterMetaData.parameterCount, parentId.id)
+          return executeAndExtract(parentId, insert, session)
+        } finally {
+          releaseStatement(insert, session)
+        }
+      } catch (sqle: SQLException) {
+        throw session.factory.serviceRegistry.getService(
+            JdbcServices::class.java).sqlExceptionHelper.convert(
+            sqle,
+            "could not insert: " + MessageHelper.infoString(persister),
```
swankjesse: nit: string template?

misk-hibernate/src/main/kotlin/misk/hibernate/Cid.kt
mmihic: 👍

misk-hibernate/src/main/kotlin/misk/hibernate/DbChild.kt
mmihic: This is specifically a child entity, no? So `DbChild` seems appropriate. `DbMember` is the type that I think could be renamed to `DbEntity`, since it covers both root entities and child entities.

misk-hibernate/src/main/kotlin/misk/hibernate/DbChild.kt
```diff
@@ -0,0 +1,10 @@
+package misk.hibernate
+
+/**
+ * Marker interface for persistent entities. Ensures that only persistent entities can be passed
```
mmihic: nit: comment is duplicative with `DbMember`. Probably want to specify a comment more specific to `DbChild` and its relationship with `DbRoot` and other entities.

misk-hibernate/src/main/kotlin/misk/hibernate/DbChild.kt
tirsen: I'm thinking:
`DbMember` -> `DbSharded`
`DbEntity` -> `DbUnsharded`

I want to name and shame anything that is not sharded by calling this out with the hope that this does not become the default choice.

I'd like to keep this names short and sweet because they're going to be everywhere. It seems an `Entity` suffix doesn't add anything above what the `Db` prefix already says.

misk-hibernate/src/test/kotlin/misk/hibernate/CidGenerator.kt
tirsen: ah yes this was inherited from the Square service container

misk-hibernate/src/test/kotlin/misk/hibernate/CidGenerator.kt
tirsen: I blame the IntelliJ Java->Kotlin converter.

misk-hibernate/src/main/kotlin/misk/hibernate/DbChild.kt
tirsen: It's a bit awkward though because that means `DbSharded` will extend `DbUnsharded`... and all Session APIs will accept a `DbUnsharded` parameter.

I don't think we can have the base class of all persistent entities be called `DbUnsharded`.

Instead I will try this:
* Add a `DbUnsharded` subclass to `DbEntity`.
* In `Session::save` method ensure enforce that an entity either subclasses `DbUnsharded` or `DbSharded`.

Not perfect but we will iterate on this.

misk-hibernate/src/main/kotlin/misk/hibernate/DbChild.kt
tirsen: comments updated

misk-hibernate/src/main/kotlin/misk/hibernate/DbMember.kt
tirsen: How?

misk-hibernate/src/main/kotlin/misk/hibernate/DbEntity.kt
```diff
@@ -3,6 +3,12 @@ package misk.hibernate
...
  * Marker interface for persistent entities. Ensures that only persistent entities can be passed
  * into [Session] methods.
+ *
+ * You can't implement this interface directly because you need to decide on the scaling strategy:
+ *   * Use [DbSharded] subclasses [DbRoot] or [DbChild] if the entity will have unbounded growth
+ *   with usage.
+ *   * Use [DbUnsharded] if your entity has bounded growth like metadata, configuration, static
+ *   data etc.
  */
 interface DbEntity<T : DbEntity<T>> {
```
ryanhall07: why don't we make this `internal` then?

PR #467, opened 2018-09-17
WebActionMetadata Endpoint

adrw:
* New endpoint that returns metadata for bound webactions:

```Kotlin
  name: String,
  function: Function<*>,
  functionAnnotations: List<Annotation>,
  acceptedMediaRanges: List<MediaRange>,
  responseContentType: MediaType?,
  parameterTypes: List<KType>,
  returnType: KType,
  pathPattern: PathPattern,
  applicationInterceptors: List<ApplicationInterceptor>,
  networkInterceptors: List<NetworkInterceptor>,
  dispatchMechanism: DispatchMechanism
```
* Follow up WebActionMetadata Tab coming soon...

### DIFF
--- a/misk/src/main/kotlin/misk/web/actions/WebActionMetadata.kt
+++ b/misk/src/main/kotlin/misk/web/actions/WebActionMetadata.kt
@@ -1,0 +1,51 @@
+package misk.web.actions
+
+import misk.ApplicationInterceptor
+import misk.web.DispatchMechanism
+import misk.web.NetworkInterceptor
+import misk.web.PathPattern
+import misk.web.mediatype.MediaRange
+import okhttp3.MediaType
+import kotlin.reflect.KType
+
+data class WebActionMetadata(
+  val name: String,
+  val function: String,
+  val functionAnnotations: List<String>,
+  val requestMediaTypes: List<String>,
+  val responseMediaType: String?,
+  val parameterTypes: List<String>,
+  val returnType: String,
+  val pathPattern: String,
+  val applicationInterceptors: List<String>,
+  val networkInterceptors: List<String>,
+  val dispatchMechanism: DispatchMechanism
+)
+
+internal fun WebActionMetadata(
+  name: String,
+  function: Function<*>,
+  functionAnnotations: List<Annotation>,
+  acceptedMediaRanges: List<MediaRange>,
+  responseContentType: MediaType?,
+  parameterTypes: List<KType>,
+  returnType: KType,
+  pathPattern: PathPattern,
+  applicationInterceptors: List<ApplicationInterceptor>,
+  networkInterceptors: List<NetworkInterceptor>,
+  dispatchMechanism: DispatchMechanism
+) : WebActionMetadata {
+  return WebActionMetadata(
+      name = name,
+      function = function.toString(),
+      functionAnnotations = functionAnnotations.map { it.toString() },
+      requestMediaTypes = acceptedMediaRanges.map { it.toString() },
+      responseMediaType = responseContentType.toString(),
+      parameterTypes = parameterTypes.map { it.toString() },
+      returnType = returnType.toString(),
+      pathPattern = pathPattern.toString(),
+      applicationInterceptors = applicationInterceptors.map { it::class.qualifiedName.toString() },
+      networkInterceptors = networkInterceptors.map { it::class.qualifiedName.toString() },
+      dispatchMechanism = dispatchMechanism
+  )
+}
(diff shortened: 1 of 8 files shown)


misk/src/main/kotlin/misk/web/actions/WebActionMetadata.kt
```diff
@@ -0,0 +1,51 @@
...
+data class WebActionMetadata(
+  val name: String,
+  val function: String,
+  val functionAnnotations: List<String>,
+  val requestMediaTypes: List<String> = listOf(),
+  val responseMediaType: String?,
+  val parameterTypes: List<String> = listOf(),
+  val returnType: String,
+  val pathPattern: String,
+  val interceptorsApplication: List<String> = listOf(),
```
swankjesse: interceptorsApplication –> applicationInterceptors

misk/src/main/kotlin/misk/web/actions/WebActionMetadata.kt
```diff
@@ -0,0 +1,51 @@
...
+  val interceptorsNetwork: List<String> = listOf(),
```
swankjesse: same

misk/src/main/kotlin/misk/web/actions/WebActionMetadata.kt
```diff
@@ -0,0 +1,51 @@
...
+  val dispatchMechanism: DispatchMechanism
+)
+
+internal fun WebActionMetadata(
+  name: String,
+  function: Function<*>,
+  functionAnnotations: List<Annotation> = listOf(),
```
swankjesse: worth having defaults for these? Are they ever used?

misk/src/main/kotlin/misk/web/actions/WebActionMetadataAction.kt
```diff
@@ -0,0 +1,26 @@
...
+import javax.inject.Singleton
+
+@Singleton
+class WebActionMetadataAction : WebAction {
+  @Inject internal lateinit var servlet: WebActionsServlet
+
+  @Get("/api/webactionmetadata")
+  @RequestContentType(MediaTypes.APPLICATION_JSON)
+  @ResponseContentType(MediaTypes.APPLICATION_JSON)
+  @Unauthenticated
```
swankjesse: we’ll need to fix these ASAP

misk/src/main/kotlin/misk/web/actions/WebActionMetadataAction.kt
```diff
@@ -0,0 +1,26 @@
...
+  @Suppress("UNUSED_PARAMETER")
```
swankjesse: ?

misk/src/main/kotlin/misk/web/actions/WebActionMetadataAction.kt
```diff
@@ -0,0 +1,26 @@
...
+import misk.web.RequestContentType
+import misk.web.ResponseContentType
+import misk.web.jetty.WebActionsServlet
+import misk.web.mediatype.MediaTypes
+import javax.inject.Inject
+import javax.inject.Singleton
+
+@Singleton
+class WebActionMetadataAction : WebAction {
+  @Inject internal lateinit var servlet: WebActionsServlet
```
swankjesse: mind injecting this as a `Provider<WebActionsServlet>` ? As-is there’s a little dependency cycle and this seems like the most natural place to break it.

```
@Inject internal lateinit var servletProvider: Provider<WebActionsServlet>
```

misk/src/main/kotlin/misk/web/jetty/WebActionsServlet.kt
```diff
@@ -102,6 +105,26 @@ internal class WebActionsServlet @Inject constructor(
       realWebSocket.adapter
     }
   }
+
+  internal fun buildWebActionMetadata(): MutableList<WebActionMetadata> {
+    var webActionMetadata: MutableList<WebActionMetadata> = mutableListOf()
+    boundActions.map {
+      webActionMetadata.add(WebActionMetadata(
```
swankjesse: should this be an extension on `BoundAction`, like so?

```
val BoundWebAction.metadata = ...
```

misk/src/main/kotlin/misk/web/actions/WebActionMetadata.kt
adrw: `name` will always be used and can always be deduced by the action name which is a required field
`function` same for function
`functionAnnotations` might be used to discern roles/services but can be empty so `listOf()` seems to be a useful default

misk/src/main/kotlin/misk/web/actions/WebActionMetadataAction.kt
adrw: to add an annotation for admin tab permissions?

misk/src/main/kotlin/misk/web/actions/WebActionMetadata.kt
adrw: jk, I'll remove all the defaults

PR #475, opened 2018-09-19
Config Yaml Parsing

adrw:
* Yaml parsing output fixed for effective config flattened yaml
* Remove `@misk/tabs` dependency on Loader and Config tabs


adrw:
Replaced with #476

PR #476, opened 2018-09-19
Config Yaml Server Parsing

adrw:
* Yaml parsing done on server for standardized return value (object of name -> file Yaml strings)
* Remove `@misk/tabs` dependency on Loader and Config tabs

### DIFF
--- a/misk/src/main/kotlin/misk/config/ConfigAdminAction.kt
+++ b/misk/src/main/kotlin/misk/config/ConfigAdminAction.kt
@@ -1,5 +1,7 @@
 package misk.config
 
+import com.fasterxml.jackson.databind.ObjectMapper
+import com.fasterxml.jackson.dataformat.yaml.YAMLMapper
 import misk.environment.Environment
 import misk.security.authz.Unauthenticated
 import misk.web.Get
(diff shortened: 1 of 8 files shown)


misk/src/main/kotlin/misk/config/ConfigAdminAction.kt
```diff
@@ -23,20 +25,22 @@ class ConfigAdminAction : WebAction {
   @Unauthenticated
   fun getAll(): Response {
     // TODO(mmihic): Need to figure out how to get the overrides.
-    val yamlFiles = MiskConfig.loadConfigYamlMap(appName, environment, listOf())
-    val effectiveYaml = MiskConfig.flattenYamlMap(yamlFiles).toString()
+    val rawYamlFiles = MiskConfig.loadConfigYamlMap(appName, environment, listOf())
+    val effectiveConfigJsonString = MiskConfig.flattenYamlMap(rawYamlFiles).toString()
+    val effectiveConfigJsonNodeTree = ObjectMapper().readTree(effectiveConfigJsonString)
+    val effectiveConfigYaml = YAMLMapper().writeValueAsString(effectiveConfigJsonNodeTree).drop(4)
```
swankjesse: nice

misk/src/main/kotlin/misk/config/ConfigAdminAction.kt
```diff
@@ -23,20 +25,22 @@ class ConfigAdminAction : WebAction {
...
+    val yamlFiles = linkedMapOf<String, String?>("classpath:/live-config.yaml" to effectiveConfigYaml)
```
swankjesse: maybe replace `classpath:/live-config.yaml` with `effective config` ?

swankjesse: no sense impersonating a file name when there isn’t one!

misk/web/@misk/common/src/yaml.ts
```diff
@@ -0,0 +1,78 @@
...
+    result += "\ \ "
+  }
+  return result
+}
+
+/**
+ * @param json 
+ * Stringified Json to Yaml string
+ */
+export const stringifiedJsonToYaml = (json: string) => {
```
swankjesse: obsolete?

PR #487, opened 2018-09-21
Refactored clustering to DRY between k8s and fakes

mmihic:
The only difference between a Kubernetes cluster and a fake cluster is
the source for the events that change the cluster. Refactored clustering
into a DefaultCluster implementation that manages the ready member and
watch lists and computes changes to the cluster in response to incoming
events, and a KubernetesClusterWatcher which converts incoming k8s watch
events into the corresponding DefaultCluster changes.


misk/src/main/kotlin/misk/clustering/Cluster.kt
```diff
@@ -23,23 +23,27 @@ interface Cluster {
...
-    /** @property Boolean true if the current service instance is ready as perceived by the cluster manager */
-    val selfReady: Boolean,
-
     /** @property Set<Member> of the ready members of the cluster */
     val readyMembers: Set<Cluster.Member>,
 
-    /** @property ClusterHashRing built from the ready members of this cluster */
-    val hashRing: ClusterHashRing = ClusterHashRing(readyMembers)
-  ) {
+    /** @property Boolean true if the current service instance is ready as perceived by the cluster manager */
```
swankjesse: is the `@property Boolean` part useful? Presumably people reading the source code don’t need that. If/when we use Dokka (Kotlin’s Javadoc) I don’t think it’ll be useful either?

misk/src/main/kotlin/misk/clustering/ClusterResourceMapper.kt
```diff
@@ -0,0 +1,13 @@
...
+
+/**
+ * A [ClusterResourceMapper] maps string based resource IDs onto members of a cluster for the
+ * purposes of resource ownership. The default [ClusterResourceMapper] is a [ClusterHashRing]
+ * which performs a consistent hash across the cluster member, but [Cluster]s can supply their
+ * own mappings if there is a mechanism specific to that cluster (or to supply a fake)
+ */
+interface ClusterResourceMapper {
+  /** @return The [Cluster.Member] that should own the given resource id */
+  fun mapResourceToMember(resourceId: String): Cluster.Member
```
swankjesse: maybe `operator fun get` ? That lets me do this:

```
val mapper: ClusterResourceMapper = ...
val memberForFoo = mapper["foo"]
```

misk/src/main/kotlin/misk/clustering/DefaultCluster.kt
```diff
@@ -0,0 +1,156 @@
...
+import javax.inject.Singleton
+
+/**
+ * A [DefaultCluster] is the default implementation of the [Cluster], which relies on an outside
+ * source such as a cluster watcher to inform it as members become ready or not ready. The
+ * [DefaultCluster] handles computing the actual cluster changes, and managing application watches
+ * that need to be triggered as the cluster changes.
+ */
+@Singleton
+internal class DefaultCluster(
```
swankjesse: Dig! Makes me contemplate dropping the `Cluster` interface and just having this

swankjesse: ... probably not right though ’cause this needs a thread and fakes shouldn’t need a thread!

misk/src/main/kotlin/misk/clustering/DefaultCluster.kt
```diff
@@ -0,0 +1,156 @@
...
+      { ClusterHashRing(it) }
+) : AbstractExecutionThreadService(), Cluster, DependentService {
+  private val snapshotRef = AtomicReference<Cluster.Snapshot>(Cluster.Snapshot(
+      self = self,
+      selfReady = false,
+      readyMembers = setOf(),
+      resourceMapper = newResourceMapper(setOf())
+  ))
+  private val running = AtomicBoolean(false)
+  private val actions = BlockingArrayQueue<(MutableSet<ClusterWatch>) -> Unit>()
```
swankjesse: neat

misk/src/main/kotlin/misk/clustering/DefaultCluster.kt
```diff
@@ -0,0 +1,156 @@
...
+      watch(Cluster.Changes(snapshot))
+    }
+  }
+
+  /** Triggers a change to the cluster in response members becoming ready or not ready */
+  fun clusterChanged(
+    membersBecomingReady: Set<Cluster.Member> = setOf(),
+    membersBecomingNotReady: Set<Cluster.Member> = setOf()
+  ) {
+    actions.add { watches ->
```
swankjesse: I really like how you’ve structured this . . . it’s all behavior and very little ceremony

misk/src/main/kotlin/misk/clustering/DefaultCluster.kt
```diff
@@ -0,0 +1,156 @@
...
+        }
+      }
+    }
+  }
+
+  /**
+   * Triggers a callback once all of the actions on the queue have been processed. Useful
+   * for writing deterministic tests
+   */
+  internal fun syncPoint(callback: () -> Unit) {
```
swankjesse: instead of taking a callback should this just create its own CountDownLatch?

misk/src/main/kotlin/misk/clustering/DefaultCluster.kt
```diff
@@ -0,0 +1,156 @@
...
+          self = snapshot.self,
+          selfReady = readyMembers.containsKey(snapshot.self.name),
+          readyMembers = newReadyMembers,
+          resourceMapper = cluster.newResourceMapper(newReadyMembers)
+      ), added = membersAdded.toSet(), removed = membersRemoved.toSet())
+    }
+  }
+
+  companion object {
+    val log = getLogger<DefaultCluster>()
```
swankjesse: best place to declare these? I’m indifferent but I know we also do `logger` up top. I think I prefer this because of how it looks in the bytecodes

misk/src/main/kotlin/misk/clustering/fake/ExplicitClusterResourceMapper.kt
```diff
@@ -0,0 +1,36 @@
...
+import misk.clustering.Cluster
+import misk.clustering.ClusterResourceMapper
+import java.util.concurrent.ConcurrentHashMap
+import java.util.concurrent.atomic.AtomicReference
+
+/**
+ * An [ExplicitClusterResourceMapper] is a [ClusterResourceMapper] where the mapping
+ * is explicit managed.
+ */
+class ExplicitClusterResourceMapper : ClusterResourceMapper {
```
swankjesse: nice

misk/src/main/kotlin/misk/clustering/fake/ExplicitClusterResourceMapper.kt
```diff
@@ -0,0 +1,36 @@
...
+  private val mappings = ConcurrentHashMap<String, Cluster.Member>()
```
swankjesse: necessary? will callers actually be single-threaded in practice

misk/src/main/kotlin/misk/clustering/fake/FakeCluster.kt
```diff
@@ -0,0 +1,73 @@
...
+ * is only intended to be used for fakes.
+ *
+ * NB(mmihic): I'd prefer for this to be in the misk-testing module, but it can't since it
+ * relies on [DefaultCluster] and we want to leave [DefaultCluster] client
+ */
+@Singleton
+class FakeCluster(
+  self: Cluster.Member = Cluster.Member(SELF_NAME, SELF_IP),
+  newResourceMapperFn: (Set<Cluster.Member>) -> ClusterResourceMapper = { ClusterHashRing(it) }
+) : Cluster, Service, DependentService {
```
swankjesse: does something break implementation delegation here?
https://kotlinlang.org/docs/reference/delegation.html#implementation-by-delegation

misk/src/main/kotlin/misk/clustering/fake/ExplicitClusterResourceMapper.kt
ryanhall07: does anything use this? would it make sense for the FakeCluster to use it so the test is more explicit?

misk/src/main/kotlin/misk/clustering/fake/FakeCluster.kt
```diff
@@ -0,0 +1,73 @@
...
+import java.util.concurrent.Executor
+import java.util.concurrent.TimeUnit
+import javax.inject.Singleton
+
+/**
+ * A [FakeCluster] is a [Cluster] that is a fake. It delegates entirely to [DefaultCluster],
+ * but allows us to keep [DefaultCluster] internal and make it clear that [FakeCluster]
+ * is only intended to be used for fakes.
+ *
+ * NB(mmihic): I'd prefer for this to be in the misk-testing module, but it can't since it
```
ryanhall07: fwiw I'm starting to find separate testing modules not really worth it and a `Fake` or `Mock` prefix is good enough and sometimes easier to discover when in the real module

misk/src/main/kotlin/misk/clustering/Cluster.kt
mmihic: you mean is the `@property` tag useful? just following the KDoc convention for it (and it bitches if the `Boolean` isn't specified. i don't really care though so happy to remove the tags

misk/src/main/kotlin/misk/clustering/ClusterResourceMapper.kt
mmihic: ah yeah i like that

misk/src/main/kotlin/misk/clustering/DefaultCluster.kt
mmihic: I think I prefer leaving it open ended for the tests, but don't feel strongly about it

misk/src/main/kotlin/misk/clustering/DefaultCluster.kt
mmihic: Yeah, I'm just not really a fan of free-standing private variables, and particularly don't like having the first thing you see in the file be a minor implementation var. 

I guess looking at the bytecodes is an Android thing? I really don't care what server bytecodes look like, unless I'm writing some obj-asm code.

misk/src/main/kotlin/misk/clustering/fake/FakeCluster.kt
mmihic: nope, just wasn't aware of this. nice feature

misk/src/main/kotlin/misk/clustering/fake/FakeCluster.kt
mmihic: yeah, not sure what drove the original testing module split tbh

misk/src/main/kotlin/misk/clustering/fake/ExplicitClusterResourceMapper.kt
mmihic: Things _will_ use it, yes.

misk/src/main/kotlin/misk/clustering/fake/FakeCluster.kt
mmihic: Ah ok so it requires that the delegate be a constructor parameter, but I don't want to expose that type externally, so won't work here

mmihic: eh i made it work

misk/src/main/kotlin/misk/clustering/Cluster.kt
swankjesse: I think we’re better off without the tags. I’ll remove ’em as I go.

misk/src/main/kotlin/misk/clustering/DefaultCluster.kt
swankjesse: Yeah, I’m thinking too hard about Android stuff where minimizing the number of `.class` files in the .jar is a thing.

misk/src/main/kotlin/misk/clustering/fake/FakeCluster.kt
swankjesse: I’m on board with fakes in the main package. More discoverable, and fewer things to put in my .gradle files. Plus we occasionally want to use fake things in say, the development server.

PR #488, opened 2018-09-21
Support for testing Services with external deps

mmihic:
Occasionally a Service requires access to an external resource during
startup. When testing those Services, we often use another Service to
startup that external resource within a test (cf
StartVitessService and StartEtcdService). For the application Service to
work properly, the test service that starts the external resource needs
to run first to make the external resource available. The best way to do
this is to have the application Service specify the test Service as a
dependency via DependentService.consumedKeys. ServiceTestingModule
provides a mechanism for doing this - you install a ServiceTestingModule
withExtraDependencies to specify your test specific dependencies, and
the module binds a Service that combines the application Service
dependencies with the extra dependencies to ensure that test specific
Services start before the application Service.

### DIFF
--- a/misk-testing/src/main/kotlin/misk/service/ServiceTestingModule.kt
+++ b/misk-testing/src/main/kotlin/misk/service/ServiceTestingModule.kt
@@ -1,0 +1,74 @@
+package misk.service
+
+import com.google.common.util.concurrent.Service
+import com.google.inject.Key
+import com.google.inject.Provider
+import misk.DependentService
+import misk.inject.KAbstractModule
+import javax.inject.Singleton
+import kotlin.reflect.KClass
+
+/**
+ * [ServiceTestingModule] provides additional help for testing service, notably allowing existing
+ * services to be bound with additional dependencies specifically required for testing. Typically
+ * used when a [Service] requires an external service (e.g. zookeeper, etcd, vitess, etc)
+ * that is being spun up within the test itself
+ */
+class ServiceTestingModule<T : Service> internal constructor(
+  private val wrappedServiceType: KClass<T>,
+  private val extraDependencies: Set<Key<*>>
+) : KAbstractModule() {
+
+  override fun configure() {
+    // Register a wrapper around the service with the service manager. The wrapper delegates
+    // to the underlying service for all calls except [DependentService.consumedKeys], which
+    // it overrides to extend the consumed service list with the extra dependencies
+    multibind<Service>().toProvider(ServiceWithTestDependenciesProvider(
+        getProvider(wrappedServiceType.java),
+        extraDependencies
+    ))
+
+    // Bind the wrapped service directly so that application code can inject it
+    bind(wrappedServiceType.java)
+  }
+
+  companion object {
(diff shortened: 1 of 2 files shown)


misk-testing/src/main/kotlin/misk/service/ServiceTestingModule.kt
```diff
@@ -0,0 +1,92 @@
...
+    // Bind the wrapped service directly so that application code can inject it
+    bind(wrappedServiceType.java)
+  }
+
+  companion object {
+    /** @return A [Module] binding the service with extra dependencies */
+    fun <T : Service> withExtraDependencies(
+      wrappedServiceType: KClass<T>,
+      vararg extraDependencies: Key<*>
+    ): com.google.inject.Module =
```
ryanhall07: why the fully qualified name?

misk-testing/src/test/kotlin/misk/service/ServiceTestingModuleTest.kt
```diff
@@ -0,0 +1,111 @@
...
+  class AppService @Inject internal constructor(
+    private val lowestLevelService: LowestLevelService,
+    private val client: FakeExternalClient
+  ) : AbstractIdleService(), DependentService {
+
+    override val consumedKeys: Set<Key<*>> = setOf(keyOf<LowestLevelService>())
+    override val producedKeys: Set<Key<*>> = setOf()
+
+    override fun startUp() {
+      System.out.println("confirming low level service")
```
wesleyk: do we typically commit in println debug statements?

mmihic: nope, that's a mistake. thanks for the catch

misk-testing/src/main/kotlin/misk/service/ServiceTestingModule.kt
mmihic: because `Module` is now a class in the `java.lang` package

misk-testing/src/test/kotlin/misk/service/ServiceTestingModuleTest.kt
```diff
@@ -0,0 +1,108 @@
...
+import javax.inject.Singleton
+
+internal class ServiceTestingModuleTest {
+  @Test fun serviceTestModuleAddsTestSpecificDependency() {
+    val injector = Guice.createInjector(object : KAbstractModule() {
+      override fun configure() {
+        install(MiskServiceModule())
+        multibind<Service>().to<LowestLevelService>()
+        install(ServiceTestingModule.withExtraDependencies<AppService>(
+            keyOf<ExternalServiceSpinupService>())
```
wesleyk: are we exercising any of the "extra dependencies" bit? Is that worthwhile?

mmihic: Not sure what you mean? This test confirms the extra dependencies

wesleyk: Derp, misread the constructor. 👍

PR #499, opened 2018-09-25
Dweisberg.prometheus histogram extract

danw5050:


misk/src/main/kotlin/misk/client/ClientMetricsInterceptor.kt
```diff
@@ -9,14 +10,15 @@ import javax.inject.Singleton
 
 class ClientMetricsInterceptor internal constructor(
   private val actionName: String,
-  private val requestDuration: Histogram
+  private val requestDuration: PrometheusHistogram
```
ryanhall07: client code (code that records metrics) should not have a reference to any specific histogram impl. This should be using some interface

misk/src/main/kotlin/misk/digester/HistogramInterface.kt
```diff
@@ -0,0 +1,11 @@
+package misk.digester
+
+import io.prometheus.client.Histogram
+
+//Skeleton interface for all histograms
+interface HistogramInterface {
```
mmihic: nit: no `Interface` in names - just call  this `Histogram`

misk/src/main/kotlin/misk/client/ClientMetricsInterceptor.kt
```diff
@@ -13,12 +13,14 @@ class ClientMetricsInterceptor internal constructor(
 ) : ClientNetworkInterceptor {
 
   override fun intercept(chain: ClientNetworkChain): Response {
+
```
ryanhall07: rm these extra new lines

misk/src/main/kotlin/misk/digester/Histogram.kt
```diff
@@ -0,0 +1,8 @@
+package misk.digester
+
+interface Histogram {
+
+    fun record(vararg labelValues: String): Any
```
ryanhall07: what does this method do if you don't provide a duration to record?

misk/src/main/kotlin/misk/digester/PrometheusHistogram.kt
```diff
@@ -0,0 +1,13 @@
+package misk.digester
+
+class PrometheusHistogram: Histogram {
+    lateinit var histogram: io.prometheus.client.Histogram
+
+    override fun record(vararg  labelValues: String): io.prometheus.client.Histogram.Child {
```
ryanhall07: oh we should remove this method. it's meant to be a Builder pattern which we no longer support with the new interface

misk/src/main/kotlin/misk/metrics/Metrics.kt
```diff
@@ -38,14 +38,12 @@ class Metrics @Inject internal constructor(val registry: CollectorRegistry) {
...
-    buckets: DoubleArray? = null
-  ): Histogram {
-    val builder = Histogram.build(name, help).labelNames(*labelNames.toTypedArray())
-    if (buckets != null) builder.buckets(*buckets)
-    return builder.register(registry)
+          name: String,
+          help: String = "",
+          labelNames: List<String> = listOf(),
+          buckets: DoubleArray? = null
+  ): misk.digester.Histogram {
```
ryanhall07: you shouldn't need the fully qualified `Histogram` name here. just `Histogram` should be fine right?

misk/src/test/kotlin/misk/client/ClientMetricsInterceptorTest.kt
```diff
@@ -63,17 +66,18 @@ internal class ClientMetricsInterceptorTest {
...
   @Test
   fun responseCodes() {
-    assertThat(requestDuration.labels("pinger.ping", "all").get().count).isEqualTo(6)
-    assertThat(requestDuration.labels("pinger.ping", "202").get().count).isEqualTo(1)
-    assertThat(requestDuration.labels("pinger.ping", "4xx").get().count).isEqualTo(2)
-    assertThat(requestDuration.labels("pinger.ping", "404").get().count).isEqualTo(1)
-    assertThat(requestDuration.labels("pinger.ping", "403").get().count).isEqualTo(1)
-    assertThat(requestDuration.labels("pinger.ping", "5xx").get().count).isEqualTo(1)
-    assertThat(requestDuration.labels("pinger.ping", "503").get().count).isEqualTo(1)
+    assertThat((requestDuration.record("pinger.ping", "all") as io.prometheus.client.Histogram.Child).get().count).isEqualTo(6)
```
ryanhall07: you shouldn't need to cast this. just add a `count` method to the base `Histogram` interface. I think that's a reasonable method to add

misk/src/test/kotlin/misk/client/ClientMetricsInterceptorTest.kt
```diff
@@ -40,17 +41,19 @@ internal class ClientMetricsInterceptorTest {
...
+  @Inject
+  private lateinit var jetty: JettyService
   private lateinit var requestDuration: Histogram
   private lateinit var clientMetrics: Metrics
   private lateinit var clientInjector: Injector
 
   @BeforeEach
   fun createClient() {
-    clientInjector = Guice.createInjector(ClientModule(jetty))
+    clientInjector = Guice.createInjector(ClientModule(jetty)) //this should be changed to the approach of MetricsInterceptorTest.kt?
```
ryanhall07: probably not. this test is trying to simulate the difference between a client and a server running into 2 different apps, which would result in 2 different injectors (1 injector per app)

misk/src/main/kotlin/misk/digester/Histogram.kt
danw5050: This was to support the tests (like the ones in MetricsInterceptorTest). I'll refactor the tests to send a duration as well.

ryanhall07: Actually I think the method existed to support a builder pattern with the prometheus client

misk/src/main/kotlin/misk/metrics/Metrics.kt
danw5050: There was arbitration from import io.prometheus.client.*, I'll change the imports to exclude prometheus.Histogram

misk/src/test/kotlin/misk/client/ClientMetricsInterceptorTest.kt
danw5050: Thanks for clarifying

misk/src/main/kotlin/misk/metrics/Metrics.kt
```diff
@@ -57,4 +56,4 @@ class Metrics @Inject internal constructor(val registry: CollectorRegistry) {
   }
 }
 
-internal val Histogram.Child.Value.count get() = buckets.max()?.toInt() ?: 0
+internal val io.prometheus.client.Histogram.Child.Value.count get() = buckets.max()?.toInt() ?: 0
```
ryanhall07: I don't think this is required anymore, since we added `count` to the `Histogram` interface

ryanhall07: either way, nothing in this class should have a reference to a prometheus Histogram

misk/src/test/kotlin/misk/client/ClientMetricsInterceptorTest.kt
```diff
@@ -40,17 +40,19 @@ internal class ClientMetricsInterceptorTest {
   @MiskTestModule
   val module = TestModule()
 
-  @Inject private lateinit var jetty: JettyService
+  @Inject
+  private lateinit var jetty: JettyService
```
ryanhall07: can you make sure you have the square formatter installed in your intellij?

https://github.com/square/java-code-styles

misk-testing/src/main/kotlin/misk/testing/MiskTestExtension.kt
```diff
@@ -24,6 +25,7 @@ internal class MiskTestExtension : BeforeEachCallback, AfterEachCallback {
         }
         for (module in context.getActionTestModules()) {
           install(module)
+          install(PrometheusHistogramRegistryModule())
```
ryanhall07: I don't think this is the right place for this. the test should be providing this module. in the future when we don't want to use prometheus histograms, there will be no way to prevent a duplicate binding.

ryanhall07: `MiskServiceModule` is probably a good place for it

misk/src/main/kotlin/misk/prometheus/PrometheusHistogram.kt
```diff
@@ -0,0 +1,18 @@
+package misk.prometheus
+
+import misk.metrics.Histogram
+
+class PrometheusHistogram: Histogram {
+    lateinit var histogram: io.prometheus.client.Histogram
+    var bucketCount: Int = 0
```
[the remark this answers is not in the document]
danw5050: You need to maintain a reference to the child object in order to get this count.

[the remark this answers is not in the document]
danw5050: You need to maintain a reference to the child that you made. All access to buckets otherwise seems to be restricted within the Histogram class.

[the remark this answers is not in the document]
danw5050: We don't seem to use the record function to return anything. Would it make sense to have it return the bucketcount of the child it just made?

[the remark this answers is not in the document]
ryanhall07: well what is count even representing? it looks to be scoped to the data point level, which include labels. so having a single count per histogram doesn't even make sense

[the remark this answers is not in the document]
danw5050: count is representing the number of buckets in a child. A child is the value of a single histogram. It would actually make more sense to return the buckets of a child once you create the child (in record) rather then having a separate method within the prometheus histogram class.

misk/src/main/kotlin/misk/metrics/Histogram.kt
```diff
@@ -0,0 +1,19 @@
...
+ *
+ * A histogram samples observations (usually things like request durations or response sizes)
+ * and counts them in configurable buckets.
+ *
+ * A sample implementation can be found in PrometheusHistogram
+ */
+
+interface Histogram {
+
+    //records a new set of labels and accompanying duration
```
ryanhall07: this should be in java doc form, like `/** */`

PR #500, opened 2018-09-25
Fix builds so that we can release to Maven central

swankjesse:


gradle.properties
```diff
@@ -1,6 +1,6 @@
 GROUP=com.squareup.misk
 
-VERSION_NAME=18.9.10-51cc9bc3
+VERSION_NAME=0.5.0-SNAPSHOT
```
adrw: We're moving back to Snapshots?

swankjesse: Nope! It’s just the most convenient thing to put in git since most builds aren’t release builds.

adrw: Sweet, makes more sense!

PR #511, opened 2018-09-27
Remove coroutines experimental flag from Gradle

adrw:
- No longer needed to run `WebProxyActionTest`
- Causing errors in Groovy syntax checks in IntelliJ


adrw:
jk, looks like it still needs the flag

PR #517, opened 2018-09-30
Add lease APIs

mmihic:
Along with fakes for cluster and lease management

### DIFF
--- a/misk/src/main/kotlin/misk/clustering/lease/Lease.kt
+++ b/misk/src/main/kotlin/misk/clustering/lease/Lease.kt
@@ -1,0 +1,21 @@
+package misk.clustering.lease
+
+/**
+ * A [Lease] is a cluster-wide time-based lock on a given resource. Leases are retrieved via
+ * [LeaseManager.requestLease]. A ready service instance will automatically attempt to acquire
+ * leases that it thinks it should own (typically based on the resource being leased consistently
+ * hashing to the service instance), and will continue to maintain the lease for as long as it is
+ * still ready. Leases should be released if the service transitions into not ready; to avoid
+ * flapping, service lease owners may want to delay releasing leases until they've been not ready
+ * for a particular amount of time.
+ */
+interface Lease {
+  /** @property String the name of the resource being leased */
+  val name: String;
+
+  /**
+   * @return true if the lease is owned by this process instance. This may involve remote calls,
+   * so it is marked as a function rather than a property to make the potential expense clearer
+   */
+  fun checkHeld(): Boolean
+}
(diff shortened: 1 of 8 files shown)


misk/src/main/kotlin/misk/clustering/lease/Lease.kt
```diff
@@ -0,0 +1,21 @@
...
+ * flapping, service lease owners may want to delay releasing leases until they've been not ready
+ * for a particular amount of time.
+ */
+interface Lease {
+  /** @property String the name of the resource being leased */
+  val name: String;
+
+  /**
+   * @return true if the lease is owned by this process instance. This may involve remote calls,
+   * so it is marked as a function rather than a property to make the potential expense clearer
```
swankjesse: 👍🏻

misk/src/main/kotlin/misk/clustering/lease/LeaseManager.kt
```diff
@@ -0,0 +1,14 @@
...
+
+interface LeaseManager {
+  /**
+   * Registers interest in the lease with the given name. Service instances should register their
+   * interest in leases as soon as they know about them; this gives the process an opportunity
+   * to notify the underlying cluster system that it is interested in the lease, or to setup
+   * other background tasks to acquire the lease if necessary. Note that registering interest
+   * in a lease doesn't necessarily mean that this service instance will attempt to acquire
+   * the lease; many clustering systems will only try to acquire leases for resources that
+   * they think they should own based on some consistent hashing system
```
swankjesse: nit: missed a `.`

also https://nedbatchelder.com/blog/201401/comments_should_be_sentences.html

misk/src/main/kotlin/misk/clustering/lease/LeaseManager.kt
```diff
@@ -0,0 +1,14 @@
...
+   */
+  fun requestLease(name: String): Lease
```
swankjesse: I’m not sure on this API. As a lease user I’m expecting methods like `acquire()` and `release()`, this API has a weaker `acquire()` and no `release()`.

Should we call this something else?

mmihic: That's intentional. You don't explicitly `acquire` and `release` a lease - you register interest in it, and if the clustering system decides you are the instance that it should be assigned to, then in the background it is acquired. We're modeling load balanced leases here, where all processes in the cluster indicate that they are potential candidates for owning a lease, and the actual distribution of named lease ownership is handled by the clustering system.

mmihic: That said, I'm open to other names.

mmihic: I merged with the current name, but I'm also not happy with it so if you have other suggestions, I'm down to change it.

swankjesse: Thumbs up. I don’t have better names either.

PR #518, opened 2018-10-01
Use Guice bindings to check service singletons

mmihic:
Previously we were looking for explicitly @Singleton annotations on
bound Service classes. This misses bindings for classes that are not
annotated as @Singleton, but are nevertheless singletons as a result of
explicitly specifying a scope during binding or being bound as an
instance.

### DIFF
--- a/misk/src/test/kotlin/misk/MiskServiceModuleTest.kt
+++ b/misk/src/test/kotlin/misk/MiskServiceModuleTest.kt
@@ -4,6 +4,9 @@ import com.google.common.util.concurrent.AbstractIdleService
 import com.google.common.util.concurrent.Service
 import com.google.common.util.concurrent.ServiceManager
 import com.google.inject.Guice
+import com.google.inject.Provides
+import com.google.inject.Scopes
+import com.google.inject.Singleton
 import misk.inject.KAbstractModule
 import misk.inject.getInstance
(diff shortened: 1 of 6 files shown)


misk/src/main/java/misk/ServiceBindings.java
```diff
@@ -0,0 +1,12 @@
+package misk;
+
+import com.google.common.util.concurrent.Service;
+import com.google.inject.Key;
+import com.google.inject.TypeLiteral;
+import java.util.Set;
+
+final class ServiceBindings {
```
swankjesse: why Java?

misk/src/test/kotlin/misk/MiskServiceModuleTest.kt
```diff
@@ -33,14 +34,50 @@ internal class MiskServiceModuleTest {
     override fun shutDown() {}
   }
 
+  class ExplicitEagerSingletonService : AbstractIdleService() {
```
swankjesse: Nice!

misk/src/main/java/misk/ServiceBindings.java
mmihic: I couldn't figure out the required magic incantations to produce the proper `TypeLiteral` in Kotlin (`@JvmSuppressWildcards` didn't do it)

PR #519, opened 2018-10-02
RepeatedTaskQueue for performing recurring tasks

mmihic:
With support for backoff on task failures and lack of work to do

### DIFF
--- a/misk/src/main/kotlin/misk/concurrent/ExplicitReleaseBlockingQueue.kt
+++ b/misk/src/main/kotlin/misk/concurrent/ExplicitReleaseBlockingQueue.kt
@@ -1,0 +1,80 @@
+package misk.concurrent
+
+import java.util.concurrent.BlockingQueue
+import java.util.concurrent.Delayed
+import java.util.concurrent.LinkedBlockingQueue
+import java.util.concurrent.PriorityBlockingQueue
+import java.util.concurrent.TimeUnit
+
+/**
+ * An [ExplicitReleaseBlockingQueue] is a [BlockingQueue] that only returns elements from
+ * [BlockingQueue.take], [BlockingQueue.poll], and [BlockingQueue.peek] after a call to
+ * [ExplicitReleaseBlockingQueue.release]. Used by tests that want to explicitly control when
+ * pollers receive queued items.
+ */
+class ExplicitReleaseBlockingQueue<T> internal constructor(
+  private val visible: BlockingQueue<T>,
+  private val pending: BlockingQueue<T>
+) : BlockingQueue<T> {
+  constructor() : this(LinkedBlockingQueue<T>(), LinkedBlockingQueue<T>())
+
+  fun release(n: Int) {
+    for (i in (0 until n)) {
+      val e = pending.poll() ?: return
+      visible.add(e)
+    }
+  }
+
+  override fun contains(element: T) = visible.contains(element)
+
+  override fun addAll(elements: Collection<T>) = pending.addAll(elements)
+
+  override fun clear() {
+    visible.clear()
+    pending.clear()
+  }
+
+  override fun removeAll(elements: Collection<T>): Boolean {
+    val removedFromReleased = visible.removeAll(elements)
+    val removedFromPending = pending.removeAll(elements)
+    return removedFromReleased || removedFromPending
+  }
+
+  override fun add(element: T) = pending.add(element)
+  override fun offer(e: T) = pending.offer(e)
+  override fun offer(e: T, timeout: Long, unit: TimeUnit) = pending.offer(e, timeout, unit)
+  override fun put(e: T) = pending.put(e)
+  fun peekPending(): T? = pending.peek()
+
+  override fun element(): T = visible.element()
+  override fun take(): T = visible.take()
+  override fun iterator() = visible.iterator()
+  override fun peek(): T? = visible.peek()
+  override fun isEmpty() = visible.isEmpty()
+  override fun poll(timeout: Long, unit: TimeUnit): T? = visible.poll(timeout, unit)
+  override fun poll(): T? = visible.poll()
+  override fun drainTo(c: MutableCollection<in T>) = visible.drainTo(c)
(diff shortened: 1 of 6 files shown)


misk/src/main/kotlin/misk/concurrent/ExplicitReleaseBlockingQueue.kt
```diff
@@ -0,0 +1,80 @@
...
+import java.util.concurrent.BlockingQueue
+import java.util.concurrent.Delayed
+import java.util.concurrent.LinkedBlockingQueue
+import java.util.concurrent.PriorityBlockingQueue
+import java.util.concurrent.TimeUnit
+
+/**
+ * An [ExplicitReleaseBlockingQueue] is a [BlockingQueue] that only returns elements from
+ * [BlockingQueue.take], [BlockingQueue.poll], and [BlockingQueue.peek] when explicitly
+ * released by the application. Useful for tests that want to explicitly control when
```
wesleyk: Just curious, is this only meant for tests? Should we be more clear about that?

mmihic: Sure - is there wording you'd want to see beyond the comment above?

misk/src/main/kotlin/misk/tasks/RepeatedTaskQueue.kt
```diff
@@ -0,0 +1,126 @@
...
+          }
+          Status.NO_WORK -> {
+            failureBackoff.reset()
+            Result(status, noWorkBackoff.nextRetry())
+          }
+          Status.FAILED -> {
+            noWorkBackoff.reset()
+            Result(status, failureBackoff.nextRetry())
+          }
+          Status.OK_NO_RESCHEDULE, Status.FAILED_NO_RESCHEDULE ->
```
wesleyk: is the distinction in these statuses mainly for metrics/observability?

misk/src/main/kotlin/misk/tasks/RepeatedTaskQueue.kt
```diff
@@ -0,0 +1,126 @@
...
+            Result(status, Duration.ofMillis(0))
+        }
+      } catch (th: Throwable) {
+        noWorkBackoff.reset()
```
wesleyk: will this exception be logged in any way?

misk/src/main/kotlin/misk/tasks/RepeatedTaskQueue.kt
mmihic: Yes. We could just change it to `NO_RESCHEDULE` 🤷‍♂️

misk/src/main/kotlin/misk/tasks/RepeatedTaskQueue.kt
mmihic: It's up to the task to log - we lack sufficient context to do so here

mmihic: We'll backup log here; I'm generally opposed to low level libraries logging because they don't have context and tend to make things extra chatty, but in this place it's probably important enough to do

misk/src/main/kotlin/misk/concurrent/ExplicitReleaseBlockingQueue.kt
```diff
@@ -0,0 +1,80 @@
...
+import java.util.concurrent.Delayed
+import java.util.concurrent.LinkedBlockingQueue
+import java.util.concurrent.PriorityBlockingQueue
+import java.util.concurrent.TimeUnit
+
+/**
+ * An [ExplicitReleaseBlockingQueue] is a [BlockingQueue] that only returns elements from
+ * [BlockingQueue.take], [BlockingQueue.poll], and [BlockingQueue.peek] after a call to
+ * [ExplicitReleaseBlockingQueue.release]. Used by tests that want to explicitly control when
+ * pollers receive queued items.
```
mmihic: @wesleyk updated the comment

misk/src/test/kotlin/misk/tasks/RepeatedTaskQueueTest.kt
```diff
@@ -0,0 +1,486 @@
...
+      // Wait for the task to schedule its follow up
+      waitForNextPendingTask()
+    }
+
+    // Next task should be scheduled at the no-work backoff
+    val nextTask = waitForNextPendingTask()
+    assertThat(nextTask.getDelay(TimeUnit.SECONDS)).isEqualTo(100L)
+  }
+
+  @Test fun resetFailureBackoffOnSuccess() {
```
wesleyk: /nit could these resetXonY tets be DRYed up a bit? Seems like they follow a similar pattern of setup -> perform 5 times in state X -> perform once in state Y -> perform once more in state X -> confirm final backoff

misk/src/main/kotlin/misk/tasks/RepeatedTaskQueue.kt
ryanhall07: agreed. it seems at this point there is a bug in the task handler since it did not properly return `FAILED`

misk/src/main/kotlin/misk/tasks/RepeatedTaskQueue.kt
```diff
@@ -0,0 +1,127 @@
...
+          Status.NO_WORK -> {
+            failureBackoff.reset()
+            Result(status, noWorkBackoff.nextRetry())
+          }
+          Status.FAILED -> {
+            noWorkBackoff.reset()
+            Result(status, failureBackoff.nextRetry())
+          }
+          Status.NO_RESCHEDULE ->
+            Result(status, Duration.ofMillis(0))
```
ryanhall07: it took me awhile to realize that the delay of 0 has no impact, since `NO_RESCHEDULE` is explicitly handled. makes sense, but probably worth a comment that the delay value here has no impact

PR #520, opened 2018-10-02
Docker Web Builds

adrw:
* New `squareup/misk-node` base image that includes hydrated `yarn cache` of all `@misk/` packages for faster clean builds of tabs. Used as base for `squareup/misk-build` and as a daemon environment to `docker exec` individual tab builds by Gradle.
* New `squareup/misk-build` docker image that will kick off clean build of all sub tabs and `@misk/` packages in a mounted `web/` directory
* Updated Gradle tasks to spin up `squareup/misk-node` container and run any web build tasks with `docker exec` in the container
* Improved uniformity across tabs and packages of build commands in `package.json`. `yarn gradle` as a command can now be run in any package or tab to kick off a clean build.
* Upcoming PRs
  * [ ] add support for running `webpack-dev-server` in Docker container
  * [ ] bring back support for using Gradle cache to not run clean builds every time `./gradlew jar` is run


docker/misk-build/Dockerfile
```diff
@@ -0,0 +1,6 @@
+FROM squareup/misk-node:0.0.1
+
+COPY misk-build.sh /usr/local/bin/
+RUN ln -s usr/local/bin/misk-build.sh
+ENTRYPOINT ["misk-build.sh"]
```
gflarity: Think you want CMD here although in the case it may not matter too much:
http://goinbigdata.com/docker-run-vs-cmd-vs-entrypoint/

The default entry point would be bash and the command would be misk-build.

gflarity: I'd just merge these two docker files into one unless there's something I'm missing? Do you not want that misk-build.sh in the base for some reason.

docker/misk-node/Dockerfile
```diff
@@ -0,0 +1,16 @@
...
+    --pure-lockfile \
+    --non-interactive \
+      @misk/common \
+      @misk/components \
+      @misk/dev \
+      @misk/tslint \
+  && rm -rf /tmp/web/node_modules \
+  && yarn global add webpack webpack-cli webpack-dev-server
+
+CMD tail -f /dev/null
```
gflarity: Guess this is needed because you go to have a command? Merging the two would solve this.

docker/misk-node/Dockerfile
```diff
@@ -0,0 +1,16 @@
+FROM node:alpine
+
+WORKDIR /tmp/web/
+
+# Populate yarn cache by installing all @misk/ packages
+RUN yarn add \
```
gflarity: Keep in mind that to update these packages you'll need to use --no-cache with the docker build command. Otherwise the result of this RUN statement just gets cached as an image (layer). That cache will be used instead of running yarn again next time.
 
https://stackoverflow.com/questions/35594987/how-to-force-docker-for-a-clean-build-of-an-image

README.md
```diff
@@ -4,3 +4,20 @@ Misk is a new open source application container from Square.
...
+- [Template Exemplar Web Tab](/misk/web/tabs/example)
+
+Workflow
+---
+1. Fork into your personal Github account
+1. Clone repo from your fork
+1. Add square as a remote with `git remote add square git@github.com:square/misk.git`
+1. Pull any new changes with `git pull square master`
+1. Push any new changes to a new branch in your personal fork
+1. Open PR from personal fork -> square/master
```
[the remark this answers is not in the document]
adrw: It removes the current requirement for anyone working on Misk to have write access to the repository to push a new "personal" branch with their changes. It doesn't need to change how write-access contributors work but provides a workflow for new contributors to open PRs before they're given write access. https://blog.scottlowe.org/2015/01/27/using-fork-branch-git-workflow/

docker/misk-node/Dockerfile
adrw: Added to the `build-tag-push.sh` script.

docker/misk-build/Dockerfile
adrw: We have another planned image to spin up dev servers so we may keep them as separate docker files for now.

docker/misk-node/Dockerfile
adrw: Main purpose is to allow this image to be run as an arbitrary environment that someone can run with `-it` flags or `docker exec` arbitrary commands into without it dying between commands.

README.md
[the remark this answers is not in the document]
swankjesse: I think this is good advice, but we should eventually move it off the readme into a contributors section or something. Totally fine for now.

PR #523, opened 2018-10-03
Test support for code using RepeatedTaskQueue

mmihic:
Allows upstream module code to create RepeatedTaskQueues that are backed
by an ExplicitReleaseDelayQueue, which in turn allows writing tests that
deterministically release tasks for execution.


misk/src/main/kotlin/misk/concurrent/ExplicitReleaseBlockingQueue.kt
```diff
@@ -18,11 +18,23 @@ class ExplicitReleaseBlockingQueue<T> internal constructor(
 ) : BlockingQueue<T> {
   constructor() : this(LinkedBlockingQueue<T>(), LinkedBlockingQueue<T>())
 
-  fun release(n: Int) {
+  /**
+   * releases up to n items from the pending queue, making them visible to [take], [poll], [peek],
+   * etc. This method does not "extend credit"; if n > number of pending items on the queue, every
+   * currently pending item is made visible, but any items added in the future continue to go
+   * onto the pending queue
```
swankjesse: thanks

PR #530, opened 2018-10-04
More eventing APIs

mmihic:


misk-events/src/main/kotlin/misk/events/Producer.kt
```diff
@@ -17,12 +17,18 @@ interface Producer {
    *
    * Transactions remaining outstanding until the application calls commit, or rollback, or until
    * a producer specific timeout occurs.
+   *
+   * Transactions may only be used from the thread that began them; accessing a transaction
+   * in multiple threads or handing it off between threads has undefined behavior.
+   *
+   * Having multiple outstanding transactions on the same thread also has undefined behavior.
```
wesleyk: is this something we can/want to actually restrict out-right (similar to how we attempt to block nested db transactions)

misk/src/main/kotlin/misk/concurrent/ExplicitReleaseBlockingQueue.kt
```diff
@@ -37,6 +37,16 @@ class ExplicitReleaseBlockingQueue<T> internal constructor(
     return numReleased
   }
 
+  /** releases all items from the pending queue, returning the number of items released */
+  fun releaseAll() : Int {
+    var numReleased = 0
+    while (true) {
+      val e = pending.poll() ?: return numReleased
+      visible.add(e)
+      numReleased ++
```
wesleyk: is it koitlin style to have a space between var and ++?

swankjesse: strange

PR #534, opened 2018-10-05
Make web_proxy_action httpClient optionally injectable

adrw:
* Paired on with jwilson
* Defaults to use localhost for proxy requests
* Allows for overriding injection of configured client that runs web proxy action requests over a different address


misk/web/README.md
```diff
@@ -1,6 +1,30 @@
-Misk Admin
+Misk Web
+===
+
+Installing the Admin Dashboard on a new service
 ---
 
+For the purpose of this guide, the service will be called DinoService. If this is an internal service, then the dashboard is included in SkimModule.
```
wesleyk: let's not mention Skim here

misk/src/main/kotlin/misk/web/proxy/WebProxyAction.kt
```diff
@@ -99,3 +99,12 @@ class WebProxyAction @Inject constructor(
   }
 }
 
+/**
+ * https://github.com/google/guice/wiki/FrequentlyAskedQuestions#how-can-i-inject-optional-parameters-into-a-constructor
+ */
+@Singleton
+class OptionalProxyClient {
+  @Inject(optional = true)
```
wesleyk: prefer OptionalBinder? https://groups.google.com/forum/#!topic/google-guice/zvhQiZozcDU

adrw: yeah, I like that.

PR #537, opened 2018-10-08
Implemented Merging and Veneur Digest

danw5050:
The work is done on VeneurDigest, MergingDiges, VeneurDigestTest, and build-wire.sh. The rest is autogenerated protos.


danw5050:
The work is done on VeneurDigest, MergingDiges,  VeneurDigestTest, and build-wire.sh. The rest is autogenerated protos.

misk/src/main/kotlin/misk/digester/VeneurDigest.kt
```diff
@@ -0,0 +1,79 @@
...
+  fun newDefaultVeneurDigest(): VeneurDigest {
+    return VeneurDigest(mergingDigest.newMerging(50.0), 0, 0.0)
+  }
+
+  /**
+   *  NewVeneurDigestFromProto creates a VeneurDigest from a DigestData proto
+   *  The DigestData proto must have veneur_digest set correctly
+   */
+  fun newVeneurDigestFromProto(digestData: DigestData): VeneurDigest {
+    var mergingDigestData: MergingDigestData = MergingDigestData.ADAPTER.decode(digestData.veneur_digest)
```
danw5050: Should the error from decode failing be handled somewhere here to still return some kind of VeneurDigest with a default mergingDigest?

misk/src/test/kotlin/misk/digester/VeneurDigestTest.kt
```diff
@@ -0,0 +1,108 @@
...
+import org.assertj.core.api.Assertions
+import org.junit.jupiter.api.Test
+
+class VeneurDigestTest {
+
+  @Test
+  fun testVeneurDigest() {
+
+    var digest = VeneurDigest().newDefaultVeneurDigest()
+    Assertions.assertThat(digest.mergingDigest.quantile( 0.5)).isEqualTo(Double.NaN)
```
tirsen: import the assertThat method

misk/src/main/kotlin/misk/digester/MergingDigest.kt
```diff
@@ -0,0 +1,351 @@
...
+
+import misk.proto.tdigest.Centroid
+import misk.proto.tdigest.MergingDigestData
+import kotlin.math.PI
+import kotlin.math.asin
+
+class MergingDigest {
+
+    //amount of compression to data
+    var compression: Double = 0.0
```
alecholmes: Make this and other member variables private since they are internal to the class's behavior. `private var compression: Double = 0.0`

By default in Kotlin, omitting a visibility modifier implies `public`.

misk/src/main/kotlin/misk/digester/MergingDigest.kt
```diff
@@ -0,0 +1,351 @@
+package misk.digester
+
+import misk.proto.tdigest.Centroid
+import misk.proto.tdigest.MergingDigestData
+import kotlin.math.PI
+import kotlin.math.asin
+
+class MergingDigest {
+
+    //amount of compression to data
```
alecholmes: Style nit: space at the beginning of comments (`// Amount of compression to data`)

misk/src/main/kotlin/misk/digester/MergingDigest.kt
```diff
@@ -0,0 +1,351 @@
...
+
+    //centroids that have been added but not yet merged into main list
+    var tempCentroids: MutableList<Centroid> = mutableListOf()
+    //total weight of unmerged temp centroids
+    var tempWeight: Double = 0.0
+
+    var min: Double = Double.POSITIVE_INFINITY
+    var max: Double = Double.NEGATIVE_INFINITY
+
+    constructor()
```
alecholmes: I don't think this should have a default constructor. Since compression is a core part of the data structure, the caller should specify it.

misk/src/main/kotlin/misk/digester/MergingDigest.kt
```diff
@@ -0,0 +1,351 @@
...
+
+    constructor(compression: Double) {
+      this.compression = compression
+    }
+
+    constructor (
```
alecholmes: This shouldn't be a public constructor. Callers should not be able to set internals.

misk/src/main/kotlin/misk/digester/MergingDigest.kt
```diff
@@ -0,0 +1,351 @@
...
+      this.max = max
+    }
+
+  /**
+   * Initializes a new merging t-digest using the given compression parameter.
+   * Lower compression values result in reduced memory consumption and less
+   * precision, especially at the median. Values from 20 to 1000 are recommended
+   * in Dunning's paper.
+   */
+  fun newMerging(compressionValue: Double): MergingDigest {
```
alecholmes: Remove this. Callers should create instances using the constructor:
```
val digest = MergingDigest(123)
```

misk/src/main/kotlin/misk/digester/MergingDigest.kt
```diff
@@ -0,0 +1,351 @@
...
+
+    return MergingDigest(compressionValue)
+  }
+
+  /**
+   *  NewMergingFromData returns a MergingDigest with values initialized from
+   * MergingDigestData.  This should be the way to generate a MergingDigest
+   * from a serialized protobuf.
+   */
+  fun newMergingFromData(mergingDigestData: MergingDigestData): MergingDigest {
```
alecholmes: Change this to a constructor:

```
val digest = MergingDigest(mergingDigestData)
```

Alternatively, you could have a companion object instead:
```
class MergingDigest {
  companion object {
    fun fromData(mergingDigestData: MergingDigestData) : MergingDigest {
      return MergingDigest(mergingDigestData)
    }
}
```
```
val digest = MergingDigest.fromData(mergingDigestData)
```

misk/src/main/kotlin/misk/digester/MergingDigest.kt
```diff
@@ -0,0 +1,351 @@
+package misk.digester
+
+import misk.proto.tdigest.Centroid
+import misk.proto.tdigest.MergingDigestData
+import kotlin.math.PI
+import kotlin.math.asin
+
+class MergingDigest {
```
alecholmes: Add a comment here describing what this type is and making it very clear that it's a direct port of the Veneur digest. Link to the Go code too.

misk/src/main/kotlin/misk/digester/MergingDigest.kt
```diff
@@ -0,0 +1,351 @@
...
+    return (7.5 + 0.37*tempCompression - 2e-4*tempCompression*tempCompression).toInt()
+  }
+
+  /**
+   * Adds a new value to the t-digest, with a given weight that must be positive.
+   * Infinities and NaN cannot be added.
+   */
+  fun add(value: Double, weight: Double) {
+    if (Double.NaN == value || Double.POSITIVE_INFINITY == value || weight <= 0) {
+      Exception("invalid value added")
```
alecholmes: InvalidArgumentException

misk/src/main/kotlin/misk/digester/MergingDigest.kt
```diff
@@ -0,0 +1,351 @@
...
+    }
+
+    if (tempCentroids.size == estimateTempBuffer(compression)) {
+          mergeAllTemps() //<- will we still need this?
```
alecholmes: Yes, I think we will still need it.

misk/src/main/kotlin/misk/digester/MergingDigest.kt
```diff
@@ -0,0 +1,351 @@
...
+
+    tempCentroids.add(next)
+    tempWeight += weight
+
+  }
+
+  /**
+   * combine the mainCentroids and tempCentroids in-place into mainCentroids
+   */
+  fun mergeAllTemps() {
```
alecholmes: `private`

misk/src/main/kotlin/misk/digester/MergingDigest.kt
```diff
@@ -0,0 +1,351 @@
...
+    }
+
+    return td
+  }
+
+  /**
+   * function taken from stripes implementation of merging_digest
+   * https://github.com/stripe/veneur/blob/master/tdigest/merging_digest.go
+   */
+  fun estimateTempBuffer(compressionValue: Double): Int {
```
alecholmes: `private`

Only the upper-cased methods from the Go implementation (like `Add`) should be ported as public.

misk/src/main/kotlin/misk/digester/MergingDigest.kt
```diff
@@ -0,0 +1,351 @@
...
+      mainCentroids.add(next)
+      // return the last index that was merged into the previous centroid
+      return indexEstimate(beforeWeight / totalWeight)
+    } else {
+      // the new index fits into the range of the current centroid, so we
+      // combine it into the current centroid's values
+      // this computation is known as welford's method, the order matters
+      // weight must be updated before mean
+
+      //TODO: is this thread safe? doublecheck that this successfully updates all properties
```
alecholmes: This class isn't thread-safe in general. That's ok, though. We'll just need to make sure that whatever uses it deals with concurrency.

misk/src/main/kotlin/misk/digester/MergingDigest.kt
```diff
@@ -0,0 +1,351 @@
...
+      return beforeIndex
+    }
+  }
+
+  /**
+   * given a quantile, estimate the index of the centroid that contains it using
+   * the given compression
+   */
+  fun indexEstimate(quantile: Double) : Double {
+    return compression * ((asin(2*quantile-1) / PI) + 0.5).toFloat()
```
alecholmes: `toDouble`. Using `toFloat` will lose precision.

alecholmes: Is `toDouble` even necessary though?

misk/src/main/kotlin/misk/digester/MergingDigest.kt
```diff
@@ -0,0 +1,351 @@
...
+  }
+
+  /**
+   * Returns a value such that the fraction of values in td below that value is
+   * approximately equal to quantile. Returns NaN if the digest is empty.
+   */
+  fun quantile(quantile: Double): Double {
+    if (quantile < 0 || quantile > 1) {
+      throw IndexOutOfBoundsException("quantile out of bounds")
```
alecholmes: `IllegalArgumentException`. `IndexOutOfBoundsException` is used for lists/arrays.

misk/src/main/kotlin/misk/digester/MergingDigest.kt
mmihic: imo these should be `val`s and specified in the ctor

misk/src/main/kotlin/misk/digester/MergingDigest.kt
mmihic: nit: use kotlin `require` for this. e.g.

```kotlin
require(value != Double.NaN && value != Double.POSITIVE_INFINITY && weight > 0) {
   "invalid value added"
}
```

misk/src/main/kotlin/misk/digester/MergingDigest.kt
```diff
@@ -0,0 +1,351 @@
...
+
+    var next = Centroid(
+      value,
+      weight,
+      mutableListOf<Double>()
+    )
+
+    tempCentroids.add(next)
+    tempWeight += weight
+
```
mmihic: nit: no blank line

misk/src/main/kotlin/misk/digester/MergingDigest.kt
mmihic:
```kotlin
require(quantile >= 0 && quantile <= 1) { "quantile must be between 0 and 1" }
```

misk/src/main/kotlin/misk/digester/MergingDigest.kt
```diff
@@ -0,0 +1,351 @@
...
+        // distribution
+        return lowerBound + (proportion * (upperBound - lowerBound))
+      }
+
+      // the quantile is above this centroid, so sum the weight and carry on
+      weightSoFar += c.weight
+      lowerBound = upperBound
+    }
+
+    // should never be reached unless empty, since the final comparison is
```
mmihic: if you should "never" reach this, throw an exception

misk/src/main/kotlin/misk/digester/MergingDigest.kt
```diff
@@ -0,0 +1,351 @@
...
+    } else {
+      return max
+    }
+  }
+
+  /**
+   * merge another digest into this one. Neither td nor other can be shared
+   * concurrently during the execution of this method.
+   */
+  fun merge(other: MergingDigest) {
```
mmihic: This should either return a new `MergingDigest` which is the result of the merge (my preference, so that digests can be made immutable) or should indicate in the method name which digest is being modified (e.g. `mergeFrom` if the receiver is being modified, or `mergeInto` if the parameter is being modified)

misk/src/main/kotlin/misk/digester/VeneurDigest.kt
```diff
@@ -0,0 +1,79 @@
...
+    other.sum += sum
+  }
+
+  /**proto returns a representation fo the t-digest that can be later be reconstituted into an instance of the same type */
+  fun proto(): DigestData {
+    val encode: ByteArray = MergingDigestData.ADAPTER.encode(mergingDigest.data())
+    return DigestData(count, sum, encode.toByteString())
+  }
+
+  /** NewVeneurDigest creates a TDigest backed by a VeneurDigest */
```
mmihic: method docs in kotlin shouldn't have the method names (differs from golang style comments)

misk/src/test/kotlin/misk/digester/VeneurDigestTest.kt
```diff
@@ -0,0 +1,108 @@
...
+  interface TestCasesStructure {
+    var sourceVals: Array<Double>
+    var destVals: Array<Double>
+    var expectedMedian: Double
+    var expectedSum: Double
+  }
+
+  @Test
+  fun testVeneurDigest_MergeInto() {
+    val testCases: Array<TestCasesStructure>
```
alecholmes: I don't know Kotlin well enough to know the best conventions for table/parameterized tests. I suspect it's notably different since this will report as one unit test, not unit test per case in the `testCases` array.

That said, a few things I would change:
* Rename `TestCasesStructure` to make it less generic-sounding.
* Make `TestCasesStructure` private.
* Make `TestCasesStructure` a data class (https://kotlinlang.org/docs/reference/data-classes.html) instead of an interface.

misk/src/main/kotlin/misk/digester/proto/build-wire.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/sh
+
+# Stolen from https://github.com/square/misk/blob/master/misk-grpc-tests/build-wire.sh
+
+# TODO(mmihic): Totally temporary script to build wire files. To be replaces with
+# proper gradle support
+# Build protos captured in test resources
+java -jar ~/Downloads/wire-compiler-2.3.0-RC1-jar-with-dependencies.jar \
```
mmihic: p sure there is a gradle task for this now; @drashkov ?

misk/src/main/kotlin/misk/digester/proto/service.proto
```diff
@@ -0,0 +1,85 @@
+syntax = "proto3";
+
+option java_multiple_files = true;
+//option java_package = "com.squareup.digester.protos.service";
+option java_outer_classname = "DigesterServiceProtos";
+
+package misk.proto.service;
+
+// Service that exposes t-digest-based histogram data.
+service DigestProvider {
```
mmihic: hmm - misk core shouldn't be exposing service interfaces. This should go into a separate module

misk/src/main/kotlin/misk/digester/MergingDigest.kt
alecholmes: We can keep this mutable since `merge` is an internal method; I left a separate comment about making a bunch of methods private.

Since `MergingDigests` will have many values added to them, I think we should keep them mutable like other metrics-oriented collector types.

misk/src/main/kotlin/misk/digester/MergingDigest.kt
danw5050: I dont think that an exception should be the expected behaviour in this case. Calling quantile when there is no populated data should still run and be valid, but should return NaN.

misk/src/main/java/digester/service.proto
```diff
@@ -0,0 +1,84 @@
+syntax = "proto3";
+
+option java_multiple_files = true;
+option java_outer_classname = "DigesterServiceProtos";
```
alecholmes: Keep the `option java_package = "com.squareup.digester.protos.service";` from the original proto file.

misk/src/main/java/tdigest/tdigest.proto
```diff
@@ -0,0 +1,24 @@
+
+syntax = "proto3";
+
+package tdigest;
```
alecholmes: Let's package this near the other protos:`option java_package = "com.squareup.digester.protos.tdigest";`

misk/src/main/kotlin/misk/digester/MergingDigest.kt
```diff
@@ -0,0 +1,323 @@
...
+
+  // centroids that have been added but not yet merged into main list
+  private var tempCentroids: MutableList<Centroid> = mutableListOf()
+  // total weight of unmerged temp centroids
+  private var tempWeight: Double = 0.0
+
+  private var min: Double = Double.POSITIVE_INFINITY
+  private var max: Double = Double.NEGATIVE_INFINITY
+
+  constructor(compression: Double) {
```
mmihic: nit: make this the default constructor.

```
class MergingDigest(private val compression: Double) {
   constructor(mergingDigestData: MergingDigestData) : this(mergingDigestData.compression) {
      // other initializers
   } 
}
```

misk/src/main/kotlin/misk/digester/MergingDigest.kt
```diff
@@ -0,0 +1,323 @@
...
+ * Direct port of Veneur digest created by Stripe
+ * https://github.com/stripe/veneur/blob/master/tdigest/merging_digest.go
+ */
+class MergingDigest {
+
+  // amount of compression to data
+  private val compression: Double
+
+  // main list of centroids
+  private var mainCentroids: MutableList<Centroid> = mutableListOf()
```
mmihic: Why is this both a `var` and `mutable`? Should likely either be a mutable `val` (if you are changing the contents of the list in place) or a `var) (if you are changing the reference to the list but not updating the contents in place)

misk/src/main/kotlin/misk/digester/MergingDigest.kt
```diff
@@ -0,0 +1,323 @@
...
+  /**
+   * Adds a new value to the t-digest, with a given weight that must be positive.
+   * Infinities and NaN cannot be added.
+   */
+  fun add(value: Double, weight: Double) {
+    require (value != Double.NaN && value != Double.POSITIVE_INFINITY && weight > 0) {
+      "invalid value added"
+    }
+
+    if (tempCentroids.size == estimateTempBuffer(compression)) {
```
mmihic: Formatting is off on all of these files (should be two space indent). Make sure you to use the Square Intellij code style settings (https://github.com/square/java-code-styles)

misk/src/main/kotlin/misk/digester/MergingDigest.kt
danw5050: Both properties are used:
`mainCentroids = mergingDigestData.main_centroids.toMutableList()`
`mainCentroids.add(next)`

alecholmes:
@danw5050 is there anything left to do on this PR? Before merging, please squash all the commits.

danw5050:
@alecholmes Somebody with write access still needs to approve the PR like @ryanhall07

misk/src/main/kotlin/misk/digester/MergingDigest.kt
```diff
@@ -0,0 +1,315 @@
...
+      mergeAllTemps()
+    }
+
+    min = minOf(min, value)
+    max = maxOf(max, value)
+
+    var next = Centroid(
+        value,
+        weight,
+        mutableListOf<Double>()
```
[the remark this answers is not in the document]
danw5050: Centroid is autogenerated by proto.

ryanhall07:
I think this should be in a separate misk module, like `misk-tdigests`. Also I don't think the name `digester` should really be in any of this, that's an internal SQ service that consumes `t-digests`. I'd be confused as an outside user.

`misk-tdigests` should be a module, that if installed, provides a binding from `Histogram` to `TDigest` and additionally installs a server that allows `TDigest`s to be collected by some service that misk does not care about. SQ uses the `digester` app to scrape that server.

PR #544, opened 2018-10-11
Add support for setting jetty acceptQueueSize, acceptors, and selectors

jjestrel:
Adds support for setting `acceptQueueSize`, `acceptors`, and `selectors` in jetty's `ServerConnector`

### DIFF
--- a/misk/src/main/kotlin/misk/web/WebConfig.kt
+++ b/misk/src/main/kotlin/misk/web/WebConfig.kt
@@ -8,7 +8,10 @@ data class WebConfig(
   val port: Int,
   val idle_timeout: Long,
   val host: String? = null,
-  val ssl: WebSslConfig? = null
+  val ssl: WebSslConfig? = null,
+  val selectors: Int? = null,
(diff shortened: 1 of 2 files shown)


misk/src/main/kotlin/misk/web/WebConfig.kt
```diff
@@ -8,7 +8,10 @@ data class WebConfig(
   val port: Int,
   val idle_timeout: Long,
   val host: String? = null,
-  val ssl: WebSslConfig? = null
+  val ssl: WebSslConfig? = null,
+  val selectors: Int? = null,
```
swankjesse: I’d like for us to be really strongly opinionated on tuning parameters like this. If more than a few special services need to tune things like this we’re probably doing a bad job as infrastructure provider.

mmihic: IMO we should be using the current configuration to figure out proper tuning parameters, then make those the defaults.

PR #545, opened 2018-10-11
Track jetty connection and thread pool metrics

mmihic:
* jetty_connections_total    - total # of connections accepted
* jetty_connections_active   - current # of open connections
* jetty_bytes_recvd_total    - total # of bytes received
* jetty_bytes_sent_total     - total # of bytes sent
* jetty_msgs_recvd_total     - total # of messages received
* jetty_msgs_sent_total      - total # of messages sent
* jetty_thread_pool_size     - current # of threads in the jetty pool
* jetty_thread_pool_busy     - current # of busy threads
* jetty_thread_pool_utilization - % utilization of the pool
* jetty_thread_pool_utilization_max - % utilization relative to the max
pool size

### DIFF
--- a/misk/src/main/kotlin/misk/web/jetty/ConnectionMetrics.kt
+++ b/misk/src/main/kotlin/misk/web/jetty/ConnectionMetrics.kt
@@ -1,0 +1,53 @@
+package misk.web.jetty
+
+import com.google.common.annotations.VisibleForTesting
+import misk.metrics.Metrics
+import javax.inject.Inject
+import javax.inject.Singleton
+
+@Singleton
+internal class ConnectionMetrics @Inject internal constructor(metrics: Metrics) {
+  val acceptedConnections = metrics.counter(
+      "jetty_connections_total",
+      "total number of connections accepted by jetty",
+      labels
+  )
+  val connectionDurations = metrics.summary(
+      "jetty_connection_duration_ms",
+      "average duration a incoming jetty connection is held open (in ms)",
+      labels
+  )
+  val activeConnections = metrics.gauge(
+      "jetty_connections_active",
+      "number of currently active connections in jetty",
+      labels
+  )
+  val bytesReceived = metrics.counter(
+      "jetty_bytes_recvd_total",
+      "total count of bytes received by jetty",
+      labels
+  )
+  val bytesSent = metrics.counter(
+      "jetty_bytes_sent_total",
+      "total count of bytes sent through jetty",
+      labels
+  )
+  val messagesReceived = metrics.counter(
+      "jetty_msgs_recvd_total",
+      "total count of HTTP messages received by jetty",
+      labels
+  )
+  val messagesSent = metrics.counter(
+      "jetty_msgs_sent_total",
+      "total count of HTTP messages sent by jetty",
+      labels
+  )
+
+  companion object {
+    @VisibleForTesting internal val labels = listOf("protocol", "port")
+
+    fun forPort(protocol: String, port: Int): Array<String> {
+      return arrayOf(protocol, port.toString())
+    }
+  }
+}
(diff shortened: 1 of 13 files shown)


misk/src/main/kotlin/misk/web/jetty/ConnectionMetricsCollector.kt
```diff
@@ -0,0 +1,37 @@
...
+    val bytesOut = connection.bytesOut
+    if (bytesOut > 0) metrics.bytesSent.labels(*labels).inc(bytesOut.toDouble())
+
+    val messagesIn = connection.messagesIn
+    if (messagesIn > 0) metrics.messagesReceived.labels(*labels).inc(messagesIn.toDouble())
+
+    val messagesOut = connection.messagesOut
+    if (messagesOut > 0) metrics.messagesSent.labels(*labels).inc(messagesOut.toDouble())
+
+    val connectionDuration = (System.currentTimeMillis() - connection.createdTimeStamp).toDouble()
```
swankjesse: createdTimeStamp ugh

swankjesse: interesting that we don’t count connections ’til they’re closed

swankjesse: (no idea on how to do better here, but it could be surprising)

misk/src/main/kotlin/misk/web/jetty/JettyThreadPoolMetricsCollector.kt
```diff
@@ -0,0 +1,31 @@
...
+import misk.logging.getLogger
+import java.util.Timer
+import javax.inject.Inject
+import javax.inject.Singleton
+import kotlin.concurrent.scheduleAtFixedRate
+
+@Singleton
+internal class JettyThreadPoolMetricsCollector @Inject internal constructor(
+  private val metrics: ThreadPoolMetrics
+) : AbstractIdleService() {
```
swankjesse: https://github.com/google/guava/wiki/ServiceExplained#abstractscheduledservice ?

misk/src/test/kotlin/misk/web/JettyServiceMetricsTest.kt
```diff
@@ -0,0 +1,127 @@
...
+import okhttp3.HttpUrl
+import okhttp3.OkHttpClient
+import okhttp3.Request
+import org.assertj.core.api.Assertions.assertThat
+import org.assertj.core.data.Percentage
+import org.junit.jupiter.api.Test
+import javax.inject.Inject
+
+@MiskTest(startService = true)
+internal class JettyServiceMetricsTest {
```
swankjesse: I love these tests

misk/src/main/kotlin/misk/web/jetty/ConnectionMetricsCollector.kt
mmihic: Yeah it's weird (although also what we have today). We could potentially do something like track the connections and have a periodic refresh timer that updates the counters based on this. Let me add that.

misk/src/main/kotlin/misk/web/jetty/ConnectionMetrics.kt
```diff
@@ -0,0 +1,53 @@
...
+import javax.inject.Singleton
+
+@Singleton
+internal class ConnectionMetrics @Inject internal constructor(metrics: Metrics) {
+  val acceptedConnections = metrics.counter(
+      "jetty_connections_total",
+      "total number of connections accepted by jetty",
+      labels
+  )
+  val connectionDurations = metrics.summary(
```
ryanhall07: I think we want a `histogram`

ryanhall07: @alecholmes  we should actually just get rid of `summary` since we can't aggregate it

mmihic: And we'll get one...as soon as histograms are finished :)

ryanhall07: got it. you prefer the summary for now. makes sense

PR #549, opened 2018-10-12
Implemented windower for digester

danw5050:


misk/src/main/kotlin/misk/digester/Windower.kt
```diff
@@ -0,0 +1,108 @@
+package misk.digester
+import java.util.Calendar
```
alecholmes: Formatting seems off. Did you run the autoformatter?

misk/src/main/kotlin/misk/digester/Windower.kt
```diff
@@ -0,0 +1,108 @@
...
+/** Window represents a time range */
+class Window (private val start: Calendar, private val end: Calendar) {
```
alecholmes: Java time libraries have changed a lot over the years. The preferred current way of representing instants in time is `java.time.Instant`

alecholmes: Use a data class instead, which gives you equals, toString, and hashCode for free.

```
data class Window (val start: Instant, val end: Instant) {
  fun contains(t: Instant): Boolean { ...  }
}
``` 

This also make `start` and `end` public.

misk/src/main/kotlin/misk/digester/Windower.kt
```diff
@@ -0,0 +1,108 @@
...
+    return start.time == other.start.time && end.time == other.end.time
+  }
+  /** Returns string representing the start and end time */
+  fun string(): String {
+    return "[${start.time}, ${end.time})"
+  }
+}
+/** Windower contains multiple windows with a specified duration */
+class Windower {
+  private var windowSize: Int = 0
```
alecholmes: Both `windowSize` and `startSecs` should be initialized in the constructor since they aren't known prior to that point.

misk/src/main/kotlin/misk/digester/Windower.kt
```diff
@@ -0,0 +1,108 @@
...
+  val startSecs: MutableList<Int> = mutableListOf()
```
alecholmes: The `startSecs` list should be immutable.

misk/src/main/kotlin/misk/digester/Windower.kt
```diff
@@ -0,0 +1,108 @@
...
+    while (start < 60) {
+      startSecs.add(start)
+      for (i in 1..(stagger-1)) {
+        startSecs.add(start + i*windowSecs/stagger)
+      }
+      start += windowSecs
+    }
+    this.windowSize = windowSecs
+  }
+  /**
```
alecholmes: Add a blank line between each method. I think the autoformatter will do this for you.

misk/src/main/kotlin/misk/digester/Windower.kt
```diff
@@ -0,0 +1,108 @@
...
+   * Returns all windows that the given time falls into.
+   * The returned slice will be ordered by window start time, and
+   * the number of windows in the returned slice will the same as the stagger
+   * given when the Windower was created.
+   */
+  fun windowContaining(t: Calendar): MutableList<Window> {
```
alecholmes: `windowsContaining`

alecholmes: `List`. Generally prefer immutability.

alecholmes: `Instant` and other `java.time.*` types here and in the method.

misk/src/main/kotlin/misk/digester/Windower.kt
```diff
@@ -0,0 +1,108 @@
...
+    // Find the earliest possible time the window could start,
+    // then round up to the nearest second boundary
+    var windowStart: Calendar = t.clone() as Calendar
+    windowStart.add(Calendar.SECOND, -windowSize)
+    var boundaryIndex = 0
+    for (i in 0..(startSecs.count()-1)) {
```
alecholmes:
```
for (i in 0 until startSecs.count())
```

misk/src/test/kotlin/misk/digester/WindowerTest.kt
```diff
@@ -0,0 +1,215 @@
...
+import kotlin.test.assertEquals
+class WindowerTest {
+  @Test
+  fun testWindowerInitialBoundaries() {
+    assertThat(Windower(60, 1).startSecs).isEqualTo(mutableListOf(0))
+    assertThat(Windower(30, 1).startSecs).isEqualTo(mutableListOf(0, 30))
+    assertThat(Windower(60, 2).startSecs).isEqualTo(mutableListOf(0, 30))
+    assertThat(Windower(60, 3).startSecs).isEqualTo(mutableListOf(0, 20, 40))
+  }
+  private data class WindowerTestClass(
```
alecholmes: Can you look into https://junit.org/junit5/docs/current/user-guide/#writing-tests-parameterized-tests? Seems like the canonical JUnit way to do table tests.

misk/src/test/kotlin/misk/digester/WindowerTest.kt
```diff
@@ -0,0 +1,215 @@
...
+    val windowSecs: Int,
+    val count: Int,
+    val at: WindowerTestClassCalendar,
+    val expected: Array<WindowTestClass>
+  )
+  private class WindowTestClass(
+    val start: WindowerTestClassCalendar,
+    val end: WindowerTestClassCalendar
+  )
+  private data class WindowerTestClassCalendar(
```
alecholmes: You shouldn't need this once you move to `Instant`s.

misk/src/test/kotlin/misk/digester/WindowerTest.kt
danw5050: Using parameterized tests would not make the testsuite more clear or readable. I don't think there is an advantage to using them in this situation as a custom data class would still be needed.

misk/src/main/kotlin/misk/digester/Windower.kt
```diff
@@ -0,0 +1,117 @@
...
+    var currWindow = Window(windowStart, windowEnd)
+    while (true) {
+      if (currWindow.contains(t)) {
+        windows.add(currWindow)
+      } else if (windows.count() > 0) {
+        // if len(windows) == 0 that means the first potential window started too early.
+        // otherwise, len(windows) > 0 means that all possible windows were already appended
+        break
+      }
+      val (start, index) = nextTime(windowStart, boundaryIndex)
```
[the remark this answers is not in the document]
danw5050: There is no way to do that as of yet. https://stackoverflow.com/questions/39273283/reassigning-variables-via-destructuring

misk/src/test/kotlin/misk/digester/WindowerTest.kt
```diff
@@ -0,0 +1,192 @@
...
+                )
+            )
+        )
+    )
+
+    testCases.forEachIndexed { i, t ->
+      // Get list of returned windows
+      val windowContaining = Windower(t.windowSecs, t.count).windowsContaining(t.at)
+      // Check number of windows to equal to expected
+      assertEquals(t.expected.count(), windowContaining.count(),
```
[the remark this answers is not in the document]
danw5050: What is the benefit of using hasSize over assertEquals? Also, hasSize cannot be used in this way.

PR #550, opened 2018-10-15
Do not attempt to parse args after block and await termination

ryanhall07:


misk/src/main/kotlin/misk/MiskApplication.kt
```diff
@@ -49,6 +49,7 @@ class MiskApplication(private val modules: List<Module>, commands: List<MiskComm
   internal fun doRun(args: Array<String>) {
     if (args.isEmpty()) {
       startServiceAndAwaitTermination()
+      return
```
mmihic: lulz

PR #564, opened 2018-10-18
Change misk tagging signature to String -> Any?

jjestrel:


misk/src/main/kotlin/misk/logging/Logging.kt
```diff
@@ -5,7 +5,7 @@ import mu.KotlinLogging
 import org.slf4j.MDC
 import org.slf4j.event.Level
 
-typealias Tag = Pair<String, String>
+typealias Tag = Pair<String, Any?>
```
mmihic: Probably shouldn't be nullable.

swankjesse: Should we just skip if the tag’s value is null?

jjestrel: I think seeing something like `physicalCard: null` might be helpful

PR #566, opened 2018-10-18
Dweisberg.implement window slider

danw5050:


misk/src/main/kotlin/misk/digester/FakeDigest.kt
```diff
@@ -0,0 +1,61 @@
...
+  override fun sum(): Double {
+    if (count > 0) {
+      return sum
+    }
+    return Double.NaN
+  }
+
+  /** Merges this t-digest into another t-digest */
+  override fun mergeInto(other: TDigest) {
+    require(other is FakeDigest) {
```
danw5050: Is there a better way of overriding the TDigest mergeInto function to accept only FakeDigest? Maybe through guice?

misk/src/main/kotlin/misk/digester/SlidingWindowDigest.kt
```diff
@@ -0,0 +1,188 @@
...
+        if (existing.window.equals(wd.window)) {
+          wd.Digest.mergeInto(existing.Digest)
+          found = true
+          break
+        }
+      }
+      if (!found) {
+        val newDigest = WindowDigest(
+            wd.window,
+            FakeDigest() //should this be some kind of custom new tDigest function?
```
danw5050: Did this to pass the tests, but this will eventually need to be abstracted to TDigest(). Guice again?

misk/src/main/kotlin/misk/digester/SlidingWindowDigest.kt
```diff
@@ -0,0 +1,188 @@
...
+        if (existingWindow.window.equals(newWindow)) {
+          found = true
+          digests.add(existingWindow)
+          break
+        }
+      }
+      if (!found) {
+        val newDigest = WindowDigest(
+            newWindow,
+            FakeDigest()
```
danw5050: Same issue of abstraction to TDigest() as above.

misk/src/main/kotlin/misk/digester/FakeDigest.kt
alecholmes: You can use the type system. Here's an example in Java (apologies for not being able to write Kotlin yet).

```
import java.io.*;
import java.util.*;

/*
 * To execute Java, please define "static void main" on a class
 * named Solution.
 *
 * If you need more classes, simply define them inline.
 */

class Solution {
  static interface TDigest<T extends TDigest<T>> {
    void mergeInto(T other);
  }
  
  static class FakeTDigest implements TDigest<FakeTDigest> {
    public void mergeInto(FakeTDigest other) {}
  }
  
  static class RealTDigest implements TDigest<RealTDigest> {
    public void mergeInto(RealTDigest other) {}
  }
  
  public static void main(String[] args) {
    FakeTDigest fake = new FakeTDigest();
    RealTDigest real = new RealTDigest();
    
    real.mergeInto(real);
    // real.mergeInto(fake); // won't compile
  }
}
```

misk/src/main/kotlin/misk/digester/SlidingWindowDigest.kt
```diff
@@ -0,0 +1,188 @@
...
+ * The following example creates a 10 second sliding window where there are 2 overlapping windows at a given time.
+ * Reported quantiles are at most 5 seconds out of date:
+ * NewSlidingWindowDigest(time.Now, NewDefaultVeneurDigest, NewWindower(10, 2))
+ */
+class SlidingWindowDigest constructor(
+  internal var utcNowClock: Clock,
+  internal val windower: Windower
+) {
+
+  internal var windows: MutableList<WindowDigest> = mutableListOf()
```
alecholmes: Why internal? Keep as much private as possible.

misk/src/main/kotlin/misk/digester/SlidingWindowDigest.kt
```diff
@@ -0,0 +1,188 @@
...
+ * The following example creates a 1 minute sliding window where there are 6 overlapping windows at a given time.
+ * Reported quantiles are at most 10 seconds out of date.
+ * NewSlidingWindowDigest(time.Now, NewDefaultVeneurDigest, NewWindower(60, 6))
+ *
+ * The following example creates a 10 second sliding window where there are 2 overlapping windows at a given time.
+ * Reported quantiles are at most 5 seconds out of date:
+ * NewSlidingWindowDigest(time.Now, NewDefaultVeneurDigest, NewWindower(10, 2))
+ */
+class SlidingWindowDigest constructor(
+  internal var utcNowClock: Clock,
```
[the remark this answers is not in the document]
danw5050: The clocks available for testing in Kotlin (Clock.fixed) are immutable in the sense that it is not possible to update the time on them. My approach is to simply create a new Clock instance each time the date/time changes and reassign it to `utcNowClock`. I did find external libraries like https://github.com/robfletcher/test-clock that could accomplish it but nothing native.

misk/src/main/kotlin/misk/digester/SlidingWindowDigest.kt
danw5050: Testcases that need windows cannot access it when private

misk/src/main/kotlin/misk/digester/SlidingWindowDigest.kt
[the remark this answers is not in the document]
alecholmes: I'd ask what other teams are doing for fake/mutable clocks in the #java or #kotlin channels. The reference to the clock shouldn't change (as in, don't replace the clock). Instead, use a single clock that returns different but controllable times.

misk/src/main/kotlin/misk/digester/TDigest.kt
```diff
@@ -0,0 +1,32 @@
...
+  fun count(): Long
+
+  /** Returns the sum of all values added into the digest, or NaN if no values have been added. */
+  fun sum(): Double
+
+  /**
+   * MergeInto merges the data in this digest into the other digest.
+   * The other digest is mutated and must be of the same TDigest type.
+   */
+  fun mergeInto(other: TDigest<*>)
```
alecholmes: This still lets you pass a different digest implementation. Instead: `fun mergeInto(other: T)`

misk/src/main/kotlin/misk/digester/SlidingWindowDigest.kt
```diff
@@ -0,0 +1,182 @@
...
+  }
+
+  /**
+   * Returns estimated value for a quantile. The returned value
+   * may not include recently observed values due to how sliding windows are approximated.
+   * If no data has been observed then NaN is returned.
+   */
+  @Synchronized fun quantile(quantile: Double): Double {
+    val now = ZonedDateTime.now(utcNowClock)
+    for (window in windows.reversed()) {
```
alecholmes: Does this copy the list? If so, better to iterate in reverse.

misk/src/test/kotlin/misk/digester/SlidingWindowDigestTest.kt
```diff
@@ -0,0 +1,232 @@
...
+    )
+  }
+
+  fun expectWindowDigests(actual: List<WindowDigest>, expected: List<WindowDigest>) {
+    Assertions.assertThat(expected.count()).isEqualTo(actual.count())
+    for (i in 0 until actual.count()) {
+      if (i >= expected.count()) {
+        break
+      }
+      //Compare window of each WindowDigest
```
alecholmes: Nit: Comments should have spaces: `// Compare` instead of `//Compare` (or `* xyz` instead of `*xyz`)

misk/src/test/kotlin/misk/digester/SlidingWindowDigestTest.kt
```diff
@@ -0,0 +1,232 @@
...
+    Assertions.assertThat(snapshot.count).isEqualTo(count)
+    assertEqualish(sum, snapshot.sum)
+    quantileVals.keys.forEachIndexed { i, q ->
+      assertEqualish(quantileVals[q], snapshot.quantileVals[i])
+    }
+  }
+
+  fun assertEqualish(a: Double?, b: Double?) {
+    if (a == Double.NaN) {
+      Assertions.assertThat(b).isEqualTo(Double.NaN)
```
alecholmes: Static import so that the full line is `assertThat(b).isEqualTo(Double.NaN)`. Also, `assertThat(b).isNan()` since NaN != NaN.

misk/src/test/kotlin/misk/digester/SlidingWindowDigestTest.kt
```diff
@@ -0,0 +1,232 @@
...
+  ) {
+    val snapshot = digest.snapshot(quantileVals.keys.toList()) //should this be keys or values?
+    Assertions.assertThat(snapshot.count).isEqualTo(count)
+    assertEqualish(sum, snapshot.sum)
+    quantileVals.keys.forEachIndexed { i, q ->
+      assertEqualish(quantileVals[q], snapshot.quantileVals[i])
+    }
+  }
+
+  fun assertEqualish(a: Double?, b: Double?) {
```
alecholmes: Why `Double?` and not `Double`?

misk/src/main/kotlin/misk/digester/SlidingWindowDigest.kt
```diff
@@ -0,0 +1,182 @@
...
+   * as this; if they do not then quantiles reported by this sliding
+   * window digest may be incorrect.
+   */
+  @Synchronized fun mergeIn(windowDigests: List<WindowDigest>) {
+    deleteOlderDigests()
+    for (wd in windowDigests) {
+      var found = false
+      for (existing in windows) {
+        if (existing.window.equals(wd.window)) {
+          wd.Digest.mergeInto(existing.Digest)
```
danw5050: @alecholmes The changes you are proposing would break this call as the Digest here has an abstract type of `TDigest<*>` so the mergeInto does not know which type it can allow to merge. I think we might need the way we abstract in this case?

misk/src/test/kotlin/misk/digester/SlidingWindowDigestTest.kt
danw5050: /**
     * Returns the value corresponding to the given [key], or `null` if such a key is not present in the map.
     */
    public operator fun get(key: K): V?

null safety from a map which is passed into assertEqualish

misk/src/main/kotlin/misk/digester/SlidingWindowDigest.kt
```diff
@@ -0,0 +1,182 @@
...
+
+/** WindowDigest holds a t-digest whose data points are scoped to a specific time window. */
+data class WindowDigest<T : TDigest<T>>(
+  val window: Window,
+  val Digest: T
+)
+
+/** Snapshot is the state of a SlidingWindowDigest at a point in time. */
+data class Snapshot(
+  val quantileVals: DoubleArray,  // Values of specific quantiles.
```
alecholmes: The compiler gives a warning about equality/hashing not working correctly with arrays. This is because in the JVM `arrayX == arrayY` does not compare elements.

`List<Double>` is safe, though.

misk/src/test/kotlin/misk/digester/SlidingWindowDigestTest.kt
```diff
@@ -0,0 +1,234 @@
...
+  ) {
+    val snapshot = digest.snapshot(quantileVals.keys.toList()) //should this be keys or values?
+    assertThat(snapshot.count).isEqualTo(count)
+    assertEqualish(sum, snapshot.sum)
+    quantileVals.keys.forEachIndexed { i, q ->
+      assertEqualish(quantileVals[q], snapshot.quantileVals[i])
+    }
+  }
+
+  fun assertEqualish(a: Double?, b: Double) {
```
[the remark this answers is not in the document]
alecholmes: Bump

misk/src/test/kotlin/misk/digester/SlidingWindowDigestTest.kt
```diff
@@ -0,0 +1,234 @@
...
+  fun expectQuantiles(
+    digest: SlidingWindowDigest<FakeDigest>,
+    count: Long,
+    sum: Double,
+    quantileVals: Map<Double, Double>
+  ) {
+    val snapshot = digest.snapshot(quantileVals.keys.toList()) //should this be keys or values?
+    assertThat(snapshot.count).isEqualTo(count)
+    assertEqualish(sum, snapshot.sum)
+    quantileVals.keys.forEachIndexed { i, q ->
```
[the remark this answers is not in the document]
alecholmes: Bump

misk/src/main/kotlin/misk/digester/SlidingWindowDigest.kt
```diff
@@ -0,0 +1,155 @@
...
+        val quantileVals = DoubleArray(quantiles.count()) { Double.NaN }
+        val digest = windows[i].digest
+        quantiles.forEachIndexed { ii, quantile ->
+          quantileVals[ii] = digest.quantile(quantile)
+        }
+        return Snapshot(quantileVals.toList(), digest.count(), digest.sum())
+      }
+    }
+    return Snapshot(
+        DoubleArray(quantiles.count()) { Double.NaN }.toList(),
```
alecholmes: `List(quantiles.count()) { Double.NaN }`

misk/src/test/kotlin/misk/digester/SlidingWindowDigestTest.kt
```diff
@@ -0,0 +1,232 @@
...
+  fun expectQuantiles(
+    count: Long,
+    sum: Double,
+    quantileVals: SortedMap<Double, Double>
+  ) {
+    val snapshot =
+        slidingWindowDigest.snapshot(quantileVals.keys.toList())
+    assertThat(snapshot.count).isEqualTo(count)
+    assertThat(snapshot.sum).isEqualTo(sum)
+    assertThat(snapshot.quantileVals.toDoubleArray()).isEqualTo(quantileVals.values.toDoubleArray())
```
alecholmes: Why convert both to arrays? You can compare two lists.

misk/src/test/kotlin/misk/digester/SlidingWindowDigestTest.kt
```diff
@@ -0,0 +1,232 @@
...
+  }
+
+  fun advanceWindows(n: Int): List<Window> {
+    repeat(n) {
+      setClock(windows()[0].end)
+    }
+    return windows()
+  }
+
+  fun expectQuantiles(
```
alecholmes: Since line is fine unless it's not readable.

```
fun expectQuantiles(count: Long, sum: Double, quantileVals: SortedMap<Double, Double>) {
```

PR #573, opened 2018-10-23
Dweisberg.update histogram to summary

danw5050:
This moves the usage of the histograms to summaries within the misk app.

### DIFF
--- a/misk/src/main/kotlin/misk/metrics/Metrics.kt
+++ b/misk/src/main/kotlin/misk/metrics/Metrics.kt
@@ -16,6 +16,8 @@ import javax.inject.Inject
 class Metrics @Inject internal constructor(val registry: CollectorRegistry) {
   @Inject private lateinit var histogramRegistry: HistogramRegistry
 
+  private val defaultQuantiles = mapOf(0.5 to 0.05, 0.75 to 0.02, 0.95 to 0.01, 0.99 to 0.001, 0.999 to 0.0001)
+
   init {
     registry.register(StandardExports())
     registry.register(MemoryPoolsExports())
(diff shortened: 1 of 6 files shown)


misk/src/main/kotlin/misk/metrics/Metrics.kt
```diff
@@ -34,17 +34,12 @@ class Metrics @Inject internal constructor(val registry: CollectorRegistry) {
...
-  fun summary(name: String, help: String = "", labelNames: List<String> = listOf()): Summary {
-    return Summary.build(name, help).labelNames(*labelNames.toTypedArray()).register(registry)
-  }
-
   fun histogram(
     name: String,
     help: String = "",
-    labelNames: List<String> = listOf(),
-    buckets: DoubleArray? = null
+    labelNames: List<String> = listOf()
```
alecholmes: The caller will also need to decide which quantiles will be ultimately emitted as metric, so also `quantiles: DoubleList = DefaultQuantiles`

misk/src/main/kotlin/misk/prometheus/PrometheusHistogramRegistry.kt
```diff
@@ -10,12 +10,10 @@ class PrometheusHistogramRegistry : HistogramRegistry {
...
     name: String,
     help: String,
-    labelNames: List<String>,
-    buckets: DoubleArray?
+    labelNames: List<String>
   ): PrometheusHistogram {
     val builder =
-        io.prometheus.client.Histogram.build(name, help).labelNames(*labelNames.toTypedArray())
-    if (buckets != null) builder.buckets(*buckets)
+        io.prometheus.client.Summary.build(name, help).labelNames(*labelNames.toTypedArray())
```
alecholmes: The user-defined quantiles need to be added here. The Prometheus client library does not have default quantiles, so without defining any you only get sum and count.

misk/src/main/kotlin/misk/prometheus/PrometheusHistogramRegistry.kt
```diff
@@ -7,15 +7,18 @@ import javax.inject.Inject
...
     name: String,
     help: String,
     labelNames: List<String>,
-    buckets: DoubleArray?
+    quantiles: Map<Double, Double>
   ): PrometheusHistogram {
     val builder =
-        io.prometheus.client.Histogram.build(name, help).labelNames(*labelNames.toTypedArray())
-    if (buckets != null) builder.buckets(*buckets)
+        io.prometheus.client.Summary.build(name, help).labelNames(*labelNames.toTypedArray())
```
alecholmes: import io.prometheus.client.Summary instead of using fully-qualified name.

PR #575, opened 2018-10-23
Web Documentation

adrw:


misk/build.gradle
```diff
@@ -111,7 +111,7 @@ task web(type: Exec) {
   def generateDockerRunCommand = { service, project, projectPath, path, cmd, daemon="-d", image = "squareup/misk-web:0.0.1" ->
     def containerName = "${generateDockerContainerName(service, project, path)}"
     def port = ""
-    if (path.startsWith("tabs")) {
+    if (path.startsWith("tabs") || path.startsWith("/tabs")) {
```
jjestrel: Shouldn't we just need one of these?

misk/web/tabs/example/src/ducks/example.ts
```diff
@@ -220,5 +220,5 @@ export default function ExampleReducer (state = initialState, action: IAction<st
  */
 export interface IExampleState {
   [key: string]: any
-  toJS: () => any
+  toJS?: () => any
```
jjestrel: More for my own curiosity but when would state have an optional `toJS`?

adrw: The ducks module uses ImmutableJS to manage merging state objects which means that to access the data from the state, a `toJS()` function is called on the ImmutableJS object in the state to access the data as a regular Javascript object. Since this interface is used in identifying props passed in to Containers and Components, the `toJS()` is only called in Containers but by making it optional the same interface can be used for the Component.

PR #578, opened 2018-10-25
Add read only transactions

jjestrel:
https://github.com/square/misk/issues/223

### DIFF
--- a/misk-hibernate/src/main/kotlin/misk/hibernate/RealTransacter.kt
+++ b/misk-hibernate/src/main/kotlin/misk/hibernate/RealTransacter.kt
@@ -10,6 +10,7 @@ import misk.jdbc.map
 import misk.jdbc.uniqueResult
 import misk.logging.getLogger
 import misk.tracing.traceWithSpan
+import org.hibernate.FlushMode
 import org.hibernate.SessionFactory
 import org.hibernate.StaleObjectStateException
 import org.hibernate.exception.LockAcquisitionException
@@ -114,13 +115,15 @@ internal class RealTransacter private constructor(
 
   override fun noRetries(): Transacter = withOptions(options.copy(maxAttempts = 1))
 
+  override fun readOnly(): Transacter = withOptions(options.copy(readOnly = true))
+
   private fun withOptions(options: TransacterOptions): Transacter =
     RealTransacter(qualifier, sessionFactory, config, threadLocalSession, options, tracer)
 
   private fun <T> withSession(lambda: (session: Session) -> T): T {
     check(threadLocalSession.get() == null) { "Attempted to start a nested session" }
 
-    val realSession = RealSession(sessionFactory.openSession(), config)
+    val realSession = RealSession(sessionFactory.openSession(), config, options.readOnly)
     threadLocalSession.set(realSession)
 
     try {
@@ -153,7 +156,8 @@ internal class RealTransacter private constructor(
     val maxAttempts: Int = 2,
     val minRetryDelayMillis: Long = 100,
     val maxRetryDelayMillis: Long = 100,
-    val retryJitterMillis: Long = 400
+    val retryJitterMillis: Long = 400,
+    val readOnly: Boolean = false
   )
 
   companion object {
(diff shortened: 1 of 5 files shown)


misk-hibernate/src/main/kotlin/misk/hibernate/RealTransacter.kt
```diff
@@ -162,12 +170,24 @@ internal class RealTransacter private constructor(
...
-    val config: DataSourceConfig
-  ) : Session {
+  internal class ReadOnlySession(
+    session: org.hibernate.Session,
+    config: DataSourceConfig
+  ) : BaseSession(session, config) {
     override val hibernateSession = session
+    init {
+      hibernateSession.isDefaultReadOnly = true
+      hibernateSession.hibernateFlushMode = FlushMode.MANUAL
```
jjestrel: Hibernate says to do this for performance benefit when being read only

misk-hibernate/src/main/kotlin/misk/hibernate/RealTransacter.kt
```diff
@@ -162,12 +170,24 @@ internal class RealTransacter private constructor(
...
+    }
+    override fun <T : DbEntity<T>> save(entity: T): Id<T> {
+      throw IllegalStateException("Saving isn't permitted in a read only session.")
+    }
+  }
 
+  internal class WritableSession(
```
swankjesse: lets chat about this . . .  I think readonly/readwrite is a great idea, but not a good fit for our type hierarchy

misk-hibernate/src/test/kotlin/misk/hibernate/TransacterTest.kt
```diff
@@ -186,6 +186,35 @@ class TransacterTest {
     assertThat(callCount.get()).isEqualTo(1)
   }
 
+  @Test
+  fun readOnlyWontSave() {
```
swankjesse: nice

misk-hibernate/src/test/kotlin/misk/hibernate/TransacterTest.kt
```diff
@@ -186,6 +186,35 @@ class TransacterTest {
...
+        session.save(DbMovie("Star Wars", LocalDate.of(1977, 5, 25)))
+      }
+    }
+    transacter.transaction {session ->
+      assertThat(queryFactory.newQuery<MovieQuery>().list(session)).isEmpty()
+    }
+  }
+
+  @Test
+  fun readOnlyWontModify() {
```
swankjesse: nice

ryanhall07:
we should include documentation that this is different from using a datasource that is in readOnly mode. that seems to always cause confusion with developers.

misk-hibernate/src/main/kotlin/misk/hibernate/RealTransacter.kt
jjestrel: Switched. We talked about layers of code that need to be modified if something is readonly/readwite, and List.remove.

jjestrel:
Added doc for `readOnly()` with `@throws` inside too

misk-hibernate/src/main/kotlin/misk/hibernate/Transacter.kt
```diff
@@ -23,6 +23,14 @@ interface Transacter {
   fun retries(): Transacter
 
   fun noRetries(): Transacter
+
+  /**
+   * Marks the sessions produced by this transacter as read only. This does not mean the underlying
+   * datasource is read only. Only that the session produced won't modify the database.
```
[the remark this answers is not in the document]
jjestrel: We create a new Transacter like with the other options, will update

misk-hibernate/src/main/kotlin/misk/hibernate/Transacter.kt
```diff
@@ -23,6 +23,14 @@ interface Transacter {
...
+   *
+   * @throws IllegalStateException when save is called on sessions produced by this Transacter
```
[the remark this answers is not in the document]
jjestrel: Moved to Session.kt's `save`

misk-hibernate/src/main/kotlin/misk/hibernate/RealTransacter.kt
```diff
@@ -164,12 +168,22 @@ internal class RealTransacter private constructor(
...
+    init {
+      if (readOnly) {
+        hibernateSession.isDefaultReadOnly = true
+        hibernateSession.hibernateFlushMode = FlushMode.MANUAL
+      }
+    }
 
     @Suppress("UNCHECKED_CAST")
     override fun <T : DbEntity<T>> save(entity: T): Id<T> {
+      if (readOnly) {
```
ryanhall07: just wanted to double check that this check is only for redundancy and providing a nice error message. I'm hoping the above hibernate settings are properly flipping the readOnly bit on the jdbc connection. Is that true?

PR #581, opened 2018-10-25
createIndex and createApp in @misk/components

### LINKED ISSUE #563: Move more of src/index.tsx to @misk/common or @misk/dev
Most of it is boilerplate that only depends on changing the tab slug, gives more flexibility in the future for changing functionality across all tabs.

adrw:
* Uniform builder functions for generating the `<App>` and bootstrapping objects usually found in `index.tsx` for all Misk Web Tabs
* Eliminates ~100 lines of boilerplate from every Misk Web Tab
* `@misk/components 0.0.68`
* [CLOSES #563]


misk/web/@misk/components/src/builders/indexBuilder.tsx
```diff
@@ -0,0 +1,68 @@
...
+import { createBrowserHistory, History } from "history"
+import * as React from "react"
+import * as ReactDOM from "react-dom"
+import { AppContainer } from "react-hot-loader"
+import { Provider } from "react-redux"
+import { AnyAction, applyMiddleware, compose, createStore, Reducer } from "redux"
+import createSagaMiddleware from "redux-saga"
+import { AllEffect } from "redux-saga/effects"
+
+export const indexBuilder = (
```
jjestrel: Is it typical to name something like this a builder?

adrw: I've seen some functions like `createSagaMiddleware` that use `create` as prefix so could move it to that.

PR #590, opened 2018-10-29
Remove jar/shadowJar dependsOn references

adrw:
These should be explicitly denoted in the including file in case jar or shadowJar doesn't exist in the service.

### DIFF
--- a/misk/build.gradle
+++ b/misk/build.gradle
@@ -9,6 +9,7 @@ kotlin {
 }
 apply plugin: 'org.junit.platform.gradle.plugin'
 apply plugin: 'com.vanniktech.maven.publish'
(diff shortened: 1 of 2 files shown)


misk/build.gradle
```diff
@@ -9,6 +9,7 @@ kotlin {
 }
 apply plugin: 'org.junit.platform.gradle.plugin'
 apply plugin: 'com.vanniktech.maven.publish'
+apply from: "https://raw.githubusercontent.com/adrw/misk/4e5ce9df6d3057efcee9dcd185623c69bfe39599/gradle/web.gradle"
```
jjestrel: Can we reference the master branch of the `square` user instead?

adrw: Yeah I was going to put that in a follow up PR but I don't have a good hash in Misk yet to reference.

jjestrel: I would lean to doing two prs for this then

PR #593, opened 2018-10-29
Fix java.lang.IllegalArgumentException in ConnectionListener

jjestrel:
Add check for messagesReceivedDiff instead of messagesReceivedSnapshot for jetty metrics collection

### DIFF
--- a/misk/src/main/kotlin/misk/web/jetty/ConnectionListener.kt
+++ b/misk/src/main/kotlin/misk/web/jetty/ConnectionListener.kt
@@ -90,7 +90,7 @@ internal class ConnectionListener(
       metrics.messagesSent.labels(*labels).inc(messagesSentDiff.toDouble())
     }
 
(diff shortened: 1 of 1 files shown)


ryanhall07:
shouldn't there also be a change to gracefully handle exceptions in a finally block and close the connection?

mmihic:
This code doesn't control the lifecycle of the connection; it just monitors metrics

jjestrel:
Are you saying that you want to wrap all of the ConnectionListener code in a try catch?

ryanhall07:
well if Jetty is not going to gracefully handle a listener throwing an exception, then yes. Most likely we could create our own composite listener that does the exception handling for all our custom listeners.

mmihic:
oh woah ok I didn't realize this was the source of the connection leaks. Then definitely yes we should have a wrapper connection listener that handles exceptions.

PR #609, opened 2018-11-02
[CLOSES #559] LocalDB Gradle Task

### LINKED ISSUE #559: Productionize Gradle task for docker for local MySql
```
docker run --name cash-mysql-56 -d -p 3306:3306 -e MYSQL_ALLOW_EMPTY_PASSWORD=true mysql:5.6
```

```
mysql --host=0.0.0.0 -P 3306 -u root
```

adrw:
* Common task for spinning up a local MySql:5.7 image for DB dependent services
* Default port of 3306 with optional override through `./gradlew localdb -Pdbport=####`
* [CLOSES #559] 
* Long term solution: #610

### DIFF
--- a/gradle/localdb.gradle
+++ b/gradle/localdb.gradle
@@ -1,0 +1,10 @@
+task localdb(type: Exec) {
+  // -Pdbport='3306': pass in a db port, it is default 3306
+  
+  def dbport = "3306"
+  if (project.hasProperty("dbport")) {
+    dbport = project.dbport
+  }
+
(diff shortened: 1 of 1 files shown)


gradle/localdb.gradle
```diff
@@ -0,0 +1,3 @@
+task localdb(type: Exec) {
+  commandLine "sh", "-c", "docker run -d --rm --name cash-mysql-57 -p 3306:3306 -e MYSQL_ALLOW_EMPTY_PASSWORD=true mysql:5.7"
```
jjestrel: If you use 3306 this will fail for people running mysql locally already, I would pick a new port if you're going to do this

jjestrel: Also I don't think this will be typical development flow for most people, I think most will have their own version running already. I don't think this needs to be packed as something into misk

adrw: Good point, I'll update it so a port can be passed in.

PR #612, opened 2018-11-02
Remove Bundle Analyzer Running on Build

adrw:
* Remove Bundle Analyzer running by default but leave it as an option for debugging large compiled code bundle sizes
* It was taking too much time for builds inside Docker

### DIFF
--- a/misk/web/@misk/common/webpack.config.js
+++ b/misk/web/@misk/common/webpack.config.js
@@ -1,6 +1,8 @@
 const path = require("path")
 const { BundleAnalyzerPlugin } = require("webpack-bundle-analyzer")
 
+const bundleAnalyzer = false
+
 module.exports = {
   name: "library",
   mode: "production",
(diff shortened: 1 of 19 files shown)


misk/web/@misk/common/webpack.config.js
```diff
@@ -1,6 +1,8 @@
 const path = require("path")
 const { BundleAnalyzerPlugin } = require("webpack-bundle-analyzer")
 
+const bundleAnalyzer = false
```
jjestrel: Do we want these to be shared from one spot?

adrw: I don't think so at this point since you'd only need to use this to debug the compiled code size per package. So, if you wanted to debug `@misk/components` being > 5mb, you'd only want to turn it on for `@misk/components` not for all `@misk/` packages since their sizes don't contribute to the problem.

PR #613, opened 2018-11-02
Add envoy client support via unix sockets

wesleyk:
r @nbroyles @swankjesse @mmihic @ryanhall07

### DIFF
--- a/misk/src/main/java/misk/client/README.md
+++ b/misk/src/main/java/misk/client/README.md
@@ -1,0 +1,2 @@
+We include two java classes here as they're direct copies from okhttp's
+samples.
--- a/misk/src/main/kotlin/misk/client/HttpClientConfigUrlProvider.kt
+++ b/misk/src/main/kotlin/misk/client/HttpClientConfigUrlProvider.kt
@@ -1,0 +1,21 @@
+package misk.client
+
+import com.google.inject.Inject
+
+/**
+ * Calculates the url for an http client config,
+ * which can differ depending on if the client
+ * is envoy-based or connects directly.
+ */
+class HttpClientConfigUrlProvider {
+  @Inject(optional = true) lateinit var envoyClientEndpointProvider: EnvoyClientEndpointProvider
+
+  fun getUrl(endpointConfig: HttpClientEndpointConfig): String {
+    when {
+      endpointConfig.url != null -> return endpointConfig.url
+      endpointConfig.envoy != null -> return envoyClientEndpointProvider.url(endpointConfig.envoy)
+      else -> throw IllegalArgumentException(
+          "One of url or envoy configuration must be set for clients")
+    }
+  }
+}
--- a/misk/src/main/kotlin/misk/client/HttpClientFactory.kt
+++ b/misk/src/main/kotlin/misk/client/HttpClientFactory.kt
@@ -1,7 +1,7 @@
 package misk.client
 
-import misk.security.ssl.SslLoader
 import misk.security.ssl.SslContextFactory
+import misk.security.ssl.SslLoader
 import okhttp3.OkHttpClient
 import java.net.Proxy
 import java.util.concurrent.TimeUnit
@@ -13,6 +13,7 @@ import javax.net.ssl.X509TrustManager
 class HttpClientFactory {
   @Inject lateinit var sslLoader: SslLoader
   @Inject lateinit var sslContextFactory: SslContextFactory
+  @com.google.inject.Inject(optional = true) lateinit var envoyClientEndpointProvider: EnvoyClientEndpointProvider
 
   /** Returns a client initialized based on `config`. */
   fun create(config: HttpClientEndpointConfig): OkHttpClient {
@@ -32,7 +33,12 @@ class HttpClientFactory {
       val sslContext = sslContextFactory.create(it.cert_store, it.trust_store)
       builder.sslSocketFactory(sslContext.socketFactory, x509TrustManager)
     }
+    config.envoy?.let {
+      builder.socketFactory(
+          UnixDomainSocketFactory(envoyClientEndpointProvider.unixSocket(config.envoy)))
+    }
     builder.proxy(Proxy.NO_PROXY)
+
     return builder.build()
   }
 
--- a/misk/src/test/kotlin/misk/client/HttpClientEnvoyTest.kt
+++ b/misk/src/test/kotlin/misk/client/HttpClientEnvoyTest.kt
@@ -1,0 +1,81 @@
+package misk.client
+
+import com.google.common.util.concurrent.Service
+import com.google.inject.Provides
+import com.google.inject.name.Named
+import com.google.inject.name.Names
+import helpers.protos.Dinosaur
+import misk.MiskServiceModule
+import misk.inject.KAbstractModule
+import misk.testing.MiskTest
+import misk.testing.MiskTestModule
+import okhttp3.mockwebserver.MockResponse
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.jupiter.api.Test
+import retrofit2.Call
+import retrofit2.http.Body
+import retrofit2.http.POST
+import java.io.File
+import javax.inject.Inject
+import javax.inject.Singleton
+
+/** Demonstrates ability to bind up an envoy-based http client and connect via unix sockets. */
+@MiskTest(startService = true)
+internal class HttpClientEnvoyTest {
+  @MiskTestModule val module = TestModule()
+
+  @Inject private lateinit var webServerService: MockWebServerService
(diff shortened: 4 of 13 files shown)


misk/src/main/java/misk/client/README.md
```diff
@@ -0,0 +1,2 @@
+We include two java classes here as they're direct copies from okhttp's
```
wesleyk: I had issues porting these over as kotlin, I think primarily due to an issue with the jnr-unixsocket library. This change I think should help, but in the meantime I think these direct copies are fine (https://github.com/jnr/jnr-unixsocket/pull/62)

misk/src/test/kotlin/misk/client/HttpClientEnvoyTest.kt
```diff
@@ -0,0 +1,93 @@
...
+    assertThat(recordedRequest.getHeader("Host"))
+        .isEqualTo(String.format("%s:%s", server.hostName, server.port))
+  }
+
+  interface ReturnADinosaur {
+    @POST("/cooldinos")
+    fun getDinosaur(@Body request: Dinosaur): Call<Void>
+  }
+
+  class TestEnvoyClientConfiguration(
```
wesleyk: an example envoy config implementation. In our actual use case in skim, we'll provide the unix socket via an ENV variable, and we'll calculate based on a square-specific formula + the app name.

misk/src/test/kotlin/misk/client/HttpClientEnvoyTest.kt
```diff
@@ -0,0 +1,93 @@
...
+
+/** Demonstrates ability to bind up an envoy-based http client and connect via unix sockets. */
+internal class HttpClientEnvoyTest {
+  private lateinit var socketFile: File
+  private lateinit var server: MockWebServer
+  private lateinit var clientInjector: Injector
+
+  @BeforeEach
+  fun createClientAndServer() {
+    socketFile = File("/tmp/ClientAndServer.sock")
```
nbroyles: Not sure if it matters, but worth using an abstract socket file? (e.g. `@whatevers.sock`)

misk/src/main/kotlin/misk/client/EnvoyClientConfiguration.kt
```diff
@@ -0,0 +1,13 @@
+package misk.client
+
+import java.io.File
+
+/**
+ * Envoy configuration provider that misk clients can customize to their needs.
+ */
+interface EnvoyClientConfiguration {
```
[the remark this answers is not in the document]
ryanhall07: I feel like `Configuration` is misleading, because it's not a misk `Config`. why not `EnvoyClientEndpointProvider`?

wesleyk:
Hmm, tests in travis-ci don't seem able to shutdown the mock web server. Locally this works fine. Looking into it.

misk/src/main/kotlin/misk/client/HttpClientConfigUrlProvider.kt
```diff
@@ -0,0 +1,24 @@
...
+/**
+ * Calculates the url for an http client config,
+ * which can differ depending on if the client
+ * is envoy-based or connects directly.
+ */
+class HttpClientConfigUrlProvider {
+  @Inject(optional = true) lateinit var envoyClientConfiguration: EnvoyClientConfiguration
+
+  fun getUrl(endpointConfig: HttpClientEndpointConfig): String {
+    endpointConfig.url?.let {
```
ryanhall07: nit: might want to use a `when`

misk/src/test/kotlin/misk/client/HttpClientEnvoyTest.kt
```diff
@@ -0,0 +1,93 @@
...
+    server.enqueue(MockResponse())
+
+    val client: ReturnADinosaur = clientInjector.getInstance(Names.named("dinosaur"))
+    val response = client.getDinosaur(dinoRequest).execute()
+    assertThat(response.code()).isEqualTo(200)
+
+    val recordedRequest = server.takeRequest()
+    assertThat(recordedRequest.path).isEqualTo("/cooldinos")
+    assertThat(recordedRequest.getHeader("Host"))
+        .isEqualTo(String.format("%s:%s", server.hostName, server.port))
```
nbroyles: I'm curious, what do these get set to? Unix sockets don't have a port?

misk/src/test/kotlin/misk/client/HttpClientEnvoyTest.kt
```diff
@@ -0,0 +1,93 @@
...
+  private lateinit var server: MockWebServer
+  private lateinit var clientInjector: Injector
+
+  @BeforeEach
+  fun createClientAndServer() {
+    socketFile = File("/tmp/ClientAndServer.sock")
+    server = MockWebServer()
+    server.setServerSocketFactory(UnixDomainServerSocketFactory(socketFile))
+
+    clientInjector = Guice.createInjector(
```
ryanhall07: why not just use a `MiskTest` and let it do the injection for you?

ryanhall07: also you could mode the `before` and `after` as a `Service` which is the misky way

misk/src/main/kotlin/misk/client/EnvoyClientConfiguration.kt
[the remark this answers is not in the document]
nbroyles: Yeah, I like that or snowp's suggestion of `EnvoyEndpointConfiguration`

misk/src/main/kotlin/misk/client/HttpClientConfigUrlProvider.kt
mmihic: or elvis

```kotlin 
return endpointConfig.url ?: 
   endpoingConfig.envoy?.let { endpointProvider.url(it) } ?: 
   throw IllegalArgumentException("blah")
```

misk/src/main/kotlin/misk/client/HttpClientFactory.kt
```diff
@@ -13,6 +13,7 @@ import javax.net.ssl.X509TrustManager
 class HttpClientFactory {
   @Inject lateinit var sslLoader: SslLoader
   @Inject lateinit var sslContextFactory: SslContextFactory
+  @com.google.inject.Inject(optional = true) lateinit var envoyEndpointConfiguration: EnvoyEndpointConfiguration
```
nbroyles: Is there any weirdness that can occur from us using two different Inject annotations?

wesleyk: mixing doesn't cause any issues AFAIK

misk/src/test/kotlin/misk/client/HttpClientEnvoyTest.kt
wesleyk: yeah, it's not meaningful, but a value is still required to be passed in from okhttp. Checkout https://github.com/square/misk/pull/613/files#diff-92980dcb978415a24adb016b23166945R67

misk/src/test/kotlin/misk/client/HttpClientEnvoyTest.kt
wesleyk: done

misk/src/main/kotlin/misk/client/HttpClientConfigUrlProvider.kt
wesleyk: went with the `when`

misk/src/test/kotlin/misk/client/MockWebServerService.kt
```diff
@@ -0,0 +1,27 @@
...
+    file.delete()
+
+    server = MockWebServer()
+    socketFile?.let {
+      server!!.setServerSocketFactory(UnixDomainServerSocketFactory(file))
+    }
+  }
+
+  override fun shutDown() {
+    server!!.shutdown()
```
ryanhall07: should probably gracefully handle this being null, in the case startUp wasn't called for whatever reason

misk/src/test/kotlin/misk/client/MockWebServerService.kt
```diff
@@ -0,0 +1,27 @@
+package misk.client
+
+import com.google.common.util.concurrent.AbstractIdleService
+import okhttp3.mockwebserver.MockWebServer
+import java.io.File
+import javax.inject.Singleton
+
+@Singleton
+class MockWebServerService(val socketFile: String?) : AbstractIdleService() {
```
ryanhall07: should this be named to make it clear it uses a unix file socket? that would not be obvious to me

PR #616, opened 2018-11-02
Implemented registry for digester

danw5050:
This will be used to handle custom digester histograms

### DIFF
--- a/misk-metrics-digester/src/main/kotlin/misk/metrics/digester/Registry.kt
+++ b/misk-metrics-digester/src/main/kotlin/misk/metrics/digester/Registry.kt
@@ -34,16 +34,16 @@ class TDigestHistogram<T : TDigest<T>> constructor(
 ) : Histogram {
 
   private val metrics = ConcurrentHashMap<Int, DigestMetric>()
-
-  /** Add all labels passed to the histogram */
-  init {
-    labelNames.forEach { label ->
-      labels(label)
-    }
-  }
+  private val labelRegex = "[a-zA-Z0-9_]*".toRegex()
 
   /** Records a new metric within the histogram */
   override fun labels(vararg labelValues: String): TDigestHistogramRecordMetric {
+    labelValues.forEach { label ->
+      require (labelRegex.matches(label)) {
+        "label name is invalid"
+      }
+    }
+
     val metric = metrics.getOrPut(key(labelValues)) {
       DigestMetric(tDigest(), labelValues.asList())
     }
@@ -51,14 +51,10 @@ class TDigestHistogram<T : TDigest<T>> constructor(
   }
 
   /** Returns the number of windows within the histogram */
-  override fun count(vararg labelValues: String): Int {
-    return metrics.get(key(labelValues))?.digest?.windows?.count() ?: 0
-  }
+  override fun count(vararg labelValues: String) = metrics.get(key(labelValues))?.digest?.windows?.count() ?: 0
 
   /** Returns a metric of the histogram. Order of labels matters */
-  fun getMetric(vararg metric: String): DigestMetric {
-    return metrics.getValue(key(metric))
-  }
+  internal fun metric(vararg metric: String) = metrics[key(metric)]
 
   /** Creates a hash for a list of labels */
   private fun key(labels: Array<out String>) = Objects.hash(labels.sorted())
@@ -84,26 +80,24 @@ class TDigestHistogramRegistry<T : TDigest<T>> constructor(
   }
 
   private val histograms = ConcurrentHashMap<String, TDigestHistogram<T>>()
+  private val metricRegex =  "[a-zA-Z_:][a-zA-Z0-9_:]*".toRegex()
 
   /** Creates and returns a new histogram. Histogram is registered within the Registry */
-  override fun newHistogram(
+  @Synchronized override fun newHistogram(
     name: String,
     help: String,
     labelNames: List<String>,
     quantiles: Map<Double, Double>
   ): TDigestHistogram<T> {
-    val histogram = TDigestHistogram(name, help, quantiles.keys.toList(), labelNames, fun() = newDigestFn())
-    return register(histogram)
-  }
 
-  /** Registers a histogram within the Registry */
-  @Synchronized private fun register(histogram: TDigestHistogram<T>): TDigestHistogram<T> {
+    require(metricRegex.matches(name)) {
+      "metrics name is invalid"
+    }
+
+    val histogram = TDigestHistogram(name, help, quantiles.keys.toList(), labelNames, fun() = newDigestFn())
     return histograms.getOrPut(histogram.name) { histogram }
   }
 
   /** Returns a list of all Histograms registered by the Registry */
   fun allHistograms() = histograms.values
-
-  /** Returns a histogram set under the given labels */
-  fun getHistogram(labels: String) = histograms[labels]
 }
--- a/misk-metrics-digester/src/main/kotlin/misk/metrics/digester/Registry.kt
+++ b/misk-metrics-digester/src/main/kotlin/misk/metrics/digester/Registry.kt
@@ -2,12 +2,11 @@ package misk.metrics.digester
 
 import misk.metrics.Histogram
 import misk.metrics.HistogramRegistry
-import java.lang.StringBuilder
 
 /** DigestMetric contains the contents of an individual metrics within a histogram */
 class DigestMetric(
-  val digest: SlidingWindowDigest<*>,
-  val labels: List<String>
+  internal val digest: SlidingWindowDigest<*>,
+  private val labels: List<String>
 ) {
   /** Adds an observation to the metric */
   fun observe(value: Double) {
@@ -93,7 +92,7 @@ class TDigestHistogramRegistry <T : TDigest<T>> constructor(
             fun() = VeneurDigest())
   ): HistogramRegistry {
 
-  val histograms: MutableMap<String, TDigestHistogram<T>> = mutableMapOf()
+  private val histograms: MutableMap<String, TDigestHistogram<T>> = mutableMapOf()
 
   /** Creates and returns a new histogram. Histogram is registered within the Registry */
   override fun newHistogram(
--- a/misk-metrics-digester/src/main/kotlin/misk/metrics/digester/Registry.kt
+++ b/misk-metrics-digester/src/main/kotlin/misk/metrics/digester/Registry.kt
@@ -35,7 +35,7 @@ class TDigestHistogram<T : TDigest<T>> constructor(
   private val metrics = ConcurrentHashMap<Int, DigestMetric>()
 
   /** Records a new metric within the histogram */
-  override fun record(vararg labelValues: String): TDigestHistogramRecordMetric {
+  override fun labels(vararg labelValues: String): TDigestHistogramRecordMetric {
     val metric = metrics.getOrPut(key(labelValues)) {
       DigestMetric(tDigest(), labelValues.asList())
     }
@@ -48,8 +48,8 @@ class TDigestHistogram<T : TDigest<T>> constructor(
   }
 
   /** Returns a metric of the histogram. Order of labels matters */
-  fun getMetric(vararg metric: String): DigestMetric? {
-    return metrics[key(metric)]
+  fun getMetric(vararg metric: String): DigestMetric {
+    return metrics.getValue(key(metric))
   }
 
   /** Creates a hash for a list of labels */
--- a/misk-metrics-digester/src/main/kotlin/misk/metrics/digester/Registry.kt
+++ b/misk-metrics-digester/src/main/kotlin/misk/metrics/digester/Registry.kt
@@ -1,98 +1,78 @@
 package misk.metrics.digester
 
+import misk.logging.getLogger
 import misk.metrics.Histogram
 import misk.metrics.HistogramRegistry
+import java.util.Objects
+import java.util.concurrent.ConcurrentHashMap
+import java.util.concurrent.ConcurrentMap
+import kotlin.reflect.KClass
+import java.lang.reflect.InvocationTargetException
+import java.lang.reflect.Method
 
 /** DigestMetric contains the contents of an individual metrics within a histogram */
 class DigestMetric(
-  internal val digest: SlidingWindowDigest<*>,
-  private val labels: List<String>
+  val digest: SlidingWindowDigest<*>,
+  val labels: List<String>
 ) {
   /** Adds an observation to the metric */
   fun observe(value: Double) {
     digest.observe(value)
   }
-
-  /** Returns the labels for the metric */
-  fun labels(): List<String> {
-    return labels
-  }
-
-  /** Returns the SlidingWindowDigest registered for the metric */
-  fun digest(): SlidingWindowDigest<*> {
-    return digest
-  }
 }
 
 /** TDigestHistogram stores histograms which records metrics */
-class TDigestHistogram <T : TDigest<T>> constructor(
-  private val name: String,
-  private val help: String,
-  private val quantiles: List<Double>,
-  private val tDigest: () -> SlidingWindowDigest<*>
+class TDigestHistogram<T : TDigest<T>> constructor(
+  val name: String,
+  val help: String,
+  val quantiles: List<Double>,
+  private val tDigest: () -> SlidingWindowDigest<T>
 ) : Histogram {
 
-  val metrics: MutableMap<Long, DigestMetric> = mutableMapOf()
+  private val metrics = ConcurrentHashMap<Int, DigestMetric>()
 
   /** Records a new metric within the histogram */
   override fun record(duration: Double, vararg labelValues: String) {
-    if (!metrics.containsKey(key(labelValues))) {
-      val metric = DigestMetric(tDigest(), labelValues.asList())
-      metrics.put(key(labelValues), metric)
+    val metric = metrics.getOrPut(key(labelValues)) {
+      DigestMetric(tDigest(), labelValues.asList())
     }
-    metrics.getValue(key(labelValues)).observe(duration)
+    metric.observe(duration)
   }
 
   /** Returns the number of windows within the histogram */
   override fun count(vararg labelValues: String): Int {
-    if (metrics.containsKey(key(labelValues))) {
-      return metrics.getValue(key(labelValues)).digest.windows.count()
-    }
-    return  0
+    return metrics.get(key(labelValues))?.digest?.windows?.count() ?: 0
   }
 
   /** Returns a metric of the histogram. Order of labels matters */
-  fun getMetric(vararg labelValues: String): DigestMetric? {
-    return metrics[key(labelValues)]
+  fun getMetric(vararg metric: String): DigestMetric? {
+    return metrics[key(metric)]
   }
 
-  /** Returns the name of the histogram */
-  fun getName(): String {
-    return name
-  }
-
-  /** Returns the help of the histogram */
-  fun getHelp(): String {
-    return help
-  }
-
-  /** Returns the quantiles of the histogram */
-  fun getQuantiles(): List<Double> {
-    return quantiles
-  }
-
-  /** Generates a unique key to store each metric under */
-  fun key(labels: Array<out String>): Long {
-    var hash: Long = 0
-    labels.forEach { key ->
-      hash += key.hashCode()
-    }
-    return hash
-  }
+  /** Creates a hash for a list of labels */
+  private fun key(labels: Array<out String>) = Objects.hash(labels.sorted())
 }
 
 /** Default quantiles that can be used to register with a new histogram */
-val defaultQuantiles = mapOf(0.5 to 0.05, 0.75 to 0.02, 0.95 to 0.01, 0.99 to 0.001, 0.999 to 0.0001)
+val defaultQuantiles =
+    mapOf(0.5 to 0.05, 0.75 to 0.02, 0.95 to 0.01, 0.99 to 0.001, 0.999 to 0.0001)
 
 /** TDigestHistogramRegistry registers all TDigestHistograms */
-class TDigestHistogramRegistry <T : TDigest<T>> constructor(
-    private val newDigestFn: () -> SlidingWindowDigest<*> =
-        fun() = SlidingWindowDigest(
-            Windower(windowSecs = 30, stagger = 3),
-            fun() = VeneurDigest())
-  ): HistogramRegistry {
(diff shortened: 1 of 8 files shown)


misk-metrics-digester/src/main/kotlin/misk/metrics/digester/Registry.kt
```diff
@@ -0,0 +1,126 @@
...
+  private val histograms: MutableMap<String, TDigestHistogram<T>> = mutableMapOf()
+
+  /** Creates and returns a new histogram. Histogram is registered within the Registry */
+  override fun newHistogram(
+    name: String,
+    help: String,
+    labelNames: List<String>,
+    quantiles: Map<Double, Double>
+  ): TDigestHistogram<T> {
+    val histogram = TDigestHistogram<T>(name, help, quantiles.keys.toList(), fun() = newDigestFn())
```
alecholmes: The `TDigestHistogram` constructor expects quantiles to be ordered but the order of `quantiles.keys` is non-deterministic.

misk-metrics-digester/src/main/kotlin/misk/metrics/digester/Registry.kt
```diff
@@ -0,0 +1,126 @@
...
+    return register(histogram)
+  }
+
+  /** Registers a histogram within the Registry */
+  @Synchronized fun register(histogram: TDigestHistogram<T>): TDigestHistogram<T> {
```
alecholmes: This should be private. It's an implementation detail.

misk-metrics-digester/src/main/kotlin/misk/metrics/digester/Registry.kt
```diff
@@ -0,0 +1,126 @@
...
+    }
+    return hash
+  }
+}
+
+/** Default quantiles that can be used to register with a new histogram */
+val defaultQuantiles = mapOf(0.5 to 0.05, 0.75 to 0.02, 0.95 to 0.01, 0.99 to 0.001, 0.999 to 0.0001)
+
+/** TDigestHistogramRegistry registers all TDigestHistograms */
+class TDigestHistogramRegistry <T : TDigest<T>> constructor(
```
alecholmes: There need to be some interface tweaks to match how this will be used in practice.

Current state:
```
    val registry = TDigestHistogramRegistry<VeneurDigest>()

    val histogram = registry.newHistogram(
      name = "request_latency_ms",
      help = "hello",
      labelNames = listOf("foo", "bar"),
      quantiles = mapOf(0.75 to 999.999) // why is this a map?
    )

    // Why does get metric return a nullable?
    histogram.getMetric("x", "y")?.observe(1.23)
```

Expected usage, I think:
```
    val registry = TDigestHistogramRegistry<VeneurDigest>()

    val histogram = registry.newHistogram(
      name = "request_latency_ms",
      help = "hello",
      labelNames = listOf("foo", "bar"),
      quantiles = defaultQuantiles, // or omitted to get defaults automatically
    )

    histogram.metric("x", "y").observe(1.23)
    // or _additional_ possibility
    histogram.metric(mapOf("foo" to "x", "bar" to "y")).observe(1.23)
```

misk-metrics-digester/src/main/kotlin/misk/metrics/digester/Registry.kt
```diff
@@ -0,0 +1,126 @@
...
+    labelNames: List<String>,
+    quantiles: Map<Double, Double>
+  ): TDigestHistogram<T> {
+    val histogram = TDigestHistogram<T>(name, help, quantiles.keys.toList(), fun() = newDigestFn())
+    return register(histogram)
+  }
+
+  /** Registers a histogram within the Registry */
+  @Synchronized fun register(histogram: TDigestHistogram<T>): TDigestHistogram<T> {
+    if  (!histograms.containsKey(histogram.getName())) {
```
alecholmes:
```
  /** Registers a histogram within the Registry */
  @Synchronized
  private fun register(histogram: TDigestHistogram<T>): TDigestHistogram<T> {
    return histograms.getOrPut(histogram.getName()) { histogram }
  }
```

misk-metrics-digester/src/main/kotlin/misk/metrics/digester/Registry.kt
```diff
@@ -0,0 +1,126 @@
...
+
+/** TDigestHistogramRegistry registers all TDigestHistograms */
+class TDigestHistogramRegistry <T : TDigest<T>> constructor(
+    private val newDigestFn: () -> SlidingWindowDigest<*> =
+        fun() = SlidingWindowDigest(
+            Windower(windowSecs = 30, stagger = 3),
+            fun() = VeneurDigest())
+  ): HistogramRegistry {
+
+  private val histograms: MutableMap<String, TDigestHistogram<T>> = mutableMapOf()
```
alecholmes: Java's concurrent maps are richer than what Go offers. `ConcurrentHashMap` is a good fit here too. https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ConcurrentHashMap.html

`private val histograms = ConcurrentHashMap<String, TDigestHistogram<T>>()`

It also makes the class much simpler:
```
  // ...

  /** Registers a histogram within the Registry */
  private fun register(histogram: TDigestHistogram<T>): TDigestHistogram<T> {
    return histograms.getOrPut(histogram.getName()) { histogram }
  }

  /** Returns a list of all Histograms registered by the Registry */
  fun allHistograms() = histograms.values

  /** Returns a histogram set under the given labels */
  fun getHistogram(labels: String) = histograms[labels]
}
```

misk-metrics-digester/src/test/kotlin/misk/metrics/digester/RegistryTest.kt
```diff
@@ -0,0 +1,32 @@
...
+            Windower(windowSecs = 30, stagger = 3), // this does not matter
+            fun() = FakeDigest()))
+
+    val histogram = registry.newHistogram("name", "help", listOf(), mapOf(0.1 to 0.1, 0.2 to 0.2))
+    histogram.record(1.23, "test")
+
+    val firstWindow = histogram.getMetric("test")!!.digest.windows[0].digest as FakeDigest
+
+    // Check that recorded value gets added to the metric
+    Assertions.assertThat(firstWindow.addedValues).isEqualTo(arrayListOf(1.23))
```
alecholmes: Static import `assertThat` so that you just have `assertThat instead of `Assertions.assertThat`

misk-metrics-digester/src/test/kotlin/misk/metrics/digester/RegistryTest.kt
```diff
@@ -0,0 +1,32 @@
...
+
+    histogram.record(4.56, "test")
+    Assertions.assertThat(firstWindow.addedValues).isEqualTo(arrayListOf(1.23, 4.56))
+
+    // Add a different metric name
+    histogram.record(2.34, "another_test")
+
+    // Check that a new metric gets created for the new key
+    Assertions.assertThat(histogram.metrics.keys.count()).isEqualTo(2)
```
alecholmes: `assertThat(histogram.metrics.keys).hasSize(2)`

misk-metrics-digester/src/test/kotlin/misk/metrics/digester/RegistryTest.kt
```diff
@@ -0,0 +1,32 @@
...
+    val histogram = registry.newHistogram("name", "help", listOf(), mapOf(0.1 to 0.1, 0.2 to 0.2))
+    histogram.record(1.23, "test")
+
+    val firstWindow = histogram.getMetric("test")!!.digest.windows[0].digest as FakeDigest
+
+    // Check that recorded value gets added to the metric
+    Assertions.assertThat(firstWindow.addedValues).isEqualTo(arrayListOf(1.23))
+
+    histogram.record(4.56, "test")
+    Assertions.assertThat(firstWindow.addedValues).isEqualTo(arrayListOf(1.23, 4.56))
```
alecholmes: Generally don't need specific types. `listOf` is sufficient; don't need `arrayListOf.

misk-metrics-digester/src/test/kotlin/misk/metrics/digester/RegistryTest.kt
```diff
@@ -0,0 +1,32 @@
+package misk.metrics.digester
```
alecholmes: Increase the test coverage to at least have parity with the Go implementation.

misk-metrics-digester/src/main/kotlin/misk/metrics/digester/Registry.kt
```diff
@@ -0,0 +1,126 @@
...
+  /** Returns the number of windows within the histogram */
+  override fun count(vararg labelValues: String): Int {
+    if (metrics.containsKey(key(labelValues))) {
+      return metrics.getValue(key(labelValues)).digest.windows.count()
+    }
+    return  0
+  }
+
+  /** Returns a metric of the histogram. Order of labels matters */
+  fun getMetric(vararg labelValues: String): DigestMetric? {
```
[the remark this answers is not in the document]
danw5050: Developers would not be using getMetrics, but how would this make it more straightforward? It makes writing tests easier and it is in the same format as count.

misk-metrics-digester/src/main/kotlin/misk/metrics/digester/Registry.kt
danw5050: Its a map because prometheus summaries require an error of quantiles estimation (https://prometheus.io/docs/practices/histograms/#errors-of-quantile-estimation). It looks like the prometheus go client provides the algorithm for this:

"A summary would have had no problem calculating the correct percentile value in both cases, at least if it uses an appropriate algorithm on the client side (like the one used by the Go client)."

misk-metrics-digester/src/main/kotlin/misk/metrics/digester/Registry.kt
```diff
@@ -0,0 +1,109 @@
...
+    help: String,
+    labelNames: List<String>,
+    quantiles: Map<Double, Double>
+  ): TDigestHistogram<T> {
+    val histogram = TDigestHistogram(name, help, quantiles.keys.toList(), labelNames, fun() = newDigestFn())
+    return register(histogram)
+  }
+
+  /** Registers a histogram within the Registry */
+  @Synchronized private fun register(histogram: TDigestHistogram<T>): TDigestHistogram<T> {
```
alecholmes: Not used or needed.

misk-metrics-digester/src/main/kotlin/misk/metrics/digester/Registry.kt
```diff
@@ -0,0 +1,109 @@
...
+    return histograms.getOrPut(histogram.name) { histogram }
+  }
+
+  /** Returns a list of all Histograms registered by the Registry */
+  fun allHistograms() = histograms.values
+
+  /** Returns a histogram set under the given labels */
+  fun getHistogram(labels: String) = histograms[labels]
```
alecholmes: Not used.

misk-metrics-digester/src/main/kotlin/misk/metrics/digester/Registry.kt
```diff
@@ -0,0 +1,109 @@
...
+  val help: String,
+  val quantiles: List<Double>,
+  private val labelNames: List<String>,
+  private val tDigest: () -> SlidingWindowDigest<T>
+) : Histogram {
+
+  private val metrics = ConcurrentHashMap<Int, DigestMetric>()
+
+  /** Add all labels passed to the histogram */
+  init {
```
alecholmes: Remove this. It is unintentionally creating metrics.

Say label names given are `[a, b]`. This creates two metrics with label values `{a->a}` and `{b->b}`

misk-metrics-digester/src/main/kotlin/misk/metrics/digester/Registry.kt
```diff
@@ -0,0 +1,109 @@
...
+    return TDigestHistogramRecordMetric(metric)
+  }
+
+  /** Returns the number of windows within the histogram */
+  override fun count(vararg labelValues: String): Int {
+    return metrics.get(key(labelValues))?.digest?.windows?.count() ?: 0
+  }
+
+  /** Returns a metric of the histogram. Order of labels matters */
+  fun getMetric(vararg metric: String): DigestMetric {
```
alecholmes: This doesn't need to be public. Also, Kotlin by convention avoids `get`/`set` prefixes.

```
internal fun metric(vararg metric: String) = metrics[key(metric)]
```

This also changes from `getValue` to normal map get. `getValue` never returns null but your unit tests expect nullable values to come back from this method (`assertThat(histogram.getMetric("label1")).isNotNull()`)

misk-metrics-digester/src/main/kotlin/misk/metrics/digester/Registry.kt
```diff
@@ -0,0 +1,109 @@
...
+    return metrics.getValue(key(metric))
+  }
+
+  /** Creates a hash for a list of labels */
+  private fun key(labels: Array<out String>) = Objects.hash(labels.sorted())
```
alecholmes: Don't sort here. This is sorting label _values_, which you definitely don't want to do.

The Go implementation is based on a map, which is why it sorts based on the map keys.

misk-metrics-digester/src/main/kotlin/misk/metrics/digester/Registry.kt
```diff
@@ -0,0 +1,109 @@
...
+  /** Records a new metric within the histogram */
+  override fun labels(vararg labelValues: String): TDigestHistogramRecordMetric {
+    val metric = metrics.getOrPut(key(labelValues)) {
+      DigestMetric(tDigest(), labelValues.asList())
+    }
+    return TDigestHistogramRecordMetric(metric)
+  }
+
+  /** Returns the number of windows within the histogram */
+  override fun count(vararg labelValues: String): Int {
```
alecholmes: Avoid adding public methods just for unit tests. Remove this.

Eventually you'll need something like `fun allMetrics() = metrics.metrics`. Your unit test can use that to assert count.

misk-metrics-digester/src/test/kotlin/misk/metrics/digester/RegistryTest.kt
```diff
@@ -0,0 +1,59 @@
...
+
+  @Test
+  fun registryDigestWithLabels() {
+    val registry = histogramRegistryFactory()
+
+    val histogram = registry.newHistogram("name", "help", listOf("label1", "label2"),
+        mapOf(0.1 to 0.1, 0.2 to 0.2))
+
+    //Check that passed labels exist
+    assertThat(histogram.getMetric("label1")).isNotNull()
```
alecholmes: This should not pass. The code should throw an user-friendly exception, as the user must pass a label value for each label name.

`histogram.getMetric()` and `histogram.getMetric("a","b","c","d")` should also throw.

alecholmes: Code also allows invalid metric names like `foo-bar`.

misk-metrics-digester/src/test/kotlin/misk/metrics/digester/RegistryTest.kt
alecholmes: Ping. Please take a look at the test cases covered in the Go implementation. There are edge cases that are not tested here, and the implementation needs some fixes for them.

misk-metrics-digester/src/test/kotlin/misk/metrics/digester/RegistryTest.kt
```diff
@@ -0,0 +1,59 @@
...
+
+    val firstWindow = histogram.getMetric("test").digest.windows[0].digest as FakeDigest
+
+    // Check that recorded value gets added to the metric
+    assertThat(firstWindow.addedValues).isEqualTo(listOf(1.23))
+    metric.observe(4.56)
+
+    // Check that both values are added to the window
+    assertThat(firstWindow.addedValues).isEqualTo(listOf(1.23, 4.56))
+    assertThat(histogram.count("test")).isEqualTo(3)
```
alecholmes: `assertThat(histogram.allMetrics()).hasSize(3)`

misk-metrics-digester/src/test/kotlin/misk/metrics/digester/RegistryTest.kt
```diff
@@ -0,0 +1,59 @@
...
+
+    // Add a different metric name
+    histogram.labels("another_test", "another_test_2").observe(2.34)
+
+    // Check that a new metric gets created for the new key
```
alecholmes: The asserts below don't check that anything is created, only that values are not null.

misk-metrics-digester/src/main/kotlin/misk/metrics/digester/Registry.kt
danw5050: Do we not want this at all? This is to maintain a similar approach to Prometheus histogram metrics.
```  init {
      labels(*labelNames as Array<out String>)
  } ```

Comment (alecholmes) [reply misk-metrics-digester/src/main/kotlin/misk/metrics/digester/Registry.kt]: No, because we don't want to mix up label names with label values.

```
// This does not create any metrics _yet_.
// However, it means that every metric derived from it will have exactly two labels
// and the _names_ of those labels will always be label1 and label2.
val hist = registry.newHistogram("yourmetric", "help", listOf("label1", "label2"), mapOf(...))

// This means a metric named "yourmetric" with labels {"label1":"value1", "label2":"value2"} is sent to Prometheus
hist.labels("value1", "value2").observe(1.23)

// This means a metric named "yourmetric" with labels {"label1":"x", "label2":"y"} is sent to Prometheus
hist.labels("x", "y").observe(1.23)
```

Your init here is effectively doing:
```
// Invalid because no observed metric has a _value_ of label1 and also because a value for label2 is omitted.
hist.labels("label1")

// Invalid because no observed metric has a _value_ of label2 and also because a value for label1 is omitted.
hist.labels("label2")
```

File: misk-metrics-digester/src/main/kotlin/misk/metrics/digester/Registry.kt (hunk 16 above)
Comment (danw5050) [reply misk-metrics-digester/src/main/kotlin/misk/metrics/digester/Registry.kt]: Not sure if it is possible https://stackoverflow.com/questions/13160672/java-getting-an-error-for-implementing-interface-method-with-weaker-access.

File: misk-metrics-digester/src/main/kotlin/misk/metrics/digester/Registry.kt (hunk 15 above)
Comment (danw5050) [reply misk-metrics-digester/src/main/kotlin/misk/metrics/digester/Registry.kt]: Don't we want labels ["label1",label2"] to have the same key as ["label2",label1"]?

File: misk-metrics-digester/src/main/kotlin/misk/metrics/digester/Registry.kt (hunk 16 above)
Comment (alecholmes) [reply misk-metrics-digester/src/main/kotlin/misk/metrics/digester/Registry.kt]: We can change the interface here  since it's misk.metrics.Histogram.

File: misk-metrics-digester/src/main/kotlin/misk/metrics/digester/Registry.kt (hunk 15 above)
Comment (alecholmes) [reply misk-metrics-digester/src/main/kotlin/misk/metrics/digester/Registry.kt]: No, since the key is based on label _values_ and not label names. So for this you need `Objects.hash(*labels)` (note the important `*` to expand the varargs).

Each `TDigestHistogram` instance represents a metric with a fixed name, a set of labels with fixed names and differing values. They key in the metrics map for this class represents a unique combination of ordered label values.

Let's say there's an instance named `hist` of `TDigestHistogram` with `labelNames == ["a-label", "b-label"]`. Calling
`hist.labels("x", "y")` and `hist.labels("y", "x")` creates two different metric values in Prometheus:
1. `{"a-label":"x", "b-label":"y"}` where the key in your metrics map is `Objects.hash("x", "y")`
2. `{"a-label":"y", "b-label":"x"}` where the key in your metric map here is `Objects.hash("y", "x")`

Even though both examples use `x` and `y` as values for labels, they use them as values for different label names. `Objects.hash("x", "y") != Objects.hash("y, "x")`. If you were to sort label values, they two metric values would end up being incorrectly squished into one metric value.

File: misk-metrics-digester/src/test/kotlin/misk/metrics/digester/RegistryTest.kt (hunk 18 above)
Comment (danw5050) [reply misk-metrics-digester/src/test/kotlin/misk/metrics/digester/RegistryTest.kt]: hasSize is for lists/arrays etc, count() returns int.

File: misk-metrics-digester/src/main/kotlin/misk/metrics/digester/Registry.kt [hunk 20]
```diff
@@ -0,0 +1,103 @@
...
+  val quantiles: List<Double>,
+  private val labelNames: List<String>,
+  private val tDigest: () -> SlidingWindowDigest<T>
+) : Histogram {
+
+  private val metrics = ConcurrentHashMap<Int, DigestMetric>()
+  private val labelRegex = "[a-zA-Z0-9_]*".toRegex()
+
+  /** Records a new metric within the histogram */
+  override fun labels(vararg labelValues: String): TDigestHistogramRecordMetric {
```
alecholmes: This still needs to check that the number label values is the same as the number of label names. This was a comment earlier and needs test coverage.

misk-metrics-digester/src/main/kotlin/misk/metrics/digester/Registry.kt
alecholmes: This still can't be sorted. Let's talk offline to clarify.

misk-metrics-digester/src/main/kotlin/misk/metrics/digester/Registry.kt
```diff
@@ -0,0 +1,103 @@
...
+              fun() = VeneurDigest())
+      )
+    }
+  }
+
+  private val histograms = ConcurrentHashMap<String, TDigestHistogram<T>>()
+  private val metricRegex =  "[a-zA-Z_:][a-zA-Z0-9_:]*".toRegex()
+
+  /** Creates and returns a new histogram. Histogram is registered within the Registry */
+  @Synchronized override fun newHistogram(
```
alecholmes: This shouldn't be synchronized. The use of `histograms` here gives the necessary amount of concurrency control without the overhead of `@Synchronized` being needed.

danw5050:
@alecholmes Yes, there were a couple things I wanted to clarify before continuing. I was waiting until our wed meeting to do that.

adrw:

chris-ryan-square:

chris-ryan-square:

PR #618, opened 2018-11-05
Cleanup AWS related bindings

mmihic:
Introduce AwsRegion and AwsAccountId types, and AwsEnvironmentModule to
pull the values for these from well-known environment variables. We
intentionally do _not_ consult the AWS Metadata Service, so that this
service can be locked down to prevent credential hijacking attacks.

Eliminate cloudwarch logging - this is no longer needed.


misk-aws/src/main/kotlin/misk/cloud/aws/Types.kt
```diff
@@ -0,0 +1,7 @@
+package misk.cloud.aws
+
+/** [AwsRegion] is the region in which the service is running */
+data class AwsRegion(val name: String)
```
swankjesse: nice

PR #621, opened 2018-11-05
Clarify params in MockWebServerService and add graceful shutdown logic

wesleyk:
r @ryanhall07

### DIFF
--- a/misk/src/test/kotlin/misk/client/MockWebServerService.kt
+++ b/misk/src/test/kotlin/misk/client/MockWebServerService.kt
@@ -6,22 +6,22 @@ import java.io.File
 import javax.inject.Singleton
 
 @Singleton
-class MockWebServerService(val socketFile: String?) : AbstractIdleService() {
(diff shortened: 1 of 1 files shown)


misk/src/test/kotlin/misk/client/MockWebServerService.kt
```diff
@@ -6,22 +6,22 @@ import java.io.File
 import javax.inject.Singleton
 
 @Singleton
-class MockWebServerService(val socketFile: String?) : AbstractIdleService() {
+class MockWebServerService(val unixSocketFile: String?) : AbstractIdleService() {
```
ryanhall07: I actually meant we should rename the class to make it clear it uses a socket file. although that would be a total mouthful of a name, so whatever.

wesleyk: well the idea is that the socket file specification is optional. You can run this as a mock server without specifying a unix socket.

PR #623, opened 2018-11-05
Jobqueue interfaces + SQS implementation

mmihic:
TransactionalJobQueue still to come


misk-jobqueue/src/main/kotlin/misk/jobqueue/JobConsumer.kt
```diff
@@ -0,0 +1,28 @@
...
+   * handler. A service may only have one subscription outstanding per queue
+   *
+   * @return a [Subscription] that can be closed to stop receiving jobs
+   */
+  fun subscribe(queueName: QueueName, handler: JobHandler): Subscription
+
+  interface Subscription {
+    /**
+     * Closes the  subscription and stops receiving jobs from the underlying queue. Jobs
+     * that are already in flight will continue to be dispatched to the handler, but
```
wesleyk: incomplete thought?

misk-aws/src/main/kotlin/misk/jobqueue/sqs/SqsJobConsumer.kt
```diff
@@ -0,0 +1,117 @@
...
+
+@Singleton
+internal class SqsJobConsumer @Inject internal constructor(
+  private val config: AwsSqsJobQueueConfig,
+  private val sqs: AmazonSQS,
+  private val queueUrls: QueueUrlMapping,
+  @ForSqsConsumer private val dispatchThreadPool: ExecutorService,
+  private val tracer: Tracer,
+  private val metrics: SqsMetrics
+) : AbstractIdleService(), JobConsumer {
```
ryanhall07: any opinion if we should use the other types of guava Services that help with thread management?

https://github.com/google/guava/wiki/ServiceExplained

misk-aws/src/main/kotlin/misk/jobqueue/sqs/SqsJobConsumer.kt
```diff
@@ -0,0 +1,117 @@
...
+              // incorrectly include the execution time of the job in the execution time of the
+              // action that triggered the job
+              message.attributes[SqsJob.ORIGINAL_TRACE_ID_ATTR]?.let {
+                ORIGINAL_TRACE_ID_TAG.set(span, it)
+              }
+
+              // Run the handler and record timing
+              try {
+                val (duration, _) = timed { handler.handleJob(message) }
+                metrics.handlerDispatchTime.record(duration.toMillis().toDouble(), queue.value)
```
ryanhall07: it would be nice to measure total time from enqueue to processing. do we have an enqueued_at timestamp anywhere?

misk-aws/src/main/kotlin/misk/jobqueue/sqs/SqsJobConsumer.kt
```diff
@@ -0,0 +1,117 @@
...
+              } catch (th: Throwable) {
+                log.error(th) { "error handling job from ${queue.value}" }
+                metrics.handlerFailures.labels(queue.value).inc()
```
ryanhall07: we probably want to include the shard number as a label, so we can identify slow/stuck receivers

misk-aws/src/test/kotlin/misk/jobqueue/sqs/DockerSqs.kt
```diff
@@ -0,0 +1,104 @@
...
+
+    val clientPort = ExposedPort.tcp(CLIENT_PORT)
+
+    log.info { "starting fake SQS service" }
+
+    val cmd = listOf("goaws")
+
+    // NB(mmihic): Because the client port is embedded directly into the queue URLs, we have to use
+    // the same external port as we do for the internal port
+    containerId = docker.createContainerCmd("pafortin/goaws")
```
ryanhall07: this is cool!

misk-jobqueue/src/main/kotlin/misk/jobqueue/JobQueue.kt
```diff
@@ -0,0 +1,28 @@
+package misk.jobqueue
+
+import java.time.Duration
+
+/**
+ * A [JobQueue] enqueues jobs for asynchronous execution, possibly in another process. Jobs
+ * are enqueued immediately and may involve an RPC to the underlying job queueing system, so should
+ * not be done from within a database transaction. Applications that need to enqueue jobs
+ * atomically with a local database transaction should use the [TransactionalJobQueue] interface
```
ryanhall07: that's coming later?

misk-aws/src/main/kotlin/misk/jobqueue/sqs/SqsJobConsumer.kt
mmihic: Not sure what this means. There's no real "shard" per se - just a pool of receivers. Which messages the receivers receive have no sharding implications, and the receiver doesn't wait for the current batch to finish dispatch before asking for more (unless the dispatch thread pool is full and applying backpressure)

misk-jobqueue/src/main/kotlin/misk/jobqueue/JobQueue.kt
mmihic: Yep

mmihic:
@ryanhall07 Correct there is no explicit ordering - it is best effort.

misk-aws/src/main/kotlin/misk/jobqueue/sqs/SqsJobConsumer.kt
ryanhall07: k. yea per-thread metrics is probably not interesting

misk-aws/src/main/kotlin/misk/jobqueue/sqs/SqsJobQueue.kt
```diff
@@ -0,0 +1,63 @@
...
+    deliveryDelay: Duration?,
+    attributes: Map<String, String>
+  ) {
+    tracer.traceWithSpan("enqueue-job-${queueName.value}") { span ->
+      metrics.jobsEnqueued.labels(queueName.value).inc()
+      try {
+        sqs.sendMessage(SendMessageRequest().apply {
+          queueUrl = queueUrlMapping[queueName]
+          messageBody = body
+          if (deliveryDelay != null) delaySeconds = (deliveryDelay.toMillis() / 1000).toInt()
```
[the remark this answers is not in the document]
wesleyk: err, but that thread seems to suggest there's a getSeconds(): https://docs.oracle.com/javase/8/docs/api/java/time/Duration.html#getSeconds--. (I don't really care about this, just curious)

PR #630, opened 2018-11-08
Partly transaction job queues

mmihic:
Enough of a transactional job queue to unblock development and initial staging testing. Honors transaction demarcations, but isn't resilient to process failures occurring after the local database commits.

### DIFF
--- a/misk-aws/src/main/kotlin/misk/jobqueue/sqs/SqsTransactionalJobQueue.kt
+++ b/misk-aws/src/main/kotlin/misk/jobqueue/sqs/SqsTransactionalJobQueue.kt
@@ -1,0 +1,46 @@
+package misk.jobqueue.sqs
+
+import misk.hibernate.Gid
+import misk.hibernate.Session
+import misk.jobqueue.JobQueue
+import misk.jobqueue.QueueName
+import misk.jobqueue.TransactionalJobQueue
+import misk.logging.getLogger
+import java.time.Duration
+import javax.inject.Inject
+import javax.inject.Singleton
+
+// Implements the TransactionalJobQueue interface by registering a post-commit hook that
+// forwards the message to SQS. It is not truly transactional, and is intended only for
+// development / staging while we build proper job forwarding from the replication stream,
+// but works so long as there is no application failure between the commit and the forward
+@Singleton
+internal class SqsTransactionalJobQueue @Inject internal constructor(
+  private val jobQueue: JobQueue
+): TransactionalJobQueue {
+  override fun enqueue(
+    session: Session,
+    queueName: QueueName,
+    body: String,
+    deliveryDelay: Duration?,
+    attributes: Map<String, String>
+  ) {
+    session.onPostCommit {
+      log.info { "forwarding to ${queueName.value}" }
+      jobQueue.enqueue(queueName, body, deliveryDelay, attributes)
+    }
+  }
+
+  override fun enqueue(
+    session: Session,
+    gid: Gid<*, *>,
+    queueName: QueueName,
+    body: String,
+    deliveryDelay: Duration?,
+    attributes: Map<String, String>
+  ) = enqueue(session, queueName, body, deliveryDelay, attributes)
+
+  companion object {
+    private val log = getLogger<SqsTransactionalJobQueue>()
+  }
+}
--- a/misk-hibernate/src/main/kotlin/misk/hibernate/RealTransacter.kt
+++ b/misk-hibernate/src/main/kotlin/misk/hibernate/RealTransacter.kt
@@ -30,8 +30,12 @@ internal class RealTransacter private constructor(
   private val tracer: Tracer?
 ) : Transacter {
 
-  constructor(qualifier: KClass<out Annotation>, sessionFactory: SessionFactory, config: DataSourceConfig, tracer: Tracer?) :
-      this(qualifier, sessionFactory, config, ThreadLocal(), TransacterOptions(), tracer)
+  constructor(
+    qualifier: KClass<out Annotation>,
+    sessionFactory: SessionFactory,
+    config: DataSourceConfig,
+    tracer: Tracer?
+  ) : this(qualifier, sessionFactory, config, ThreadLocal(), TransacterOptions(), tracer)
 
   override val inTransaction: Boolean
     get() = threadLocalSession.get() != null
@@ -50,9 +54,9 @@ internal class RealTransacter private constructor(
     assert(options.maxAttempts > 0)
 
     val backoff = ExponentialBackoff(
-      Duration.ofMillis(options.minRetryDelayMillis),
-      Duration.ofMillis(options.maxRetryDelayMillis),
-      Duration.ofMillis(options.retryJitterMillis)
+        Duration.ofMillis(options.minRetryDelayMillis),
+        Duration.ofMillis(options.maxRetryDelayMillis),
+        Duration.ofMillis(options.retryJitterMillis)
     )
     var attempt = 0
 
@@ -64,11 +68,15 @@ internal class RealTransacter private constructor(
         if (!isRetryable(e)) throw e
 
         if (attempt >= options.maxAttempts) {
-          logger.info(e) { "${qualifier.simpleName} recoverable transaction exception (attempt: $attempt), max attempts exceeded" }
+          logger.info(e) {
+            "${qualifier.simpleName} recoverable transaction exception (attempt: $attempt), max attempts exceeded"
+          }
           throw e
         }
 
-        logger.info(e) { "${qualifier.simpleName} recoverable transaction exception (attempt: $attempt), retrying" }
+        logger.info(e) {
+          "${qualifier.simpleName} recoverable transaction exception (attempt: $attempt), retrying"
+        }
         val sleepDuration = backoff.nextRetry()
         if (!sleepDuration.isZero) {
           try {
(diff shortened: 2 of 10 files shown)


misk-hibernate/src/main/kotlin/misk/hibernate/RealTransacter.kt
```diff
@@ -245,10 +256,39 @@ internal class RealTransacter private constructor(
...
+      }
+    }
+
+    override fun onPreCommit(work: () -> Unit) {
+      preCommitHooks.add(work)
+    }
+
+    internal fun postCommit() {
+      postCommitHooks.forEach {
+        // At this point the transaction has committed, so simply log errors and don't propagate
```
ryanhall07: I think I would still want the exception to propagate, even if the tx has been committed and can't be rolled back. For example it would be nice for the current WebAction to terminate with a 500 and be allowed to be retried by the caller, hopefully then the postCommit works

misk-hibernate/src/main/kotlin/misk/hibernate/RealTransacter.kt
```diff
@@ -82,23 +91,22 @@ internal class RealTransacter private constructor(
...
 
   private fun <T> transactionInternal(lambda: (session: Session) -> T): T {
-    if (tracer != null) {
-      return tracer.traceWithSpan(DB_TRANSACTION_SPAN_NAME) { span ->
-        Tags.COMPONENT.set(span, TRANSACTER_SPAN_TAG)
-        transactionInternalSession(lambda)
-      }
-    }
-    return transactionInternalSession(lambda)
+    return if (tracer != null) tracer.traceWithSpan(DB_TRANSACTION_SPAN_NAME) { span ->
```
nbroyles: nit: formatting seems odd here?

mmihic: how so?

misk-hibernate/src/main/kotlin/misk/hibernate/RealTransacter.kt
mmihic: I don't think I agree. If `Transacter.transaction` throws an exception, most code is going to assume that the transaction failed - which it didn't, and re-running the work in the transaction may not be possible. We could potentially use a specific exception to differentiate these cases (`PostCommitHookFailedException`) but I think this might be too subtle for application code.

nbroyles: +1, if there's important work happening in a postCommit, you probably want a heads up that it didn't happen.

mmihic: Ok, I'll wrap in a specific exception.

misk-hibernate/src/main/kotlin/misk/hibernate/RealTransacter.kt
```diff
@@ -245,10 +256,39 @@ internal class RealTransacter private constructor(
       return session.doReturningWork(work)
     }
 
+    internal fun preCommit() {
+      preCommitHooks.forEach {
+        // Propagate hook exceptions up to the transacter so that the the transaction is rolled
+        // back and the error gets returned to the applicatiob
```
wesleyk: /nit application

misk-hibernate/src/test/kotlin/misk/hibernate/TransacterTest.kt
```diff
@@ -239,6 +240,137 @@ class TransacterTest {
     tracingAssertions()
   }
 
+  @Test
```
wesleyk: really easy to follow tests!

tirsen:
Another random observation that you're probably already aware of: Afaict SQS doesn't have "two phase" send like Jobsy. That increases probability of a post commit send failure. I haven't entirely thought through all the implications of that but it seems wise to keep message size small. Fortunately we already do that, just sending a small ID and maybe an enum.

misk-aws/src/main/kotlin/misk/jobqueue/sqs/SqsTransactionalJobQueue.kt
```diff
@@ -0,0 +1,46 @@
...
+  override fun enqueue(
+    session: Session,
+    queueName: QueueName,
+    body: String,
+    deliveryDelay: Duration?,
+    attributes: Map<String, String>
+  ) {
+    session.onPostCommit {
+      log.info { "forwarding to ${queueName.value}" }
+      jobQueue.enqueue(queueName, body, deliveryDelay, attributes)
```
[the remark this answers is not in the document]
mmihic: This is totally temporary code - the "real" code will always write to to local spool, this is just enough to get raven unblocked in development / testing.  The class level comment I hope makes that clear, but maybe not?

misk-aws/src/main/kotlin/misk/jobqueue/sqs/SqsTransactionalJobQueue.kt
```diff
@@ -0,0 +1,46 @@
...
+import misk.jobqueue.QueueName
+import misk.jobqueue.TransactionalJobQueue
+import misk.logging.getLogger
+import java.time.Duration
+import javax.inject.Inject
+import javax.inject.Singleton
+
+// Implements the TransactionalJobQueue interface by registering a post-commit hook that
+// forwards the message to SQS. It is not truly transactional, and is intended only for
+// development / staging while we build proper job forwarding from the replication stream,
```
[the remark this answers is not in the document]
mmihic: For it to be actually transactional (in particular, atomic + durable wrt the local database transaction) you always have to write to a local spool. We could then attempt an enqueue followed by a delete, but then you still run into the issue that the replication stream observer might see the insert before you finished the enqueue+delete. So I think once we decide to use replication stream as a forwarding system, it's kindof all or nothing

misk-hibernate/src/main/kotlin/misk/hibernate/RealTransacter.kt
```diff
@@ -161,9 +169,9 @@ internal class RealTransacter private constructor(
   )
 
   companion object {
-    val APPLICATION_TRANSACTION_SPAN_NAME = "app-db-transaction"
-    val DB_TRANSACTION_SPAN_NAME = "db-session"
-    val TRANSACTER_SPAN_TAG = "hibernate-transacter"
+    const val APPLICATION_TRANSACTION_SPAN_NAME = "app-db-transaction"
```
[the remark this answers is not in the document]
mmihic: https://stackoverflow.com/questions/37482378/static-data-in-kotlin/37485356#37485356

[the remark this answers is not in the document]
mmihic: Generally prefer `const val` for primitive types + strings. Intellij warns about this (pretty much all the non-functional changes in this file are the result of accepting Intellij's warnings)

misk-hibernate/src/test/kotlin/misk/hibernate/TransacterTest.kt
```diff
@@ -239,6 +240,141 @@ class TransacterTest {
...
+    assertThat(cause).isInstanceOf(IllegalStateException::class.java)
+
+    // Only the first hook should complete - the others should be the skipped due to error...
+    assertThat(postCommitHooksTriggered).containsExactly("first")
+
+    // ...but the transaction itself should have completed
+    assertThat(transacter.transaction { it.load(swid) }.name).isEqualTo("Star Wars")
+
+  }
+
```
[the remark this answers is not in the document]
mmihic: The "real" transactional queue won't use `postCommit` hooks - it'll write to the local spool directly when you call `enqueue`. And yeah once the actual spooling behavior exists there will be cowrite tests.

misk-aws/src/main/kotlin/misk/jobqueue/sqs/SqsTransactionalJobQueue.kt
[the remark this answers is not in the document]
tirsen: Jobsy had this two phase enqueue and now that I think about it I think we can use it here too.

The first phase we just enqueue directly to Jobsy and I think we even did that inside the transaction. We then also did a transaction local write to a "ready ticket" table inside the transaction. On the processing side we checked the ready ticket table to see if the transaction that posted actually committed. If it didn't commit (yet) we don't ack the message and retry for a configurable amount of time (until the transaction is assumed to have rolled back).

(We also did an optimized "happy path" which cached the ready ticket in Redis no a post commit hook. If the ready ticket wasn't in Redis we had to fall back to checking the ready ticket table. This dramatically drops qps on the ready ticket table.)

The advantage is that we don't need a polling job or streaming the binlogs.

The main drawback is that we call out to Jobsy inside the transaction so you gotta have a pretty short timeout on that call. I don't believe we've had issues with this in Franklin.

Of course this only works if the sender and receiver of a job is the same application so that they can both read and write the ready ticket table (which should always be the case for job queueing but won't work for a pubsub).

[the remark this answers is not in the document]
mmihic: Yeah I explicitly rejected the jobsy approach of doing an RPC with the database transaction open. We get away with this in datacenter because Franklin and jobsy are quite close so the median latency is very low and the latency distribution is pretty tight. Neither of these hold for SQS. p50 on an SQS send is ~100ms and the range of latencies is much broader. Even though median is well below the timeout, adding an extra 100-200ms to every database transaction is going to increase lock contention and cause transactions to pile up (as we saw when we had the few hundred extra ms of cross-shard latency awhile back). 

Besides, the jobsy model won't work for evently and we need to deliver real transactional publishes for both at around the same time.

Issue #422, opened 2018-08-28
Add JS logging library


adrw:

adrw:
Moved to https://github.com/square/misk-web/issues/16

PR #633, opened 2018-11-12
Environment Navbar

### LINKED ISSUE #462: Add coloured bar to navbar for environment

### LINKED ISSUE #634: Enable scrolling of menu in mobile or when too long

### LINKED ISSUE #637: Split TopbarComponent up into subcomponents for easier maintainability / modularity

adrw:
* [CLOSES #462] Environment Navbar added to `@misk/components:TopbarComponents`
* Fixed Refresh local Node Modules functionality in `misk-build`
* Add ability to override build Docker image with `dockerImage` field in package.json
* Add color Enum to `@misk/common` to have common SQ colors
* Add additional Responsive containers
* Responsive spot in Navbar for additional menu items
* Additional menu items also show in pull down menu for cases of overflow / mobile view
* Size awareness allows responsive pruning of top navbar additional header items to the dropdown menu instead
* New `TextHTMLOrElementComponent` renders anything passed into it
* New navbarItems props for Topbar Component allows dynamic passing in of top level navbar links and React components
* New status props for Topbar Component allows dynamic setting of status text / link / React Component in coloured status bar that shows by default in Dev, Testing, and Staging environments
* [CLOSES #634] Scrollable Misk Menu (for when it is too long and goes off screen)
* `misk-web@0.0.6`
* [CLOSES #637] Split up Topbar into feature with many sub components
* Rename TopbarComponent -> Navbar

screen shot 2018-11-15 at 5 16 53 pm


misk/web/@misk/common/webpack.config.js
```diff
@@ -7,7 +7,7 @@ module.exports = {
   name: "library",
   mode: "production",
   entry: {
-    common: path.resolve(__dirname, "src/index.ts")
+    common: path.resolve(__dirname, "src/index.ts"),
```
jjestrel: Not for this PR, but if we're using a `js` linter anywhere else, we could run it through our config files to keep them consistent?

misk/web/@misk/components/src/components/TextHTMLOrElementComponent.tsx
```diff
@@ -0,0 +1,30 @@
...
+ */
+
+export interface ITextHTMLOrElementProps {
+  content: string | Element | JSX.Element
+}
+
+export const TextHTMLOrElementComponent = (
+  props: ITextHTMLOrElementProps
+): JSX.Element => {
+  const { content }: any = props
```
jjestrel: Can this be `const { content }: string | Element | JSX.Element`?

misk/web/@misk/components/src/components/TextHTMLOrElementComponent.tsx
```diff
@@ -0,0 +1,30 @@
...
+  let FormattedContent: any = content
+
+  if (typeof content === "string" && !content.startsWith("<")) {
+    FormattedContent = <span>{content}</span>
+  } else if (typeof content === "string" && content.startsWith("<")) {
+    FormattedContent = <span dangerouslySetInnerHTML={{ __html: content }} />
+  }
```
jjestrel: If we receive a `JSX.Element` here, do we need to wrap it in a `span` too?

misk/web/@misk/components/src/components/TextHTMLOrElementComponent.tsx
```diff
@@ -0,0 +1,30 @@
...
+
+export const TextHTMLOrElementComponent = (
+  props: ITextHTMLOrElementProps
+): JSX.Element => {
+  const { content }: any = props
+  let FormattedContent: any = content
+
+  if (typeof content === "string" && !content.startsWith("<")) {
+    FormattedContent = <span>{content}</span>
+  } else if (typeof content === "string" && content.startsWith("<")) {
```
jjestrel: When would we want to pass in an html string instead of a `JSX.Element`?

docker/misk-web/misk-web
```diff
@@ -21,10 +21,7 @@ usage() {
 
   -l    'lint' run Prettier linting
 
-  -o    compute a md5 hash over the current directory and stores in .hash file
-
-  -m    compute a md5 hash over the current directory and stores in .hash file, creates .hashdiff file if there is an existing hash
-        exclude patterns: cachedUrls, lib, node_modules, .hash
+  -m    compute a md5 hash over the current directory and stores in .hash file
```
jaredwindover: Does it store it on its own, or do you have to redirect it? It looks like it's being redirected above.

docker/misk-web/misk-web
```diff
@@ -64,15 +62,28 @@ hashDir() {
...
+        echo "${oldHash} ${newHash} false"
+    else
+        echo "${oldHash} ${newHash} true"
+    fi
+}
+
+hashDiffFiles() {
+    oldHashFile="$1"
+    newHashFile="$2"
+    [ -f "$oldHashFile" ] || echo "true"
```
jaredwindover: What are these two lines doing? Early exit with true if either doesn't exist?

gradle/web.gradle
```diff
@@ -129,7 +154,7 @@ task web(type: Exec) {
     }
   }
 
-  def sleepUntilDockerExits = "while docker ps --format '{{.Names}}' | grep -q '${projectName}-${subProjectName}'; do echo -e \"\\r\\033[1A\\033[0K[\$(date +'%Y-%m-%d %T')] waiting for [\$(docker ps --format '{{.Names}}' | grep '${projectName}-${subProjectName}' | wc -l | tr -d '\t')] ${projectName}-${subProjectName} running containers to finish...\" && sleep 5; done"
+  def sleepUntilDockerExits = "while docker ps --format '{{.Names}}' | grep -q '${projectName}-${subProjectName}-web'; do echo -e \"\\r\\033[1A\\033[0K[\$(date +'%Y-%m-%d %T')] waiting for [\$(docker ps --format '{{.Names}}' | grep '${projectName}-${subProjectName}' | wc -l | tr -d '\t')] *${projectName}-${subProjectName}-web* running containers to finish...\" && sleep 2; done"
```
jaredwindover: Is this "-web" the same as the prefix? Should it be a variable as well?

misk/web/@misk/components/src/components/TextHTMLOrElementComponent.tsx
```diff
@@ -0,0 +1,30 @@
+import * as React from "react"
+
+/**
+ * <TextHTMLOrElementComponent content={<div>item</div>} length={35} />
```
jaredwindover: Is there a reason to not use the element's children?

misk/web/@misk/components/src/components/TopbarComponentDimensionAware.tsx
```diff
@@ -0,0 +1,388 @@
...
+  Icon,
+  Navbar,
+  NavbarGroup
+} from "@blueprintjs/core"
+import { IconNames } from "@blueprintjs/icons"
+import { color, Environment, IDashboardTab } from "@misk/common"
+import { groupBy, sortBy } from "lodash"
+import * as React from "react"
+import { Link } from "react-router-dom"
+import styled from "styled-components"
```
jaredwindover: This is pretty cool

misk/web/@misk/components/src/components/TopbarComponentDimensionAware.tsx
```diff
@@ -0,0 +1,388 @@
...
+    text-decoration: underline;
+    letter-spacing: 1px;
+    &:hover {
+      color: ${color.white};
+      text-decoration: underline;
+    }
+  }
+`
+
+const environmentToColor = (environment: Environment) => {
```
jaredwindover: nit: I think a map would be better here than a switch.

misk/web/@misk/components/src/components/TopbarComponentDimensionAware.tsx
```diff
@@ -0,0 +1,388 @@
...
+    case Environment.PRODUCTION:
+      return color.red
+  }
+}
+
+export class TopbarComponentDimensionAware extends React.Component<
+  IDimensionAwareProps & ITopbarProps,
+  {}
+> {
+  public state = {
```
jaredwindover: This is probably fine, and this mostly comes from my tendency to avoid state whenever possible, but a nice pattern can be to use a HOC to pass in the state and its handlers to the component. Totally unnecessary, but I like the separation of concerns you get.

misk/web/@misk/components/src/components/TopbarComponentDimensionAware.tsx
```diff
@@ -0,0 +1,388 @@
...
+            </ResponsiveContainer>
+          </MiskMenu>
+        </MiskCollapse>
+        {this.renderBanner(status, environment, environmentBannerVisible)}
+      </MiskNavbar>
+    )
+  }
+
+  private NavbarItems(environment?: Environment) {
+    console.log(this.props.navbarItems)
```
jaredwindover: Should this be here?

misk/web/@misk/components/src/components/TopbarComponentDimensionAware.tsx
```diff
@@ -0,0 +1,388 @@
...
+    return [this.renderNavbarEnvironment(environment)].concat(
+      this.props.navbarItems
+        ? this.props.navbarItems.map(item => (
```
jaredwindover: Why are navbar items so heterogeneous to need a component to marshall them into spans?

misk/web/@misk/components/src/components/TopbarComponentDimensionAware.tsx
```diff
@@ -0,0 +1,388 @@
...
+    environment?: Environment,
+    environmentBannerVisible: Environment[] = [
+      Environment.DEVELOPMENT,
+      Environment.STAGING,
+      Environment.TESTING
+    ]
+  ) {
+    if (environmentBannerVisible.includes(environment)) {
+      if (typeof status === "string" && !status.startsWith("<")) {
+        console.log(`[STATUS] ${status}`)
```
jaredwindover: What's the convention on console messages in Misk? Is this for debugging?

misk/web/@misk/components/src/components/TopbarComponentDimensionAware.tsx
```diff
@@ -0,0 +1,388 @@
...
+      <div>
+        <MiskMenuCategory>{categoryName}</MiskMenuCategory>
+        <MiskMenuDivider />
+        <MiskMenuLinks>
+          {sortedCategoryLinks.map((link: IDashboardTab) => (
+            <MiskMenuLink
+              key={link.slug}
+              onClick={this.handleClick}
+              to={link.url_path_prefix}
+            >
```
jaredwindover: nit: is it intentional to have this on its own line?

gradle/web.gradle
```diff
@@ -18,47 +19,68 @@ task web(type: Exec) {
     if (slug.startsWith('-')) {
       slug = slug.drop(1)
     }
+    if (slug == "") {
+      slug = "xyz"
```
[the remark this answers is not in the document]
adrw: An empty slug is generated if the project isn't tab-specific so this is a placeholder, I'll change it to `misc` or leave it empty for less confusion.

misk/web/@misk/common/webpack.config.js
adrw: Good idea, I'll update the linting filters so the config files get picked up 👍

misk/web/@misk/components/src/components/TextHTMLOrElementComponent.tsx
adrw: Thanks! That's a good catch!

misk/web/@misk/components/src/components/TextHTMLOrElementComponent.tsx
adrw: If it's already a `JSX.Element` then it's ready to be rendered. Wrapping it in an additional `span` can end up messing with `FlexBox` CSS arrangement unfortunately.

misk/web/@misk/components/src/components/TextHTMLOrElementComponent.tsx
adrw: An unlikely use case but would allow for example server side rendering of navbar items to come from Misk server, static definitions, and other places that aren't in the same Typescript project.

docker/misk-web/misk-web
adrw: It automatically uses `tee` to store to `.hash` in the current directory but also outputs to standard out the hash for logging and so it can be redirected if the hash should go to another location.

docker/misk-web/misk-web
adrw: Correct. If the oldHash file `$(pwd)/.hash` doesn't exist, then the comparison of the directory contents can't be checked at all so automatically falsify the existing cache.

jaredwindover: Seems like this may echo "hash1 hash2 true/false" or just "true" depending on whether or not both files exist. Is that fine?

misk/web/@misk/components/src/components/TextHTMLOrElementComponent.tsx
adrw: Changed it to use children 👍

misk/web/@misk/components/src/components/TopbarComponentDimensionAware.tsx
adrw: Good call https://jsperf.com/closure-switch-vs-object/3

misk/web/@misk/components/src/components/TopbarComponentDimensionAware.tsx
adrw: Yeah I've tried to minimize any local state to simple UI methods. I've further reduced this massive component into smaller ones that should help limit the scope that the State is needed. HOC is definitely a good pattern.

misk/web/@misk/components/src/components/TopbarComponentDimensionAware.tsx
adrw: Oops!

misk/web/@misk/components/src/components/TopbarComponentDimensionAware.tsx
adrw: Long term we want the ability to arbitrarily inject potentially complex full elements such as search bar and others that may or may not ship in the same `@misk/components` package.

misk/web/@misk/components/src/components/TopbarComponentDimensionAware.tsx
adrw: No convention yet but it is on our todo list #422

misk/web/@misk/components/src/components/TopbarComponentDimensionAware.tsx
adrw: `jsx-alignment` rule in TSLint https://github.com/palantir/tslint-react

misk/web/@misk/components/src/containers/DesktopWideOnlyContainer.tsx
```diff
@@ -0,0 +1,13 @@
+import styled from "styled-components"
+
+/**
+ * <DesktopWideOnlyContainer.1>
+ *    <span>Stuff</span>
+ * </DesktopWideOnlyContainer.1>
+ */
+
+export const DesktopWideOnlyContainer = styled.div`
+  @media (max-width: 1200px) {
```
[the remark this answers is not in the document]
adrw: Added an issue for this #645

gradle/web.gradle
adrw: Just to denote that it is a web container as opposed to others that may be generated by the service that aren't web related.

PR #641, opened 2018-11-16
Revert "Add ability for clients to specify their own Aws ClientConfig…

wesleyk:
…uration"

This reverts commit 128cf1762c4717da30d3f9785d4c67c1b79a1671.

No longer needed since we'll use default java proxy properties instead.


ryanhall07:
> @ryanhall07 it's unique to our setup. 1) Our proxy setup will be detailed in our container contract 2) Misk devs won't have to think about this because we share dockerfiles.

I mean outside of SQ, how is a misk user going to know misk relies on these Java system vars

wesleyk:
> I mean outside of SQ, how is a misk user going to know misk relies on these Java system vars

Eh yeah, you mean like, the only way misk users can configure a proxy is to set these vars? Yeah true. Misk honestly probably needs some sort of container expectations as well.

wesleyk:
> > I mean outside of SQ, how is a misk user going to know misk relies on these Java system vars
> 
> Eh yeah, you mean like, the only way misk users can configure a proxy is to set these vars? Yeah true. Misk honestly probably needs some sort of container expectations as well.

Like the ENVIRONMENT env variable

PR #642, opened 2018-11-16
Initial Schema Validation

acostama:
- compare hibernate entities against db schema to:
 - find missing tables
 - find missing columns
 - find duplicate columns
 - check for bad identifiers.. enforce lower_snake_case
 - check that identifiers match exactly


adrw:

misk-hibernate/src/main/kotlin/misk/hibernate/SchemaValidator.kt
```diff
@@ -0,0 +1,301 @@
...
+    }
+
+    validate(hibernateOnly.isEmpty()) {
+      "Database table \"${dbTable.name}\" is missing columns ${hibernateOnly.map { it.name }} found in hibernate \"${hibernateTable.name}\""
+    }
+
+    // TODO (maacosta) how strict should we be here? If `DEFAULT NULL` column exists in the Db and hibernate
+    //                does not know about it hibernate can still do writes to db. However if we are going to do
+    //                lookups on this column it might not work.. do we look at queries?.
+    validate(dbOnly.isEmpty()) {
```
frojasg: what do you mean with 
> However if we are going to do lookups on this column it might not work.. do we look at queries?.
How are you going to do lookups if the app doesn't know about it? 

This validation is dangerous if I change the schema of my table before deploying the app to production. I won't be able to restart the application until the new version of the app is deployed.

misk-hibernate/src/main/kotlin/misk/hibernate/SchemaValidator.kt
```diff
@@ -0,0 +1,301 @@
...
+
+    for ((dbTable, hibernateTable) in intersectionPairs) {
+      withDeclaration(hibernateTable.snakeCaseName) {
+        validateTables(dbTable, hibernateTable)
+      }
+    }
+  }
+
+  private fun validateTables(dbTable: TableDeclaration, hibernateTable: TableDeclaration) {
+    val (dbOnly, hibernateOnly, intersectionPairs) = splitChildren(
```
wesleyk: maybe dbOnlyColumns and hibernateOnlyColumns?

PR #651, opened 2018-11-19
Ensure use of http/2 when communicating w/ Envoy sidecar

nbroyles:


misk/src/main/kotlin/misk/client/HttpClientFactory.kt
```diff
@@ -39,6 +39,7 @@ class HttpClientFactory {
           UnixDomainSocketFactory(envoyClientEndpointProvider.unixSocket(config.envoy)))
       // No DNS lookup needed since we're just sending the request over a socket.
       builder.dns(NoOpDns())
+      builder.protocols(listOf(Protocol.H2_PRIOR_KNOWLEDGE))
```
wesleyk: can you leave a comment why we only want this set for envoy?

adrw:

swankjesse:
cool!

PR #653, opened 2018-11-20
Lazy SessionFactory binding

tirsen:
Injecting a raw SessionFactory triggers an eager call to
SessionFactoryService.get() at injection time if you have a service that
(directly or indirectly) injects a Transacter. The SessionFactoryService
hasn't been started at that point which causes an IllegalStateException.
Using a Provider<SessionFactory> makes this call lazy which solves the
issue.

### DIFF
--- a/misk-hibernate/src/main/kotlin/misk/hibernate/RealTransacter.kt
+++ b/misk-hibernate/src/main/kotlin/misk/hibernate/RealTransacter.kt
@@ -16,6 +16,7 @@ import org.hibernate.StaleObjectStateException
 import org.hibernate.exception.LockAcquisitionException
 import java.sql.Connection
 import java.time.Duration
+import javax.inject.Provider
 import javax.persistence.OptimisticLockException
 import kotlin.reflect.KClass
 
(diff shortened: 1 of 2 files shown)


misk-hibernate/src/main/kotlin/misk/hibernate/RealTransacter.kt
```diff
@@ -32,10 +33,13 @@ internal class RealTransacter private constructor(
...
     qualifier: KClass<out Annotation>,
-    sessionFactory: SessionFactory,
+    sessionFactoryProvider: Provider<SessionFactory>,
     config: DataSourceConfig,
     tracer: Tracer?
-  ) : this(qualifier, sessionFactory, config, ThreadLocal(), TransacterOptions(), tracer)
+  ) : this(qualifier, sessionFactoryProvider, config, ThreadLocal(), TransacterOptions(), tracer)
+
+  private val sessionFactory
+    get() = sessionFactoryProvider.get()
```
tirsen: This is only run when the `sessionFactory` property is accessed. It's like a getter.

PR #654, opened 2018-11-20
Don't check for migrations on Vitess yet

tirsen:


misk-hibernate/src/main/kotlin/misk/hibernate/SchemaMigrator.kt
```diff
@@ -2,12 +2,17 @@ package misk.hibernate
 
 import com.google.common.base.Stopwatch
 import misk.jdbc.DataSourceConfig
+import misk.jdbc.DataSourceType
 import misk.logging.getLogger
 import misk.resources.ResourceLoader
 import org.hibernate.SessionFactory
 import org.hibernate.query.Query
 import java.sql.Connection
-import java.util.*
```
tirsen: I assume we don't want to wildcard import right? Or do we alllow this in Kotlin?

adrw:

misk-hibernate/src/main/kotlin/misk/hibernate/SchemaMigrator.kt
wesleyk: I don't think so, I think some folks don't have the same java import rules setup for Kotlin in IntelliJ (myself included until recently).

PR #655, opened 2018-11-20
Pull using docker pull CLI instead of the API

tirsen:
The CLI doesn't time out and gives much better UI feedback.

### DIFF
--- a/misk-hibernate-testing/src/main/kotlin/misk/jdbc/StartVitessService.kt
+++ b/misk-hibernate-testing/src/main/kotlin/misk/jdbc/StartVitessService.kt
@@ -1,7 +1,6 @@
 package misk.jdbc
 
 import com.github.dockerjava.api.DockerClient
-import com.github.dockerjava.api.exception.InternalServerErrorException
 import com.github.dockerjava.api.model.Bind
 import com.github.dockerjava.api.model.ExposedPort
 import com.github.dockerjava.api.model.Frame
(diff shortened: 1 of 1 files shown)


misk-hibernate-testing/src/main/kotlin/misk/jdbc/StartVitessService.kt
```diff
@@ -245,20 +244,8 @@ class StartVitessService(val config: DataSourceConfig) : AbstractIdleService() {
...
-          .withTag("latest")
-          .exec(PullImageResultCallback())
-          .awaitSuccess()
-    } catch (e: InternalServerErrorException) {
-      if (e.message?.contains("Service Unavailable") == true) {
-        logger.info("You do not have an internet connection. Skipping pull.")
-      } else {
-        throw e
-      }
+    if (runCommand("docker pull vitess/base:latest") != 0) {
```
adrw: Should this be pinned to a version?

.travis.yml
```diff
@@ -7,7 +7,6 @@ jdk:
 # Run Docker Pulls in parallel, esp for vitess/base with size > 1GB
 install:
   - docker pull squareup/misk-web &
-  - docker pull vitess/base &
```
adrw: I'd recommend leaving this in so it can pull in the background and keep minimal CI build times as opposed to synchronously waiting for it to pull >1GB image down when it hits the `misk-hibernate` code.

PR #663, opened 2018-11-23
Allow sessions to load entities via Gid

A-Tran:


misk-hibernate/src/main/kotlin/misk/hibernate/InvalidChildEntityLoadException.kt
```diff
@@ -0,0 +1,13 @@
...
+import org.hibernate.TypeMismatchException
+
+/**
+ * [InvalidChildEntityLoadException] is raised when there is an attempt to retrieve a child entity
+ * using an id. Naturally, loading a child by id without a lookup will result in a scatter query,
+ * which we want to avoid. Hibernate will catch these with a `TypeMismatchException` as the entity
+ * will expect a gid, but we'll rewrite those as `InvalidChildEntityLoadException` for clarity. The
+ * expected way to load child entities is to use `loadByGid()`
+ */
+class InvalidChildEntityLoadException(e: TypeMismatchException) : Exception(e) {
```
swankjesse: I think we can probably get away without this? Fewer exception types is preferable, unless there’s somebody who wants to catch one exception and not another

A-Tran: I think the main incentive here is clarity - something like `session.load(entityId)` feels pretty natural to do, but will blow up if you use a child entity. Rather than seeing `TypeMismatchException: Expected gid but got id` and not necessarily understanding the context around the exception - I think highlighting the context of `InvalidChildEntityLoad` makes it clear where the error is rooted. Though I suppose I eventually found my way, so I guess other people could too? Not too attached to this idea - but thought it could be useful.

tirsen: I agree. We definitely want the errors to be entirely self explanatory and `TypeMismatchException` is not. Adding an extra exception here with a nice explanatory JavaDoc is a big win for engineers self sufficiency. Self sufficiency increases independence and independence increases job satisfaction.

misk-hibernate/src/main/kotlin/misk/hibernate/RealTransacter.kt
```diff
@@ -209,7 +210,15 @@ internal class RealTransacter private constructor(
     }
 
     override fun <T : DbEntity<T>> load(id: Id<T>, type: KClass<T>): T {
-      return session.get(type.java, id)
+      try {
+        return session.get(type.java, id)
+      } catch (e : TypeMismatchException) {
```
tirsen: Instead of catching and rethrowing can you instead check for the specific error as a precondition and throw the error there. I suspect `TypeMismatchException` could also be caused by other errors which aren't explained by this specific exception.

misk-hibernate/src/main/kotlin/misk/hibernate/RealTransacter.kt
```diff
@@ -209,7 +210,15 @@ internal class RealTransacter private constructor(
...
+        throw InvalidChildEntityLoadException(e)
+      }
+    }
+
+    override fun  <R : DbRoot<R>, T : DbSharded<R, T>> loadByGid(gid: Gid<R, T>, type: KClass<T>): T {
```
tirsen: Love the idea but not sold on the name. It's a bit long for something that is going to be one of the most commonly used methods. `loadSharded` rolls a bit easier on the tongue but is still long on the screen. Damn Java and its type erasure, otherwise we could have overloaded the method. :-( I'm almost tempted to rename the current `load` method to `loadUnsharded` and make this the short `load` method. This will be more common. Anyway we can iterate on the name.

A-Tran: I'm going to use `loadSharded` for now - For a little while (in development) I had it working by overloading `load`, i'm unsure what Java's type erasure is, or how that would've surfaced a problem

misk-hibernate/src/main/kotlin/misk/hibernate/RealTransacter.kt
A-Tran: I'm a bit unclear on what you're suggesting here:
```check for specific error as a precondition, and throw the error there```

This exception is effectively a subset of `TypeMismatchException`, so it sounds like what you want to do is check that we have a `InvalidChildEntityException` before even making the `session.get` call, so that we don't falsely group a `TypeMismatch` as a `InvalidChildEntity`. Does this sound about right?

A-Tran:
Updated to exclude the custom exception, will follow-up with a way to prevent `load` from being called with a `DbChild` or `DbSharded`.

misk-hibernate/src/main/kotlin/misk/hibernate/RealTransacter.kt
tirsen: Correct.

misk-hibernate/src/main/kotlin/misk/hibernate/RealTransacter.kt
tirsen: Oh yeah that's true. `Gid` won't be erased so it should work by overloading. I would suggest you overload `load`!

PR #664, opened 2018-11-23
Add travis badge to README

jjestrel:


swankjesse:
Generally we’ve tried to avoid these badges in our user-facing open source projects. This one’s fine ’cause we’re not stable enough. But I’d like to get to the point where we’re always green and we can remove this.

PR #672, opened 2018-11-28
Load secrets as referenced in config files

acostama:
as per https://github.com/square/misk/issues/660


misk/src/main/kotlin/misk/config/ConfigModule.kt
```diff
@@ -18,11 +23,25 @@ class ConfigModule<T : Config>(
...
     bindChildConfigClass(configClass)
+
+    val configLoaderProvider = object : Provider<SecretConfigLoader> {
+      @Inject lateinit var resourceLoader: ResourceLoader
+      @Inject lateinit var moshi: Moshi
+      override fun get(): SecretConfigLoader =
+          SecretConfigLoader(resourceLoader, moshi, configClass, config)
+    }
+
+    multibind<Service>().toProvider(Provider<SecretConfigService> {
```
mmihic: It's a little strange for config to register a service - can we just do the loading in-line here, or on-demand when someone calls `value`

misk/src/main/kotlin/misk/config/MiskConfig.kt
```diff
@@ -34,7 +45,14 @@ object MiskConfig {
     environment: Environment,
     overrideFiles: List<File> = listOf()
   ): T {
-    val mapper = ObjectMapper(YAMLFactory()).registerModules(KotlinModule(), JavaTimeModule())
+
+    checkConfigStructure(configClass)
+
+    val simpleModule = SimpleModule()
```
mmihic: super minor language pedantry, but kotlin has a nice `apply` syntax that lets you operate on this like a builder pattern

```kotlin
val simpleModule = SimpleModule().apply { addDeserializer(Secret::class.java, SecretDeserializer()) }
```

misk/src/main/kotlin/misk/config/Secret.kt
```diff
@@ -0,0 +1,44 @@
...
+ * limit: 5
+ * ```
+ *
+ * Lastly, this secret information would be accessed using:
+ * ```
+ * superSecretConfig.secret_information.value.answer_to_universe
+ * ```
+ */
+
+interface Secret<T>: Config{
```
mmihic: Do we necessarily want these to always be a top-level `Config`?

misk/src/main/kotlin/misk/config/SecretConfigService.kt
```diff
@@ -0,0 +1,25 @@
...
+import javax.inject.Provider
+import javax.inject.Singleton
+
+@Singleton
+class SecretConfigService internal constructor(
+  private val secretConfigLoaderProvider: Provider<SecretConfigLoader>
+) : AbstractIdleService(), DependentService {
+
+  override val consumedKeys = setOf<Key<*>>()
+  override val producedKeys = setOf<Key<*>>(SecretConfigService::class.toKey())
```
mmihic: This is where it gets a little strange for me - config is something I expect to be available during bootstrap, not at some point after the server has already started, and I think there will be a lot of confusion when someone tries to access a secret value prior to this service running (either because they have forgotten to declare a dependency on this service, or because they are in code that is outside of a service)

 I think ideally we'd load the secrets at the time a secret consumer calls `value`; I get that due to the dependency on `ResourceLoader` this has to happen after Guice binding time, but that still allows unfettered access to secrets at any point during server start without the need to interact with the service dependency graph.

misk/src/main/kotlin/misk/config/MiskConfig.kt
```diff
@@ -113,7 +131,58 @@ object MiskConfig {
...
+      val reference = ctxt?.readValue(p, String::class.java) as String
+      return DeferredSecret(reference, thisType)
+    }
+  }
+
+  class DeferredSecret<T>(val reference: String, val type: Class<T>) : Secret<T> {
+    var deferredValue: T? = null
+
+    override val value: T
+      get() = deferredValue!!
```
mmihic: We probably want to throw something other than `NullPointerException` if the deferred value is not null, so that it's clear that the caller tried to reference a secret before it was loaded.

mmihic: Also this isn't thread safe.

misk/src/main/kotlin/misk/config/MiskConfig.kt
```diff
@@ -113,7 +131,58 @@ object MiskConfig {
...
+
+    override fun createContextual(
+      ctxt: DeserializationContext?,
+      property: BeanProperty?
+    ): JsonDeserializer<*> {
+      thisType = property!!.type.bindings.getBoundType(0).rawClass
+      return this
+    }
+
+    override fun deserialize(p: JsonParser?, ctxt: DeserializationContext?): Secret<*> {
```
swankjesse: `p` -> `jsonParser`
`ctxt` -> `deserializationContext`

mmihic:
This is v cool. My main comment is re: when we want to load secrets.

misk/src/main/kotlin/misk/config/SecretConfigService.kt
swankjesse: Okay I think I know where we went sideways. We wanted to be able to decode secrets of arbitrary types: YAMLs, JSONs, etc.

I think we avoid a lot of complexity by saying secrets are only in two forms, YAMLs and Strings? Then we don‘t need the injector to provide us with a configured Moshi.

Well, maybe YAML, Strings, and PEMs. But not anything that requires Injector-layer configuration.

misk/src/main/kotlin/misk/config/MiskConfig.kt
```diff
@@ -115,4 +131,67 @@ object MiskConfig {
...
+      jsonParser: JsonParser,
+      deserializationContext: DeserializationContext
+    ): Secret<*>? {
+      if (type == null) {
+        // This only happens if ObjectMapper does not call createContextual fo this property.
+        throw JsonMappingException.from(jsonParser,
+            "Attempting to deserialize an object with no type")
+      }
+      val reference = deserializationContext.readValue(jsonParser, String::class.java) as String
+      return SecretImplementation(loadSecret(reference, type))
```
swankjesse: typically we call this `RealSecret`

misk/src/main/kotlin/misk/config/MiskConfig.kt
```diff
@@ -115,4 +131,67 @@ object MiskConfig {
...
+    }
+
+    private fun loadSecret(reference: String, type: JavaType): Any {
+      val source = requireNotNull(ResourceLoader.SYSTEM.utf8(reference)) {
```
swankjesse: nice!

misk/src/main/kotlin/misk/config/Secret.kt
```diff
@@ -0,0 +1,44 @@
...
+ * limit: 5
+ * ```
+ *
+ * Lastly, this secret information would be accessed using:
+ * ```
+ * superSecretConfig.secret_information.value.answer_to_universe
+ * ```
+ */
+
+interface Secret<T>{
```
swankjesse: nice!

PR #674, opened 2018-11-28
Delete Example Tab and misk-web documentation

adrw:
- Migrate both to `misk-web`
- misk/Example Tab -> misk-web/Palette Tab


adrw:
Yes, it's already in misk-web

PR #678, opened 2018-11-29
Increase ServiceManager wait time to 60s

ryanhall07:
This allows test services to pull remote docker images, which may take
awhile. This is obviously far from an exact science, but works in
practice.


ryanhall07:
we discussed over slack to add a timeout to `DependentService`. I think that is overkill and awkward for a couple of reasons.

1. This is only needed for testing since we wait forever in production and a slow service at the moment only exists in tests when attempting to pull an image.
2. the `ServiceManager` doesn't actually enforce per Service timeouts. So at best an individual timeout could be used to help calculate an aggregate timeout which is awkward.

another option is to add an optional param to `@MiskTest(timeout= 5)` if we really care about making it configurable and opt in for "slow" tests.

PR #682, opened 2018-11-30
ktlint misk

jjestrel:
Autoformat with ktlint -a -F and ignore violations that can't be autoresolved


swankjesse:
Nice. I’d love it if you could disable the “dangling newline at end of file” check. Having those isn’t something we currently spend any time thinking about but adding this tool means we’ll suddenly be spending energy on the least interesting whitespace there is.

PR #686, opened 2018-12-03
Add a simple misk-redis module

hamdanjaveed:
* Uses the `Jedis` redis client
* Permits 4 operations, `get`, `set`, `setex` and `del` for strings only


ryanhall07:
looks reasonable. you should create a `Fake` impl and some tests. also eventually you'll want to test the real with a test container that spins up a real redis

hamdanjaveed:
Added a `FakeRedis` along with a `FakeRedisTest`

misk-redis/src/main/kotlin/misk/redis/FakeRedis.kt
```diff
@@ -0,0 +1,52 @@
...
+import java.time.Instant
+import javax.inject.Inject
+
+class FakeRedis : Redis {
+  @Inject lateinit var clock: FakeClock
+
+  // The value type stored in our key-value store
+  private data class Value (
+    val value: String,
+    val expiryInstant: Instant = Instant.MAX
```
ryanhall07: is this a typical default for the real thing? you typically want your fakes to behave like the real thing

misk-redis/src/main/kotlin/misk/redis/FakeRedis.kt
```diff
@@ -0,0 +1,52 @@
...
+  )
+
+  // Acts as the Redis key-value store
+  private val keyValueStore = HashMap<String, Value>()
```
ryanhall07: you probably want this to be thread safe, since a test might use multiple threaads

ryanhall07: a simple `ConcurrentHashMap` is fine to use

ryanhall07: also FYI for the future. you typically want to use `mutableMap`, instead of `HashMap` directly. It's more idiomatic and It will use a `LinkedHashMap` under the hood, which is more test friendly. 

https://publicobject.com/2016/02/08/linkedhashmap-is-always-better-than-hashmap/

misk-redis/src/main/kotlin/misk/redis/FakeRedis.kt
```diff
@@ -0,0 +1,52 @@
...
+
+  override fun del(key: String): Long {
```
ryanhall07: why does `del` return 0 or 1? seems like a boolean is more appropriate

misk-redis/src/main/kotlin/misk/redis/FakeRedis.kt
ryanhall07: actually this is a private class, so probably doesn't really matter. nit: I would just remove the default since it's easy to set it explicitly

misk-redis/src/main/kotlin/misk/redis/Redis.kt
```diff
@@ -0,0 +1,19 @@
+package misk.redis
+
+/** A Redis client. */
+interface Redis {
+  /** Deletes a single key. */
```
ryanhall07: these methods should document what is being returned, especially when it's not totally obvious, like `del`

misk-redis/src/main/kotlin/misk/redis/Redis.kt
```diff
@@ -0,0 +1,19 @@
...
+  fun del(vararg keys: String): Long
+
+  /** Retrieves a key. */
+  fun get(key: String): String?
+
+  /** Sets a key. */
+  fun set(key: String, value: String): String
+
+  /** Sets a key with an expiration date. */
+  fun setex(key: String, seconds: Int, value: String): String
```
ryanhall07: consider using a higher order type like `Duration` for the expiration. that way somebody won't accidentally pass milliseconds or something

misk-redis/src/main/kotlin/misk/redis/RedisModule.kt
```diff
@@ -0,0 +1,14 @@
+package misk.redis
+
+import misk.inject.KAbstractModule
+import redis.clients.jedis.Jedis
+
+class RedisModule(val config: RedisConfig): KAbstractModule() {
+  override fun configure() {
+    val jedis = Jedis(config.host_name, config.port)
```
ryanhall07: does this actually attempt to bind to the redis? If so you'll want to move it out to a `Service`

ryanhall07: either way, you'll eventually want a `Service` to manage the lifecycle connection to the `Redis`

misk-redis/src/main/kotlin/misk/redis/FakeRedis.kt
hamdanjaveed: Yeah it's only used by `set` and `setex` which set the expiry appropriately, I'll remove the default

misk-redis/src/main/kotlin/misk/redis/FakeRedis.kt
hamdanjaveed: Good call, using `ConcurrentHashMap`

misk-redis/src/main/kotlin/misk/redis/FakeRedis.kt
hamdanjaveed: Its how Jedis' `del` command works (http://xetorthio.github.io/jedis/redis/clients/jedis/Jedis.html#del-java.lang.String-), it returns an integer telling you how many keys were deleted. I'll add a comment to make this clearer.

ryanhall07: can it ever be more than 1 if you pass a single key?

misk-redis/src/main/kotlin/misk/redis/RedisModule.kt
hamdanjaveed: Added a `RedisService` to manage the connection lifecycle for redis

misk-redis/src/main/kotlin/misk/redis/FakeRedis.kt
ryanhall07: k. to me a `fun del(key: String): boolean` makes more sense then

misk-redis/src/main/kotlin/misk/redis/RedisService.kt
```diff
@@ -0,0 +1,24 @@
...
+  private val redisProvider: Provider<Redis>
+) : AbstractIdleService() {
+  private lateinit var redis: Redis
+
+  override fun startUp() {
+    redis = redisProvider.get()
+  }
+
+  override fun shutDown() {
+    if (::redis.isInitialized) redis.close()
```
ryanhall07: you should move this check into the `redis.close()`. ideally `close()` should be idempotent

misk-redis/src/main/kotlin/misk/redis/FakeRedis.kt
```diff
@@ -0,0 +1,60 @@
...
+  private val keyValueStore = ConcurrentHashMap<String, Value>()
+
+  override fun del(key: String): Long {
+    if (!keyValueStore.containsKey(key)) return 0
+
+    keyValueStore.remove(key)
+    return 1
+  }
+
+  override fun del(vararg keys: String): Long {
```
mmihic: `Long` here is a bit weird - unlikely there will be greater than 2^32 keys, and `Int` doesn't require typecasting in most cases.

misk-redis/src/main/kotlin/misk/redis/Redis.kt
```diff
@@ -0,0 +1,55 @@
...
+   */
+  fun del(vararg keys: String): Long
+
+  /**
+   * Retrieves a key.
+   *
+   * @param key the key to retrieve
+   * @return a string value if the key was found, null if the key was not found
+   */
+  fun get(key: String): String?
```
mmihic: might be a good use case for [`operator get` and `operator set`](https://kotlinlang.org/docs/reference/operator-overloading.html#indexed)

misk-redis/src/main/kotlin/misk/redis/Redis.kt
```diff
@@ -0,0 +1,55 @@
...
+
+  /**
+   * Sets a key with an expiration date.
+   *
+   * @param key the key to set
+   * @param expiryDuration the amount of time before the key expires
+   * @param value the value to set
+   * @return the value that was set
+   */
+  fun setex(key: String, expiryDuration: Duration, value: String): String
```
mmihic: make the expiration an optional parameter on `set`?

misk-redis/src/main/kotlin/misk/redis/Redis.kt
```diff
@@ -0,0 +1,55 @@
...
+
+  /**
+   * Closes the connection to the redis instance.
+   */
+  fun close()
```
mmihic: Does this need to be exposed to application code? I'd think the need to close a is client specific to the `Jedis` implementation and handled for applications automatically by the `RedisService`, in which case this method should just be an `internal` method on `RealRedis`

misk-redis/src/main/kotlin/misk/redis/RedisService.kt
```diff
@@ -0,0 +1,24 @@
...
+ * Controls the connection lifecycle for Redis.
+ */
+@Singleton
+internal class RedisService @Inject internal constructor(
+  private val redisProvider: Provider<Redis>
+) : AbstractIdleService() {
+  private lateinit var redis: Redis
+
+  override fun startUp() {
+    redis = redisProvider.get()
```
mmihic: Why do we need to lazy init the `redis` instance?

ryanhall07: it connects to the instance on the Constructor

misk-redis/src/main/kotlin/misk/redis/FakeRedis.kt
hamdanjaveed: Agreed, changed to `Int`

misk-redis/src/main/kotlin/misk/redis/RedisService.kt
hamdanjaveed: Yeah, instantiating `Jedis` connects to the Redis instance so we can't initialize the property as we declare it, adding a comment to clarify

misk-redis/src/main/kotlin/misk/redis/Redis.kt
hamdanjaveed: Ah good point, `get` and `set` can be called with the indexing operator

misk-redis/src/main/kotlin/misk/redis/Redis.kt
hamdanjaveed: I opted for having `set` be an operator, which means we can't have default variables in the function signature. What do you think about keeping `setex` a separate function?

mmihic: I would still just call it `set` - you can have overloaded methods in Kotlin.

misk-redis/src/main/kotlin/misk/redis/Redis.kt
hamdanjaveed: You're right, this shouldn't be exposed to users of the client, I'll make it an `internal` method on `RealRedis`

misk-redis/src/main/kotlin/misk/redis/FakeRedis.kt
hamdanjaveed: No arguments against that, changed `del` to return `Boolean`

misk-redis/src/main/kotlin/misk/redis/Redis.kt
hamdanjaveed: Changed `setex` to `set`

PR #692, opened 2018-12-05
Don't upload artifacts for samples

ryanhall07:
We were never uploading these anyways, but a previous change in
#691 accidentally forced all
subprojects into uploading. So we now we need to explicitly opt out for
samples.


ryanhall07:
i'm going to merge this since it's fixing a bug I introduced and blocking our internal build

PR #693, opened 2018-12-06
Use env proxy for HttpClients

acostama:


ryanhall07:
I don't think we want this and should instead be using a `ProxySelector`.

```

    /**
     * Sets the HTTP proxy that will be used by connections created by this client. This takes
     * precedence over {@link #proxySelector}, which is only honored when this proxy is null (which
     * it is by default). To disable proxy use completely, call {@code proxy(Proxy.NO_PROXY)}.
     */
    public Builder proxy(@Nullable Proxy proxy) {
      this.proxy = proxy;
      return this;
    }
```

By default the system proxy selector is used, which seems like a reasonable default to me that uses env variables

```
     * <p>If unset, the {@link ProxySelector#getDefault() system-wide default} proxy selector will
     * be used.
```

PR #694, opened 2018-12-06
Don't bind RedisConfig in RedisModule

hamdanjaveed:
RedisConfig will already be bound if it's present in the service's config.

### DIFF
--- a/misk-redis/src/main/kotlin/misk/redis/RedisModule.kt
+++ b/misk-redis/src/main/kotlin/misk/redis/RedisModule.kt
@@ -6,9 +6,8 @@ import com.google.inject.Singleton
 import misk.inject.KAbstractModule
 import redis.clients.jedis.Jedis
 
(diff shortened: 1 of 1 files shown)


misk-redis/src/main/kotlin/misk/redis/RedisModule.kt
```diff
@@ -6,9 +6,8 @@ import com.google.inject.Singleton
 import misk.inject.KAbstractModule
 import redis.clients.jedis.Jedis
 
-class RedisModule(val config: RedisConfig): KAbstractModule() {
+class RedisModule : KAbstractModule() {
   override fun configure() {
-    bind<RedisConfig>().toInstance(config)
```
ryanhall07: did result in a double binding error? can we add a test that catches this bug?

hamdanjaveed: For sure, I'll make sure to add a test that uses this module to catch this

PR #695, opened 2018-12-10
Fix vitess version

jjestrel:
Set to `sha256:3cef0e042dee2312e5e0d80c56a6bf1581b52063ad0441551241f6785b110c38`


tirsen:
Shouldn't we also just pull only this image version? Wouldn't it otherwise fail?

PR #698, opened 2018-12-11
remove docker pull vitess from travis, since we pull during tests anyway

jjestrel:


.travis.yml
```diff
@@ -6,8 +6,7 @@ jdk:
 
 # Run Docker Pulls in parallel, esp for vitess/base with size > 1GB
```
ryanhall07: update the comment

acostama: Not too familiar with the travis structure but is this comment still relevant after this change?

PR #700, opened 2018-12-13
Bind Prometheus Collectors with a multiset

ryanhall07:
This allows other modules/libraries to contribute Collectors, without
having to inject the global CollectorRegistry. Seems a more natural way
to contribute a Collector.

### DIFF
--- a/misk/src/main/kotlin/misk/metrics/MetricsModule.kt
+++ b/misk/src/main/kotlin/misk/metrics/MetricsModule.kt
@@ -1,11 +1,26 @@
 package misk.metrics
 
+import io.prometheus.client.Collector
 import io.prometheus.client.CollectorRegistry
+import io.prometheus.client.hotspot.BufferPoolsExports
+import io.prometheus.client.hotspot.ClassLoadingExports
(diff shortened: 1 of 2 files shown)


misk/src/main/kotlin/misk/metrics/MetricsModule.kt
```diff
@@ -1,11 +1,26 @@
...
+import io.prometheus.client.hotspot.StandardExports
+import io.prometheus.client.hotspot.ThreadExports
+import io.prometheus.client.hotspot.VersionInfoExports
 import misk.inject.KAbstractModule
 
 internal class MetricsModule : KAbstractModule() {
   override fun configure() {
     bind<Metrics>().asEagerSingleton()
     bind<CollectorRegistry>().asEagerSingleton()
+    multibind<Collector>().toInstance(StandardExports())
```
wesleyk: so we expect other modules to start to bind their own collectors? But these are considered "standard" enough to live here?

ryanhall07: yea. they are all jvm metrics, so every misk app will want them

PR #705, opened 2018-12-18
Use jetty alpn that relies on java9+

jjestrel:
This makes us java 8 incompatible since jetty-alpn-java-server is built on java 9. In the spirit for staying on LTS versions, we'll jump from 8 to 11.


misk/src/main/kotlin/misk/web/jetty/JettyService.kt
```diff
@@ -98,11 +99,13 @@ class JettyService @Inject internal constructor(
       val httpsConfig = HttpConfiguration(httpConfig)
       httpsConfig.addCustomizer(SecureRequestCustomizer())
 
-      val ssl = SslConnectionFactory(sslContextFactory, "http/1.1")
+      val alpn = ALPNServerConnectionFactory("h2", "http/1.1")
```
ryanhall07: what is going on here? looks like a real change

jjestrel: It's a revert of this: https://github.com/square/misk/pull/703/files#diff-b02f31595f44a790e04711ffe3a62570 same code on this side even though we aren't using conscrypt, just different dependency

PR #706, opened 2018-12-18
Bump misk to java 11

jjestrel:


build.gradle
```diff
@@ -52,13 +52,13 @@ subprojects {
   compileKotlin {
     kotlinOptions {
       jvmTarget = "1.8"
-      allWarningsAsErrors = true
+      // TODO(jayestrella): re-add allWarningsAsErrors = true when we (+okio) bump kotlin versions
```
jjestrel: Java 11 has `String.isBlank` and `String.repeat` which is causing `"This member is not fully supported by Kotlin compiler, so it may be absent or have different signature in next major version"` as a warning. We can bump kotlin versions to 1.3 and it handles this, but then we get warnings from okio for enabling experimental coroutines on version 1.3 (since okio is on 1.2).
see also: https://jar-download.com/artifacts/org.jetbrains.kotlin/kotlin-compiler-embeddable/1.1.3-2/source-code/org/jetbrains/kotlin/load/kotlin/JvmBuiltInsSettings.kt

tl;dr warnings because versions are hard.

PR #707, opened 2018-12-19
Bind CuratorFramework as Singleton in production

ryanhall07:
@Singleton on the provider does not provide a Singleton. Also shared the
same module between tests and prod to prevent a similar bug in the
future. (tests were working because it was bound as a singleton)


misk-zookeeper/src/test/kotlin/misk/clustering/zookeeper/ZkTestModule.kt
```diff
@@ -17,14 +17,11 @@ import javax.inject.Singleton
...
     bind<ZookeeperConfig>().toInstance(ZookeeperConfig(zk_connect = "127.0.0.1:$zkPortKey"))
-    bind<CuratorFramework>().toProvider(CuratorFrameworkProvider::class.java).asSingleton()
-    bind<LeaseManager>().to<ZkLeaseManager>()
     bind<String>().annotatedWith<AppName>().toInstance("my-app")
 
-    multibind<Service>().to<ZkService>()
     multibind<Service>().to<StartZookeeperService>()
-    multibind<Service>().to<ZkLeaseManager>()
     install(FakeClusterModule())
+    install(ZookeeperModule())
```
wesleyk: nice

misk-zookeeper/src/main/kotlin/misk/clustering/zookeeper/ZookeeperModule.kt
```diff
@@ -4,14 +4,15 @@ import com.google.common.util.concurrent.Service
...
+import misk.inject.asSingleton
 import org.apache.curator.framework.CuratorFramework
 
 class ZookeeperModule : KAbstractModule() {
   override fun configure() {
     multibind<Service>().to<ZkService>()
     multibind<Service>().to<ZkLeaseManager>()
     bind<LeaseManager>().to<ZkLeaseManager>()
-    bind<CuratorFramework>().toProvider(CuratorFrameworkProvider::class.java)
+    bind<CuratorFramework>().toProvider(CuratorFrameworkProvider::class.java).asSingleton()
```
wesleyk: ah that's subtle...

PR #710, opened 2018-12-20
Misk module for easy jobqueue install

acostama:


misk-aws/src/main/kotlin/misk/jobqueue/sqs/MiskJobModule.kt
```diff
@@ -0,0 +1,124 @@
...
+ *  @Inject @JobQueueA lateinit var jobQueueA: MiskJobQueue`
+ *  ```
+ */
+
+class MiskJobModule(private val environment: Environment) : KAbstractModule() {
+  private lateinit var jobQueueProvider: Provider<JobQueue>
+  private val jobQueuesMap =
+      mutableListOf<JobQueueBinding>()
+
+  inline fun <reified T : JobHandler> withHandler(
```
mmihic: can we make the annotation be a `reified` type parameter as well?

misk-aws/src/main/kotlin/misk/jobqueue/sqs/MiskJobModule.kt
```diff
@@ -0,0 +1,124 @@
...
+
+/**
+ * Simple wrapper to avoid carrying [QueueName] everywhere.
+ *
+ * Assuming a jobqueue is bound to annotation @DinoQueue. Inject using the annotation:
+ *
+ *  `@Inject @DinoQueue lateinit var dinoQueue: MiskJobQueue`
+ *
+ */
+interface MiskJobQueue {
```
mmihic: Rather than introducing a new interface, maybe we should just push the `QueueName` beneath `JobQueue` and leave that as the main interface. Having two slightly similar interfaces is likely to cause confusion.

misk-aws/src/main/kotlin/misk/jobqueue/sqs/MiskJobService.kt
```diff
@@ -0,0 +1,24 @@
...
+
+import com.google.common.util.concurrent.AbstractIdleService
+import misk.jobqueue.JobConsumer
+import misk.jobqueue.JobHandler
+import misk.jobqueue.QueueName
+import javax.inject.Provider
+import javax.inject.Singleton
+
+@Singleton
+class MiskJobService constructor(
```
mmihic: This isn't specific to sqs - move to the general `misk-jobs` module?

misk-aws/src/main/kotlin/misk/jobqueue/sqs/MiskJobModule.kt
acostama: yes! didn't know this was possible lol

misk-aws/src/main/kotlin/misk/jobqueue/sqs/MiskJobModule.kt
```diff
@@ -0,0 +1,124 @@
...
+    configureJobQueues()
+    val boundHandlers = bindHandlers()
+
+    bind(jobQueueServiceKey).toProvider(Provider<MiskJobService> {
+      MiskJobService(jobConsumerProvider.get(), boundHandlers)
+    }).asSingleton()
+
+    multibind<Service>().to(jobQueueServiceKey)
+
+    if (environment == Environment.DEVELOPMENT || environment == Environment.TESTING) {
```
mmihic: This will prevent applications from installing multiple `MiskJobQueueModule`s, which in turn prevents you from using `MiskJobQueue` and job consumer auto-registration from within components that might get installed by the application. The pattern we've generally taken in misk is to have one modules for global singletons which gets installed by the application, and another module for registering "handlers" that get used by those global singletons.  That way a component provider can provide a top-level module that installs the component specific handlers, without it also installing duplicate bindings for the global singletons.

cf. `MiskWebModule`

misk-aws/src/main/kotlin/misk/jobqueue/sqs/FakeAwsSqsModule.kt
```diff
@@ -0,0 +1,108 @@
...
+}
+
+/**
+ * Fake for testing and development. Doesn't handle [misk.jobqueue.TransactionalJobQueue]s
+ */
+
+@Singleton
+class FakeSqs @Inject internal constructor(
+  private val tokenGenerator: TokenGenerator
+) : AbstractScheduledService(), JobQueue, JobConsumer {
```
mmihic: Usually with fakes like this, you want to have explicit control over when tasks get dequeued and processed so that you can write tests that are deterministic and avoid race conditions.  There's a class `ExplicitReleaseBlockingQueue` specifically for this purpose; I'd recommend having the `FakeSqs` use this blocking queue and expose a method to explicit release 1 / n / all of the jobs currently on the queue. That way tests have full control over when jobs will run.

PR #712, opened 2018-12-27
Allows hibernate config SHOW_SQL to be configurable via DataSourceConfig

frojasg:


misk-hibernate/src/main/kotlin/misk/hibernate/SessionFactoryService.kt
```diff
@@ -90,7 +90,7 @@ internal class SessionFactoryService(
     registryBuilder.run {
       applySetting(AvailableSettings.DATASOURCE, dataSource.get())
       applySetting(AvailableSettings.DIALECT, config.type.hibernateDialect)
-      applySetting(AvailableSettings.SHOW_SQL, "false")
+      applySetting(AvailableSettings.SHOW_SQL, config.show_sql)
```
wesleyk: the logged output is stripped of query params, correct? Just making sure we don't beginning potentially logging PII

frojasg: that's correct.

PR #713, opened 2018-12-27
Fixes SchemaValidator breadcrumb by removing last element

frojasg:


misk-hibernate/src/main/kotlin/misk/hibernate/SchemaValidator.kt
```diff
@@ -237,7 +237,7 @@ internal class SchemaValidator {
     try {
       lambda()
     } finally {
-      pathBreadcrumbs.dropLast(1)
+      pathBreadcrumbs.removeAt(pathBreadcrumbs.size - 1)
```
ryanhall07: I'm not following what the difference between these 2 methods. can you add a test to show the bug being fixed?

frojasg: `dropLast` returns a new list instead of mutating `pathBreadcrumbs`. 

I changed an existing test to verify the path was fixed. Before this change this was the path generated
```
ERROR at schemavalidation.bad_identifier_table.id.tbl6_camel_case.tbl6_database_camelcase.tbl6_hibernate_camelcase.tbl6_lower_hyphen.tbl6_mixed_underscore.tbl6_upper_underscore.missing_columns_table.id.tbl4_int_column.tbl4_string_column.nullable_mismatch_table.id.tbl5_both_notnull.tbl5_both_null.tbl5_database_null.tbl5_hibernate_null:
  Column tbl5_hibernate_null is NOT NULL in database but tbl5_hibernate_null is nullable in hibernate
```
I could create a new test to try to make this more obvious, but the path is generated only when there's an error and check for a random path without checking the error was not very intuitive for me.

ryanhall07: ah got it

PR #714, opened 2018-12-27
SchemaValidator read default value of the column

frojasg:
Changes: 

* SchemaValidator accepts nullable hibernate columns if the column has a default. This change is useful for fields that are set by Misk listeners, like `created_at` and `updated_at`.
* We should ignore fields that are unknown by Hibernate, that doesn't break the insertion or updates. This change is useful to avoid coupling between application deployments and database schema migrations.

### DIFF
--- a/misk-hibernate/src/main/kotlin/misk/hibernate/SchemaValidator.kt
+++ b/misk-hibernate/src/main/kotlin/misk/hibernate/SchemaValidator.kt
@@ -136,11 +136,8 @@ internal class SchemaValidator {
       "Database table \"${dbTable.name}\" is missing columns ${hibernateOnly.map { it.name }} found in hibernate \"${hibernateTable.name}\""
     }
 
-    // TODO (maacosta) how strict should we be here? If `DEFAULT NULL` column exists in the Db and hibernate
-    //                does not know about it hibernate can still do writes to db. However if we are going to do
-    //                lookups on this column it might not work. Do we look at queries?.
-    validate(dbOnly.isEmpty()) {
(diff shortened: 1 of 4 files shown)


misk-hibernate/src/main/kotlin/misk/hibernate/SchemaValidator.kt
```diff
@@ -134,11 +136,8 @@ internal class SchemaValidator {
       "Database table \"${dbTable.name}\" is missing columns ${hibernateOnly.map { it.name }} found in hibernate \"${hibernateTable.name}\""
     }
 
-    // TODO (maacosta) how strict should we be here? If `DEFAULT NULL` column exists in the Db and hibernate
-    //                does not know about it hibernate can still do writes to db. However if we are going to do
-    //                lookups on this column it might not work.. do we look at queries?.
-    validate(dbOnly.isEmpty()) {
-      "Hibernate entity \"${hibernateTable.name}\" is missing columns ${dbOnly.map { it.name }} expected in table \"${dbTable.name}\""
+    validate(dbOnly.isEmpty() || dbOnly.all { it.hasDefaultValue || it.nullable }) {
```
swankjesse: cool!

misk-hibernate/src/main/kotlin/misk/hibernate/SchemaValidator.kt
```diff
@@ -157,8 +156,8 @@ internal class SchemaValidator {
     }
 
     // We have that the hibernate column only needs to be null if the database is null.
-    // It's okay if hibernate is more strict.
-    validate(dbColumn.nullable || !hibernateColumn.nullable) {
+    // It's okay if hibernate is more strict. However, we shouldn't care that much if the column has a default value
```
swankjesse: nice comment

PR #715, opened 2019-01-02
 Bump okhttp and update unix socket factories to latest okhttp samples

wesleyk:
r @ryanhall07 @swankjesse @nbroyles


swankjesse:
I’m also going to try to upstream these to the jnr-unixsocket library.

PR #717, opened 2019-01-04
EnvVarLoader interface

ryanhall07:
This allows loading Environment Variables from the system or from an in
memory map for tests.

### DIFF
--- a/misk-aws/src/main/kotlin/misk/cloud/aws/AwsEnvironmentModule.kt
+++ b/misk-aws/src/main/kotlin/misk/cloud/aws/AwsEnvironmentModule.kt
@@ -1,21 +1,30 @@
 package misk.cloud.aws
 
+import com.google.inject.Provides
+import misk.environment.EnvVarLoader
+import misk.environment.ForEnvVars
 import misk.inject.KAbstractModule
-import java.lang.System.getenv
 
 /** [AwsEnvironmentModule] pulls region and account information from installed env vars */
-class AwsEnvironmentModule(
-  private val region: String = getRequiredEnv("REGION"),
-  private val accountId: String = getRequiredEnv("ACCOUNT_ID")
-) : KAbstractModule() {
-  override fun configure() {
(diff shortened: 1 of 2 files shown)


misk-aws/src/main/kotlin/misk/cloud/aws/AwsEnvironmentModule.kt
```diff
@@ -1,21 +1,30 @@
...
+}
 
-  companion object {
-    private fun getRequiredEnv(name: String): String {
-      return getenv(name) ?: throw IllegalStateException("$name env var not set")
-    }
+/**
+ * [FakeAwsEnvironmentModule] pulls region and account information from an in memory map.
+ */
+class FakeAwsEnvironmentModule : KAbstractModule() {
```
wesleyk: so we'll install this in our `SkimFakesModule`?

misk-aws/src/main/kotlin/misk/cloud/aws/AwsEnvironmentModule.kt
```diff
@@ -1,21 +1,30 @@
...
+  override fun configure() {
+    newMapBinder<String, String>(ForEnvVars::class)
+        .addBinding("REGION").toInstance("us-east-1")
+    newMapBinder<String, String>(ForEnvVars::class)
+        .addBinding("ACCOUNT_ID").toInstance("8675309")
```
wesleyk: nice

nbroyles: Classic

misk-aws/src/main/kotlin/misk/cloud/aws/AwsEnvironmentModule.kt
ryanhall07: yea probably. although my first use case is to use in the `InjectorTest` when creating injectors for all envs.

PR #721, opened 2019-01-04
Split MiskServiceModule into real and testing

ryanhall07:
This allows unit testing envs to use MiskTestingServiceModule, which
does not rely on environment vars or the underlying filesystem.

### DIFF
--- a/misk/src/main/kotlin/misk/MiskTestingServiceModule.kt
+++ b/misk/src/main/kotlin/misk/MiskTestingServiceModule.kt
@@ -18,24 +18,59 @@ import com.google.inject.spi.InstanceBinding
 import com.google.inject.spi.LinkedKeyBinding
 import com.google.inject.spi.ProviderInstanceBinding
 import com.google.inject.util.Types
+import misk.environment.FakeEnvVarModule
+import misk.environment.RealEnvVarModule
 import misk.healthchecks.HealthCheck
 import misk.inject.KAbstractModule
 import misk.metrics.MetricsModule
 import misk.moshi.MoshiModule
 import misk.prometheus.PrometheusHistogramRegistryModule
 import misk.resources.ResourceLoaderModule
+import misk.resources.TestingResourceLoaderModule
 import misk.time.ClockModule
 import misk.tokens.TokenGeneratorModule
 import javax.inject.Singleton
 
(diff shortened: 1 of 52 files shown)


misk/src/main/kotlin/misk/MiskTestingServiceModule.kt
```diff
@@ -18,24 +18,51 @@ import com.google.inject.spi.InstanceBinding
...
 import misk.prometheus.PrometheusHistogramRegistryModule
 import misk.resources.ResourceLoaderModule
+import misk.resources.TestingResourceLoaderModule
 import misk.time.ClockModule
 import misk.tokens.TokenGeneratorModule
 import javax.inject.Singleton
 
-class MiskServiceModule : KAbstractModule() {
+/**
+ * [MiskRealServiceModule] should be installed in real environments.
```
ryanhall07: this is the real change. the rest is just changing existing code to use either Real or Testing

misk/src/main/kotlin/misk/MiskTestingServiceModule.kt
```diff
@@ -18,24 +18,51 @@ import com.google.inject.spi.InstanceBinding
...
+class MiskRealServiceModule : KAbstractModule() {
+  override fun configure() {
+    install(ResourceLoaderModule())
+    install(RealEnvVarModule())
+    install(MiskCommonServiceModule())
+  }
+}
+
+/**
+ * [MiskTestingServiceModule] should be installed in unit testing environments.
```
wesleyk: does this suggest that there are tests that would use `MiskRealServiceModule`? Can't think of any that we would actually use in practice.

wesleyk: perhaps some sort of true end-to-end system test that actually provides a real env.

ryanhall07: I don't think so. It would be too much a hassle to setup a "real env" on your host machine (unless you were using docker)

ryanhall07: I added some more javadoc to help explain this should only fake out things about the env, not include application layer fakes

PR #724, opened 2019-01-05
Add tracing on BEGIN; COMMIT; ROLLBACK;

jjestrel:
Add more tracing

### DIFF
--- a/misk-hibernate/src/test/kotlin/misk/hibernate/TransacterTest.kt
+++ b/misk-hibernate/src/test/kotlin/misk/hibernate/TransacterTest.kt
@@ -4,6 +4,9 @@ import io.opentracing.mock.MockTracer
 import io.opentracing.tag.Tags
 import misk.exceptions.UnauthorizedException
 import misk.hibernate.RealTransacter.Companion.APPLICATION_TRANSACTION_SPAN_NAME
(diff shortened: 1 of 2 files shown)


misk-hibernate/src/test/kotlin/misk/hibernate/TransacterTest.kt
```diff
@@ -374,16 +377,30 @@ class TransacterTest {
...
-    assertThat(tracer.finishedSpans().get(1).tags())
+
+    assertThat(orderedSpans.get(2).operationName())
+        .isEqualTo(DB_BEGIN_SPAN_NAME)
+    assertThat(orderedSpans.get(2).tags())
+        .containsEntry(Tags.COMPONENT.getKey(), TRANSACTER_SPAN_TAG)
+
+
+    assertThat(orderedSpans.get(3).operationName())
+        .isEqualTo(if (committed) { DB_COMMIT_SPAN_NAME } else { DB_ROLLBACK_SPAN_NAME })
```
wesleyk: not a fan of conditionals inside the actual equal assertion. How about moving the conditional out, and having two assertThat's instead?

jjestrel: Changed

PR #725, opened 2019-01-07
Propagate Exception with jdbc url for debugging

ryanhall07:


misk-hibernate/src/main/kotlin/misk/jdbc/PingDatabaseService.kt
```diff
@@ -9,7 +9,7 @@ import misk.backoff.retry
 import misk.environment.Environment
 import misk.inject.toKey
 import java.time.Duration
-import java.util.*
+import java.util.Properties
```
jjestrel: Nice

PR #728, opened 2019-01-08
Revert "Require DatasourceConfig host and port"

ryanhall07:
This reverts commit c015e22be631389f61df445ecde160ed32c1411e.


ryanhall07:
This ended up breaking sq apps internally, since we don't require them to provide the host/port for staging/prod, since we automatically provision the config in k8s.  I still think we want to do this, but need to think of a better way. For now reverting

PR #732, opened 2019-01-09
Add namespace support to SchemaMigrator

drashkov:
If .sql migrations are found in a subdirectory, the subdirectory path becomes the namespace. 

Addresses https://github.com/square/misk/issues/636


misk-hibernate/src/main/kotlin/misk/hibernate/SchemaMigrator.kt
```diff
@@ -28,26 +23,55 @@ private val logger = getLogger<SchemaMigrator>()
...
+ * Applied schema migrations are tracked in the database in a `schema_versions` table. Migrations may
  * be applied either by [SchemaMigrator.applyAll] or manually. When you applying schema changes
- * manually you must add a row to the `schema_version` table to record which version was applied.
+ * manually you must add a row to the `schema_versions` table to record which version was applied.
  */
+
+data class NamedspacedMigration(val namespace: String, val version: Int) {
+  // We don't store the path from which schema changes came, so we don't use it for comparison
+  // between what's in the database and what's available.
+  var path = ""
```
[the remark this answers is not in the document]
drashkov: Yeah, for data classes only the parameters in the primary constructor are used in the generated methods, so using this field gets ignored.

misk-hibernate/src/main/kotlin/misk/hibernate/SchemaMigrator.kt
```diff
@@ -28,26 +23,55 @@ private val logger = getLogger<SchemaMigrator>()
...
+          }
+        }
+      }
+    }
+    return resources
+  }
+
   /** Returns a map from version to path. */
-  fun availableMigrations(): NavigableMap<Int, String> {
+  private fun availableMigrations(): Set<NamedspacedMigration> {
```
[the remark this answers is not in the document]
drashkov: Yes, I will change this to a sorted set and explicitly sort the migrations.

misk-hibernate/src/main/kotlin/misk/hibernate/SchemaMigrator.kt
```diff
@@ -60,59 +84,81 @@ internal class SchemaMigrator(
...
             |  version varchar(50) NOT NULL,
             |  installed_by varchar(30) DEFAULT NULL,
-            |  UNIQUE KEY (version)
+            |  UNIQUE KEY (namespace, version)
             |);
             |""".trimMargin())
         statement.executeBatch()
       }
+
+      // Temporary migration to new schema - can be removed after all services upgraded.
```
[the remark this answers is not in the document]
drashkov: Yeah, it though it would be roughly similar, since it would have to be conditional alter, which has to be either a stored procedure or code like this.

wesleyk:
hmm, what's the rollout strategy here, given that the current schema_version schema is baked into our container contract?

misk-hibernate/src/main/kotlin/misk/hibernate/SchemaMigrator.kt
```diff
@@ -23,31 +20,105 @@ private val logger = getLogger<SchemaMigrator>()
...
+      return version.compareTo(other.version)
+    } else {
+      return namespace.compareTo(other.namespace)
+    }
+  }
+
+  fun toNamespacedVersion() = "$namespace$version"
+
+  companion object {
+    fun fromNamespacedVersion(namespacedVersion: String): NamedspacedMigration {
```
swankjesse: could you doc a couple example strings?

misk-hibernate/src/main/kotlin/misk/hibernate/SchemaMigrator.kt
```diff
@@ -23,31 +20,105 @@ private val logger = getLogger<SchemaMigrator>()
...
+    fun fromResourcePath(resource: String, migrationsResource: String): NamedspacedMigration {
+      val matcher = MIGRATION_PATTERN.matcher(resource)
+      require(matcher.matches()) { "unexpected resource: $resource" }
+      val cleanNamespace = matcher.group(1).removePrefix(migrationsResource).removePrefix("/")
+      val namedspacedMigration = NamedspacedMigration(cleanNamespace, matcher.group(2).toInt())
+      namedspacedMigration.path = resource
+      return namedspacedMigration
+    }
+
+    /** Matches file names like `exemplar/migrations/v100__exemplar.sql`. */
```
swankjesse: nice

misk-hibernate/src/main/kotlin/misk/hibernate/SchemaMigrator.kt
```diff
@@ -23,31 +20,105 @@ private val logger = getLogger<SchemaMigrator>()
...
+    val MIGRATION_PATTERN = Pattern.compile("(.*/?)v(\\d+)__[^/]+\\.sql")!!
```
swankjesse: I think you need a slash preceding the `v`

I think this matches: `foo/luv1__franklin.sql`

swankjesse: ... looks like this bug was already here (my fault!)

misk-hibernate/src/test/kotlin/misk/hibernate/SchemaMigratorTest.kt
```diff
@@ -143,35 +176,52 @@ internal class SchemaMigratorTest {
...
+    resourceLoader.put("${config.data_source.migrations_resources!![0]}/v1001__bar.sql", """
+        |CREATE TABLE table_2 (name varchar(255))
+        |""".trimMargin())
+
+    assertThat(assertFailsWith<IllegalArgumentException> {
+      schemaMigrator.requireAll()
+    }).hasMessageContaining("Duplicate migrations found")
   }
 
   @Test fun resourceVersionParsing() {
```
swankjesse: nice

misk-hibernate/src/test/kotlin/misk/hibernate/SchemaMigratorTest.kt
```diff
@@ -110,31 +122,52 @@ internal class SchemaMigratorTest {
...
+    schemaMigrator.applyAll("SchemaMigratorTest", sortedSetOf(
+        NamedspacedMigration(1001),
+        NamedspacedMigration(1002),
+        NamedspacedMigration("name/space/", 1001)))
+    assertThat(schemaMigrator.appliedMigrations()).containsExactly(
+        NamedspacedMigration(1001),
+        NamedspacedMigration(1002),
+        NamedspacedMigration(1003),
+        NamedspacedMigration(1004),
+        NamedspacedMigration("name/space/", 1001),
```
ryanhall07: does the namespace provide any value other than allowing migrations with the same name to avoid conflicts?

drashkov: No, it's just libraries and projects can keep their own independent version 1..n without copying or conflicts

misk-hibernate/src/test/kotlin/misk/hibernate/SchemaMigratorTest.kt
```diff
@@ -110,31 +122,52 @@ internal class SchemaMigratorTest {
...
+    resourceLoader.put("${mainSource}/v1003__movies.sql", """
         |CREATE TABLE table_3 (name varchar(255))
         |""".trimMargin())
-    resourceLoader.put("${config.data_source.migrations_resource}/v1004__movies.sql", """
+    resourceLoader.put("${mainSource}/v1004__movies.sql", """
         |CREATE TABLE table_4 (name varchar(255))
         |""".trimMargin())
-    schemaMigrator.applyAll("SchemaMigratorTest", setOf(1001, 1002))
-    assertThat(schemaMigrator.appliedMigrations()).containsExactly(1001, 1002, 1003, 1004)
+    resourceLoader.put("${mainSource}/namespace/v1001__props.sql", """
```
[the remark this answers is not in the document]
drashkov: It's supposed to show a library with a merged migrations path. I'll rename it for clarity to `merged_library_table`

misk-hibernate/src/main/kotlin/misk/hibernate/SchemaMigrator.kt
drashkov: I'll add a test case for it

misk-hibernate/src/main/kotlin/misk/hibernate/SchemaMigrator.kt
```diff
@@ -23,31 +20,105 @@ private val logger = getLogger<SchemaMigrator>()
...
+
+  override fun compareTo(other: NamedspacedMigration): Int {
+    if (namespace == other.namespace) {
+      return version.compareTo(other.version)
+    } else {
+      return namespace.compareTo(other.namespace)
+    }
+  }
+
+  fun toNamespacedVersion() = "$namespace$version"
```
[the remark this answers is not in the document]
drashkov: I agree it'll look a bit cleaner, but I think this form plays better with the migration tools, so I'd like to keep it.

PR #734, opened 2019-01-11
Add support for using ZooKeeper with mTLS

nbroyles:


misk-zookeeper/src/main/kotlin/misk/clustering/zookeeper/CuratorFrameworkProvider.kt
```diff
@@ -12,6 +12,15 @@ internal class CuratorFrameworkProvider @Inject internal constructor(
 ) : Provider<CuratorFramework> {
 
   override fun get(): CuratorFramework {
+    if (config.cert_store != null && config.trust_store != null) {
```
ryanhall07: I think we should require either both are set or neither are set

misk-zookeeper/src/main/kotlin/misk/clustering/zookeeper/CuratorFrameworkProvider.kt
```diff
@@ -12,6 +12,15 @@ internal class CuratorFrameworkProvider @Inject internal constructor(
...
+      System.setProperty("zookeeper.clientCnxnSocket", "org.apache.zookeeper.ClientCnxnSocketNetty")
```
ryanhall07: you _have_ to configure via system properties? you can't configure the zkclient directly?

misk-zookeeper/src/main/kotlin/misk/clustering/zookeeper/CuratorFrameworkProvider.kt
```diff
@@ -12,6 +12,15 @@ internal class CuratorFrameworkProvider @Inject internal constructor(
...
+      System.setProperty("zookeeper.client.secure", "true")
+      System.setProperty("zookeeper.ssl.keyStore.location", config.cert_store.resource)
+      System.setProperty("zookeeper.ssl.keyStore.password", config.cert_store.passphrase ?: "changeit")
```
ryanhall07: I don't think we should promote this as a default value in misk, even though we do it in SQ.

misk-zookeeper/src/main/kotlin/misk/clustering/zookeeper/CuratorFrameworkProvider.kt
```diff
@@ -12,6 +12,15 @@ internal class CuratorFrameworkProvider @Inject internal constructor(
...
+      System.setProperty("zookeeper.ssl.trustStore.location", config.trust_store.resource)
+      System.setProperty("zookeeper.ssl.trustStore.password", config.trust_store.passphrase ?: "changeit")
```
ryanhall07: make this default value on the Config object instead?

misk-zookeeper/src/main/kotlin/misk/clustering/zookeeper/CuratorFrameworkProvider.kt
ryanhall07: from a quick search, I _think_ you can configure it via the `ZkClientConfig` using the `setProperty`

misk-zookeeper/src/main/kotlin/misk/clustering/zookeeper/CuratorFrameworkProvider.kt
nbroyles: Why not? This is java's default password: https://docs.oracle.com/cd/E19798-01/821-1841/gjrgy/

ryanhall07: but is that really a best practice default? at the very least if we go this route, we should clearly document why we believe `changeit` is fine as default. at first glance to a developer it would seem extremely insecure. for example it is recommended to only us `changeit` if you store your keystore (or keystore contents)  in a secure manner and a password would just be redundant anyways.

misk-zookeeper/src/main/kotlin/misk/clustering/zookeeper/CuratorFrameworkProvider.kt
nbroyles: Yeah, I looked into this for a bit previously and the problem is that Curator wraps the ZK client and doesn't give you anyway to set these configuration options.

misk-zookeeper/src/main/kotlin/misk/clustering/zookeeper/CuratorFrameworkProvider.kt
ryanhall07: well we could have a `JksTrustStoreConfig` that provides this default. anywhere we load a jks trust store is going to have this logic

misk-zookeeper/src/main/kotlin/misk/clustering/zookeeper/CuratorFrameworkProvider.kt
nbroyles: Oh, I misunderstood what your initial question was in reference to. Yeah, I'll just move this to the config file itself and not provide a default here.

misk-zookeeper/src/main/kotlin/misk/clustering/zookeeper/CuratorFrameworkProvider.kt
ryanhall07: it looks like you can set a `ZookeeperFactory` when building the `CuratorFrameworkFactory`. This should allow configuring the `ZkClientConfig`

misk-zookeeper/src/main/kotlin/misk/clustering/zookeeper/CuratorFrameworkProvider.kt
nbroyles: (Default here being a default in Misk, leaving the client to specify what password they want via their config)

misk-zookeeper/src/main/kotlin/misk/clustering/zookeeper/CuratorFrameworkProvider.kt
ryanhall07:
```
    return CuratorFrameworkFactory.builder()
        .connectString(config.zk_connect)
        .zookeeperFactory { connectString, sessionTimeout, watcher, canBeReadOnly ->
          val zk = DefaultZookeeperFactory().newZooKeeper(connectString, sessionTimeout, watcher, canBeReadOnly)
          zk.clientConfig.setProperty("foo", "bar")
          zk
        }
```

misk-zookeeper/src/main/kotlin/misk/clustering/zookeeper/CuratorFrameworkProvider.kt
nbroyles: Eh, feels overkill and then odd that we have a JksTrustStore config and then a general TrustStore config where you set a format.

misk-zookeeper/src/main/kotlin/misk/clustering/zookeeper/CuratorFrameworkProvider.kt
nbroyles: Hah, wow. Nice find.

misk-zookeeper/src/main/kotlin/misk/clustering/zookeeper/CuratorFrameworkProvider.kt
```diff
@@ -12,6 +13,15 @@ internal class CuratorFrameworkProvider @Inject internal constructor(
 ) : Provider<CuratorFramework> {
 
   override fun get(): CuratorFramework {
+    var tlsEnabled = false
+    if ((config.cert_store != null && config.trust_store == null) ||
+        (config.cert_store == null && config.trust_store != null)) {
+      throw IllegalArgumentException("only one of ZookeeperConfig cert_store and trust_store" +
```
ryanhall07: nit: can use `require` instead of if-throw

ryanhall07: nit: use `require`

PR #736, opened 2019-01-14
Remove Producer.Transaction interface

mightyguava:
Removing the `Producer.Transaction` interface and publishing events directly. We will offer `SpooledPublisher` for publishing events atomically, and must be backed by a data store. `Producer.Transaction` is limited in that while the events are published as one atomic block, they are not a part of a database transaction, and thus only provide partial atomicity when used as a step business logic that modifies database state. We prefer only one way of supporting atomic transactional event publishes.

### DIFF
--- a/misk-events/src/main/kotlin/misk/events/Producer.kt
+++ b/misk-events/src/main/kotlin/misk/events/Producer.kt
@@ -5,30 +5,10 @@ package misk.events
  */
 interface Producer {
   /**
-   * A producer [Transaction] is a unit of publishing. Producer transactions adhere to the
-   * transactional concepts of isolation (events publishing within a transaction are only
-   * visible to consumers once the transaction is committed) and atomicity (once committed, all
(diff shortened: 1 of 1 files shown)


misk-events/src/main/kotlin/misk/events/Producer.kt
```diff
@@ -5,30 +5,10 @@ package misk.events
...
-   * as part of the local database transaction.
-   *
-   * Transactions remaining outstanding until the application calls commit, or rollback, or until
-   * a producer specific timeout occurs.
-   *
-   * Transactions may only be used from the thread that began them; accessing a transaction
-   * in multiple threads or handing it off between threads has undefined behavior.
-   *
-   * Having multiple outstanding transactions on the same thread also has undefined behavior.
+   * If multiple events need to be produced published atomically, or if end-to-end ordering is required, use a
```
ryanhall07: nit: remove produced or published in the sentence

PR #737, opened 2019-01-14
Use mTLS with ZK in tests

nbroyles:


misk-zookeeper/src/test/kotlin/misk/clustering/zookeeper/ZkTestModule.kt
```diff
@@ -9,11 +9,19 @@ import misk.config.AppName
...
 import misk.logging.getLogger
+import misk.security.ssl.CertStoreConfig
+import misk.security.ssl.SslLoader.Companion.FORMAT_JKS
+import misk.security.ssl.TrustStoreConfig
 import javax.inject.Singleton
 
 internal class ZkTestModule : KAbstractModule() {
   override fun configure() {
-    bind<ZookeeperConfig>().toInstance(ZookeeperConfig(zk_connect = "127.0.0.1:$zkPortKey"))
+    val keystorePath = this::class.java.getResource("/zookeeper/keystore.jks").path
```
ryanhall07: you should use the `ResourceLoader` to get these. should be able to bind the `ZookeeperConfig` to a provider that uses the `ResourceLoader`

misk-aws/src/test/kotlin/misk/jobqueue/sqs/DockerSqs.kt
```diff
@@ -20,13 +20,13 @@ internal class DockerSqs {
...
   private val clientPort = ExposedPort.tcp(CLIENT_PORT)
-  private val composer = Composer("e-sqs", Container {
+  private val composer = Composer("e-sqs", Container({
     withImage("pafortin/goaws")
         .withName("sqs")
         .withCmd(listOf("goaws"))
         .withExposedPorts(clientPort)
         .withPortBindings(Ports().apply { bind(clientPort, Ports.Binding.bindPort(CLIENT_PORT)) })
-  })
+  }, {}))
```
ryanhall07: maybe don't require this arg? or create multiple constructors?

nbroyles: I'll do the multiple constructors. Couldn't find a way to not require with a previous lambda there as an arg.

ryanhall07: there is also `ResourceLoader.SYSETM`, but i'd actually like to kill that since it makes `InjectorTest`s impossible. would be _fine_ in tests though.

misk-zookeeper/src/test/kotlin/misk/clustering/zookeeper/ZkTestModule.kt
nbroyles: Blerg, decided to leave this as-is since using the `ResourceLoader` to get the absolute path of a resource isn't really supported.

ryanhall07: ah, you don't need the contents

PR #738, opened 2019-01-14
Hibernate.field can find inherited properties

mightyguava:
This was throwing errors on startup for inherited properties due to a usage in https://github.com/square/misk/blob/3fb0b899e5ae6402ceccb345fee606b8bdf3b6a7/misk-hibernate/src/main/kotlin/misk/hibernate/SessionFactoryService.kt#L150 which iterates through all properties in all Entities. It only looked for fields declared by entity class, not any inherited fields.

### DIFF
--- a/misk-hibernate/src/test/kotlin/misk/hibernate/HibernateTest.kt
+++ b/misk-hibernate/src/test/kotlin/misk/hibernate/HibernateTest.kt
@@ -1,0 +1,32 @@
+package misk.hibernate
+
+import org.hibernate.mapping.Property
+import org.junit.jupiter.api.Assertions.assertEquals
+import org.junit.jupiter.api.Test
+import org.junit.jupiter.api.assertThrows
+
+internal class HibernateTest {
+  @Test
+  fun inheritedFieldTest() {
+    val parentProperty = Property()
+    parentProperty.name = "parentVal"
+    val childProperty = Property()
+    childProperty.name = "childVal"
+
+    assertEquals("parentVal", field(Parent::class.java, parentProperty).name)
+    assertEquals("parentVal", field(Child::class.java, parentProperty).name)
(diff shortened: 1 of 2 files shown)


misk-hibernate/src/test/kotlin/misk/hibernate/HibernateTest.kt
```diff
@@ -0,0 +1,27 @@
...
+  @Test
+  fun inheritedFieldTest() {
+    val parentProperty = Property()
+    parentProperty.name = "parentVal"
+    val childProperty = Property()
+    childProperty.name = "childVal"
+
+    assertEquals("parentVal", field(Parent::class.java, parentProperty).name)
+    assertEquals("parentVal", field(Child::class.java, parentProperty).name)
+    assertEquals("childVal", field(Child::class.java, childProperty).name)
```
wesleyk: can we have a failing case as well?

mightyguava: There isn't really a failing case per-se. The `assertEquals` is just paranoia to confirm that it returned the right property. If the property doesn't exist, it's an exception.

wesleyk: to clarify, that's what I mean by "failing"; a case that throws an exception

wesleyk: (and asserting that an exception is indeed thrown)

mightyguava: Ok, done

wesleyk: thanks!

PR #742, opened 2019-01-23
Vitess enable RBR and clean up volumes properly

mightyguava:
- Enable RBR for binlogs, probably what we'll always use in production
- Without the `withRemoveVolumes()` flag docker doesn't clean up Vitess's data volume after shutdown

### DIFF
--- a/misk-hibernate/src/main/kotlin/misk/vitess/StartVitessService.kt
+++ b/misk-hibernate/src/main/kotlin/misk/vitess/StartVitessService.kt
@@ -195,6 +195,7 @@ class DockerVitessCluster(
     grantExternalAccessToDbaUser()
 
     turnOnGeneralLog()
+    setRowBasedBinaryLogging()
   }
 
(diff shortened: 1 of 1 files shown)


misk-hibernate/src/main/kotlin/misk/vitess/StartVitessService.kt
```diff
@@ -255,13 +256,24 @@ class DockerVitessCluster(
...
+  }
+
   fun stop() {
     if (!isRunning) {
       return
     }
     isRunning = false
 
-    docker.removeContainerCmd(containerId!!).withForce(true).exec()
+    docker.removeContainerCmd(containerId!!).withForce(true).withRemoveVolumes(true).exec()
```
ryanhall07: we should port vitess to use the container `Composer` in `misk-testing` and add this flag so other containers don't make the same mistakes (like zk and kafka)

PR #743, opened 2019-01-23
StartVitessService produces itself as a service dependency

mightyguava:
Allows other services to depend on Vitess being available, e.g. for testing.

### DIFF
--- a/misk-testing/src/main/kotlin/misk/service/ServiceTestingModule.kt
+++ b/misk-testing/src/main/kotlin/misk/service/ServiceTestingModule.kt
@@ -27,9 +27,6 @@ class ServiceTestingModule<T : Service> internal constructor(
         getProvider(wrappedServiceType.java),
         extraDependencies
     )).asSingleton()
-
-    // Bind the wrapped service directly so that application code can inject it
-    bind(wrappedServiceType.java)
   }
 
   companion object {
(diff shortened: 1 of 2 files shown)


misk-testing/src/main/kotlin/misk/service/ServiceTestingModule.kt
```diff
@@ -27,9 +27,6 @@ class ServiceTestingModule<T : Service> internal constructor(
         getProvider(wrappedServiceType.java),
         extraDependencies
     )).asSingleton()
-
-    // Bind the wrapped service directly so that application code can inject it
-    bind(wrappedServiceType.java)
```
ryanhall07: why did this have to be removed?

mightyguava: For my use case, the wrapped service is already bound via a custom provider, since it has qualified injections. This binding conflicts. I also don't think it's common that an application will need to inject the Service, and even if it does the app could just call `bind()` on the type itself.

ryanhall07: yea I probably agree that it's not common. you are probably going to break downstream repos though :)

mightyguava: I checked, we seem to only have 1 usage of this method and it doesn't break.

PR #746, opened 2019-01-25
Add ACL support to ZooKeeper

nbroyles:
/services/\<app-name>/... and /leases/\<app-name>/... allow CRUD access to any client who identifies as the app itself based on the Distinguished Name in the cert provided.

/services and /leases allow CRU access to any client with a cert so that they can create the parent node if need be, but cannot remove it.

Also provide a client that helps apps put their service data in the correct location.

### DIFF
--- a/misk-zookeeper/src/main/kotlin/misk/clustering/zookeeper/ZookeeperModule.kt
+++ b/misk-zookeeper/src/main/kotlin/misk/clustering/zookeeper/ZookeeperModule.kt
@@ -5,6 +5,7 @@ import com.google.inject.Key
 import misk.clustering.lease.LeaseManager
 import misk.inject.KAbstractModule
 import misk.inject.asSingleton
+import misk.zookeeper.CuratorFrameworkProvider
 import org.apache.curator.framework.CuratorFramework
 
 class ZookeeperModule : KAbstractModule() {
--- a/misk-zookeeper/src/main/kotlin/misk/zookeeper/CuratorFrameworkProvider.kt
+++ b/misk-zookeeper/src/main/kotlin/misk/zookeeper/CuratorFrameworkProvider.kt
@@ -1,10 +1,10 @@
-package misk.clustering.zookeeper
+package misk.zookeeper
 
 import com.google.common.util.concurrent.ThreadFactoryBuilder
+import misk.clustering.zookeeper.ZookeeperConfig
 import org.apache.curator.framework.CuratorFramework
 import org.apache.curator.framework.CuratorFrameworkFactory
 import org.apache.curator.retry.ExponentialBackoffRetry
-import org.apache.curator.utils.DefaultZookeeperFactory
 import org.apache.zookeeper.ZooKeeper
 import org.apache.zookeeper.client.ZKClientConfig
 import javax.inject.Inject
@@ -33,7 +33,7 @@ internal class CuratorFrameworkProvider @Inject internal constructor(
         .sessionTimeoutMs(config.session_timeout_msecs)
         .canBeReadOnly(false)
         .threadFactory(ThreadFactoryBuilder()
-            .setNameFormat("zk-clustering-${config.zk_connect}")
+            .setNameFormat("zk-${config.zk_connect}")
             .build())
         .zookeeperFactory { connectString, sessionTimeout, watcher, canBeReadOnly ->
           val clientConfig = ZKClientConfig()
--- a/misk-zookeeper/src/main/kotlin/misk/zookeeper/CuratorFrameworkProvider.kt
+++ b/misk-zookeeper/src/main/kotlin/misk/zookeeper/CuratorFrameworkProvider.kt
@@ -2,13 +2,34 @@ package misk.zookeeper
 
 import com.google.common.util.concurrent.ThreadFactoryBuilder
 import misk.clustering.zookeeper.ZookeeperConfig
+import misk.clustering.zookeeper.asZkPath
 import org.apache.curator.framework.CuratorFramework
 import org.apache.curator.framework.CuratorFrameworkFactory
+import org.apache.curator.framework.api.ACLProvider
 import org.apache.curator.retry.ExponentialBackoffRetry
+import org.apache.zookeeper.ZooDefs
+import org.apache.zookeeper.ZooDefs.Ids.ANYONE_ID_UNSAFE
+import org.apache.zookeeper.ZooDefs.Ids.AUTH_IDS
 import org.apache.zookeeper.ZooKeeper
 import org.apache.zookeeper.client.ZKClientConfig
+import org.apache.zookeeper.data.ACL
+import java.util.Collections
 import javax.inject.Inject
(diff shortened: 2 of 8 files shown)


misk-zookeeper/src/main/kotlin/misk/zookeeper/CuratorFrameworkProvider.kt
```diff
@@ -1,15 +1,36 @@
...
+import org.apache.zookeeper.ZooDefs.Ids.AUTH_IDS
 import org.apache.zookeeper.ZooKeeper
 import org.apache.zookeeper.client.ZKClientConfig
+import org.apache.zookeeper.data.ACL
+import java.util.Collections
 import javax.inject.Inject
 import javax.inject.Provider
 
+// Directory where ZooKeeper stores the leases that it manages on a service's behalf
+const val LEASES_NODE = "leases"
```
ryanhall07: what if a service wants to use zk for more than just leases?

misk-zookeeper/src/main/kotlin/misk/zookeeper/CuratorFrameworkProvider.kt
```diff
@@ -47,6 +68,24 @@ internal class CuratorFrameworkProvider @Inject internal constructor(
...
         }
+        .aclProvider(object : ACLProvider {
+          override fun getDefaultAcl(): List<ACL> {
+            // Default ACL allows clients CRUD operations on znodes this service has created.
+            // Service identification comes from the client cert
+            return Collections.singletonList(ACL(DEFAULT_PERMS, AUTH_IDS))
+          }
+
+          override fun getAclForPath(path: String?): List<ACL> {
+            // Shared directories need to be created in such a way that anyone can create them if
```
ryanhall07: i don't understand why this is necessary to  allow any client

misk-zookeeper/src/main/kotlin/misk/zookeeper/CuratorFrameworkProvider.kt
nbroyles: Then it can use the `ZkClientFactory` and put whatever it wants in `/services/<app-name>/`. The client factory will give them a client configured to do just that.

misk-zookeeper/src/main/kotlin/misk/zookeeper/CuratorFrameworkProvider.kt
nbroyles: The directory might not exist. For example, Larry could be trying to write to `/leases/larry/` but `/leases/` may not exist. We need to ensure Larry creates `/leases` in a way so that in the future Moe can create `/leases/moe` without any issues

misk-zookeeper/src/main/kotlin/misk/zookeeper/CuratorFrameworkProvider.kt
ryanhall07: so why the need for leases and services ?

ryanhall07: hmm because leases are a different namespace...

why aren't leases scoped to the service?

nbroyles: I don't think there's a good reason, tbh. I think that's a relic of the fact that the LeaseManager code predated the ZK work. I think we can move that into the `/services` namespace with little problem unless there's something I'm missing. If we want to do that, it's better to do it now rather than later.

ryanhall07: my vote is to move it, but i'm not familiar with the problem space.

would it simplify the ACL logic if everything was nested under `/services`?

ryanhall07: i don't know enough about zk to have an opinion.

nbroyles: There's nothing really special around these paths. They're just namespaces so we can set them up however we see fit (highly analogous to organize files on disk)

I'm going to go with `/services/<app-name>/leases` and `/services/<app-name>/data` since it seems to be the most logical to me.

misk-zookeeper/src/main/kotlin/misk/clustering/zookeeper/ZookeeperModule.kt
```diff
@@ -5,6 +5,7 @@ import com.google.inject.Key
 import misk.clustering.lease.LeaseManager
 import misk.inject.KAbstractModule
 import misk.inject.asSingleton
+import misk.zookeeper.CuratorFrameworkProvider
```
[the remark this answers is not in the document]
nbroyles: No. I moved the provider out of the `clustering` package as it's used for more than just that.

misk-zookeeper/src/main/kotlin/misk/zookeeper/CuratorFrameworkProvider.kt
```diff
@@ -47,6 +68,24 @@ internal class CuratorFrameworkProvider @Inject internal constructor(
...
+            // they're missing and read or write to them, but anyone should not be able to delete
```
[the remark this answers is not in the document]
nbroyles: lol, all good! An individual service shouldn't be able to delete a shared directory like `services` -- because that would delete all the service-specific subdirectories.

misk-zookeeper/src/main/kotlin/misk/zookeeper/CuratorFrameworkProvider.kt
```diff
@@ -47,6 +68,24 @@ internal class CuratorFrameworkProvider @Inject internal constructor(
...
+            // them.
+            if (path == LEASES_NODE.asZkPath || path == SERVICES_NODE.asZkPath) {
```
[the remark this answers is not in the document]
nbroyles: By defining the `aclProvider` on the `CuratorFrameworkFactory.Builder`, we are sharing this behavior everywhere since the CuratorFramework wraps all interactions with ZooKeeper.

If the path is neither of these then the default ACL is used meaning the node will be created with the default permissions of "the service that created this node can perform CRUD operations".

misk-zookeeper/src/main/kotlin/misk/zookeeper/CuratorFrameworkProvider.kt
ryanhall07: is it safe to rollout this change? is it going to break clustering during the rollout? probably not a big deal since nobody in prod depends on clustering yet

nbroyles: Yeah, exactly. Plus, I was planning on blowing away ZK anyway when I do this so that we can pick up the encrypted volumes changes.

PR #747, opened 2019-01-25
Install the FakeClockModule in MiskTestingServiceModule, move MiskTestingServiceModule into misk-testing

hamdanjaveed:
The `MiskTestingServiceModule` currently binds a real system clock to `Clock`, which means that in order for tests to use a `FakeClock`, you have to override `MiskTestingServiceModule` with a `FakeClockModule`.

The main change this PR makes is that `MiskTestingServiceModule` now installs a `FakeClockModule`, and `MiskRealServiceModule` installs a `ClockModule`.

However, this would cause a circular dependency between the `:misk` and `:misk-testing` projects. In order to access `FakeClockModule`, the `:misk` project needs to depend on the `:misk-testing` project, which it can't do because `:misk-testing` depends on `:misk`.

The solution I opted for in this PR is to move `MiskTestingServiceModule` into `:misk-testing`. This removes the need for `:misk` to depend on `:misk-testing`. Existing misk apps should already have a dependency on `:misk-testing` for their testing classes, and thus no changes should be required.

Perhaps a better solution could be to make a `:misk-app`-type project and have `MiskTestingServiceModule` and similar files moved into there. Then the `:misk-app` project could depend on both `:misk` and `:misk-testing`. This also resolves the circular dependency. Existing misk apps will have to have their imports updated and add a dependency to `:misk-app`.

### DIFF
--- a/misk-testing/src/main/kotlin/misk/MiskTestingServiceModule.kt
+++ b/misk-testing/src/main/kotlin/misk/MiskTestingServiceModule.kt
@@ -1,0 +1,23 @@
+package misk
+
+import misk.MiskCommonServiceModule
+import misk.environment.FakeEnvVarModule
+import misk.inject.KAbstractModule
+import misk.resources.TestingResourceLoaderModule
+import misk.time.FakeClockModule
+
+/**
+ * [MiskTestingServiceModule] should be installed in unit testing environments.
+ *
+ * This should not contain application level fakes for testing. It includes a small, selective
+ * set of fake bindings to replace real bindings that cannot exist in a unit testing environment
+ * (e.g system env vars and filesystem dependencies).
+ */
+class MiskTestingServiceModule : KAbstractModule() {
+  override fun configure() {
+    install(TestingResourceLoaderModule())
+    install(FakeEnvVarModule())
+    install(FakeClockModule())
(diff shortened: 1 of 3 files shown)


misk-aws/src/test/kotlin/misk/jobqueue/sqs/SqsJobQueueTest.kt
```diff
@@ -10,7 +10,7 @@ import com.amazonaws.services.sqs.model.CreateQueueRequest
 import com.google.common.util.concurrent.Service
 import com.google.inject.Provides
 import com.google.inject.util.Modules
-import misk.MiskTestingServiceModule
+import misk.service.MiskTestingServiceModule
```
ryanhall07: why the package name change?

hamdanjaveed: I moved it into the service package to group it with `ServiceTestingModule`, could move it out of `misk.service` now that i think about it
image

hamdanjaveed: Moved `MiskTestingServiceModule ` into `misk` so no package name change.

PR #748, opened 2019-01-28
Support verifying migrations on Vitess

tirsen:


misk-hibernate/src/main/kotlin/misk/hibernate/SchemaMigrator.kt
```diff
@@ -198,21 +205,16 @@ internal class SchemaMigrator(
...
     try {
-      val missingMigrations = availableMigrations() - appliedMigrations()
-      check(missingMigrations.isEmpty()) {
-        "${qualifier.simpleName} is missing migrations:\n  " +
-            missingMigrations.map { it.path }.joinToString(separator = "\n  ")
+      transacter.get().shards().forEach { shard ->
+        val missingMigrations = availableMigrations(shard.keyspace) - appliedMigrations(shard)
+
+        check(missingMigrations.isEmpty()) {
+          "${qualifier.simpleName} is missing migrations:\n  " + missingMigrations.map { it.path }.joinToString(
```
wesleyk: will this clearly denote which shard is missing the migration?

misk-hibernate/src/main/kotlin/misk/hibernate/SchemaMigratorService.kt
```diff
@@ -22,8 +24,11 @@ class SchemaMigratorService internal constructor(
   override fun startUp() {
     val schemaMigrator = schemaMigratorProvider.get()
     if (environment == Environment.TESTING || environment == Environment.DEVELOPMENT) {
-      val appliedMigrations = schemaMigrator.initialize()
-      schemaMigrator.applyAll("SchemaMigratorService", appliedMigrations)
+      if (config.type != DataSourceType.VITESS) {
+        // vttestserver automatically applies migrations
```
wesleyk: are there particular reasons why we shouldn't set it up instead of `vttestserver`? Does `vttestserver` do other special setup that existing tables would mess with? I ask because it would be nice to exercise applyAll for tests for vitess-based apps

misk-hibernate/src/test/kotlin/misk/hibernate/VitessSchemaMigratorTest.kt
```diff
@@ -0,0 +1,80 @@
...
+  }
+
+  @Test fun requireAll() {
+    schemaMigrator.initialize()
+
+    assertThat(schemaMigrator.appliedMigrations(shard).map { it.version }).isEmpty()
+
+    insertSchemaMigration("1")
+    insertSchemaMigration("2")
+    schemaMigrator.requireAll(shard)
```
wesleyk: anything to assert here? Or is it enough that it doesn't throw an exception? Could we leave a comment indicating that?

misk-hibernate/src/test/kotlin/misk/hibernate/VitessSchemaMigratorTest.kt
```diff
@@ -0,0 +1,80 @@
...
+      schemaVersion.setString(1, version)
+      schemaVersion.setString(2, "test")
+      schemaVersion.executeUpdate()
+    }
+  }
+
+  /** Open a direct connection to the Vitess MySQL instance. */
+  private fun openDirectConnection(): Connection {
+    val cluster = vitessService.cluster()
+    val connection = cluster.openMysqlConnection()
```
wesleyk: /nit in-line this with the return?

misk/src/main/kotlin/misk/resources/ResourceLoader.kt
```diff
@@ -126,6 +126,7 @@ class ResourceLoader @Inject constructor(
    */
   private fun parseAddress(path: String): Address {
     val colon = path.indexOf(':')
+    check(colon != -1) { "address scheme not specified in: $path"}
```
wesleyk: nice

misk-hibernate/src/main/kotlin/misk/hibernate/SchemaMigratorService.kt
tirsen: Yeah doing it in vttestserver is faster and easier because it runs it in parallel and it runs it before applying the vschema. It would be a lot of hard and quite slow to do it outside of vttestserver because we would have to call in to do all the set up. This is the main reason Vitess tests are so slow in the Square monorepo.

I don't think there's a point in exercising `applyAll` in Vitess tests. It's never used in production.

PR #750, opened 2019-01-29
Set RBR mode via mycnf instead of global var

mightyguava:
Setting the global var at runtime only affects sessions created afterwards. Any sessions cached from before setting this var still use STATEMENT logging, and can result in some transactions non-deterministically logging as statements instead of rows.

I also had to change the `VitessScaleSafetyChecks` because with RBR mode turned on, it looks like Vitess no appends writes the `_stream` tag in a comment at the end of the query, breaking the transaction decorator that's extracting the keyspace ID for DML statements.

STATEMENT mode:
```sql
| 2019-01-29 18:09:10.851772 | vt_dba[vt_dba] @ localhost []  |        95 | 866341052 | Query        | /* insert misk.hibernate.DbMovie */ insert into movies(created_at, name, release_date, updated_at, id) values ('2017-12-31 16:00:00.0', 'Jurassic Park', '1993-06-09', '2017-12-31 16:00:00.0', 1) /* _stream movies (id ) (1 ); */ /* vtgate:: keyspace_id:166b40b44aba4bd6 */         |
```
ROW mode:
```sql
| 2019-01-29 18:13:33.579635 | vt_dba[vt_dba] @ localhost []  |        95 | 749452573 | Query        | /* insert misk.hibernate.DbMovie */ insert into movies(created_at, name, release_date, updated_at, id) values ('2017-12-31 16:00:00.0', 'Jurassic Park', '1993-06-09', '2017-12-31 16:00:00.0', 1) /* vtgate:: keyspace_id:166b40b44aba4bd6 */ |
```

I made the general log query a little more generic, but I'm not sure if it will catch all DML properly. The original query looks a bit fragile to start with, so I think the new query should still be acceptable.

### DIFF
--- a/misk-hibernate-testing/src/main/kotlin/misk/jdbc/VitessScaleSafetyChecks.kt
+++ b/misk-hibernate-testing/src/main/kotlin/misk/jdbc/VitessScaleSafetyChecks.kt
@@ -264,16 +264,18 @@ class VitessScaleSafetyChecks(
   private fun extractLastDmlQuery(): String? {
     return connect().let { c ->
       c.createStatement().use { s ->
-        // OFFSET 1 because this query itself is also directly logged before it's executed
-        // "/* _stream" is appended to every DML statement by Vitess so that we skip a bunch of
-        // irrelevant statements
         s.executeQuery("""
                   SELECT argument
                   FROM mysql.general_log
                   WHERE command_type = 'Query'
-                  AND argument LIKE '%/* _stream%'
+                  AND (
+                    argument LIKE '%update%'
+                    OR argument LIKE '%insert%'
+                    OR argument LIKE '%delete%'
+                  )
+                  AND NOT argument LIKE 'SELECT argument%'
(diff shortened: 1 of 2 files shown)


misk-hibernate-testing/src/main/kotlin/misk/jdbc/VitessScaleSafetyChecks.kt
```diff
@@ -264,16 +264,18 @@ class VitessScaleSafetyChecks(
...
                   WHERE command_type = 'Query'
-                  AND argument LIKE '%/* _stream%'
+                  AND (
+                    argument LIKE '%update%'
+                    OR argument LIKE '%insert%'
+                    OR argument LIKE '%delete%'
+                  )
+                  AND NOT argument LIKE 'SELECT argument%'
                   ORDER BY event_time DESC
-                  LIMIT 1 OFFSET 1
```
tirsen: Nice! We could get rid of the hacky OFFSET 1!

PR #751, opened 2019-01-30
Updates MoshiJsonAdapter to be available externally

MichaelCPell:
I think we need to make the `MoshiJsonAdapter` annotation public, so that I can inject my custom Factory.


MichaelCPell:
Closing in favor of https://github.com/square/misk/pull/754

PR #753, opened 2019-01-30
Add listener support to Leases

mightyguava:
Allow clients to register listeners for lease state changes. If the client chooses to use listeners, it should not have ever to call `checkHeld()`, since the `LeaseManager` should automatically update leases in the background.

This interface is meant for apps that need to lease long-running background processes. An app that uses listeners would instead do something like

```kotlin
leaseManager.requestLease("my-lease").addListener(object : Lease.StateChangeListener {
  override fun afterAcquire(lease: Lease) {
    startBackgroundProcess()
  }

  override fun beforeRelease(lease: Lease) {
    stopBackgroundProcess()
  }
}
```

### DIFF
--- a/misk-zookeeper/src/main/kotlin/misk/clustering/zookeeper/ZkLease.kt
+++ b/misk-zookeeper/src/main/kotlin/misk/clustering/zookeeper/ZkLease.kt
@@ -48,6 +48,7 @@ internal class ZkLease(
   }
 
   @GuardedBy("lock") private var status: Status = Status.UNKNOWN
+  @GuardedBy("lock") private val listeners = mutableListOf<Lease.StateChangeListener>()
   private val lock = ReentrantLock()
   private val leaseData = ownerName.toByteArray(Charsets.UTF_8)
   private val leaseZkPath = name.asZkPath
@@ -85,6 +86,19 @@ internal class ZkLease(
     }
   }
 
+  override fun addListener(listener: Lease.StateChangeListener) {
+    lock.withLock {
+      listeners.add(listener)
+      when (status) {
+        // We already own the lease and should tell the listener
+        Status.HELD -> listener.afterAcquire(this)
+        // We don't know if we own the lease, so check
+        Status.UNKNOWN -> checkHeld()
+        else -> {}
+      }
+    }
+  }
+
   fun close() {
     lock.withLock {
       if (status != Status.NOT_HELD) release()
@@ -104,6 +118,7 @@ internal class ZkLease(
   private fun release(): Boolean {
     try {
       if (checkLeaseNodeExists() && checkLeaseDataMatches()) {
+        notifyBeforeRelease()
         // Lease exists in zk and we own it, so delete it
         manager.client.value.delete().guaranteed().forPath(leaseZkPath)
         log.info { "released lease $name" }
@@ -129,6 +144,7 @@ internal class ZkLease(
       if (checkLeaseDataMatches()) {
         log.info { "reclaiming currently held lease $name" }
         status = Status.HELD
+        notifyAfterAcquire()
         return true
       }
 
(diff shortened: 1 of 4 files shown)


misk-zookeeper/src/test/kotlin/misk/clustering/zookeeper/ZkLeaseTest.kt
```diff
@@ -212,6 +217,48 @@ internal class ZkLeaseTest {
     assertThat(lease.checkHeld()).isFalse() // still not held since its closed
   }
 
+  @Test fun listenerNotifications() {
+    val listenerMock = Mockito.mock<Lease.StateChangeListener>()
```
mightyguava: I don't usually mocks, but this seemed appropriate since I want to verify exactly how many times the listener gets called.

misk-zookeeper/src/main/kotlin/misk/clustering/zookeeper/ZkLease.kt
```diff
@@ -85,6 +86,19 @@ internal class ZkLease(
...
   }
 
+  override fun addListener(listener: Lease.StateChangeListener) {
+    lock.withLock {
+      listeners.add(listener)
+      when (status) {
+        // We already own the lease and should tell the listener
+        Status.HELD -> listener.afterAcquire(this)
+        // We don't know if we own the lease, so check
+        Status.UNKNOWN -> checkHeld()
```
mightyguava: This is necessary since the client is no longer expected to call `checkHeld()` after requesting a lease.

misk-zookeeper/src/main/kotlin/misk/clustering/zookeeper/ZkLease.kt
```diff
@@ -155,6 +172,26 @@ internal class ZkLease(
     return false
   }
 
+  private fun notifyAfterAcquire() {
```
mightyguava: Should listeners be notified asynchronously?

mmihic: No, but we do need to be careful about possible deadlocks from app code calling back into the lease.

misk-zookeeper/src/test/kotlin/misk/clustering/zookeeper/ZkLeaseTest.kt
mmihic: Could also just increment an `AtomicInteger` and check the result...

misk/src/main/kotlin/misk/clustering/lease/Lease.kt
```diff
@@ -18,4 +18,24 @@ interface Lease {
    * so it is marked as a function rather than a property to make the potential expense clearer
    */
   fun checkHeld(): Boolean
+
+  /**
+   * Registers a listener that is called on lease state changes.
+   */
+  fun addListener(listener: StateChangeListener)
```
mightyguava: Should there be a `removeListener()` method? I didn't see the need for it and preferred not to make the interface wider.

mmihic: Agreed, don't think it's necessary. Can always add it later if we do see the need, but I think listeners will always be registered at startup and never removed.

misk-zookeeper/src/main/kotlin/misk/clustering/zookeeper/ZkLease.kt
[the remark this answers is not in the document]
mmihic: Only if we managed to acquire it and are not still in the unknown state

mightyguava:
> might want a test that makes sure the listener can call `checkHeld` without deadlocking

Added in 391132a

misk/src/main/kotlin/misk/clustering/lease/Lease.kt
```diff
@@ -18,4 +18,24 @@ interface Lease {
...
+
+  interface StateChangeListener {
+    /**
+     * Called immediately after the lease is acquired. Also called immediately if the lease is
+     * already owned by this process instance when the listener is registered.
+     * @param lease the lease that is acquired
+     */
+    fun afterAcquire(lease: Lease)
```
ryanhall07: nice interface w/ descriptive method names

PR #754, opened 2019-01-30
Adds JsonAdapter.Factory support to MoshiAdapterModule

MichaelCPell:
This will allow support for `PolymorphicJsonAdapterFactory` and any other Factory based adapter approach.


swankjesse:
Two requests:

Can you add a test? Right now it looks like these new bindings might get dropped on the floor.

Can you change this type check from being where the adapter/factory is bound to instead be where Moshi is configured? That way a heterogeneous list of adapters and factories retain their order.

MichaelCPell:
@swankjesse Updated the commit re: the requests.

PR #756, opened 2019-02-01
Modify RedisConfig and use JedisPool

hamdanjaveed:
RedisConfig updated to include a writer and reader endpoint. We also use JedisPool
instead of single Jedis instances to ensure thread safety.

### DIFF
--- a/misk-redis/src/main/kotlin/misk/redis/RedisConfig.kt
+++ b/misk-redis/src/main/kotlin/misk/redis/RedisConfig.kt
@@ -2,8 +2,18 @@ package misk.redis
 
 import misk.config.Config
 
-data class RedisConfig(
-  val host_name: String,
-  val port: Int,
-  val auth_password: String
-) : Config
+class RedisConfig: java.util.LinkedHashMap<String, RedisReplicationGroupConfig>, Config {
+  constructor(): super()
+  constructor(m: Map<String, RedisReplicationGroupConfig>): super(m)
+}
+
+data class RedisReplicationGroupConfig(
+  val writer_endpoint: RedisNodeConfig,
+  val reader_endpoint: RedisNodeConfig,
+  val redis_auth_password: String
+)
+
+data class RedisNodeConfig(
+  val hostname: String,
+  val port: Int
+)
--- a/misk-redis/src/main/kotlin/misk/redis/RedisModule.kt
+++ b/misk-redis/src/main/kotlin/misk/redis/RedisModule.kt
@@ -4,18 +4,37 @@ import com.google.common.util.concurrent.Service
 import com.google.inject.Provides
 import com.google.inject.Singleton
 import misk.inject.KAbstractModule
-import redis.clients.jedis.Jedis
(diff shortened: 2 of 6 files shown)


misk-redis/src/main/kotlin/misk/redis/RedisConfig.kt
```diff
@@ -2,8 +2,26 @@ package misk.redis
...
 import misk.config.Config
 
-data class RedisConfig(
-  val host_name: String,
-  val port: Int,
-  val auth_password: String
-) : Config
+/**
+ * The top level field is the name of the replication group.
+ * See: https://docs.aws.amazon.com/AmazonElastiCache/latest/red-ug/WhatIs.Terms.html
```
wesleyk: these are aws-specific configs/comments. Do we want that to live in the misk-redis package?

hamdanjaveed: Since we are abstracting away the aws-specific details with this layer I think you're right. Removing.

misk-redis/src/main/kotlin/misk/redis/RedisConfig.kt
```diff
@@ -2,8 +2,18 @@ package misk.redis
...
-) : Config
+class RedisConfig: java.util.LinkedHashMap<String, RedisReplicationGroupConfig>, Config {
+  constructor(): super()
+  constructor(m: Map<String, RedisReplicationGroupConfig>): super(m)
+}
+
+data class RedisReplicationGroupConfig(
+  val writer_endpoint: RedisNodeConfig,
+  val reader_endpoint: RedisNodeConfig,
+  val redis_auth_password: String
```
ryanhall07: I wonder if this should be a `Secret`. while it's not strictly necessary now because it's a single value, I would assume in the future `Secret` would allow us to redact the value instead of fuzzy matching on fields named `password`

misk-redis/src/main/kotlin/misk/redis/RedisModule.kt
```diff
@@ -4,18 +4,38 @@ import com.google.common.util.concurrent.Service
 import com.google.inject.Provides
 import com.google.inject.Singleton
 import misk.inject.KAbstractModule
-import redis.clients.jedis.Jedis
+import redis.clients.jedis.JedisPool
+import redis.clients.jedis.JedisPoolConfig
 
-class RedisModule : KAbstractModule() {
+/**
+ * Configures a JedisPool to connect to an ElastiCache replication group. The use of a JedisPool ensures thread safety.
```
wesleyk: elasticache is still a aws concept?

hamdanjaveed: Updated comment

misk-redis/src/main/kotlin/misk/redis/RedisConfig.kt
hamdanjaveed: I think a `Secret` makes sense here, I'll keep it in mind

PR #757, opened 2019-02-06
Query tracing at hibernate level

qelper:
Adds INSERT, UPDATE, DELETE and SELECT query tracing at hibernate level.

### DIFF
--- a/misk-hibernate/src/main/kotlin/misk/hibernate/QueryTracingListener.kt
+++ b/misk-hibernate/src/main/kotlin/misk/hibernate/QueryTracingListener.kt
@@ -1,0 +1,100 @@
+package misk.hibernate
+
+import io.opentracing.Scope
+import io.opentracing.Tracer
+import misk.hibernate.QueryTracingSpanNames.Companion.DB_DELETE
+import misk.hibernate.QueryTracingSpanNames.Companion.DB_INSERT
+import misk.hibernate.QueryTracingSpanNames.Companion.DB_UPDATE
+import misk.logging.getLogger
+import org.hibernate.event.spi.PostDeleteEvent
+import org.hibernate.event.spi.PostDeleteEventListener
+import org.hibernate.event.spi.PostInsertEvent
+import org.hibernate.event.spi.PostInsertEventListener
+import org.hibernate.event.spi.PostUpdateEvent
+import org.hibernate.event.spi.PostUpdateEventListener
+import org.hibernate.event.spi.PreDeleteEvent
+import org.hibernate.event.spi.PreDeleteEventListener
+import org.hibernate.event.spi.PreInsertEvent
+import org.hibernate.event.spi.PreInsertEventListener
+import org.hibernate.event.spi.PreUpdateEvent
+import org.hibernate.event.spi.PreUpdateEventListener
+import org.hibernate.persister.entity.EntityPersister
+import javax.inject.Singleton
+
+private val logger = getLogger<QueryTracingListener>()
+
+/**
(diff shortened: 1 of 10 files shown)


misk-hibernate/src/main/kotlin/misk/hibernate/RealTransacter.kt
```diff
@@ -102,6 +111,8 @@ internal class RealTransacter private constructor(
       try {
         val result = lambda(session)
 
+        // Flush any changes to the databased before commit
+        session.hibernateSession.flush()
```
wesleyk: hmm I can't quite speak to the impact of adding a flush here so I'd want some additional eyes on this point. Why is this needed? Doesn't the transaction commit take care of that for you.

Also: databased -> database on the comment

misk-hibernate/src/main/kotlin/misk/hibernate/QueryTracingListener.kt
```diff
@@ -0,0 +1,110 @@
...
+  }
+
+  private fun startSpan(spanName: String) {
+    if (tracer == null) {
+      return
+    }
+
+    val lastScope = lastScopeInThread.get()
+    lastScope?.let {
+      logger.debug { "lastScope wasn't closed" }
```
wesleyk: would we want at least a logger.info with some debugging information? seems like a bug we'd want to know is occurring/track down

misk-hibernate/src/test/kotlin/misk/hibernate/QueryTracingListenerTest.kt
```diff
@@ -0,0 +1,69 @@
...
+import misk.testing.MiskTest
+import misk.testing.MiskTestModule
+import misk.testing.MockTracingBackendModule
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.jupiter.api.Test
+import java.time.LocalDate
+import javax.inject.Inject
+
+@MiskTest(startService = true)
+class QueryTracingListenerTest {
```
wesleyk: could we test when there's more than one trace in a txn?

misk-hibernate/src/main/kotlin/misk/hibernate/QueryTracingListener.kt
ryanhall07: would just want to rate limit the logging.

misk-hibernate/src/main/kotlin/misk/hibernate/RealTransacter.kt
ryanhall07: my guess is so the query listeners fire, although I would expect them to fire when committing

misk-hibernate/src/main/kotlin/misk/hibernate/ReflectionQueryFactory.kt
```diff
@@ -68,7 +72,9 @@ internal class ReflectionQuery<T : DbEntity<T>>(
 
     val typedQuery = session.hibernateSession.createQuery(query)
     typedQuery.maxResults = effectiveMaxRows(returnList)
-    val rows = typedQuery.list()
+    val rows = traceSelect {
+      typedQuery.list()
```
wesleyk: is there a preload hook we could include in QueryTracingListener instead of having a separate place where queries are traced here?

qelper: there is, it doesn't do what we need. I have added a comment about it in `QueryTracingListener`.

PR #762, opened 2019-02-07
Only log k8s connection exceptions if they persist for 1 minute

ryanhall07:
It's not uncommon for k8s watch exceptions to occur since it uses long
polling and the connection can get interrupted. Only log an error if
this persists for a minute.


ryanhall07:
not sure how to test this, since the k8s client is not an interface and difficult to mock :(

PR #763, opened 2019-02-07
Refactor out hibernate query limits into dependency to speed up ReflectionQueryFactoryTest

wesleyk:


misk-hibernate/src/test/kotlin/misk/hibernate/ReflectionQueryFactoryTest.kt
```diff
@@ -253,63 +258,63 @@ class ReflectionQueryFactoryTest {
...
-          .hasSize(3100)
+          .hasSize(31)
     }
     assertThat(logCollector.takeMessages(loggerClass = ReflectionQuery::class)).isEmpty()
 
-    // More than 10000 rows throws an exception.
+    // More than 40 rows throws an exception.
     transacter.transaction { session ->
-      for (i in 3101..10100) {
+      for (i in 32..41) {
```
adrw: Were any of these tests (that have been significantly reduced in quantity) the stress testing ones that had been mentioned in discussions? If so, should we have a way to still run them periodically (say daily) but not on every build / PR?

wesleyk: The intention of these aren't to stress test, they're to see that the various logs are outputted

wesleyk: (I wouldn't consider 10K to necessarily be "stress testing" anyways)

PR #764, opened 2019-02-07
Add FakeSleeper and FakeScheduledExecutorService

mightyguava:
`FakeScheduledExecutorService` can be used to test delayed task execution, and is copied from an internal implementation by @davidgarth 

`FakeSleeper` can be used to test `Thread.sleep()` calls.

### DIFF
--- a/misk-testing/src/main/kotlin/misk/concurrent/FakeScheduledExecutorService.kt
+++ b/misk-testing/src/main/kotlin/misk/concurrent/FakeScheduledExecutorService.kt
@@ -1,0 +1,77 @@
+package misk.concurrent
+
+import com.google.common.collect.MultimapBuilder
+import com.google.common.util.concurrent.MoreExecutors
+import java.time.Clock
+import java.util.concurrent.Callable
+import java.util.concurrent.Delayed
+import java.util.concurrent.ExecutorService
+import java.util.concurrent.FutureTask
+import java.util.concurrent.RunnableFuture
+import java.util.concurrent.ScheduledExecutorService
+import java.util.concurrent.ScheduledFuture
+import java.util.concurrent.TimeUnit
+import javax.inject.Inject
+import javax.inject.Singleton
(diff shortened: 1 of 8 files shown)


misk-testing/src/main/kotlin/misk/concurrent/FakeScheduledExecutorService.kt
```diff
@@ -0,0 +1,77 @@
...
+import java.util.concurrent.TimeUnit
+import javax.inject.Inject
+import javax.inject.Singleton
+
+/**
+ * ScheduledExecutorService for testing that runs in the current thread and is triggered using the
+ * `tick()` method. An injected [Clock] is used to decide whether to execute a scheduled task.
+ */
+@Singleton
+class FakeScheduledExecutorService @Inject constructor(private val clock: Clock) :
```
ryanhall07: test for this?

misk-testing/src/main/kotlin/misk/concurrent/FakeSleeper.kt
```diff
@@ -0,0 +1,77 @@
...
+
+/**
+ * [Sleeper] for testing that blocks threads calling [sleep], and checks whether the threads should
+ * wake using the [tick()] method. An injected [Clock] is used to decide whether to wake a thread.
+ */
+@Singleton
+class FakeSleeper @Inject constructor(private val clock: Clock) : Sleeper {
+  private var count: Int = 0
+  private var lastDuration: Duration? = null
+  private var lock = ReentrantLock()
```
mmihic: Should be `val`s

misk-testing/src/main/kotlin/misk/concurrent/FakeScheduledExecutorService.kt
mightyguava: Added a simple test.

PR #767, opened 2019-02-07
Validate schemas across all shards in Vitess

tirsen:


misk-hibernate/src/main/kotlin/misk/hibernate/SchemaValidator.kt
```diff
@@ -20,21 +20,35 @@ internal class SchemaValidator {
...
 
-    val hibernateSchema = readDeclarationFromHibernate(hibernateMetadata)
+    transacter.shards().forEach { shard ->
+      val dbMetadata =
+          transacter.transaction { s -> s.hibernateSession.doReturningWork { it.metaData } }
+      val dbSchema = readDeclarationFromDatabase(dbMetadata, config.database)
+      withDeclaration(config.database!!) {
+        validateDatabase(dbSchema, hibernateSchema)
+      }
+      allDbTables.addAll(dbSchema.tables)
```
wesleyk: hmm would this give a false positive result in the following scenario:

Hibernate: tables A, B, C
Shard 1: tables A, B, no table C
Shard 2: table C

then it looks like allDbTables matches hibernate tables?

Or is this necessary anyways because not all tables will be present on all shards, and there's no easy way to associate the hibernate entities to which shards they belong to in a generic way?

tirsen: This is actually entirely valid if shard 1 and 2 are in different keyspaces.

There is an additional check we could do that all shards in the same keyspace have the same set of tables but the schema migration check should cover that.

wesleyk: > There is an additional check we could do that all shards in the same keyspace have the same set of tables but the schema migration check should cover that.

Cool, this is more what I was wondering. Thanks!

PR #771, opened 2019-02-08
StartVitessService can reuse running Vitess server

mightyguava:
Gives a stable name for the Vitess docker container that `StartVitessService` starts, and allows the service to reuse an already running container with the same name, so that Vitess can be started in a separate process and kept running across dev server restart and test invocations. This is intended to help with rapid iteration of code changes without having to pay the cost of starting a Vitess server each time.

`startVitessDaemon` provides a simple hook for a `main` function to start a Vitess server in a separate process.

### DIFF
--- a/misk-hibernate/src/main/kotlin/misk/vitess/StartVitessService.kt
+++ b/misk-hibernate/src/main/kotlin/misk/vitess/StartVitessService.kt
@@ -41,7 +41,7 @@ import kotlin.reflect.KClass
 import kotlin.streams.toList
 
 const val VITESS_VERSION = "sha256:3cef0e042dee2312e5e0d80c56a6bf1581b52063ad0441551241f6785b110c38"
-const val CONTAINER_NAME_PREFIX = "cash-vitess-testing"
+const val CONTAINER_NAME_PREFIX = "misk-vitess-testing"
 
 class Keyspace(val sharded: Boolean) {
   // Defaulting to 2 shards for sharded keyspaces,
@@ -71,8 +71,9 @@ class VitessCluster(
           ?: throw IllegalStateException("vitess_schema_resource_root must be specified")
       val hasVschema = resourceLoader.walk(config.vitess_schema_resource_root)
           .any { it.endsWith("vschema.json") }
-      check(
-          hasVschema) { "schema root not valid, does not contain any vschema.json: ${config.vitess_schema_resource_root}" }
+      check(hasVschema) {
+        "schema root not valid, does not contain any vschema.json: ${config.vitess_schema_resource_root}"
+      }
       // We can't use Files::createTempDirectory because it creates a directory under the path
       // /var/folders that is not possible to mount in Docker
       schemaDir = Paths.get("/tmp/vitess_schema_${System.currentTimeMillis()}")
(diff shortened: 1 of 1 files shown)


misk-hibernate/src/main/kotlin/misk/vitess/StartVitessService.kt
```diff
@@ -69,7 +71,8 @@ class VitessCluster(
           ?: throw IllegalStateException("vitess_schema_resource_root must be specified")
       val hasVschema = resourceLoader.walk(config.vitess_schema_resource_root)
           .any { it.endsWith("vschema.json") }
-      check(hasVschema) { "schema root not valid, does not contain any vschema.json: ${config.vitess_schema_resource_root}"}
+      check(
+          hasVschema) { "schema root not valid, does not contain any vschema.json: ${config.vitess_schema_resource_root}" }
```
tirsen: super nit: I would cut the line after the `{` instead.

misk-hibernate/src/main/kotlin/misk/vitess/StartVitessService.kt
```diff
@@ -35,11 +35,13 @@ import java.nio.file.Paths
 import java.time.Duration
 import java.util.Properties
 import java.util.concurrent.TimeUnit
+import java.util.concurrent.atomic.AtomicBoolean
 import kotlin.concurrent.thread
 import kotlin.reflect.KClass
 import kotlin.streams.toList
 
 const val VITESS_VERSION = "sha256:3cef0e042dee2312e5e0d80c56a6bf1581b52063ad0441551241f6785b110c38"
+const val CONTAINER_NAME_PREFIX = "cash-vitess-testing"
```
tirsen: Call it `misk-vitess-testing` instead. We want misk to be generally useful even outside of Square.

misk-hibernate/src/main/kotlin/misk/vitess/StartVitessService.kt
```diff
@@ -175,25 +179,39 @@ class DockerVitessCluster(
...
-        .awaitStarted()
+    val containerName = "$CONTAINER_NAME_PREFIX-${cluster.name}"
+
+    val runningContainer = docker.listContainersCmd()
+        .withNameFilter(listOf(containerName))
+        .withLimit(1)
+        .exec()
+        .firstOrNull()
+    if (runningContainer != null) {
+      StartVitessService.logger.info("Existing Vitess cluster named $containerName found")
```
tirsen: NICE!

misk-hibernate/src/main/kotlin/misk/vitess/StartVitessDaemon.kt
```diff
@@ -0,0 +1,42 @@
...
+ * A helper method to start the vitess cluster outside of the dev server or test process, to enable
+ * rapid iteration. This should be called directly a `main()` function, for example:
+ *
+ *  fun main() {
+ *    val config = MiskConfig.load<MyAppConfig>("myapp")
+ *    startVitessDaemon(MyAppDb::class, config.data_source_clusters.values.first().writer)
+ *  }
+ *
+ */
+fun startVitessDaemon(
```
tirsen: nit: I reckon you can put this in the "companion" object inside `StartVitessService`. It doesn't quite feel like something small like this deserves an entire file all on its own. :-)

misk-hibernate/src/main/kotlin/misk/vitess/StartVitessService.kt
```diff
@@ -134,6 +137,7 @@ class DockerVitessCluster(
   private var containerId: String? = null
 
   private var isRunning = false
+  private var stopContainerOnExit = true
```
wesleyk: which test will actually stop the container? If I'm reading this correctly, the first test will start up the container and also have `stopContainerOnExit`, meaning when it ends it'll shut down the container as well?

mightyguava: This will stop the container on JVM exit, as the `stop()` method is called by the JVM shutdown hook. `stopContainerOnExit` is set to `false` if the docker container was not started by this JVM, to allow it to keep running across test runs.

misk-hibernate/src/main/kotlin/misk/vitess/StartVitessService.kt
mightyguava: This was Intellij...

misk-hibernate/src/main/kotlin/misk/vitess/StartVitessService.kt
mightyguava: Oops, I had it as misk at one point

misk-hibernate/src/main/kotlin/misk/vitess/StartVitessService.kt
mightyguava: Actually let me add a little detail for posterity:

All Vitess clusters used by the app/test are tracked by `StartVitessService` in a global cache (look for the `CacheBuilder`). The cache is invalidated by a JVM shutdown hook. On invalidation, the cache's `removalListener` will call the each entry's `DockerVitessCluster`'s `stop()` method. If `stopContainerOnExit` is set to true, it will stop and remove the container.

wesleyk: Thanks! Want to add these details in the javadoc :-)

mightyguava: Sure.

PR #772, opened 2019-02-08
Re-use ZK service across test runs in the same VM

wesleyk:


misk-zookeeper/src/test/kotlin/misk/clustering/zookeeper/ZkTestModule.kt
```diff
@@ -29,20 +29,32 @@ internal class ZkTestModule : KAbstractModule() {
...
-  class StartZookeeperService : AbstractIdleService(), DependentService {
+  @Singleton class StartZookeeperService : AbstractIdleService(), DependentService {
     override val consumedKeys: Set<Key<*>> = setOf()
     override val producedKeys: Set<Key<*>> = setOf(startZkServiceKey)
-    private val sharedZookeeper = EmbeddedZookeeper(zkPortKey)
 
     override fun startUp() {
-      log.info { "starting zk instance " }
-      sharedZookeeper.start()
+      if (!hasStarted) {
```
mightyguava: Depending on how aggressively gradle is parallelizing, you may want to make this an `AtomicBoolean`?

misk-zookeeper/src/test/kotlin/misk/clustering/zookeeper/ZkTestModule.kt
```diff
@@ -29,20 +30,31 @@ internal class ZkTestModule : KAbstractModule() {
...
+        })
+      } else {
+        log.info { "zk instance already running" }
+      }
     }
 
     override fun shutDown() {
-      log.info { "stopping zk instance" }
-      sharedZookeeper.stop()
+      // Shutdown happens as a runtime shutdown hook
```
hamdanjaveed: This is so that ZK doesn't shutdown prematurely before all tests are done running?

wesleyk: before this change, every test spun up the service. With this change, all tests within the same vm will reuse the same zk service

wesleyk:
> LG. I bet we'll want to make this a first class concept for other test services to use

Shouldn't you be snowboarding

wesleyk:
> LG. I bet we'll want to make this a first class concept for other test services to use

Good point, will think more about this

PR #773, opened 2019-02-09
Introduce CachedTestService to speedup tests

wesleyk:
To reuse services across test runs in a given
runtime. We use this for the docker sqs service
and the zk test service.

### DIFF
--- a/misk-testing/src/main/kotlin/misk/service/CachedTestService.kt
+++ b/misk-testing/src/main/kotlin/misk/service/CachedTestService.kt
@@ -1,0 +1,73 @@
+package misk.service
+
+import com.google.common.util.concurrent.AbstractIdleService
+import misk.logging.getLogger
+import java.util.concurrent.ConcurrentHashMap
+import java.util.concurrent.atomic.AtomicBoolean
+
+/**
+ * Test services can derive from CachedTestService
+ * if they'd like to reuse the same service for the span
+ * of a given runtime. This is helpful when you want to avoid
+ * incurring the cost of service startup and shutdown with
+ * each test run.
+ *
+ * NOTE: The caching is only useful if the implementing service
+ * references a shared instance of their underlying resources.
+ * A common way to do this is to leverage a companion object.
+ *
+ * Example:
+ *
+ * @Singleton class TestService : CachedTestService() {
+ *   override fun actualStartup() {
+ *     service.start()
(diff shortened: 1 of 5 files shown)


misk-testing/src/main/kotlin/misk/service/CachedTestService.kt
```diff
@@ -0,0 +1,71 @@
+package misk.service
+
+import com.google.common.util.concurrent.AbstractIdleService
+import misk.logging.getLogger
+import java.util.concurrent.atomic.AtomicBoolean
+
+/**
+ * Test services can implement CachedTestService
```
mmihic: nit: derive from not implement.

misk-testing/src/main/kotlin/misk/service/CachedTestService.kt
```diff
@@ -0,0 +1,71 @@
...
+        log.info { "$javaClass.name already running, not starting anything" }
+      }
+  }
+
+  final override fun shutDown() {
+    log.info { "$javaClass.name is cached, will shutdown on runtime shutdown" }
+  }
+
+  companion object {
+    private var hasStarted = AtomicBoolean(false)
```
mmihic: why var?

mmihic: also making it a member of the companion object means it's shared across everything that derives from `CachedTestService` - is that the intent, or is there supposed to be one `hasStarted` per service type?

wesleyk: ah right that's not the intent, and this build is going to fail... lemme think about this some more.

wesleyk: thinking of having a concurrent map to maintain a bool per service

wesleyk: > why var?

ugh I don't write enough kotlin. Changed to val

wesleyk: the tests passed, but that's because right now the two usages are across two different modules. Good catch! Pushing fix in a sec.

wesleyk: updated and added a test case

PR #774, opened 2019-02-11
Enable gradle build caching for CI

wesleyk:
Turns out we weren't actually using build caches! The travis-CI config we had was the travis-side of storing/loading the cache results, but the cache was never being populated in the first place.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,7 @@ install:
   - docker pull squareup/misk-web
 
 script:
-  - ./gradlew test --parallel --scan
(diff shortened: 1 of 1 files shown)


wesleyk:
Sample retried build ran in 1m11s: https://scans.gradle.com/s/ghsp3xfh5tq3y/.
Sample build with one test changed in the sqs jobqueue module: https://scans.gradle.com/s/5cqcwtcxovqfy/performance/execution

The test change caused the build to be unable to leverage any of the cache... still more to look into there, but this is still a good start we want to leverage

PR #775, opened 2019-02-11
Remove eager misk/web pull

wesleyk:
@adrw: It looks like with a recent bump of misk-web, the version that should be docker-pulled is 0.1.3. Thus, this install command was causing us to perform two docker pulls of misk-web. One for latest and one for 0.1.3. I'm not really seeing why we want this in the install step anyways instead of happening as part of the gradle build, so I'm removing it for now to reduce the extra build time.

Example build log where the pull happens twice, latest master: https://travis-ci.org/square/misk/builds/491356844

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,9 +3,6 @@ language: java
 jdk:
   - openjdk11
 
-install:
-  - docker pull squareup/misk-web
-
 script:
   - ./gradlew test --parallel --scan
 


adrw:
The eager pull allows Docker to pull the common layers of the image (ie. `node:alpine10`...etc). If it happens non-blocking this can speed up but it looks like it's happening synchronously right now which doesn't help so I'm okay with the removal 👍

PR #777, opened 2019-02-12
Add functions to get/set values as ByteArrays in misk-redis.

hamdanjaveed:
These functions can be used to store arbitrary arrays of bytes in Redis. A service may
want to do this to store protos for example.

### DIFF
--- a/misk-redis/src/main/kotlin/misk/redis/FakeRedis.kt
+++ b/misk-redis/src/main/kotlin/misk/redis/FakeRedis.kt
@@ -1,6 +1,7 @@
 package misk.redis
 
 import misk.time.FakeClock
+import okio.ByteString
 import java.time.Duration
 import java.time.Instant
 import java.util.concurrent.ConcurrentHashMap
@@ -10,17 +11,19 @@ import javax.inject.Inject
 class FakeRedis : Redis {
   @Inject lateinit var clock: FakeClock
 
-  // The value type stored in our key-value store
+  // The value type stored in our key-value store.
   private data class Value (
-    val value: String,
+    val data: ByteString,
     val expiryInstant: Instant
   )
 
-  // Acts as the Redis key-value store
+  // Acts as the Redis key-value store.
   private val keyValueStore = ConcurrentHashMap<String, Value>()
 
   override fun del(key: String): Boolean {
-    if (!keyValueStore.containsKey(key)) return false
+    if (!keyValueStore.containsKey(key)) {
+      return false
+    }
 
     return keyValueStore.remove(key) != null
   }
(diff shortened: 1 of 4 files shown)


misk-redis/src/main/kotlin/misk/redis/FakeRedis.kt
```diff
@@ -10,17 +10,39 @@ import javax.inject.Inject
 class FakeRedis : Redis {
   @Inject lateinit var clock: FakeClock
 
-  // The value type stored in our key-value store
+  // The value type stored in our key-data store. Data is always stored as ByteArrays.
+  // We manually define an equals and hashcode function as kotlin uses referential equality when comparing arrays,
+  // whereas we care about structural equality.
   private data class Value (
-    val value: String,
+    val data: ByteArray,
```
ryanhall07: should we just use okio `ByteString` instead? not a fan of the odd equality nature of ByteArray

misk-redis/src/main/kotlin/misk/redis/FakeRedis.kt
```diff
@@ -31,6 +53,11 @@ class FakeRedis : Redis {
   }
 
   override fun get(key: String): String? {
```
ryanhall07: does it even make sense to support this method? if you use a `ByteString`, a client can easily call `utf8()` to do the conversion and it's explicit

misk-redis/src/main/kotlin/misk/redis/FakeRedis.kt
```diff
@@ -39,18 +66,51 @@ class FakeRedis : Redis {
...
+    keyValueStore[key] = Value(
+      data = value,
+      expiryInstant = clock.instant().plusSeconds(expiryDuration.seconds)
+    )
+
+    return StatusCode.OK.toString()
+  }
+
+  companion object {
+    // Status codes returned by redis operations
```
wesleyk: kind of strange that we return a string "OK" for successful operations. Is that useful for clients? Are they expected to assert that the response of the operation `.equals("OK")`? What does RealRedis return?

My recommendation would be to expect success if an exception is not thrown.

wesleyk: and perhaps return the set value?

misk-redis/src/main/kotlin/misk/redis/FakeRedis.kt
hamdanjaveed: Yeah I was conflicted with this, the reason I supported it is because storing a string & retrieving a string is a common use case, but I'm not against doing what you said either since it'd simplify the api quite a bit. I'll follow your suggestion.

misk-redis/src/main/kotlin/misk/redis/FakeRedis.kt
hamdanjaveed: My reasoning here was to have `FakeRedis` behave like `RealRedis`. Maybe the correct thing to do here is to modify `RealRedis` to have this behaviour instead (return the set value instead of returning the redis response as a string).

PR #780, opened 2019-02-12
Expose datasource defaults

mightyguava:
In local testing, we have a convention of providing partial configs expecting misk to fill in the defaults. This should probably change in the future...

For now though, I need access to the defaults per data source type and would like to not duplicate them, so I exposed a `withDefaults()` method. I also moved `buildJdbcUrl()` out of the enum and into the `DataSourceConfig` to make this work, and I think it's probably actually cleaner this way.


wesleyk:
Nice PR description! That would make a great commit message as well

PR #783, opened 2019-02-14
Use hibernate and kotlin magic to determine if a value is nullable at…

jjestrel:
… the kotlin level as well as the hibernate level when doing schema validation


jjestrel:
one day

PR #789, opened 2019-02-22
Poll ZK for lease changes

mightyguava:
With addition of listener support in #753, the ZK lease manager actually
now needs to poll for changes to leases in addition to receiving cluster
watch events. The node can receive cluster membership changes about
another member exiting the cluster, but before that member releases its
leases.

This change also changes the Lease check to be a bit nicer when there
are no members in the cluster. It will assume it doesn't own the lease
and log a warning, rather than logging an error and ignoring a potential
state transition. This should only have an effect on logs on service
startup/shutdown when there 0 services running.

### DIFF
--- a/misk-zookeeper/src/main/kotlin/misk/clustering/zookeeper/ZkLease.kt
+++ b/misk-zookeeper/src/main/kotlin/misk/clustering/zookeeper/ZkLease.kt
@@ -1,10 +1,10 @@
 package misk.clustering.zookeeper
 
+import misk.clustering.NoMembersAvailableException
 import misk.clustering.lease.Lease
 import misk.logging.getLogger
 import org.apache.zookeeper.CreateMode
 import org.apache.zookeeper.KeeperException
-import org.apache.zookeeper.ZooDefs
 import java.util.concurrent.locks.ReentrantLock
 import javax.annotation.concurrent.GuardedBy
 import kotlin.concurrent.withLock
@@ -94,7 +94,8 @@ internal class ZkLease(
         Status.HELD -> listener.afterAcquire(this)
         // We don't know if we own the lease, so check
         Status.UNKNOWN -> checkHeld()
-        else -> {}
+        else -> {
+        }
       }
     }
   }
@@ -195,8 +196,13 @@ internal class ZkLease(
   /** @return true if we should hold the lease per the cluster membership */
   private fun shouldHoldLease(): Boolean {
     val clusterSnapshot = manager.cluster.snapshot
-    val desiredLeaseOwner = clusterSnapshot.resourceMapper[leaseResourceName]
-    return desiredLeaseOwner.name == clusterSnapshot.self.name
+    val desiredLeaseOwner = try {
+      clusterSnapshot.resourceMapper[leaseResourceName]
+    } catch (e: NoMembersAvailableException) {
+      log.warn { e.message }
+      null
+    }
+    return desiredLeaseOwner?.name == clusterSnapshot.self.name
   }
 
   /** @return true if the lease node exists in zk */
--- a/misk-zookeeper/src/main/kotlin/misk/clustering/zookeeper/ZkLeaseManager.kt
+++ b/misk-zookeeper/src/main/kotlin/misk/clustering/zookeeper/ZkLeaseManager.kt
@@ -2,19 +2,21 @@ package misk.clustering.zookeeper
 
 import com.google.common.annotations.VisibleForTesting
 import com.google.common.util.concurrent.AbstractExecutionThreadService
-import com.google.common.util.concurrent.ThreadFactoryBuilder
 import misk.DependentService
 import misk.clustering.Cluster
 import misk.clustering.lease.LeaseManager
 import misk.config.AppName
 import misk.inject.keyOf
 import misk.logging.getLogger
+import misk.tasks.RepeatedTaskQueue
+import misk.tasks.Result
+import misk.tasks.Status
 import misk.zookeeper.SERVICES_NODE
 import org.apache.curator.framework.CuratorFramework
 import org.apache.curator.framework.state.ConnectionStateListener
+import java.time.Duration
 import java.util.concurrent.ConcurrentHashMap
 import java.util.concurrent.CountDownLatch
-import java.util.concurrent.Executors
 import java.util.concurrent.LinkedBlockingQueue
 import java.util.concurrent.TimeUnit
 import java.util.concurrent.atomic.AtomicBoolean
@@ -29,6 +31,7 @@ import javax.inject.Singleton
 @Singleton
 internal class ZkLeaseManager @Inject internal constructor(
   @AppName appName: String,
+  @ForZkLease private val taskQueue: RepeatedTaskQueue,
   internal val cluster: Cluster,
   curator: CuratorFramework
 ) : AbstractExecutionThreadService(), LeaseManager, DependentService {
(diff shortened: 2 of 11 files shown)


misk/src/main/kotlin/misk/clustering/fake/ExplicitClusterResourceMapper.kt
```diff
@@ -30,6 +31,9 @@ class ExplicitClusterResourceMapper : ClusterResourceMapper {
   }
 
   override fun get(resourceId: String): Cluster.Member {
+    if (mappings.isEmpty()) {
+      throw NoMembersAvailableException(resourceId)
```
mightyguava: Didn't know if this should just return null or throw exception. Throwing exception seemed better here since Kotlin doesn't like nulls

misk-zookeeper/src/test/kotlin/misk/clustering/zookeeper/ZkTestModule.kt
```diff
@@ -26,10 +32,24 @@ internal class ZkTestModule : KAbstractModule() {
...
+        return RepeatedTaskQueue.forTesting("zk-lease-poller", clock, delayQueue)
+      }
+
+      @Provides @ForZkLease
+      fun provideDelayQueue(): ExplicitReleaseDelayQueue<DelayedTask> = ExplicitReleaseDelayQueue()
+    }))
   }
 
-  @Singleton class StartZookeeperService : CachedTestService(), DependentService {
+  @Singleton
```
mightyguava: Intellij autoformat, based on our shared style settings.

misk-zookeeper/src/main/kotlin/misk/clustering/zookeeper/ZkLeaseManager.kt
```diff
@@ -65,19 +66,17 @@ internal class ZkLeaseManager @Inject internal constructor(
...
-
     // When we are disconnected from zk, inform all of the leases that their state is now unknown
     client.value.connectionStateListenable.addListener(ConnectionStateListener { _, state ->
       actionQueue.put {
         handleConnectionStateChanged(state.isConnected)
       }
     })
+
+    taskQueue.schedule(leasePollInterval) {
+      checkAllLeases()
```
mmihic: Can this throw? Since you aren't scheduling with backoff it'll just re-poll at the same interval, but will log an error with an exception which might page which I'm not sure that we want?

misk-zookeeper/src/main/kotlin/misk/clustering/zookeeper/ZkLeaseManager.kt
```diff
@@ -65,19 +66,17 @@ internal class ZkLeaseManager @Inject internal constructor(
...
-    }
-
     // When we are disconnected from zk, inform all of the leases that their state is now unknown
     client.value.connectionStateListenable.addListener(ConnectionStateListener { _, state ->
       actionQueue.put {
         handleConnectionStateChanged(state.isConnected)
       }
     })
+
+    taskQueue.schedule(leasePollInterval) {
```
mmihic: Downside here is that we've made lease checking sequential whereas before on cluster change we would check for changes in ownership for each lease in parallel. Probably doesn't matter much given that only a small number of leases are likely to change ownership state (thus necessitating calls to zk) on each pass.

misk/src/main/kotlin/misk/clustering/NoMembersAvailableException.kt
```diff
@@ -0,0 +1,7 @@
+package misk.clustering
+
+/**
+ * Thrown if the cluster does not have any members available.
+ */
+class NoMembersAvailableException(val resourceId: String) :
```
mmihic: 👍

misk/src/main/kotlin/misk/clustering/fake/ExplicitClusterResourceMapper.kt
mmihic: Agree, it's better to have an explicit exception vs a null which most callers would just convert into a generic and uninformative NPE

misk-zookeeper/src/main/kotlin/misk/clustering/zookeeper/ZkLeaseManager.kt
mightyguava: This cannot throw. There's blanket exception handling here https://github.com/square/misk/blob/master/misk-zookeeper/src/main/kotlin/misk/clustering/zookeeper/ZkLease.kt#L80

misk-zookeeper/src/main/kotlin/misk/clustering/zookeeper/ZkLease.kt
```diff
@@ -195,8 +196,13 @@ internal class ZkLease(
   /** @return true if we should hold the lease per the cluster membership */
   private fun shouldHoldLease(): Boolean {
     val clusterSnapshot = manager.cluster.snapshot
-    val desiredLeaseOwner = clusterSnapshot.resourceMapper[leaseResourceName]
-    return desiredLeaseOwner.name == clusterSnapshot.self.name
+    val desiredLeaseOwner = try {
+      clusterSnapshot.resourceMapper[leaseResourceName]
+    } catch (e: NoMembersAvailableException) {
```
ryanhall07: can you update the javadoc for `ClusterResourceMapper` to say it throws the exception?

misk-zookeeper/src/main/kotlin/misk/clustering/zookeeper/ZkLeaseManager.kt
mightyguava: That's what I was thinking too. I could have kept the cached thread executor, but since it's polling every second and there's unlikely to be much i/o, I figured sequential is better.

ryanhall07: my first thought is we just schedule a task for every lease. however, the set of leases is not static, so that's kind of tricky.

misk-zookeeper/src/test/kotlin/misk/clustering/zookeeper/ZkLeaseTest.kt
```diff
@@ -236,8 +236,8 @@ internal class ZkLeaseTest {
     reset(listenerMock)
 
     // Further calls to checkHeld() should not trigger events because the lease does not change
-    lease.checkHeld()
-    lease.checkHeld()
+    assertThat(lease.checkHeld()).isTrue()
```
wesleyk: nice

misk-zookeeper/src/main/kotlin/misk/clustering/zookeeper/ZkLeaseManager.kt
```diff
@@ -40,9 +43,7 @@ internal class ZkLeaseManager @Inject internal constructor(
 
   private val ownerName = cluster.snapshot.self.name
   private val actionQueue = LinkedBlockingQueue<() -> Unit>()
-  private val leaseCheckExecutor = Executors.newCachedThreadPool(ThreadFactoryBuilder()
-      .setNameFormat("lease-monitor-%d")
-      .build())
+  private val leasePollInterval: Duration = Duration.ofSeconds(1L)
```
wesleyk: haven't been following too closely, is there any particular science around 1 second?

mightyguava: `checkHeld()` will check the in-memory map of cluster members to figure out if it should own the lease before attempting to talk to ZK, so this is a pretty cheap check. It should happen as frequently as possible to ensure the lease is always held by some process. The map is updated by a cluster watch and should happen infrequently.

wesleyk: thanks for the explanation! (was more wondering 1 second vs 100ms, etc)

mightyguava: It could probably be 100ms, but I think we can tolerate ~1 second of having no one owning a lease? Service startup/shutdown is on order of seconds anyways, so an ephemeral ZK lease node could take a few seconds to get deleted after shutdown is triggered anyways.

wesleyk: mm yeah, that's a good way to put it!

PR #790, opened 2019-02-22
Add ability for Vitess to specify client keystore on JDBC connections

nbroyles:


misk-hibernate/src/main/kotlin/misk/jdbc/DataSourceConfig.kt
```diff
@@ -99,7 +101,34 @@ data class DataSourceConfig(
...
+        var queryParams = ""
+        var useSSL = false
+        /**
+         * Query params for VitessJDBC driver look like default MySQL JDBC driver query params
+         * but are named slightly differently and a smaller subset are supported. See
+         * [io.vitess.jdbc.VitessJDBCUrl] for the complete list
+         */
+        if (!config.trust_certificate_key_store_url.isNullOrBlank()) {
+          require(!config.trust_certificate_key_store_password.isNullOrBlank()) {
+            "must provide a trust_certificate_key_store_password"
```
wesleyk: maybe include "if trust_certificate_key_store_url is set"

misk-hibernate/src/main/kotlin/misk/jdbc/DataSourceConfig.kt
```diff
@@ -99,7 +101,34 @@ data class DataSourceConfig(
...
+          }
+          queryParams += "${if (queryParams.isEmpty()) "?" else "&"}trustStore=${config.trust_certificate_key_store_url}"
```
wesleyk: man github struggles with this syntax highlighting 😬

misk-hibernate/src/main/kotlin/misk/jdbc/DataSourceConfig.kt
```diff
@@ -99,7 +101,34 @@ data class DataSourceConfig(
         "jdbc:hsqldb:mem:${database!!};sql.syntax_mys=true"
       }
       DataSourceType.VITESS -> {
-        "jdbc:vitess://${config.host}:${config.port}/${config.database}"
+        var queryParams = ""
```
mightyguava: I'd just do `queryParams = "?"` here. Should be legal to have a `?` with no params

nbroyles: I'm gonna play it safe for now and leave as-is as I don't have a good way to test that `?` is safe (even though I agree w/ your assessment). Going to work on getting my test app hooked up to a vitess and then I'll try it out.

PR #803, opened 2019-02-27
Add misk-crypto module

yoavamit:
This PR adds the ability to load crypto keys from a configuration file, and use these keys to encrypt and decrypt data in the app.

In order to use a key, one would have to generate it using Google's `tinkey` tool and encrypt it using a KMS key.
Keys are being loaded from the `CryptoModule` class, decrypted using the same KMS client and are used to initialize a `Cipher` object which can be injected using Guice.


misk-crypto/README.md
```diff
@@ -0,0 +1,59 @@
+This module provides Misk with the ability to store and manage cryptographic key material in a
```
wesleyk: really nice docs, thank you!

misk-crypto/src/main/kotlin/misk/crypto/Cipher.kt
```diff
@@ -0,0 +1,49 @@
...
+import com.google.crypto.tink.aead.AeadFactory
+import java.io.ByteArrayOutputStream
+import java.util.Base64
+
+/**
+ * An even simpler interface than Tink's Aead.
+ * This also makes it easier to mock Ciphers for testing purposes.
+ */
+interface Cipher {
+  fun encrypt(plaintext: ByteArray?): ByteArray?
```
wesleyk: why accept a nullable byte array as a param?

misk-crypto/src/main/kotlin/misk/crypto/Cipher.kt
```diff
@@ -0,0 +1,49 @@
...
+  fun decrypt(ciphertext: ByteArray?): ByteArray?
+  fun keyInfo(): Map<String, String>
```
wesleyk: could we document this method?

misk-crypto/src/main/kotlin/misk/crypto/CryptoModule.kt
```diff
@@ -0,0 +1,36 @@
...
+ */
+class CryptoModule(
+  private val config: CryptoConfig,
+  private val kmsClient: KmsClient
+) : KAbstractModule() {
+
+  override fun configure() {
+    AeadConfig.register()
+    for (keyConfig: Key in config.keys) {
+      val keyUri = config.gcp_key_uri ?: "aws-kms://alias/${config.aws_kms_key_alias}"
```
wesleyk: should we assert that only one is set?

misk-crypto/src/test/resources/test-config.yaml
```diff
@@ -0,0 +1,4 @@
+aws_kms_key_alias: "master_key_name"
```
wesleyk: worth mentioning here or somewhere else how you generated this?

samples/exemplar/src/test/kotlin/com/squareup/exemplar/ActionsTestModule.kt
```diff
@@ -0,0 +1,20 @@
+package com.squareup.exemplar
+
+import com.google.inject.name.Names
+import misk.crypto.Cipher
+import misk.inject.KAbstractModule
+
+class ActionsTestModule : KAbstractModule() {
```
wesleyk: fwiw in this case I'd say you could use constructor injection for the action class and then pass in the FakeCipher. IMO that'd be easier to follow

misk-crypto/src/main/kotlin/misk/crypto/Cipher.kt
wesleyk: as well as defining what keys are set? is returning a more explicit object with tink and encryptedKey fields more clear here?

misk-crypto/README.md
wesleyk: let's chat in #misk but it would be great if you could create a crypto-exemplar service in staging that demonstrated the functionality out. We've been adding exemplar services that demonstrate various platform offerings as an easy way for folks to view/reuse example code.

misk-crypto/README.md
```diff
@@ -0,0 +1,59 @@
...
+  keys:
+    - key_name: "my_payment_token_key"
+      encrypted_kek: [key encrypted using the same KMS used by the app encoded in base64] 
+``` 
+Using a key
+-----
+To use your newly created key:
+```$kotlin
+class PaymentTokenGenerator {
+  @Inject @Named("my_payment_token_key") lateinit val tokenCipher: Cipher
```
mightyguava: val can't be lateinit. Also can we move this into a primary constructor?

misk-crypto/README.md
```diff
@@ -0,0 +1,59 @@
...
+In order to initialize the module, the app must have a configured GcpKmsClient (Google),
+or an AwsKmsClient (Amazon) client in order to have access to the KMS.
+
+Setup
+-----
+In one of your app's module files:
+```$kotlin
+class MyAppModule : KAbstractModule {
+  override fun configure() {
+    val kmsClient = AwsKmsClient().withDefaultCredentials()
```
mightyguava: Can we make this provided by an AwsKmsClientModule or a GcpKmsClientModule, rather than asking the client to pass it in?

samples/exemplar/src/main/kotlin/com/squareup/exemplar/actions/HelloWebAction.kt
```diff
@@ -8,22 +11,39 @@ import misk.web.ResponseContentType
...
   @Get("/hello/{name}")
   @ResponseContentType(MediaTypes.APPLICATION_JSON)
   fun hello(
     @PathParam name: String,
     @Suppress("UNUSED_PARAMETER")
     @RequestHeaders headers: Headers,
-    @QueryParam nickName: String?,
-    @QueryParam greetings: List<String>?
+    @QueryParam greetings: List<String>?,
+    @QueryParam action: String?
```
mightyguava: Maybe the cipher example can be its own example. I think the HelloWebAction is supposed to be a bare minimum example.

misk-crypto/README.md
```diff
@@ -0,0 +1,59 @@
+This module provides Misk with the ability to store and manage cryptographic key material in a
+safe and easy way.
+
+How it works?
+-----
+The `Cipher` object is used to encrypt/decrypt data.
+
+It uses a key which is loaded from a configuration file at startup.
+The key associated with each cipher should be stored **encrypted** in the app's common configuration file.
```
alanpaulin: Does this live on the same host as the app? Does the app need to send the key to another host or service to decrypt it? That was a goal we had with the internal predecessor .

misk-crypto/src/main/kotlin/misk/crypto/Cipher.kt
mmihic: nit: does this need to be a method or should it be a `val`?

misk-crypto/README.md
mightyguava: Readability is about the same no? Downside to field injection is that the class can no longer be easily constructed without Guice.

```kotlin
class PaymentTokenGenerator @Inject constructor(
  @Named("my_payment_token_key") val tokenCipher:Cipher
) {
```

vs

```kotlin
class PaymentTokenGenerator() {
@Inject @Named("my_payment_token_key") lateinit var tokenCipher: Cipher
```

misk-crypto/src/main/kotlin/misk/crypto/Cipher.kt
```diff
@@ -0,0 +1,49 @@
...
+import java.io.ByteArrayOutputStream
+import java.util.Base64
+
+/**
+ * An even simpler interface than Tink's Aead.
+ * This also makes it easier to mock Ciphers for testing purposes.
+ */
+interface Cipher {
+  fun encrypt(plaintext: ByteArray?): ByteArray?
+  fun decrypt(ciphertext: ByteArray?): ByteArray?
```
swankjesse: Needs documentation. What does this do if the ciphertext fails to decrypt?

misk-crypto/src/main/kotlin/misk/crypto/Cipher.kt
```diff
@@ -0,0 +1,49 @@
...
+  fun keyInfo(): Map<String, String>
+}
+
+/**
+ * Real implementation of the Cipher interface.
+ * This class uses key (KeysetHandle) that's been decrypted by the KMS when the module was loaded.
+ */
+class CipherImpl internal constructor(
```
swankjesse: CipherImpl –> RealCipher

misk-crypto/src/main/kotlin/misk/crypto/FakeKmsClient.kt
```diff
@@ -0,0 +1,18 @@
...
+
+import com.google.crypto.tink.Aead
+import com.google.crypto.tink.KmsClient
+
+/**
+ * This fake implementation of a Key Management Service client is meant to be used
+ * for testing and development purposes only.
+ *
+ * When calling `getAead`, it'll provide a fake Aead object that **performs no encryption**.
+ * Instead, it only does Base64 encoding/decoding so developers could debug their apps.
```
swankjesse: Thumbs up

misk-crypto/src/main/kotlin/misk/crypto/FakeMasterEncryptionKey.kt
```diff
@@ -0,0 +1,15 @@
+package misk.crypto
+
+import com.google.crypto.tink.Aead
+import java.util.Base64
+
+class FakeMasterEncryptionKey : Aead {
```
swankjesse: Need this be public?

misk-crypto/src/main/kotlin/misk/crypto/KeyManager.kt
```diff
@@ -0,0 +1,18 @@
+package misk.crypto
+
+/**
+ * Singleton object used as a map of key name to its corresponding Cipher implementation.
+ * It's being populated by the {@link CryptoModule} class when loaded.
+ */
+object KeyManager {
```
swankjesse: No no no no . . . use a `@Singleton` class. Stateful objects are very bad because state leaks across tests

misk-crypto/src/main/kotlin/misk/crypto/KeyManager.kt
```diff
@@ -0,0 +1,18 @@
...
+
+  private var keys: HashMap<String, Cipher> = HashMap()
```
swankjesse: https://publicobject.com/2016/02/08/linkedhashmap-is-always-better-than-hashmap/

swankjesse: or `mutableMapOf<String, Cipher>()`

misk-crypto/src/test/kotlin/misk/crypto/CryptoModuleTest.kt
```diff
@@ -0,0 +1,39 @@
...
+    AeadConfig.register()
+  }
+
+  @Test
+  fun testImportKey() {
+    val keyHandle = KeysetHandle.generateNew(AeadKeyTemplates.AES256_GCM)
+    val encryptedKey = generateEncryptedKey(keyHandle)
+    val key = Key("test", encryptedKey)
+    val config = CryptoConfig(listOf(key), "doesn't matter")
+    Guice.createInjector(CryptoModule(config, FakeKmsClient()))
```
swankjesse: where did the injector go?

misk-crypto/src/test/kotlin/misk/crypto/CryptoModuleTest.kt
```diff
@@ -0,0 +1,39 @@
...
+    val testKey = KeyManager["test"]
+    Assertions.assertThat(testKey).isNotNull()
```
swankjesse: please import `Assertions.*` so this line starts with `assertThat()`

misk-crypto/src/test/resources/test-config.yaml
swankjesse: Or in the readme?

samples/exemplar/src/main/kotlin/com/squareup/exemplar/actions/HelloWebAction.kt
swankjesse: Yeah, encryption has no business here

samples/exemplar/src/main/resources/exemplar-common.yaml
```diff
@@ -3,4 +3,10 @@ web:
   idle_timeout: 30000
 
 prometheus:
-  http_port: 8081
\ No newline at end of file
+  http_port: 8081
+
+crypto:
```
swankjesse: please make a different sample specifically for encryption

samples/exemplar/src/test/kotlin/com/squareup/exemplar/HelloWebActionTest.kt
```diff
@@ -3,18 +3,31 @@ package com.squareup.exemplar
...
+  @Inject private lateinit var helloWebAction: HelloWebAction
 
   @Test
   fun happyPath() {
     assertThat(helloWebAction.hello("sandy", Headers.of(), null, null))
         .isEqualTo(HelloResponse("YO", "SANDY"))
   }
+
+  @Test
+  fun testWithEncryption() {
```
swankjesse: this test belongs in the misk-cipher package

misk-crypto/README.md
yoavamit: Good point!
That's actually what I'm working on right now, I want to be sure I can easily create and deploy an app that uses this module before I merge this PR.

misk-crypto/src/main/kotlin/misk/crypto/Cipher.kt
yoavamit: @wesleyk the `Map<String, String>` was basically just an experiment here TBH.
I plan to use this value later as  the output of an HTTP endpoint, and I'm assuming I'll add more fields to it later.
That why a map was a pretty comfortable choice here (though naive).

WDYT about putting this in a data class?

samples/exemplar/src/test/kotlin/com/squareup/exemplar/ActionsTestModule.kt
yoavamit: Agree, I'll do that in another sample app I'll generate instead of `examplar`

misk-crypto/src/main/kotlin/misk/crypto/Cipher.kt
wesleyk: yeah I'd be for using an actual data class.

wesleyk: moshi makes it easy to then expose that data class over an API if you so desire

misk-crypto/src/main/kotlin/misk/crypto/Cipher.kt
yoavamit: fixed both

misk-crypto/src/main/kotlin/misk/crypto/KeyManager.kt
yoavamit: 👍

misk-crypto/src/main/kotlin/misk/crypto/FakeKmsClient.kt
yoavamit: :)

misk-crypto/README.md
yoavamit: That's a good point, and also something I've been battling with since I started writing this module.
As I see it, the client must provide some kind of configuration/object that says "hi this where my KMS is, here's how to access it" - be it AWS credentials or GCP (or even custom thing, IDK).
We could probably automate and wrap this with an `AwsKmsClientModule` or something, but having that makes a few assumptions about the environment (like if there's an environment variable or a file in a certain path containing credentials)

I feel like it'd be a good addition to provide some kind of way to automatically get some an object, but I think it'd be better to put it in our internal misk implementation.
Because:
1. As I said, I want this to be env-agnostic as possible
2. We'll probably have better tools and infrastructure around this to make it super easy to integrate with AWS.

misk-crypto/README.md
yoavamit: I'm not sure I understand. Are you referring to how the key is being decrypted?
If so, then the way it works is:
- When the app loads, it instantiates a `Aead` object via the KMS.
- Then, for each encrypted key in the configuration file:
  - Decrypt the key using the above `Aead` and create a new `Cipher` object with the decrypted result.

The key point here is that the decrypted key is only available in memory and is not being written anywhere to disk.

misk-crypto/README.md
```diff
@@ -0,0 +1,60 @@
...
+) {
+  
+  fun encryptToken(token: String): ByteString {
+    return tokenCipher.encrypt(token.encodeUtf8())
+  }
+  
+  fun decryptToken(encryptedToken: ByteString): String {
+    // using the key manager to get the right cipher by name
+    return KeyManager["my_payment_token_key"]
+        .decrypt(encryptedToken.toByteString())
```
swankjesse: don’t need `toByteString()`

misk-crypto/src/main/kotlin/misk/crypto/Cipher.kt
```diff
@@ -0,0 +1,69 @@
...
+import java.util.Base64
+
+/**
+ * An even simpler interface than Tink's [Aead].
+ * This also makes it easier to mock Ciphers for testing purposes.
+ */
+interface Cipher {
+  /**
+   * Encrypt the given data.
+   * If anything goes wrong, throws a [java.security.GeneralSecurityException]
```
swankjesse: Is it possible for anything to go wrong?

swankjesse: This seems to me like one of those things where theoretically the VM might not have AEAD but in practice it never happens, and it’s a waste to write code to recover from that. 

(So don’t document it)

misk-crypto/src/main/kotlin/misk/crypto/Cipher.kt
```diff
@@ -0,0 +1,69 @@
...
+   */
+  fun encrypt(plaintext: ByteString): ByteString
+
+  /**
+   * Decrypt the given data
+   * If anything goes wrong, including failing to decipher the data,
+   * throws a [java.security.GeneralSecurityException]
```
swankjesse: I think you might want a different behavior for decryption failure? Maybe a better exception or it could return a `ByteString?`

yoavamit: The interface Google provides in Tink specifically says that these methods may throw a `GeneralSecurityException`.
I think it'd be a good approach to stick with those exceptions. hiding/swallowing these exceptions could be problematic and could even lead to data corruption.
I think that if the app fails to encrypt/decrypt anything, it's either going to be something widespread (like a bad key was setup somehow), or as you said, probably won't happen at all.
In any case, I think you want the app to blow-up in these cases.

misk-crypto/src/main/kotlin/misk/crypto/Cipher.kt
```diff
@@ -0,0 +1,68 @@
...
+
+/**
+ * An even simpler interface than Tink's [Aead].
+ * This also makes it easier to mock Ciphers for testing purposes.
+ */
+interface Cipher {
+  /**
+   * Encrypt the given data.
+   */
+  fun encrypt(plaintext: ByteString): ByteString
```
ddz: The methods in this interface omit the associated data arguments to encrypt and decrypt supported by Tink's Aead interface. The associated data used by AEADs can be used to add integrity validation of non-secret data in a single pass for higher performance and less moving parts than having to hash them separately. If there isn't any associated data, they can just be null for both encrypt and decrypt. What is the trade-off here for not surfacing those capabilities from misk-crypto?

ddz: Also: does using `ByteString` objects here cause additional copies of the plaintext to be created on the heap? Normally, crypto libraries try to be careful to not make additional copies of plaintext data so that the caller can overwrite the plaintext after calling `encrypt` and are done with it.

yoavamit: Sorry I didn't address this earlier.

I think that for now, having the simple encrypt/decrypt interface with just the data should be enough, but I'm definitely open to expanding this API to do validation as well.
I didn't add it here for now, because I want to keep this initial PR simple and with few features.
We can add this and also the HMAC validation as well, after we'll get a little bit of milage on this module.

As for the `ByteString` objects, I'm not really sure, but I bet @swankjesse knows about it.

misk-crypto/README.md
alanpaulin: Makes sense. I think the important thing is that if someone were to get access to the data on disk, they wouldn't be able to decrypt the encrypted data (because they don't have access to the KMS.

PR #807, opened 2019-03-01
Change MiskCallerAuthenticator from a multi to single binding

wesleyk:
since only one is used at a time.


misk/src/main/kotlin/misk/security/authz/MiskCallerAuthenticator.kt
```diff
@@ -6,13 +6,6 @@ import misk.MiskCaller
  * Interface for determining the current [MiskCaller]. Typically use an [ActionScoped] [Request],
  * [ActionScoped] [ClientCertSubject], etc to determine the caller based on request headers or
  * client certificate information.
- *
- * This interface is used with Guice multibindings. Register instances by calling `multibind()`
- * in a `KAbstractModule`:
- *
- * ```
- * multibind<MiskCallerAuthenticator>().to<MyAuthenticator>()
```
adrw: Should probably leave this documentation in but change it to bind?

PR #809, opened 2019-03-02
Web Actions Tab 0.1

adrw:
* Responsive, interactive UI for introspecting bound WebActions
* Shows following metadata
  * Name
  * Path
  * Function
  * HTTP Method
  * Access Annotation
  * Request -> Response Content Types
  * Application Interceptors
  * Network Interceptors

ezgif com-video-to-gif

screen shot 2019-03-01 at 8 10 24 pm


misk/web/tabs/web-actions/src/containers/WebActionsContainer.tsx
```diff
@@ -0,0 +1,237 @@
...
+
+const RequestResponeContentTypes = (props: { action: IWebAction }) => (
+  <span>
+    <span>{props.action.requestMediaTypes}</span>{" "}
+    <Icon icon={IconNames.ARROW_RIGHT} />{" "}
+    <span>{props.action.responseMediaType}</span>
+  </span>
+)
+
+enum HttpMethod {
```
wesleyk: seems a bit strange that we have to re-define http methods. Is this nowhere else?

misk/web/tabs/web-actions/src/containers/WebActionsContainer.tsx
```diff
@@ -0,0 +1,237 @@
...
+  PATCH: Intent.SUCCESS,
+  POST: Intent.SUCCESS,
+  PUT: Intent.SUCCESS
+}
+
+const filterAnnotations = (functionAnnotations: string[]) =>
+  functionAnnotations.filter(
+    a =>
+      !(
+        a.includes("RequestContentType") ||
```
wesleyk: what's the significance of these?

misk/web/tabs/web-actions/src/containers/WebActionsContainer.tsx
adrw: Agreed, it should be upstreamed to `@misk/core`, couldn't find it at first look in any of our dependency libraries that we could use out of the box.

misk/web/tabs/web-actions/src/containers/WebActionsContainer.tsx
adrw: It filters out annotations that contain information that will already be displayed with dedicated panels (ie. don't need to show the AccessAnnotation because there's an `Access` box).

wesleyk: ah gotcha. I think we'd probably want to filter this out server side if we think clients will find this not useful compared to other information provided by the API. Not a blocker for now!

adrw: Sounds good! Definitely should upstream to the WebAction any filters / manipulations once we figure out what we'll need in the final UI 👍

PR #813, opened 2019-03-04
Use <Spinner/> instead of error components for less alarming loading

adrw:


wesleyk:
as discussed offline, let's keep things as is until we have better load -> timeout support

adrw:
Shelving until can be upstreamed properly into Misk-Web https://github.com/square/misk-web/issues/153

PR #814, opened 2019-03-05
Don't bind the StartVitessService if the configured database is not Vitess

jmuia:


misk-hibernate/src/main/kotlin/misk/hibernate/HibernateModule.kt
```diff
@@ -159,13 +160,15 @@ class HibernateModule(
...
+
     val environment = Environment.fromEnvironmentVariable()
-    if (environment == Environment.DEVELOPMENT || environment == Environment.TESTING) {
-      val startVitessServiceKey = StartVitessService::class.toKey(qualifier)
-      multibind<Service>().to(startVitessServiceKey)
-      bind(startVitessServiceKey).toProvider(Provider<StartVitessService> {
-        StartVitessService(environment = environment, config = config, qualifier = qualifier)
-      }).asSingleton()
-    }
+    if (environment !in setOf(Environment.DEVELOPMENT, Environment.TESTING)) return
```
wesleyk: /nit mind adding some brackets on these one-liners? I find them rather confusing and am forever scared of https://nakedsecurity.sophos.com/2014/02/24/anatomy-of-a-goto-fail-apples-ssl-bug-explained-plus-an-unofficial-patch/

jmuia: good story! will change

jmuia:
A bit hairier than I expected. The `VitessScaleSafetyChecks` are also bound even if the test database isn't Vitess.

misk-hibernate-testing/src/main/kotlin/misk/hibernate/HibernateTestingModule.kt
```diff
@@ -21,49 +23,53 @@ import kotlin.reflect.KClass
  */
 class HibernateTestingModule(
   private val qualifier: KClass<out Annotation>,
+  private val config: DataSourceConfig,
```
drashkov: Looks good, but this will break services on the next update. Do we want to update everything, or make this optional?

jmuia: Yep, good call out. I was planning to do a big batch fix once merged.

PR #815, opened 2019-03-05
Add HikariCP metrics

jmuia:


misk-hibernate/src/main/kotlin/misk/jdbc/DataSourceService.kt
```diff
@@ -22,7 +24,8 @@ internal class DataSourceService(
   private val qualifier: KClass<out Annotation>,
   private val config: DataSourceConfig,
   private val environment: Environment,
-  private val dataSourceDecorators: Set<DataSourceDecorator>
+  private val dataSourceDecorators: Set<DataSourceDecorator>,
+  private val collectorRegistry: CollectorRegistry? = null
```
mightyguava: Why is this nullable?

wesleyk: I think because not every client of misk-hibernate will have metrics installed?

mightyguava: This change requires all usages of HibernateModule to have metrics installed, since the binding is here https://github.com/square/misk/pull/815/files#diff-212704a7927dd09259da3fe20775f0b0R96. It's not optional.

mightyguava: Actually yeah, this is a breaking change for HibernateModule. @jmuia you might want to make adding the registry a `DataSourceDecorator` bound in a different module instead.

wesleyk: rather surprising that the tests go green

mightyguava: Hmm I think by default you are getting an implicit binding with a new instance of CollectorRegistry, which is almost certainly not what should get injected.

mmihic:
Inject `misk.metrics.Metrics` which is bound as a singleton, and we can expose its `CollectorRegistry` - that way you are using the same registry instance as the rest of the system

misk-hibernate/src/main/kotlin/misk/jdbc/DataSourceService.kt
jmuia: I'll take a look at `DataSourceDecorator`. Misk installs metrics by default in `MiskCommonServiceModule`, which is why the tests are ok.

I made it nullable because some tests were creating the DataSourceService manually, but I can refactor a bit and have them pass in the CollectorRegistry.

jmuia:
> Inject `misk.metrics.Metrics` which is bound as a singleton, and we can expose its `CollectorRegistry` - that way you are using the same registry instance as the rest of the system

@mmihic the `CollectorRegistry` is also bound as a singleton. I can still update though so we have the higher-level interface available?

misk-hibernate/src/main/kotlin/misk/jdbc/DataSourceService.kt
jmuia: `CollectorRegistry` is explicitly bound in `MetricsModule` as a singleton.

mightyguava: Yes, but if you don't install `MetricsModule()`, then someone using `HibernateModule()` but not `MetricsModule()` will get an implicit binding which will construct a new instance. When #816 lands, it'll be an error I have to fix...

mightyguava:
@mmihic should we just HibernateModule require a binding to `misk.metrics.Metrics`? That would let us avoid having to fuddle around with optional injection or data source decorators. It wouldn't really impact internal use since Skim would bind it, but seems not so great for open source use.

mmihic:
> @mmihic should we just HibernateModule require a binding to `misk.metrics.Metrics`? That would let us avoid having to fuddle around with optional injection or data source decorators. It wouldn't really impact internal use since Skim would bind it, but seems not so great for open source use.

I think that's reasonable. Most other "upper level components" require a binding to this for their own metrics, so having `HibernateModule` do the same is imo fine. `misk.metrics.Metrics` _is_ the mechanism for accessing metrics

jmuia:
Ended up making it optional, since it was a simple change.
Updated it to use `Metrics` rather than `CollectorRegistry` directly.

Issue #794, opened 2019-02-25
Explicit config binding


mightyguava:
Currently, any subconfigs in the app's `Config` class are bound automatically by [ConfigModule](https://github.com/square/misk/blob/master/misk/src/main/kotlin/misk/config/ConfigModule.kt) via reflection, by discovering which sub-`Config`s are on the app `Config`. While this is convenient, there are a few important consequences

- It's difficult for a developer not intimately familiar with Guice and reflection to trace where a Config is bound, unless they had prior knowledge that the `ConfigModule` bound it all automatically
- It forces a heavy reliance on implicit injections for configs. If an app installs `FooModule()` and `BarModule()`, and the app does not define `FooConfig` and `BarConfig` in its app `Config`, `ConfigModule` will not bind these subconfigs, and Guice will just inject the implicit binding, if the config is fully covered by default values. This introduces a lot of inconsistency in tests, and developers end up unknowingly relying the implicit binding of the config (a "default" config basically). This can and has caused confusion, and makes it difficult/impractical to enable https://github.com/square/misk/issues/784.
- All subconfigs are available for injection, whether that is intended by the module design or not. For example, `HibernateModule` is meant to be qualified with an annotation, so it's unlikely that a top-level DataSourceClustersConfig/DataSourceConfig should be made injectable unqualified
- The name for the subconfig in the app `Config` matters. If the name of the subconfig in the YAML is not the lower_underscore case of its class name, it will get bound with an `@Named` annotation instead. Classes try try to inject this config without the qualifier will again get the implicit binding instead.

I'd like us to start explicitly binding sub-configs, making it the responsibility of the `Module` consuming the config object to choose to whether to bind it or not. Guice is magical enough. Misk should not make it more magical, but rather be an examplar of what is the best way to use Guice.

Before:
```kotlin
override func configure() {
  install(ConfigModule.create("my-app", myAppConfig)
  install(MiskWebModule())
  install(KubernetesClusterModule())
}
```

After:
```kotlin
override func configure() {
  install(ConfigModule.create("my-app", myAppConfig)
  install(MiskWebModule(myAppConfig.web_config))
  install(KubernetesClusterModule(myAppConfig.kubernetes_cluster_config))
}
```

The downside to the explicit config binding approach is of course that there will need to be additional code added to each config, and a ton of threading things through.

ryanhall07:
what does installing the `ConfigModule` bind in the after?

mightyguava:
[...]

Issue #784, opened 2019-02-14
Require explicit bindings in Guice


mightyguava:
I've run into more than a few cases where Guice injected an implicit binding rather than the qualified binding I expected, either because I forgot to bind something, or because of Kotlin magic where an `@Qualifier` that didn't have the right `@Target` annotations was removed by the compiler without warning (where you need `@field:Annotation` syntax). Seems like others have dealt with the same issues.

I think it'd be nicer if we just turned on [`requireExplicitBindings`](https://google.github.io/guice/api-docs/latest/javadoc/com/google/inject/Binder.html#requireExplicitBindings--), which requires any injected key to be explicitly bound in a module. It would save a lot of debugging headache as the codebase grows in complexity.

Unfortunately, to enable it, we'd have to go and fix implicit bindings in all usages, which might take a while.

mightyguava:
Blocked by #794

mightyguava:
We ended up using requireAtInjectionOnConstructors() instead of requireExplicitBindings(), which gets the benefit of making it difficult to accidentally inject an implicit binding, but doesn't add too much overhead with module configs.

PR #816, opened 2019-03-05
Require configs to be explicitly bound and enable requireExplicitInject()

mightyguava:
This addresses #794 and #784. See those issues for context on why.

The changes made here are:

1. c30e7ec removes the code in `ConfigModule` that walks the app's `Config` object and binds its child `Config` classes automatically. This required a small refactoring of usages of Config.
2. 891dfd6 calls `binder().requireExplictBindings()` in `MiskApplication` and `MiskTestExtension`, and fixes breakages in the misk main module.
3. The remaining commits fix usages in other modules.

A particular note here is that in addition to enabling require explict injections, there's a breaking API change to action binding. Every action was registered as a `WebActionEntry` multibinding, that required an implicit binding for the action. With explicit bindings enabled, the multibinding of 

```
multibind<WebActionEntry>().toInstance(WebActionEntry<ReturnADinosaurAction>())
```

would need an extra line

```
multibind<WebActionEntry>().toInstance(WebActionEntry<ReturnADinosaurAction>())
bind<ReturnADinosaurAction()>`
```

This makes for an unpleasant API, so instead I un-deprecated the `WebActionModule` to wrap the calls. This has the additional benefit that we no longer need the explicit binding verification for actions in Skim's `ServiceBuilder`.

This PR could probably have been split for #794 and #784, since I'll need to fix breakages in every app regardless and the config part is pretty small, I just put them together.


misk/src/test/kotlin/misk/web/resource/StaticResourceActionTest.kt
```diff
@@ -211,34 +210,28 @@ class StaticResourceActionTest {
 
   class TestModule : KAbstractModule() {
     override fun configure() {
+      install(HttpClientsConfigModule(HttpClientsConfig(
```
adrw: I'm not as experienced in Guice but would having this provided (as before) be faster in tests than being installed?

mightyguava: What do you mean faster? Like how quickly the test runs? It'll be the same

adrw: Yeah test run time. Sounds good 👍

misk/src/main/kotlin/misk/web/WebActionModule.kt
```diff
@@ -5,27 +5,77 @@ import misk.web.actions.WebAction
...
+    }
+
+    /**
+     * Registers a web action with a path prefix.
+     * @param url_path_prefix: defaults to "/". If not empty, must match pattern requirements:
+     *   - must begin with "/"
+     *   - any number of non-whitespace characters (including additional path segments or "/")
+     *   - must terminate with a non-"/" because rest of path will start with "/"
+     */
+    inline fun <reified A : WebAction> createWithPrefix(url_path_prefix: String): WebActionModule<A> =
```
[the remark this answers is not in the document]
mightyguava: There are less methods yes, but I don't think it's a simpler API. It's about what reads better:

```
WebModuleAction.create<DinosaurAction>("/foo/bar")
WebModuleAction.createWithPrefix<DinosaurAction>("/foo/bar")
```

The latter clearly indicates that the string you pass in is a prefix.

misk/src/main/kotlin/misk/web/WebActionModule.kt
```diff
@@ -5,27 +5,77 @@ import misk.web.actions.WebAction
...
 import kotlin.reflect.KClass
 
-@Suppress("DEPRECATION")
 class WebActionModule<A : WebAction> private constructor(
-  val webActionClass: KClass<A>
+  val actionClass: KClass<A>,
+  val url_path_prefix: String
 ) : KAbstractModule() {
   override fun configure() {
-    multibind<WebActionEntry>().toInstance(WebActionEntry(webActionClass))
```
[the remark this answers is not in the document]
mightyguava: The List<WebActionEntry> binding doesn't change.

misk/src/main/kotlin/misk/security/authz/FakeCallerAuthenticator.kt
```diff
@@ -40,14 +42,4 @@ class FakeCallerAuthenticator @Inject constructor(
...
-annotation class DevelopmentOnly
-
-/**
- * https://github.com/google/guice/wiki/FrequentlyAskedQuestions#how-can-i-inject-optional-parameters-into-a-constructor
- */
-@Singleton
-class OptionalBinder {
-  @com.google.inject.Inject(optional = true)
-  @DevelopmentOnly
-  var developmentCaller: MiskCaller? = null
```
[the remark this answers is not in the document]
mightyguava: Yeah, the API hasn't changed. The `OptionalBinder` pattern lets you pass an optional injection into a constructor. I moved it to be an optional injection on a field instead, which is what `OptionalBinder` does itself. In misk, there's no benefit to do constructor optional injection, since we use Guice to bind tests too.

misk/src/main/kotlin/misk/web/WebActionModule.kt
```diff
@@ -5,27 +5,77 @@ import misk.web.actions.WebAction
...
+    bind(actionClass.java)
+    multibind<WebActionEntry>().toInstance(WebActionEntry(actionClass, url_path_prefix))
```
adrw: Missed this, nice 👍

misk/src/main/kotlin/misk/web/WebActionModule.kt
```diff
@@ -5,27 +5,78 @@ import misk.web.actions.WebAction
...
 class WebActionModule<A : WebAction> private constructor(
-  val webActionClass: KClass<A>
+  val actionClass: KClass<A>,
+  val url_path_prefix: String
 ) : KAbstractModule() {
   override fun configure() {
-    multibind<WebActionEntry>().toInstance(WebActionEntry(webActionClass))
+    multibind<WebActionEntry>().toInstance(WebActionEntry(actionClass, url_path_prefix))
+    // Ensures that the action has an @Inject annotation and that its dependencies are satisfied
+    getProvider(actionClass.java)
```
swankjesse: 👍🏻

misk/src/main/kotlin/misk/web/WebActionModule.kt
```diff
@@ -5,27 +5,78 @@ import misk.web.actions.WebAction
...
   }
 
   companion object {
-    @Deprecated("use multibind<WebActionEntry>().toInstance(WebActionEntry(...))")
+    /**
+     * Registers a web action.
```
swankjesse: https://publicobject.com/2007/08/05/stop_writing_redundant_javadoc/ ?

misk/src/main/kotlin/misk/web/WebActionModule.kt
```diff
@@ -5,27 +5,78 @@ import misk.web.actions.WebAction
...
+     */
+    fun <A : WebAction> create(actionClass: KClass<A>): WebActionModule<A> {
+      return WebActionModule(actionClass, "/")
+    }
+
+    /**
+     * Registers a web action with a path prefix.
+     * @param url_path_prefix: defaults to "/". If not empty, must match pattern requirements:
```
swankjesse: rather than copy-pasting this 3x can you link to a single definition?

misk/src/test/kotlin/misk/inject/MultiBinderTest.kt
```diff
@@ -0,0 +1,42 @@
+package misk.inject
+
+import misk.testing.MiskTest
+import misk.testing.MiskTestModule
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.jupiter.api.Test
+import javax.inject.Inject
+
+@MiskTest
+class MultiBinderTest {
```
swankjesse: Nice

misk-aws/src/test/kotlin/misk/jobqueue/sqs/SqsJobQueueTest.kt
```diff
@@ -236,7 +236,7 @@ internal class SqsJobQueueTest {
       multibind<Service>().to<DockerSqs.Service>()
       install(MiskTestingServiceModule())
       install(MockTracingBackendModule())
-      install(Modules.override(AwsSqsJobQueueModule()).with(SQSTestModule()))
+      install(Modules.override(AwsSqsJobQueueModule(AwsSqsJobQueueConfig())).with(SQSTestModule()))
```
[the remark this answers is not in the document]
mightyguava: I'd prefer to not use default values for configs. It'll make it more likely for developers to forget to pass them through in production when they just install a module. If they are sure they want the default, they should instantiate it explicitly.

misk-eventrouter/src/main/kotlin/misk/eventrouter/RealEventRouterModule.kt
```diff
@@ -56,7 +56,6 @@ class RealEventRouterModule(val environment: Environment) : KAbstractModule() {
  * not run multiple enqueued tasks concurrently! Instead it should have exactly 1 thread always.
  */
 @Qualifier
-@Retention(AnnotationRetention.RUNTIME)
```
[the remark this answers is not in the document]
mightyguava: Yeah I haven't found a need for it at least.

misk/src/main/kotlin/misk/web/WebActionModule.kt
mightyguava: kotlindoc doesn't support linking to overloaded methods, but I'll just delete the redundant ones.

mightyguava:
Looks like we have overall consensus. Merging this for now since it's like a ticking merge conflict bomb. Feel free to continue commenting and I'll follow up with changes.

PR #818, opened 2019-03-05
Add support for providing trust_certificate_key_store_path and client_certificate_key_store_path

nbroyles:
Some database drivers (e.g. Vitess) don't accept urls for certs in the
JDBC url they generate. So, for the sake of clarity, allow folks to
specify paths instead of urls. Fields ending in url imply that the string
begins with a scheme, so to ensure that clients do the right thing,
provide the path fields as an alternative.

### DIFF
--- a/misk-hibernate/src/main/kotlin/misk/jdbc/DataSourceConfig.kt
+++ b/misk-hibernate/src/main/kotlin/misk/jdbc/DataSourceConfig.kt
@@ -49,6 +49,10 @@ data class DataSourceConfig(
   val trust_certificate_key_store_password: String? = null,
   val client_certificate_key_store_url: String? = null,
   val client_certificate_key_store_password: String? = null,
+  // Vitess driver doesn't support passing in URLs so support paths and prefer this for Vitess
+  // going forward
+  val trust_certificate_key_store_path: String? = null,
+  val client_certificate_key_store_path: String? = null,
   val show_sql: String? = "false"
 ) {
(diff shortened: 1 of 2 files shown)


misk-hibernate/src/main/kotlin/misk/jdbc/DataSourceConfig.kt
```diff
@@ -103,24 +124,31 @@ data class DataSourceConfig(
       DataSourceType.VITESS -> {
         var queryParams = ""
         var useSSL = false
+
+        // NOTE(nb): still support url properties in Vitess for backwards compatibility.
```
wesleyk: nb(nb)

wesleyk: IMO it's a little confusing to support both, do we plan on migrating clients and phasing support out?

wesleyk:
for my own understanding, this is purely for a more clear API?

misk-hibernate/src/main/kotlin/misk/jdbc/DataSourceConfig.kt
nbroyles: Yeah, the plan is to migrate folks and phasing out the use of the url, but we need to support both until that day comes.

nbroyles:
> for my own understanding, this is purely for a more clear API?

Correct!

PR #822, opened 2019-03-06
Improve exception logging around database failures on service start up

nbroyles:


misk-hibernate/src/main/kotlin/misk/jdbc/PingDatabaseService.kt
```diff
@@ -8,12 +8,15 @@ import misk.backoff.ExponentialBackoff
...
 import misk.environment.Environment
 import misk.inject.toKey
+import misk.logging.getLogger
 import java.time.Duration
 import java.util.Properties
 import javax.inject.Inject
 import javax.inject.Singleton
 import kotlin.reflect.KClass
 
+private val logger = getLogger<PingDatabaseService>()
```
mightyguava: This should be in an companion object

PR #823, opened 2019-03-06
Web Actions Tab 0.1.3 - 0.1.4

adrw:
* Move client side processing of metadata to Redux so it is not render blocking
* Skeleton Loading Component instead of using `<Spinner/>`

### DIFF
--- a/misk/web/tabs/web-actions/src/containers/WebActionsContainer.tsx
+++ b/misk/web/tabs/web-actions/src/containers/WebActionsContainer.tsx
@@ -16,51 +16,18 @@ import {
 } from "@blueprintjs/core"
 import { IconNames } from "@blueprintjs/icons"
 import { FlexContainer } from "@misk/core"
-import {
-  onChangeToggleFnCall,
-  simpleSelect,
-  simpleType
-} from "@misk/simpleredux"
+import { onChangeToggleFnCall, simpleSelect } from "@misk/simpleredux"
 import { HTTPMethod } from "http-method-enum"
-import { chain, flatMap } from "lodash"
 import * as React from "react"
 import { connect } from "react-redux"
 import styled from "styled-components"
-import { IDispatchProps, IState, rootDispatcher, rootSelectors } from "../ducks"
-
-export interface IWebActionAPI {
-  allowedServices: string[]
-  allowedRoles: string[]
-  applicationInterceptors: string[]
-  dispatchMechanism: HTTPMethod
-  function: string
-  functionAnnotations: string[]
-  name: string
-  networkInterceptors: string[]
-  parameterTypes: string[]
-  pathPattern: string
-  requestMediaTypes: string[]
-  responseMediaType: string
-  returnType: string
-}
-
-export interface IWebActionInternal {
-  allowedServices: string
-  allowedRoles: string
-  applicationInterceptors: string[]
-  authFunctionAnnotations: string[]
-  dispatchMechanism: HTTPMethod[]
-  function: string
-  functionAnnotations: string[]
-  name: string
-  networkInterceptors: string[]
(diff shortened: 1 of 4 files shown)


misk/web/tabs/web-actions/src/containers/WebActionsContainer.tsx
```diff
@@ -288,81 +256,64 @@ export const FilterWebActions = (props: IState & IDispatchProps) => {
     </div>
   )
 }
+const SkeletonText = () => (
+  <span className={Classes.SKELETON}>{"Lorem ipsum"}</span>
```
wesleyk: hmm will this ever show up to devs?

misk/web/tabs/web-actions/src/containers/WebActionsContainer.tsx
```diff
@@ -289,80 +257,18 @@ export const FilterWebActions = (props: IState & IDispatchProps) => {
...
-      {/* <FilterWebActions {...props} /> */}
-    </div>
-  )
+const WebActionsContainer = (props: IState & IDispatchProps) => {
+  const metadata = simpleSelect(props.webActions, "metadata")
+  if (metadata.length > 0) {
+    return (
+      <div>
+        <WebActions metadata={metadata} {...props} />
+        {/* <FilterWebActions {...props} /> */}
```
wesleyk: this is confusing, is this commented out?

misk/web/tabs/web-actions/src/containers/WebActionsContainer.tsx
adrw: If computation takes too long, usually may flash prior to being filled in with the real data. Similar to the Skeleton table we already use in SampleTableComponent in Misk-Web/Palette

https://square.github.io/misk-web/examples/tabs/palette/demo/

misk/web/tabs/web-actions/src/containers/WebActionsContainer.tsx
adrw: Yes commenting with `JSX` is a bit confusing.

misk/web/tabs/web-actions/src/containers/WebActionsContainer.tsx
wesleyk: why not put no txt?

misk/web/tabs/web-actions/src/containers/WebActionsContainer.tsx
wesleyk: so this is a comment, why are we leaving this here?

adrw: Will remove 👍

misk/web/tabs/web-actions/src/ducks/webActions.ts
```diff
@@ -90,8 +134,82 @@ function* handleDinosaur(action: IAction<WEBACTIONS, IWebActionsPayload>) {
...
+          action => action.dispatchMechanism
+        )
+        const mergedAction = actions[0]
+        mergedAction.dispatchMechanism = dispatchMechanism.sort().reverse()
+        return mergedAction
+      })
+      .sortBy(["name", "pathPattern"])
+      .value()
+    // TODO(adrw) build index of keyspace for filterable fields
+    // const index = chain(metadata).value()
```
wesleyk: same comment about not committing commented out code like this

misk/web/tabs/web-actions/src/containers/WebActionsContainer.tsx
adrw: A bit unintuitive at first glance but allows a way to mock the final UI when the data has loaded: https://blueprintjs.com/docs/#core/components/skeleton

PR #826, opened 2019-03-07
Web Actions Tab: Trigger Inline Requests

adrw:
ezgif-4-742d119b66da

### DIFF
--- a/misk/web/tabs/web-actions/src/containers/WebActionsContainer.tsx
+++ b/misk/web/tabs/web-actions/src/containers/WebActionsContainer.tsx
@@ -1,5 +1,6 @@
 import {
   Button,
+  ButtonGroup,
   Card,
   Classes,
   Collapse,
@@ -7,17 +8,24 @@ import {
   H3,
   H5,
   Icon,
+  InputGroup,
   Intent,
   Menu,
   MenuItem,
   Pre,
   Spinner,
   Tag,
+  TextArea,
   Tooltip
 } from "@blueprintjs/core"
 import { IconNames } from "@blueprintjs/icons"
 import { FlexContainer } from "@misk/core"
-import { onChangeToggleFnCall, simpleSelect } from "@misk/simpleredux"
+import {
+  onChangeFnCall,
+  onChangeToggleFnCall,
+  onClickFnCall,
+  simpleSelect
+} from "@misk/simpleredux"
 import { HTTPMethod } from "http-method-enum"
 import * as React from "react"
 import { connect } from "react-redux"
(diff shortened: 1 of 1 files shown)


misk/web/tabs/web-actions/src/containers/WebActionsContainer.tsx
```diff
@@ -322,7 +448,19 @@ const WebActionsContainer = (props: IState & IDispatchProps) => {
...
+      <div>
+        <SkeletonWebActions />
+        <br />
+        <SkeletonWebActions />
+        <br />
+        <SkeletonWebActions />
+        <br />
+        <SkeletonWebActions />
+        <br />
+        <SkeletonWebActions />
```
hamdanjaveed: Are these webaction stubs? Is this saying that in the case when there are no webactions for a service, we'll see 5 stubbed web actions? If thats the behaviour maybe some text saying "No web actions" would suffice?

hamdanjaveed:
Some comments briefly explaining what the major components are would be helpful (would help with connecting the coded representation to the DOM element that it's describing).

misk/web/tabs/web-actions/src/containers/WebActionsContainer.tsx
adrw: These are [Blueprint Skeletons](https://blueprintjs.com/docs/#core/components/skeleton) that are equivalent in purpose to a loading spinner. If the Web Actions endpoint hasn't completed yet, then this will show up to pre-write the UI. Once the data is available, it just fills in. If there are no bound Web Actions then the screen would be empty (though this won't be the case since all containers must provide readiness, liveness, and status endpoints).

hamdanjaveed: Ah got it

misk/web/tabs/web-actions/src/containers/WebActionsContainer.tsx
```diff
@@ -151,91 +196,208 @@ const MethodTag = (props: { method: HTTPMethod }) => (
...
+      "data"
+    )}
+  >
+    <InputGroup
+      onChange={onChangeFnCall(
+        props.simpleFormInput,
+        `${tag}::${props.action.pathPattern}::URL`
+      )}
+      value={
+        simpleSelect(
```
wesleyk: pretty long conditional here, is it easy to split these out into their own vars?

wesleyk:
very cool!

misk/web/tabs/web-actions/src/containers/WebActionsContainer.tsx
adrw: Fixed in #827

PR #828, opened 2019-03-07
Introduce TestAdminDashboardModule that test environments can actually install

wesleyk:


misk/src/main/kotlin/misk/web/metadata/AdminDashboardModule.kt
```diff
@@ -111,4 +112,15 @@ class AdminDashboardModule(val environment: Environment) : KAbstractModule() {
       ))
     }
   }
+}
+
+// Module that allows testing/development environments to bind up the admin dashboard
+class AdminDashboardTestingModule(val environment: Environment): KAbstractModule() {
```
adrw: Should this be moved out to its own file?

wesleyk: we've been putting testing ones in the same file (at least in the case of skim)

PR #830, opened 2019-03-08
Web Actions Tab: Fix All/None Roles/Services bug

adrw:
* `Unauthenticated` annotation wasn't correctly determining "All" or "None" for the Services and Roles metadata

### DIFF
--- a/misk/web/tabs/web-actions/src/ducks/webActions.ts
+++ b/misk/web/tabs/web-actions/src/ducks/webActions.ts
@@ -159,7 +159,7 @@ function* handleMetadata() {
             )
         )
         const emptyAllowedArrayValue =
-          authFunctionAnnotations.length > 1 &&
+          authFunctionAnnotations.length > 0 &&
           authFunctionAnnotations[0].includes("Unauthenticated")
             ? "All"
             : "None"


misk/web/tabs/web-actions/src/ducks/webActions.ts
```diff
@@ -159,7 +159,7 @@ function* handleMetadata() {
             )
         )
         const emptyAllowedArrayValue =
-          authFunctionAnnotations.length > 1 &&
+          authFunctionAnnotations.length > 0 &&
           authFunctionAnnotations[0].includes("Unauthenticated")
```
wesleyk: above we only filter `authFunctionAnnotations` to include "Access" or "authz", would this ever be true then?

adrw: I'm not sure if we have a strict one `Access` or `authz` annotation per web action, if it is enforced as strictly one then we can restructure that processing so it assumes there will only ever be one match.

wesleyk: let's sync up offline about this

PR #831, opened 2019-03-08
Web Actions Tab: More Accurate Coalescing

adrw:
* `groupBy` that allowed for coalescing of web actions with the same `pathPattern` is improved to group on all uniquely identifying metadata so that bound web actions with the same `pathPattern` and unique other metadata are shown as separate entries
* Add `uniq` filter on `dispatchMechanism` so that web actions installed in duplicate do not have multiple same HTTP Method tags displayed on the `webAction`

### DIFF
--- a/misk/web/tabs/web-actions/src/ducks/webActions.ts
+++ b/misk/web/tabs/web-actions/src/ducks/webActions.ts
@@ -6,7 +6,7 @@ import {
 } from "@misk/simpleredux"
 import axios from "axios"
 import { HTTPMethod } from "http-method-enum"
-import { chain, flatMap } from "lodash"
+import { chain } from "lodash"
 import { all, AllEffect, call, put, takeLatest } from "redux-saga/effects"
 
 export interface IWebActionAPI {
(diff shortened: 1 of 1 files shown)


misk/web/tabs/web-actions/src/ducks/webActions.ts
```diff
@@ -183,12 +200,14 @@ function* handleMetadata() {
...
       })
-      .groupBy("pathPattern")
+      .groupBy(groupByWebActionHash)
       .map((actions: IWebActionInternal[]) => {
-        const dispatchMechanism = flatMap(
-          actions,
-          action => action.dispatchMechanism
-        )
+        const dispatchMechanism = chain(actions)
+          .flatMap(action => action.dispatchMechanism)
```
wesleyk: this should be handled server side, we should leave a TODO here to remove this duplicate processing

wesleyk:
> bound web actions with the same pathPattern and unique other metadata are shown as separate entries

Why do we want this? Can you share some examples?

adrw:
@wesleyk Example would be Static Resource Action which has a binding for each dashboard tab and has both Get and Post methods on them. Realistically, we could probably remove the Post annotation.

https://github.com/square/misk/blob/a4ed6ed7f3bbf4d6c31da525aa2310cab6bff2e4/misk/src/main/kotlin/misk/web/resources/StaticResourceAction.kt#L33-L41

PR #833, opened 2019-03-08
Web Actions Tab: Remove Tooltip from MetadataCollapse

adrw:
* Confusing interaction with a redundant tooltip on a Metadata item which when clicked reveals a Collapse component

### DIFF
--- a/misk/web/tabs/web-actions/src/containers/WebActionsContainer.tsx
+++ b/misk/web/tabs/web-actions/src/containers/WebActionsContainer.tsx
@@ -120,15 +120,27 @@ const Metadata = (
   props: {
     content: string | JSX.Element
     label: string
-    tooltip: string | JSX.Element
+    tooltip?: string | JSX.Element
   } & any
-) => (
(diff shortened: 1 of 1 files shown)


misk/web/tabs/web-actions/src/containers/WebActionsContainer.tsx
```diff
@@ -120,15 +120,27 @@ const Metadata = (
...
+        label={props.label}
+        text={<Tooltip content={props.tooltip}>{props.content}</Tooltip>}
+        {...props}
+      />
+    )
+  } else {
+    return (
+      <MenuItem
+        label={props.label}
+        text={<Tooltip>{props.content}</Tooltip>}
```
hamdanjaveed: Probably misunderstanding but do we want the `Tooltip` tag here?

adrw: An empty one provides identical padding and margin of the element.

PR #834, opened 2019-03-08
Use Wire gRPC to generate proto's service

oldergod:
- Generates protos, including services, to Kotlin,
- Use Wire's gRPC APIs to do the rest.
- Beautiful.


mightyguava:
Is wire gRPC compatible with vanilla gRPC?

swankjesse:
Yes!

PR #835, opened 2019-03-08
Don't do healthcheck if not all services started

mightyguava:
Don't do healthchecks if services haven't all started. The app isn't in a good state yet,
and check health could end up triggering random errors that we don't want to flood the
logs with.


misk/src/main/kotlin/misk/web/actions/ReadinessCheckAction.kt
```diff
@@ -24,17 +24,26 @@ class ReadinessCheckAction @Inject internal constructor(
...
   fun readinessCheck(): Response<String> {
     val servicesNotRunning = services.filter { !it.isRunning }
+
+    for (service in servicesNotRunning) {
+      logger.info("Service not running: $service")
+    }
+
+    if (!servicesNotRunning.isEmpty()) {
+      // Don't do healthchecks if services haven't all started. The app isn't in a good state yet,
+      // and check health could end up triggering random errors that we don't want to flood the
```
wesleyk: "and a health check"?

PR #836, opened 2019-03-08
Web Actions Tab: Basic Filtering

adrw:
Screen Shot 2019-03-08 at 6 30 41 PM

### DIFF
--- a/misk/web/tabs/web-actions/src/containers/WebActionsContainer.tsx
+++ b/misk/web/tabs/web-actions/src/containers/WebActionsContainer.tsx
@@ -16,7 +16,8 @@ import {
   Spinner,
   Tag,
   TextArea,
-  Tooltip
+  Tooltip,
+  HTMLSelect
 } from "@blueprintjs/core"
 import { IconNames } from "@blueprintjs/icons"
 import { FlexContainer } from "@misk/core"
@@ -26,6 +27,7 @@ import {
   onClickFnCall,
   simpleSelect
 } from "@misk/simpleredux"
+import { chain } from "lodash"
 import { HTTPMethod } from "http-method-enum"
 import * as React from "react"
 import { connect } from "react-redux"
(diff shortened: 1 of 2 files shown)


misk/web/tabs/web-actions/src/ducks/webActions.ts
```diff
@@ -197,10 +209,11 @@ function* handleMetadata() {
           authFunctionAnnotations,
           dispatchMechanism: [action.dispatchMechanism],
           function: action.function.split("fun ").pop(),
+          hash: groupByWebActionHash(action),
           nonAccessOrTypeFunctionAnnotations
         }
       })
-      .groupBy(groupByWebActionHash)
+      .groupBy("hash")
```
wesleyk: this is pretty magical, is there something more deterministic to use here?

misk/web/tabs/web-actions/src/containers/WebActionsContainer.tsx
```diff
@@ -430,64 +444,91 @@ const SkeletonText = () => (
...
+          {...props}
+        />
       </div>
     )
   } else {
     // Displays mock of 5 Web Action cards which fill in when data is available
     return (
       <div>
-        <SkeletonWebActions />
+        <SkeletonWebActions {...props} />
```
wesleyk: why do the mocks need this props reference?

misk/web/tabs/web-actions/src/ducks/webActions.ts
```diff
@@ -43,6 +44,15 @@ export interface IWebActionInternal {
   returnType: string
 }
 
+export const WebActionInternal: { [key: string]: string } = {
```
wesleyk: small comment explaining what this is?

misk/web/tabs/web-actions/src/ducks/webActions.ts
adrw: The `groupByWebActionHash` function previously referenced now saves the value in a key in the object called `"hash"`. Maybe a different named key would seem less magical.

wesleyk: we're relying on the fact that the key is "hash". Can we name it something more relevant and extract that to not a magic string?

misk/web/tabs/web-actions/src/containers/WebActionsContainer.tsx
adrw: Initially had put in so the new filter form would be in the mock but realized I did it wrong so going to fix it and push.

PR #837, opened 2019-03-11
Add onSessionClose hook

qelper:
Opening up another transaction in the postCommitHook returns stale session which is already closed. onSessionClose hook allow for new transactions to be opened as part of the hook.

### DIFF
--- a/misk-hibernate/src/main/kotlin/misk/hibernate/Session.kt
+++ b/misk-hibernate/src/main/kotlin/misk/hibernate/Session.kt
@@ -29,6 +29,14 @@ interface Session {
    * hook will cause the transaction to be rolled back.
    */
   fun onPreCommit(work: () -> Unit)
+
+  /**
+   * Registers a hook that fires after a session is closed. This is called regardless if a session
+   * was successfully committed or rolled back.
+   *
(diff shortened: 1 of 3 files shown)


misk-hibernate/src/main/kotlin/misk/hibernate/Session.kt
```diff
@@ -29,6 +29,14 @@ interface Session {
    * hook will cause the transaction to be rolled back.
    */
   fun onPreCommit(work: () -> Unit)
+
+  /**
+   * Registers a hook that fires after a session is closed. This is called regardless if a session
+   * was successfully committed or rolled back.
```
wesleyk: do we think clients would want to disambiguate between successful vs failed sessions? What's an example usage of when onSessionClose is utilized?

qelper: > do we think clients would want to disambiguate between successful vs failed sessions? What's an example usage of when onSessionClose is utilized?

Yes, clients may benefit from it. So far my use case is to refresh the state from the database either on rollback or success so the distinction doesn't really matter too much for me. We can extend this to have more state in the callback if the use case grows.

PR #839, opened 2019-03-11
Add log if connecting to redis cluster failed

hamdanjaveed:


misk-redis/src/main/kotlin/misk/redis/RedisModule.kt
```diff
@@ -35,10 +37,19 @@ class RedisModule(
...
     // Authenticate with the redis server
-    jedisPool.resource.use {
-      it.auth(replicationGroup.redis_auth_password)
+    try {
+      jedisPool.resource.use { it.auth(replicationGroup.redis_auth_password) }
+    } catch(e: JedisConnectionException) {
+      log.error { "Failed to connect to redis cluster: " +
+        "${replicationGroup.writer_endpoint.hostname}:${replicationGroup.writer_endpoint.port}" }
+
+      e.printStackTrace()
```
ryanhall07: don't print the stack trace to the console, just log the exception above

misk-redis/src/main/kotlin/misk/redis/RedisModule.kt
```diff
@@ -35,10 +37,19 @@ class RedisModule(
     )
 
     // Authenticate with the redis server
-    jedisPool.resource.use {
-      it.auth(replicationGroup.redis_auth_password)
+    try {
```
ryanhall07: this behavior is different now since you swallowed the exception and did not propagate

hamdanjaveed: Re-throwing the exception now

misk-redis/src/main/kotlin/misk/redis/RedisModule.kt
```diff
@@ -35,10 +37,19 @@ class RedisModule(
...
+      jedisPool.resource.use { it.auth(replicationGroup.redis_auth_password) }
+    } catch(e: JedisConnectionException) {
+      log.error { "Failed to connect to redis cluster: " +
```
ryanhall07: an alternative approach is to throw a new exception that holds the context you need. that way you don't have a separate error message that is unconnected to the final stack trace.

```
class RedisConnectionException : RuntimeException(ReplicationGroup, Cause) {
  super("Failed to connect...", cause);
}

catch (e: JedisConnectionException) {
  throw RedisConnectionException(replicationGroup, e);
}
```

hamdanjaveed: Nice I like that, I'll switch to creating a new exception

misk-redis/src/main/kotlin/misk/redis/RedisModule.kt
```diff
@@ -35,10 +37,17 @@ class RedisModule(
...
+      jedisPool.resource.use { it.auth(replicationGroup.redis_auth_password) }
+    } catch(e: JedisConnectionException) {
+      throw RedisConnectionException(
+        "${replicationGroup.writer_endpoint.hostname}:${replicationGroup.writer_endpoint.port}", e)
     }
 
     return RealRedis(jedisPool)
   }
+
+  companion object {
```
ryanhall07: nit: remove the unnecessary log

ryanhall07:
also a general comment, you should refactor the `RedisModule` to be side effect free and not create redis connections from the constructor. I know you have it lazily provided with a Provider, but it would be even better to move the connection logic to a `start` method. This will require a wrapper object, that delays constructing the Redis object until the `start` method is called.

https://github.com/google/guice/wiki/ModulesShouldBeFastAndSideEffectFree

PR #841, opened 2019-03-11
Add module for binding AWS job queue handlers

mightyguava:
Right now, to register job queue handlers, you need to define a separate
service that will in its `startUp()` function subscribe to queues, which
are actually consumed by the `SqsJobConsumer`. This is kind of a hairy
interface.

With `AwsSqsJobQueueModule`, you just have to install the module to bind
a handler to a queue.

I spent some time trying to write tests for it, but haven't found a good
way to do so because of how difficult it is to manipulate the service
dependency graph in tests. There isn't a good way for the service that
registers the subscriptions to depend on both `DockerSqs` startup and
queue creation. Since the module is pretty simple I'm just going to
leave it without tests.

### DIFF
--- a/misk-aws/src/main/kotlin/misk/jobqueue/sqs/AwsSqsJobHandlerModule.kt
+++ b/misk-aws/src/main/kotlin/misk/jobqueue/sqs/AwsSqsJobHandlerModule.kt
@@ -1,0 +1,63 @@
+package misk.jobqueue.sqs
+
+import com.google.common.util.concurrent.AbstractIdleService
+import com.google.common.util.concurrent.Service
+import com.google.inject.Key
+import misk.DependentService
+import misk.inject.KAbstractModule
+import misk.jobqueue.JobHandler
+import misk.jobqueue.QueueName
+import javax.inject.Inject
+import javax.inject.Singleton
+import kotlin.reflect.KClass
+
+/**
+ * Install this module to register a handler for an SQS queue.
+ */
+class AwsSqsJobHandlerModule<T : JobHandler> private constructor(
+  private val queueName: QueueName,
+  private val handler: KClass<T>
+) : KAbstractModule() {
+  override fun configure() {
+    newMapBinder<QueueName, JobHandler>().addBinding(queueName).to(handler.java)
+    multibind<Service>().to<AwsSqsJobHandlerSubscriptionService>()
+  }
+
+  companion object {
+    inline fun <reified T : JobHandler> create(queueName: QueueName):
+        AwsSqsJobHandlerModule<T> = create(queueName, T::class)
+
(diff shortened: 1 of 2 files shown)


misk-aws/src/main/kotlin/misk/jobqueue/sqs/AwsSqsJobQueueModule.kt
```diff
@@ -17,6 +17,13 @@ import java.util.concurrent.Executors
 
 /** [AwsSqsJobQueueModule] installs job queue support provided by SQS */
 class AwsSqsJobQueueModule(private val config: AwsSqsJobQueueConfig) : KAbstractModule() {
+  override fun configure() {
+    multibind<Service>().to<SqsJobConsumer>()
+    install(AwsSqsJobQueueCommonModule(config))
+  }
+}
+
+internal class AwsSqsJobQueueCommonModule(private val config: AwsSqsJobQueueConfig) : KAbstractModule() {
```
wesleyk: what's the benefit of splitting this out? Doesn't seem used elsewhere

mightyguava: Ah, yeah this was me trying to add tests by splitting apart the service. Will revert

misk-aws/src/main/kotlin/misk/jobqueue/sqs/AwsSqsJobHandlerModule.kt
```diff
@@ -0,0 +1,63 @@
...
+  companion object {
+    inline fun <reified T : JobHandler> create(queueName: QueueName):
+        AwsSqsJobHandlerModule<T> = create(queueName, T::class)
+
+    @JvmStatic
+    fun <T : JobHandler> create(
+      queueName: QueueName,
+      handlerClass: Class<T>
+    ): AwsSqsJobHandlerModule<T> {
+      return create(queueName, handlerClass.kotlin)
```
ryanhall07: nit: weird to convert to KT to convert back to java. I'd just keep it java, since guice needs that anyways for binding

ryanhall07:
can you use `ServiceTestingModule` for testing?

mightyguava:
I'm thinking maybe `ServiceTestingModule` should override a service in the dependency graph, not in the service manager.

misk-aws/src/main/kotlin/misk/jobqueue/sqs/AwsSqsJobHandlerModule.kt
mightyguava: It seems like we usually convert to KClass throughout misk.

PR #842, opened 2019-03-11
Set key to file url instead of file name to allow config override files with the same names

hamdanjaveed:
For e.g., if we had the following override files when loading our test_app's misk config

/resources/db_config/test_app-config.yaml
/resources/cache_config/test_app-config.yaml

We would like to override our config with the content from both files. Currently, we only
use the filename as a key when creating a mapping of our config. This PR changes the key
to contain the entire URL.

### DIFF
--- a/misk/src/test/kotlin/misk/config/MiskConfigTest.kt
+++ b/misk/src/test/kotlin/misk/config/MiskConfigTest.kt
@@ -112,4 +112,23 @@ class MiskConfigTest {
     assertThat(filesInDir[0]).endsWith("/overrides/override-test-app1.yaml")
     assertThat(filesInDir[1]).endsWith("/overrides/override-test-app2.yaml")
   }
+
+  @Test
+  fun handlesDuplicateNamedExternalFiles() {
+    val overrides = listOf(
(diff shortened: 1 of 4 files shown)


misk/src/test/kotlin/misk/config/MiskConfigTest.kt
```diff
@@ -112,4 +112,22 @@ class MiskConfigTest {
...
+  @Test
+  fun handlesDuplicateNamedExternalFiles() {
+    val overrides = listOf(
+      MiskConfigTest::class.java.getResource("/overrides/override-test-app1.yaml"),
+      MiskConfigTest::class.java.getResource("/additional_overrides/override-test-app1.yaml"))
+      .map { File(it.file) }
+
+    val config = MiskConfig.load<TestConfig>("test_app", defaultEnv, overrides)
+    assertThat(config.consumer_a).isEqualTo(ConsumerConfig(14, 1))
+    assertThat(config.consumer_b).isEqualTo(ConsumerConfig(34, 79))
```
wesleyk: oh we should make sure these assertions check values from both resources

hamdanjaveed: Just confirmed, they do

PR #844, opened 2019-03-12
Web Actions Tab: Add null handling to Application/Network Interceptors

adrw:
* Application or Network Interceptors are appearing null when they lack a qualified class name
* New `ClassNameFormatter` that handles formatting a valid class name

### DIFF
--- a/misk/src/main/kotlin/misk/web/MiskWebModule.kt
+++ b/misk/src/main/kotlin/misk/web/MiskWebModule.kt
@@ -17,6 +17,7 @@ import misk.web.actions.LivenessCheckAction
 import misk.web.actions.NotFoundAction
 import misk.web.actions.ReadinessCheckAction
 import misk.web.actions.StatusAction
+import misk.web.actions.WebActionMetadataAction
 import misk.web.exceptions.ActionExceptionLogLevelConfig
 import misk.web.exceptions.ActionExceptionMapper
 import misk.web.exceptions.ExceptionHandlingInterceptor
@@ -28,6 +29,7 @@ import misk.web.extractors.PathPatternParameterExtractorFactory
 import misk.web.extractors.QueryStringParameterExtractorFactory
 import misk.web.extractors.RequestBodyParameterExtractor
 import misk.web.extractors.WebSocketParameterExtractorFactory
+import misk.web.formatter.ClassNameFormatter
 import misk.web.interceptors.InternalErrorInterceptorFactory
 import misk.web.interceptors.MarshallerInterceptor
 import misk.web.interceptors.MetricsInterceptor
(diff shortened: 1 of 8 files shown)


wesleyk:
let's add a test

misk/src/main/kotlin/misk/web/actions/WebActionMetadataAction.kt
```diff
@@ -72,8 +73,8 @@ internal fun WebActionMetadata(
       parameterTypes = parameterTypes.map { it.toString() },
       returnType = returnType.toString(),
       pathPattern = pathPattern.toString(),
-      applicationInterceptors = applicationInterceptors.map { it::class.qualifiedName.toString() },
-      networkInterceptors = networkInterceptors.map { it::class.qualifiedName.toString() },
+      applicationInterceptors = applicationInterceptors.map { ClassNameFormatter().format(it::class) },
```
wesleyk: let's inject this dependency

misk/src/main/kotlin/misk/web/formatter/ClassNameFormatter.kt
```diff
@@ -0,0 +1,27 @@
...
+
+class ClassNameFormatter {
+  fun <T : Any> format(kclass: KClass<T>): String {
+    return when (kclass.qualifiedName) {
+      null -> kclass.toString().split("class ").last()
+      else -> kclass.qualifiedName.toString()
+    }
+  }
+
+  fun isValid(className: String): Boolean {
```
wesleyk: do we use this? Why expose it?

misk/src/test/kotlin/misk/web/actions/WebActionMetadataActionTest.kt
```diff
@@ -27,6 +29,12 @@ class WebActionMetadataActionTest {
     assertThat(customServiceActionMetadata.pathPattern).isEqualTo("/custom_service_access")
     assertThat(customServiceActionMetadata.allowedServices).containsOnly("payments")
     assertThat(customServiceActionMetadata.allowedRoles).isEmpty()
+    assertThat(
+        customServiceActionMetadata.applicationInterceptors.all {
+          classNameFormatter.isValid(it)
```
wesleyk: seems a bit backwards to assert this here; how about instead we confirm that we return the formatted class name?

misk/src/main/kotlin/misk/web/MiskWebModule.kt
```diff
@@ -139,6 +141,10 @@ class MiskWebModule(private val config: WebConfig) : KAbstractModule() {
     install(WebActionModule.create<ReadinessCheckAction>())
     install(WebActionModule.create<LivenessCheckAction>())
     install(WebActionModule.create<NotFoundAction>())
+
+    // WebActionMetadataAction.Factory
+    bind<ClassNameFormatter>().toInstance(ClassNameFormatter())
```
wesleyk: we shouldn't need these bindings

misk/src/main/kotlin/misk/web/actions/WebActionMetadataAction.kt
```diff
@@ -30,6 +31,44 @@ class WebActionMetadataAction @Inject constructor() : WebAction {
   }
 
   data class Response(val webActionMetadata: List<WebActionMetadata>)
+
+  class Factory {
```
wesleyk: rather than an additional layer of abstraction let's have a top level method on the action

adrw:
ClassNameFormatter pulled out into #851

adrw:
Merged with #853

PR #848, opened 2019-03-13
Pull Vitess image when we know if it's a Vitess data source or not

tirsen:


misk-hibernate/src/main/kotlin/misk/vitess/StartVitessService.kt
```diff
@@ -313,14 +313,32 @@ class LogContainerResultCallback : ResultCallbackTemplate<LogContainerResultCall
...
+    this.qualifier = qualifier
+    this.environment = environment
+    this.config = config
+    this.consumedKeys = setOf()
+    this.producedKeys = setOf(Key.get(StartVitessService::class.java))
+
+    if (config.type == DataSourceType.VITESS) {
+      // Need to pull the docker image here,
+      // can't do it during service start up because it's too slow
+      StartVitessService.pullImage()
```
tirsen: I have to confess that I don't really like doing things like this in a constructor, but neither do I like doing it in a class initializer either so I guess this doesn't really make things worse either.

mightyguava: Your other option is to do it in the `configure()` function of the vitess module. That might be slightly better?

misk-hibernate/src/main/kotlin/misk/vitess/StartVitessService.kt
```diff
@@ -329,6 +347,10 @@ class StartVitessService(
       // We only start up Vitess if Vitess has been configured
       return
     }
+    check(StartVitessService.imagePulled.get()) {
+      "Docker image not pulled, we can't pull here because service start up times out, " +
```
mightyguava: If it's in the constructor, you shouldn't need this?

chris-ryan-square:

chris-ryan-square:

PR #849, opened 2019-03-13
Expose QueueName.deadletterQueue

mightyguava:
Since this is an enforced convention, we should expose it. Otherwise we'll see this everywhere:

```kotlin
      var queueName = "jobqueue-exemplar_queue"
      if (deadletter) {
        queueName += "_dlq"
      }
```


mightyguava:
CI flaked: 
```
misk.web.JettyServiceMetricsTest > threadPoolMetrics() FAILED
    java.lang.AssertionError: 
    Expecting:
      <0.6>
    to be close to:
      <0.5>
    by less than 10% but difference was 19.999999999999996%.
    (a difference of exactly 10% being considered valid)
        at misk.web.JettyServiceMetricsTest.threadPoolMetrics(JettyServiceMetricsTest.kt:99)
```

PR #850, opened 2019-03-13
Support overriding dependencies in service graph

mightyguava:
ServiceTestingModule is a hard to use because it doesn't override the Service it is adding extra dependencies to, since you can't override a Multibinding. Instead, it binds a new service with extra `consumedKeys`. It's on the developer to refactor their modules so that in tests they don't bind the original service, because otherwise the service graph will complain about having 2 services producing the same key.

This adds support to `CoordinatedService` for explicitly overriding service dependencies. There is a limitation with this approach that it targets all instances of a service, not a specific binding, since overrides can only target by Class, not Key.

### DIFF
--- a/misk-testing/src/main/kotlin/misk/service/ServiceTestingModule.kt
+++ b/misk-testing/src/main/kotlin/misk/service/ServiceTestingModule.kt
@@ -2,10 +2,8 @@ package misk.service
 
 import com.google.common.util.concurrent.Service
 import com.google.inject.Key
-import com.google.inject.Provider
-import misk.DependentService
+import misk.ServiceDependencyOverride
 import misk.inject.KAbstractModule
-import misk.inject.asSingleton
 import kotlin.reflect.KClass
 
 /**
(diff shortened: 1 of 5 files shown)


misk-testing/src/main/kotlin/misk/service/ServiceTestingModule.kt
```diff
@@ -15,63 +13,26 @@ import kotlin.reflect.KClass
  * that is being spun up within the test itself
  */
 class ServiceTestingModule<T : Service> internal constructor(
-  private val wrappedServiceKey: Key<T>,
+  private val wrappedServiceClass: KClass<T>,
```
ryanhall07: I don't understand why it has to be by Class and not Key?

mightyguava: Services are bound using multi binder, but there are binding strategies

```
multibind<Service>().to<FooService>()
multibind<Service>().toProvider(FooServiceProvider(...))
multibind<Service>().to(Key.get(FooService)
multibind<Service>().toInstance(FooService(...))
```

If the binding strategy is provider or instance, then there's no key for the linked binding. The only way I can think of to target a specific service is to check the instance's class.

PR #853, opened 2019-03-14
Format application/network interceptor names for WebActionMetadata

wesleyk:


misk/src/main/kotlin/misk/web/formatter/ClassNameFormatter.kt
```diff
@@ -4,10 +4,12 @@ import javax.inject.Inject
 import kotlin.reflect.KClass
 
 class ClassNameFormatter @Inject constructor() {
-  fun <T : Any> format(kclass: KClass<T>): String {
-    return when (kclass.qualifiedName) {
-      null -> kclass.toString().split("class ").last()
-      else -> kclass.qualifiedName.toString()
+  companion object {
```
adrw: Good call 👍

PR #859, opened 2019-03-18
Authenticate jedis pool properly

hamdanjaveed:
Instantiate the `JedisPool` with the redis password to authenticate all instances.


misk-redis/src/main/kotlin/misk/redis/RedisModule.kt
```diff
@@ -33,21 +32,11 @@ class RedisModule(
         jedisPoolConfig,
         replicationGroup.writer_endpoint.hostname,
         replicationGroup.writer_endpoint.port,
-        true
+      Protocol.DEFAULT_TIMEOUT,
```
wesleyk: align with lines above

PR #860, opened 2019-03-19
[DNM - Hack Week] Add --dev-mode flag to support special instructions in local development

nbroyles:
For hack week, adding the ability to spin up several services for local
testing. As a part of that, need to be able to reset services to a known
good state for testing. If --dev-mode is provided,
DevModeCoordinationService invokes the services required to make the app
state good. What's a good app state is determined by the service itself.
Apps use the DevModeService annotation to bind services that should
be used to clean up app state.


jjestrel:
Closed and leaving branch around

PR #862, opened 2019-03-22
Only start vitess in TESTING

ryanhall07:
Docker compose starts vitess in DEVELOPMENT now

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,9 +17,11 @@ matrix:
       services:
         - mysql
         - docker
+      env:
+        - ENVIRONMENT=TESTING
     - language: node_js
       node_js:
         - 10.15
       script:
         - npm install -g @misk/cli && miskweb ci-build -e
-          
+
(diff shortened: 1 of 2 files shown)


mightyguava:
Pretty sure this will break things. It's similar to the issue in https://github.com/square/misk/pull/814, that `VitessScaleSafetyChecks` installed by `HibernateTestingModule` has a dependency on `StartVitessService`.

ryanhall07:
> Pretty sure this will break things. It's similar to the issue in #814, that `VitessScaleSafetyChecks` installed by `HibernateTestingModule` has a dependency on `StartVitessService`.

ryanhall07:
@mightyguava I don't think so, since `HibernateTestingModule` is still installed in tests. This is different from #814 since it's not trying to remove vitess in tests, only in local development

mightyguava:
Ah ok. I thought HibernateTestingModule was being installed for local dev too...

PR #863, opened 2019-03-22
Move TruncateTablesService to misk-hibernate

adrw:
* Allows services and non-tests to use TruncateTablesService
* Pre-req for Reset Service


wesleyk:
I don't think we want this. It's a really dangerous operation and would rather have it explicitly in some sort of testing library that prod wouldn't import.

adrw:
To mitigate risk, we could add scoping in the service to never run in staging / production in environments.

The purpose for this is to use `TruncateTablesService` locally to reset development databases.

mightyguava:
`TruncateTablesService` just iterates through all the tables and wipes them, excluding any special tables like `schema_versions`. For local dev, we could achieve it with an external tool, like `devenv` maybe?

adrw:
Some of our hack week work was around letting each Misk App bind a `ResetStateService` that would run after initial DB resources were available but before the more application level code ran. This would allow a CLI option to be passed in when the service was run that would reset the state on connected resources that may have stayed running in between Misk App runs. 

Primary example of this would be running tests in the Misk App that connect out to DBs and Square services running in Docker-Compose. Each test should be run on a clean state, but if that can be accomplished without recreating every Docker Compose container than a lot of time would be saved.

mightyguava:
If you are running tests, then you should be able to include `hibernate-testing`?

Outside of tests, I don't think an application should be capable of completely resetting its own state. That does seem somewhat dangerous like Wesley said. I was suggesting in my previous comment that, rather having the app do it, have the external tool that is setting up the development environment do it. Let the ResetStateService be part of the tool that you are using set up the development environment, rather than part of the app. Let the `devenv` tool you guys are making allow a CLI option to be passed in to reset the state, and have `devenv` do it, not the app.

ryanhall07:
I still think you would want the service to bind a `ResetStateService`, which can be controlled via some endpoint like `/dev/reset`. the `devenv` tool would just be responsible for calling `/dev/reset`.

so you might have `devenv myservice restart`, which actually invokes `docker stop` then `docker start`. or you might have `devenv myservice reset` which calls `http://myservice/dev/reset` which should be faster.

with all that being said, we haven't really spent anytime prototyping this idea and proving this added complexity is even worth it (as opposed to just always restarting the service). there is potentially a lot of complexity around resetting the state of a running service (it could just blow up). I recommend we first prove this is worth it, before adding it in.

adrw:
Closing for now until we have a chance to take a concerted stab at what this functionality would look like.

PR #868, opened 2019-03-26
@MiskTest should default environment to TESTING

dhanji:


misk/src/main/kotlin/misk/environment/Environment.kt
```diff
@@ -12,13 +13,27 @@ enum class Environment {
   companion object {
     internal val logger = getLogger<Environment>()
     private const val ENV_ENVIRONMENT = "ENVIRONMENT"
+    private lateinit var env: Environment
+
+    fun setTesting() {
+      if (::env.isInitialized) {
```
wesleyk: is this a safeguard in case this is called out of band?

ryanhall07: it's to allow the env var to take precedence. I think eventually we'd want to this to be a set once property with no default. until then having it prefer the env var is probably the safest rollout

wesleyk: ah because we have scripts already passing in ENVIRONMENT=TESTING, gotcha

ryanhall07: yea. I think we'll still need that env var set, because we don't enforce _all_ tests use the `MiskTestExtension`. although those tests are probably not reading the env var anyways, since they don't use guice. so probably 99% ok :)

misk-testing/src/main/kotlin/misk/testing/MiskTestExtension.kt
```diff
@@ -18,6 +19,7 @@ internal class MiskTestExtension : BeforeEachCallback, AfterEachCallback {
   override fun beforeEach(context: ExtensionContext) {
     val module = object : KAbstractModule() {
       override fun configure() {
+        Environment.setTesting()
```
ryanhall07: why do this in the `configure` method? makes more sense outside (especially since some modules depend on this env var unfortunately)

ryanhall07:
i'm going to try and fix the tests/comments so we can land this

PR #869, opened 2019-03-26
Inject interceptors as Providers in TypedClientProvider

ryanhall07:
This allows Guice to properly detect a circular dependency when an
application injects an HTTP Client in an Interceptor. It's not clear why
Guice requires this, but oh well.

### DIFF
--- a/misk/src/main/kotlin/misk/client/TypedHttpClientModule.kt
+++ b/misk/src/main/kotlin/misk/client/TypedHttpClientModule.kt
@@ -51,11 +51,14 @@ class TypedHttpClientModule<T : Any>(
     @Inject
     private lateinit var httpClientsConfig: HttpClientsConfig
 
+    // Use Providers for the interceptors so Guice can properly detect cycles when apps inject
+    // an HTTP Client in an Interceptor.
+    // https://gist.github.com/ryanhall07/e3eac6d2d47b72a4c37bce87219d7ced
     @Inject
-    private lateinit var clientNetworkInterceptorFactories: List<ClientNetworkInterceptor.Factory>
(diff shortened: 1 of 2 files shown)


ryanhall07:
with this change we get

```
1) Found a circular dependency involving com.squareup.investcore.token.service.TokenService, and circular dependencies are disabled.
  at com.squareup.investcore.token.service.RealTokenService.class(RealTokenService.kt:21)
  while locating com.squareup.investcore.token.service.RealTokenService
  while locating com.squareup.investcore.token.service.TokenService
    for the 2nd parameter of com.squareup.investcore.clients.drivewealth.DrivewealthClientInterceptor$Factory.<init>
```

before this change we get

```
1) Error in custom provider, kotlin.UninitializedPropertyAccessException: lateinit property httpClientsConfig has not been initialized
  at misk.client.TypedHttpClientModule.configure(TypedHttpClientModule.kt:34) (via modules: com.squareup.investcore.clients.drivewealth.DrivewealthModule -> misk.client.TypedHttpClientModule)
```

misk/src/main/kotlin/misk/client/TypedHttpClientModule.kt
```diff
@@ -51,11 +51,14 @@ class TypedHttpClientModule<T : Any>(
     @Inject
     private lateinit var httpClientsConfig: HttpClientsConfig
 
+    // Use Providers for the interceptors so Guice can properly detect cycles when apps inject
+    // an HTTP Client in an Interceptor. No idea why Guice is not smart enough without with the
```
wesleyk: rm with

ryanhall07:
Example showing the issue

https://gist.github.com/ryanhall07/e3eac6d2d47b72a4c37bce87219d7ced

PR #870, opened 2019-03-27
Fix flaky tests

ryanhall07:
This dependency wiring could use love and rework. I think we should stop producing keys that map 1:1 with actually services, and instead start producing logical keys like `PersistenceLayerReady`. so `StartVitessService` would produce `PersistenceLayerReady`, but also if you're not in tests or not using vitess, you would install a no-op service that also produces `PersistenceLayerReady`. 

For now I just wanted to fix the flakes :)


ryanhall07:
going to try and fix the tests/comments so we can merge this

ryanhall07:
merging to remove flakes. can follow up if anybody disagrees with how this was done

mightyguava:
Thank you for fixing the flake!

PR #871, opened 2019-03-27
Add request param type information into WebActionMetadata for Wire messages

wesleyk:


misk/src/main/kotlin/misk/web/RequestTypes.kt
```diff
@@ -0,0 +1,106 @@
...
+
+  private fun handleField(
+      fieldType: KType,
+      fieldName: String,
+      fields: MutableList<Field>,
+      stack: LinkedList<KClass<*>>,
+      repeated: Boolean = false) {
+    val fieldClass = fieldType.classifier as KClass<*>
+    when (fieldClass) {
+      String::class -> fields.add(Field(fieldName, String::class.simpleName!!, repeated))
```
adrw: Good central location to find how each fieldClass is processed.

misk/src/test/kotlin/misk/ActionsTest.kt
```diff
@@ -1,36 +1,48 @@
...
-  fun methodReferenceNotAllowedAsAction() {
+  @Test fun noRequestBodyAction() {
+    val action = NoRequestBodyAction::myActionMethod.asAction()
+    assertThat(action.name).isEqualTo("NoRequestBodyAction")
+    assertThat(action.parameterTypes).hasSize(1)
+    assertThat(action.parameterTypes[0]).isEqualTo(Int::class.createType())
+    assertThat(action.requestType).isNull()
+  }
+
+  @Test fun methodReferenceNotAllowedAsAction() {
```
adrw: Could you add a single sentence what this is testing?

misk/src/test/kotlin/misk/ActionsTest.kt
```diff
@@ -1,36 +1,48 @@
...
     assertFailsWith<IllegalArgumentException> {
       val t = TestAction()
       t::myActionMethod.asAction()
     }
   }
 
-  @Test
-  fun freeStandingFunctionNotAllowedAsAction() {
+  @Test fun freeStandingFunctionNotAllowedAsAction() {
```
adrw: Documentation here as well?

misk/src/test/kotlin/misk/ActionsTest.kt
wesleyk: these were refactored so the annotation is on the same line as the test. I'm more a fan of the test method name explaining what it's doing. I think the test name is explanatory enough?

PR #872, opened 2019-03-28
Add ExternalDependency support to MiskTestExtension

mightyguava:
We currently spin up external dependencies like Vitess and SQS in tests,
by inserting additional services into the service dependency graph. It
has forced us to use different workarounds, like ServiceTestingModule
which modifies the service graph in tests, and StartVitessService which
is a production dependency but really doesn't belong there.

@ryanhall07 had the idea that we could have the MiskTestExtension start up
external dependencies instead, since they really are just test fixtures
and are better handled by the testing framework.

It converts the DockerSqs service, which is
simple but is a good enough demonstration.

### DIFF
--- a/misk-testing/src/main/kotlin/misk/testing/ExternalDependency.kt
+++ b/misk-testing/src/main/kotlin/misk/testing/ExternalDependency.kt
@@ -1,0 +1,34 @@
+package misk.testing
+
+/**
+ * An external dependency of the Misk Application that needs to be started for test, like Redis,
+ * Vitess, MySQL, SQS, etc.
+ */
+interface ExternalDependency {
+  /**
+   * Starts the dependency.
+   */
+  fun startup()
+
+  /**
+   * Stops the dependency.
+   */
+  fun shutdown()
+
+  /**
+   * Called before each test run
+   */
+  fun beforeEach()
+
+  /**
+   * Called before each test run
+   */
+  fun afterEach()
+
+  /**
+   * Unique ID for the dependency, used as a stable key across tests. Can be overridden if more
+   * than one instance of the dependency is supported.
+   */
+  val id: String get() = javaClass.name
+}
+
--- a/misk-testing/src/main/kotlin/misk/testing/MiskTestExtension.kt
+++ b/misk-testing/src/main/kotlin/misk/testing/MiskTestExtension.kt
@@ -4,10 +4,12 @@ import com.google.common.util.concurrent.ServiceManager
 import com.google.inject.Guice
 import com.google.inject.Injector
 import com.google.inject.Module
+import io.netty.util.internal.ConcurrentSet
 import misk.environment.Environment
 import misk.inject.KAbstractModule
 import misk.inject.getInstance
 import misk.inject.uninject
+import misk.logging.getLogger
 import org.junit.jupiter.api.extension.AfterEachCallback
 import org.junit.jupiter.api.extension.BeforeEachCallback
 import org.junit.jupiter.api.extension.ExtensionContext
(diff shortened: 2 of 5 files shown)


misk-testing/src/main/kotlin/misk/testing/ExternalDependency.kt
```diff
@@ -0,0 +1,65 @@
...
+
+/**
+ * An external dependency of the Misk Application that needs to be started for test, like Redis,
+ * Vitess, MySQL, SQS, etc.
+ */
+interface ExternalDependency {
+  /**
+   * Starts the dependency.
+   */
+  fun startup()
```
ryanhall07: it might be nice to have some kind of `beforeTest` hook on each external dependency. something like `TruncateTables` for a mysql/vitess external dependency

misk-testing/src/main/kotlin/misk/testing/MiskTestExtension.kt
```diff
@@ -19,6 +19,11 @@ internal class MiskTestExtension : BeforeEachCallback, AfterEachCallback {
 
   override fun beforeEach(context: ExtensionContext) {
     Environment.setTesting()
+
+    for (dep in context.getExternalDependencies()) {
```
ryanhall07: nit: could also extend `BeforeAllCallback` so we don't call startUp for every test (I know it's a no-op).

mightyguava: It gets kinda ugly to make this a `BeforeAllCallback`. Currently you do

```kotlin
  @MiskExternalDependency private val dockerSqs = DockerSqs()
```

to be used in a `BeforeAllCallback` it would have to be

```kotlin
companion object {
  @JVMField
  @MiskExternalDependency private val dockerSqs = DockerSqs()
}
```

What do you think?

ryanhall07: oh gross! yea no need for that

misk-testing/src/main/kotlin/misk/testing/ExternalDependency.kt
mightyguava: Refactored this so that ExternalDependency is always JVM scoped. I don't think we'll actually have any that are supposed to have a test class or test method lifecycle. Makes the API simpler and now it's an interface and has `beforeEach()` and `afterEach()` methods. I'm using `afterEach()` to delete queues after every test run

misk-testing/src/main/kotlin/misk/testing/ExternalDependency.kt
```diff
@@ -0,0 +1,34 @@
...
+
+  /**
+   * Stops the dependency.
+   */
+  fun shutdown()
+
+  /**
+   * Called before each test run
+   */
+  fun beforeEach() {}
```
ryanhall07: I'd prefer we just force impls to extend this so they have to think about it.

misk-testing/src/main/kotlin/misk/testing/MiskTestExtension.kt
```diff
@@ -101,6 +121,20 @@ internal class MiskTestExtension : BeforeEachCallback, AfterEachCallback {
       beforeEachCallbacks.forEach { it.beforeEach(context) }
     }
   }
+
+  private fun ExternalDependency.startOnce() {
```
ryanhall07: super nit: `startIfNecessary` so it's clear it doesn't blow up

ryanhall07:
LG! 🐑 it

misk-testing/src/main/kotlin/misk/testing/ExternalDependency.kt
mightyguava: This is a default function. Impls do need to extend it

misk-testing/src/main/kotlin/misk/testing/MiskTestExtension.kt
mightyguava: hmm... that's how I usually interpret `once()` in a function name. Will change though

ryanhall07: yea, super nit :)

tirsen:
This is awesome.

PR #873, opened 2019-03-28
Depend on SchemaMigrationCheckService instead

ryanhall07:
SchemaMigratorService is a test only service that prod/staging does not
install anymore.


ryanhall07:
And once @mightyguava lands the `ExternalDependency` stuff we can move `SchemaMigratorService` to a `MysqlDependency`

ryanhall07:
this doesn't work. trying something else

PR #874, opened 2019-03-28
Always run StartVitessService

ryanhall07:


misk-hibernate-testing/src/main/kotlin/misk/jdbc/VitessScaleSafetyChecks.kt
```diff
@@ -305,10 +305,10 @@ class VitessScaleSafetyChecks(
     val request = Request.Builder()
         .url("http://localhost:27000/debug/vars")
         .build()
+    val response = okHttpClient.newCall(request).execute()
```
ryanhall07: this was part of the original commit to fix flaky tests, that didn't need to be reverted.

misk-hibernate/src/main/kotlin/misk/hibernate/HibernateModule.kt
```diff
@@ -162,40 +165,14 @@ class HibernateModule(
     install(HibernateHealthCheckModule(qualifier, sessionFactoryProvider))
   }
 
-  private fun maybeBindStartVitessService() {
-    // TODO(jmuia): don't bind the StartVitessService if the datasource is not Vitess.
-
-    val environment = Environment.fromEnvironmentVariable()
```
ryanhall07: doing this breaks InjectorTests since they rely on setting the Environment

misk-hibernate/src/main/kotlin/misk/vitess/StartVitessService.kt
```diff
@@ -349,9 +343,41 @@ class StartVitessService(
...
 
+  fun shouldRunVitess() =  config.type != DataSourceType.VITESS || environment == TESTING || environment == DEVELOPMENT
+
   override fun shutDown() {
   }
 
+  /**
+   * Shut down the cached clusters on JVM exit.
+   */
+  init {
```
ryanhall07: moved from Companion.Init to StartVitessService.Init so we can check the Environment. Shouldn't impact anything, it still checks the `imagedPulled` var

misk-hibernate/src/main/kotlin/misk/vitess/StartVitessService.kt
```diff
@@ -347,9 +343,41 @@ class StartVitessService(
 
   fun cluster() = cluster!!.cluster
 
+  fun shouldRunVitess() =  config.type != DataSourceType.VITESS || environment == TESTING || environment == DEVELOPMENT
```
jjestrel: `fun shouldRunVitess() =  config.type != DataSourceType.VITESS ` is confusing imo. I'm not sure why `onfig.type != DataSourceType.VITESS` should possibly pull the docker image

mightyguava: re: docker image, it's because it takes like 5 minutes to pull and service startup will usually time out before that. There isn't a better execution path at the moment to pull the image

ryanhall07: eh it's a bug. unfortunately tests aren't going to catch it because they are always going to run it in TESTING :) thanks for catching

mightyguava: shouldn't this be `config.type == DataSourceType.VITESS && (environment == TESTING || environment == DEVELOPMENT)`?

misk-hibernate/src/main/kotlin/misk/vitess/StartVitessService.kt
```diff
@@ -347,9 +343,41 @@ class StartVitessService(
...
+   * Shut down the cached clusters on JVM exit.
+   */
+  init {
+    // We need to do this outside of the service start up because this takes a really long time
+    // the first time you do it and can cause service manager to time out.
+    if (shouldRunVitess() && imagePulled.compareAndSet(false, true) &&
+        runCommand("docker pull vitess/base@$VITESS_VERSION") != 0) {
+      logger.warn("Failed to pull Vitess docker image. Proceeding regardless.")
+    }
+    Runtime.getRuntime().addShutdownHook(Thread {
```
mightyguava: This now just gets run a ton of times? I guess it'll just be a noop

ryanhall07: hmm good point. I can move that back to the companion init

misk-hibernate/src/main/kotlin/misk/vitess/StartVitessService.kt
ryanhall07: lol yes

ryanhall07:
@mightyguava comments addressed

PR #878, opened 2019-03-29
Extend transaction timeout for debugging

tirsen:


misk-hibernate/src/main/kotlin/misk/vitess/StartVitessService.kt
```diff
@@ -177,6 +177,9 @@ class DockerVitessCluster(
         "-mysql_bind_host=0.0.0.0",
         "-data_dir=/vt/vtdataroot",
         "-schema_dir=schema",
+        // Increase the transaction timeout so you can have a breakpoint
+        // inside a transaction without it timing out
+        "-queryserver-config-transaction-timeout=${Duration.ofHours(24).toMillis()}",
```
wesleyk: (this is only for testing, right?)

ryanhall07: yes

misk-hibernate/src/main/kotlin/misk/vitess/StartVitessService.kt
```diff
@@ -352,9 +355,10 @@ class StartVitessService(
   init {
     // We need to do this outside of the service start up because this takes a really long time
     // the first time you do it and can cause service manager to time out.
-    if (shouldRunVitess() && imagePulled.compareAndSet(false, true) &&
-        runCommand("docker pull vitess/base@$VITESS_VERSION") != 0) {
-      logger.warn("Failed to pull Vitess docker image. Proceeding regardless.")
+    if (imagePulled.compareAndSet(false, true)) {
```
ryanhall07: why did you remove the `shouldRunVitess` conditional?

PR #879, opened 2019-03-29
Web Actions Tab: Typed Send Request Form

adrw:
* Typed forms for use with proto web actions
* Supports nested and repeating types

Mar 29, 2019 Web Actions Typed Forms

### DIFF
--- a/misk/web/tabs/web-actions/src/ducks/webActions.ts
+++ b/misk/web/tabs/web-actions/src/ducks/webActions.ts
@@ -9,6 +9,41 @@ import { HTTPMethod } from "http-method-enum"
 import { chain } from "lodash"
 import { all, AllEffect, call, put, takeLatest } from "redux-saga/effects"
 
+export const enum TypescriptBaseTypes {
+  "boolean" = "boolean",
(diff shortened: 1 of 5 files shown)


misk/web/tabs/web-actions/src/ducks/webActions.ts
```diff
@@ -9,6 +9,41 @@ import { HTTPMethod } from "http-method-enum"
...
+export const enum TypescriptBaseTypes {
+  "boolean" = "boolean",
+  "number" = "number",
+  "string" = "string",
+  "enum" = "enum",
+  "any" = "any",
+  "null" = "null"
+}
+
+export const enum KotlinTypes {
```
wesleyk: should we be expressing all the types here? https://github.com/square/misk/blob/1e5bb123c2e85118a205d1e040e99cb111531dbf/misk/src/main/kotlin/misk/web/RequestTypes.kt#L73-L81

adrw: Absolutely 👍 I'll add a commit that adds this.

PR #880, opened 2019-03-29
Web Actions Tab: Typed Form Kotlin Types + Boolean UI

adrw:


misk/web/tabs/web-actions/src/ducks/webActions.ts
```diff
@@ -10,26 +10,38 @@ import { chain } from "lodash"
 import { all, AllEffect, call, put, takeLatest } from "redux-saga/effects"
 
 export const enum TypescriptBaseTypes {
+  "any" = "any",
```
wesleyk: do enums in typescript have to be defined like this, given that these are all tautological? Can we simply put the single string value?

adrw: Looked further and it seems like String Enums (as we're using) are recommended for these use cases https://www.typescriptlang.org/docs/handbook/enums.html

PR #883, opened 2019-03-29
enable circleci

mightyguava:


.circleci/config.yml
```diff
@@ -0,0 +1,58 @@
...
+          name: "Start MySQL"
+          # There's a race condition here, where it's possible that MySQL hasn't finished starting
+          # up before we try to connect to it. But Kotlin builds are so damn slow it'll never
+          # happen in practice.
+          command: docker run -d -p 3306:3306 --name mysql -e MYSQL_ALLOW_EMPTY_PASSWORD=1 mysql:5.7
+      - run:
+          name: "Build and test"
+          command: ./gradlew test -i --scan --parallel --build-cache
+  node:
+    working_directory: ~/misk-web
```
adrw: Shouldn't this still be `~/misk`? This step is to build the `misk-web` tabs that are in Misk repo (config, loader, web actions)

adrw: I checked the build logs and the tabs seemed to be built but it's unintuitive to me why that would be happening in the `~/misk-web` directory.

.circleci/config.yml
```diff
@@ -0,0 +1,58 @@
+version: 2.1
+jobs:
+  java:
+    machine:
+      image: circleci/classic:latest
+    working_directory: ~/misk
+    steps:
+      - checkout # check out source code to working directory
```
adrw: nit: unnecessary comment

.circleci/config.yml
```diff
@@ -0,0 +1,58 @@
...
+      - run:
+          name: "Download and extract OpenJDK 11"
```
adrw: Is this because their `openjdk-11` isn't sufficient? Is there another docker image we could use instead of bootstrapping this ourselves? Seems like it will be very fragile and prone to require updating.

.circleci/config.yml
wesleyk: eh this is helpful for me not knowing what checkout exactly does

.circleci/config.yml
mightyguava: It doesn't matter. `checkout` loads your code into `working_directory`. I copied this from wesley's old branch, but I can change it to `misk` and it'll work the same.

.circleci/config.yml
mightyguava: The tests don't run in docker, they run in a VM. For various reasons it's difficult to get our tests to work outside of docker. Circle CI does not offer an openjdk 11 image for the VM.

This is not fragile. It is the official mirror for OpenJDK. We will just bump the version when we want to bump our Java version.

PR #884, opened 2019-03-30
Add configurable jetty max thread pool

jjestrel:


misk/src/main/kotlin/misk/web/WebConfig.kt
```diff
@@ -13,6 +13,7 @@ data class WebConfig(
   val selectors: Int? = null,
   val acceptors: Int? = null,
   val queueSize: Int? = null,
+  val jettyMaxThreadPoolSize: Int? = null,
```
mightyguava: `jetty_max_thread_pool_size`

wesleyk: hmm ideally we migrate queueSize to snake_case as well, assuming it'd be an easy migration because I imagine contacts is the only client

jjestrel: I'll change both

PR #885, opened 2019-03-30
Add .delete to ReflectionQuery CLOSES #602

### LINKED ISSUE #602: Support `delete` in `ReflectionQuery`
We should have a `@Delete` annotation, like the `@Select` we currently have, this will let us support delete with a where clause instead of using `hibernateSession.delete(entity);`

jjestrel:


misk-hibernate/src/test/kotlin/misk/hibernate/ReflectionQueryFactoryTest.kt
```diff
@@ -420,6 +420,42 @@ class ReflectionQueryFactoryTest {
...
+  @Test
+  fun delete() {
+    transacter.transaction { session ->
+      session.save(DbMovie("Jurassic Park", LocalDate.of(1993, 6, 9)))
+      session.save(DbMovie("Rocky", LocalDate.of(1976, 11, 21)))
+      session.save(DbMovie("Star Wars", LocalDate.of(1977, 5, 25)))
+    }
+
+    transacter.transaction { session ->
+      queryFactory.newQuery<OperatorsMovieQuery>()
```
alanpaulin: Can you `assertThat` the result of the delete is `2`?

jjestrel: Done

misk-hibernate/src/main/kotlin/misk/hibernate/Query.kt
```diff
@@ -11,6 +11,8 @@ interface Query<T> {
 
   fun list(session: Session): List<T>
 
+  fun delete(session: Session): Int
```
wesleyk: javadoc? (mention what is returned)

misk-hibernate/src/test/kotlin/misk/hibernate/ReflectionQueryFactoryTest.kt
```diff
@@ -420,6 +420,43 @@ class ReflectionQueryFactoryTest {
     }
   }
 
+  @Test
+  fun deleteFailsWithOrderBy() {
+    transacter.transaction { session ->
```
wesleyk: /nit is it worth even setting this data up?

misk-hibernate/src/main/kotlin/misk/hibernate/Query.kt
jjestrel: Added

misk-hibernate/src/test/kotlin/misk/hibernate/ReflectionQueryFactoryTest.kt
jjestrel: I changed the test to make sure the delete didn't go through

PR #890, opened 2019-04-01
Changed the way this module is initialized

yoavamit:
The KMS client is now initialized by its own module.
This allows `CryptoModule` to be completely independent of the KMS.

Users may choose to create their own KMS client implementation or use
one of the provided modules for AWS and GCP.

### DIFF
--- a/misk-crypto/README.md
+++ b/misk-crypto/README.md
@@ -17,7 +17,7 @@ In one of your app's module files:
 ```$kotlin
 class MyAppModule : KAbstractModule {
   override fun configure() {
-    val kmsClient = AwsKmsClient().withDefaultCredentials()
+    install(AwsKmsClientModule()) // will provide an AWS client with default credentials
     install(CryptoModule(kmsClient, cryptoConfig))
   }
 }
--- a/misk-crypto/README.md
+++ b/misk-crypto/README.md
@@ -18,7 +18,7 @@ In one of your app's module files:
 class MyAppModule : KAbstractModule {
   override fun configure() {
     install(AwsKmsClientModule()) // will provide an AWS client with default credentials
-    install(CryptoModule(kmsClient, cryptoConfig))
+    install(CryptoModule(cryptoConfig))
   }
 }
 ```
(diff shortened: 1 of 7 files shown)


misk-crypto/README.md
```diff
@@ -17,7 +17,7 @@ In one of your app's module files:
 ```$kotlin
 class MyAppModule : KAbstractModule {
   override fun configure() {
-    val kmsClient = AwsKmsClient().withDefaultCredentials()
+    install(AwsKmsClientModule()) // will provide an AWS client with default credentials
     install(CryptoModule(kmsClient, cryptoConfig))
```
adrw: Where's `kmsClient` coming from in L21 now?

yoavamit: Good catch, it should've been deleted.

Just to reiterate, the main thing I changed here is that now `CryptoModule` expects a `KmsClient` to be bounded by the user instead of receiving it is a parameter.
The `CryptoModule` will also throw an exception during `configure` if no client was provided.

misk-crypto/src/main/kotlin/misk/crypto/CryptoModule.kt
```diff
@@ -5,35 +5,50 @@ import com.google.crypto.tink.JsonKeysetReader
...
 
   override fun configure() {
+    requireBinding(KmsClient::class.java)
     check(!(config.gcp_key_uri != null && config.aws_kms_key_alias != null))
     check(!(config.gcp_key_uri == null && config.aws_kms_key_alias == null))
     AeadConfig.register()
-    val keyManager = KeyManager()
-    bind<KeyManager>().toInstance(keyManager)
+
+    bind<KeyManager>().asSingleton()
```
mightyguava: nit: @Singleton annotation is a behavior and should be on the class.

yoavamit: The `KeyManager` class is also annotated with `@Singleton`

mightyguava: Then why are you adding `bind<>()` here? If it has an @Inject you don't need to do this.

yoavamit: Yep, totally right. My bad. 
Fixed :)

PR #896, opened 2019-04-02
Database entity encryption

yoavamit:
This adds a column annotation for Hibernate to get it to encrypt and decrypt values using `misk-crypto`.
Columns marked with the `@EncryptedColumn()` annotation must have the following:
- Must be of type `ByteArray` in the entity class
- Backed by a `VARBINARY()` in the database.
- Cannot be a part of an index.

This annotation uses keys defined in `misk-crypto` and it's the user's responsibility to configure and load the keys mentioned in the annotation's `keyName` parameter.
If the specified key isn't loaded or fails to encrypt/decrypt, an exception is thrown so no undecipherable data will be written to the database.


ryanhall07:
It would be nice to include a test so it's clear how a developer would use this new annotation

misk-hibernate/src/main/kotlin/misk/hibernate/EncryptedColumn.kt
```diff
@@ -0,0 +1,4 @@
+package misk.hibernate
+
+@Target(AnnotationTarget.FIELD)
+annotation class EncryptedColumn(val keyName: String)
```
ryanhall07: seems burdensome to have to include the key for every encrypted column. does the key change per column?

yoavamit: It could.
I can imagine a scenario where you'd want to encrypt different columns with different keys, depending on their secrecy, what other systems might need access to that data and so on.

Also, having this configurable decouples the `misk-crypto` and `misk-hibernate` configuration.

ryanhall07: do you think that's the common case though? should there be a default key if the app use the same key for every column?

yoavamit: That's a good point.
I think having a default key name could be really cool in terms of user experience.

ryanhall07: it's not really clear this keyName should match the keyName in the crypto config. some javadoc would be helpful

misk-hibernate/src/test/kotlin/misk/hibernate/EncryptedColumnTest.kt
```diff
@@ -0,0 +1,111 @@
...
+    override lateinit var id: Id<DbJerryGarciaSong>
+
+    @Column(nullable = false)
+    var title: String
+
+    @Column(nullable = false)
+    var length: Int = 0
+
+    @Column(nullable = false)
+    @EncryptedColumn(keyName = "albumKey")
```
ryanhall07: does it make sense to enforce `@EncryptedColum` must be of type `ByteString`?

misk-hibernate/src/main/kotlin/misk/hibernate/EncryptedColumnType.kt
```diff
@@ -0,0 +1,80 @@
...
+    if (value == null) {
+      st.setNull(index, Types.BINARY)
+    } else {
+      // encrypt the data in set it in the prepared statement
+      val encrypted = cipher.encrypt(value as ByteString).toByteArray()
+      st.setBytes(index, encrypted)
+    }
+  }
+
+  override fun nullSafeGet(
```
ryanhall07: probably need some tests dealing with null values

misk-hibernate/src/test/kotlin/misk/hibernate/EncryptedColumnTest.kt
yoavamit: Yes. `ByteString` is currently the only type the `Cipher` object accepts.
Although, I think it'd be a cool feature to support more types like `ByteArray` and others types that can be easily de/serialized to `ByteString`

misk-hibernate/src/main/kotlin/misk/hibernate/EncryptedColumnType.kt
yoavamit: Yeah, this is still just POC code.

There are also some open questions about the behavior of this thing, for example:
- Can we support multiple annotations on a column? (what if you want to encrypt a JSON column)
- How does this play with indexing/querying on such columns?
- IDK, there's probably on kinds of serialization issues I haven't thought of...

misk-hibernate/src/main/kotlin/misk/hibernate/EncryptedColumn.kt
yoavamit: Done

yoavamit: Regarding a default key and having a single key for all columns - I think maybe we shouldn't do that.
Having a default key implies there's a key somewhere that's not managed by the owner of the app.
As for specifying a single key for all columns, it' could be a nice improvement.
But, assuming you usually encrypt one or 2 columns in a table, if at all, I'd go with the explicit approach of specifying the key name for each of these columns.

misk-hibernate/src/test/kotlin/misk/hibernate/EncryptedColumnTest.kt
yoavamit: Update - I've decided to use `ByteArray` for now.

misk-hibernate/src/main/kotlin/misk/hibernate/EncryptedColumnType.kt
yoavamit: Some answers:
- About multiple column annotations, I think that for now it's actually best to explicitly not support that. It can be really confusing, and I'm not sure how much it's needed - you usually need to encrypt a specific secret field in a proto/json blob not the entire thing. If you do need to encrypt the entire thing, you can always serialize it to a `ByteArray`.
- When implementing the `SecretColumnType.equals()` method properly, queryig on the field works (for exact matches). There's also a test for that in this PR.
- Serialization and caching is also taken care of when properly implementing the `assemble`, `disassemble` and `hashCode` methods

misk-crypto/src/main/kotlin/misk/crypto/KeyManager.kt
```diff
@@ -83,43 +70,23 @@ class MacKeyManager @Inject internal constructor(
  * ```
  */
 @Singleton
-class DigitalSignatureKeyManager @Inject internal constructor(
-  private val injector: Injector
-) {
-  private val signers: HashMap<String, DigitalSignature> = LinkedHashMap()
+class DigitalSignatureKeyManager @Inject internal constructor(injector: Injector)
```
jjestrel: I think prefer the old formatting
```
class DigitalSignatureKeyManager @Inject internal constructor(
  private val injector: Injector
) : MappedKeyManager<DigitalSignature>(injector, DigitalSignature::class.java) {
```

misk-hibernate/src/main/kotlin/misk/hibernate/SecretColumn.kt
```diff
@@ -0,0 +1,32 @@
+package misk.hibernate
+
+/**
```
jjestrel: nice docs

misk-hibernate/src/main/kotlin/misk/hibernate/SecretColumn.kt
```diff
@@ -0,0 +1,32 @@
...
+ *   keys:
+ *     - key_name: "secretColumnKey"
+ * ```
+ * Then, in an entity class:
+ * ```
+ * @Column
+ * @SecretColumn(keyName = "secretColumnKey")
+ * var secret: String
+ * ```
+ * Hibernate fields annotated with [SecretColumn] must be declared as `VARBINARY()`
```
jjestrel: Please call out that these columns can't be part of an index somewhere

misk-hibernate/src/main/kotlin/misk/hibernate/SecretColumnType.kt
```diff
@@ -0,0 +1,93 @@
...
+  private lateinit var _typeConfiguration: TypeConfiguration
+
+  override fun setTypeConfiguration(typeConfiguration: TypeConfiguration) {
+    _typeConfiguration = typeConfiguration
+  }
+
+  override fun getTypeConfiguration(): TypeConfiguration = _typeConfiguration
+
+  override fun setParameterValues(parameters: Properties) {
+    val keyManager = _typeConfiguration.metadataBuildingContext.bootstrapContext.serviceRegistry.injector
```
jjestrel: Guessing we can't directly pass this in somehow?

yoavamit: Yes, hibernate has a different injector being setup, so if you just `@Inject keyManager: AeadKeyManager` you'll get an empty one.

misk-crypto/src/main/kotlin/misk/crypto/KeyManager.kt
yoavamit: Fixed!

misk-hibernate/src/main/kotlin/misk/hibernate/SecretColumn.kt
yoavamit: Good point, I updated the docs

PR #901, opened 2019-04-04
VitessScaleSafetyChecks: Support multiple independent transactions at the same time

Syeberman:
~~**TODO**: Add tests~~

### DIFF
--- a/misk-hibernate-testing/src/main/kotlin/misk/jdbc/VitessScaleSafetyChecks.kt
+++ b/misk-hibernate-testing/src/main/kotlin/misk/jdbc/VitessScaleSafetyChecks.kt
@@ -28,9 +28,12 @@ import java.sql.ResultSet
 import java.sql.SQLException
 import java.sql.SQLSyntaxErrorException
 import java.sql.Timestamp
+import java.util.ArrayDeque
 import java.util.Locale
 import javax.inject.Singleton
 import javax.sql.DataSource
+
+private val vtgateKeyspaceIdRegex = "vtgate:: keyspace_id:([^ ]+)".toRegex()
 
 internal data class Instruction(
   val Opcode: String,
@@ -236,7 +239,7 @@ internal class Explanation {
   }
 }
 
-open class ExtendedQueryExectionListener : QueryExecutionListener, MethodExecutionListener {
+open class ExtendedQueryExecutionListener : QueryExecutionListener, MethodExecutionListener {
   override fun beforeMethod(executionContext: MethodExecutionContext) {
     if (isStartTransaction(executionContext)) {
       beforeStartTransaction()
@@ -380,8 +383,8 @@ class VitessScaleSafetyChecks(
     return proxy
   }
 
-  inner class FullScatterDetector : ExtendedQueryExectionListener() {
-    val count = ThreadLocal.withInitial { 0 }
+  inner class FullScatterDetector : ExtendedQueryExecutionListener() {
+    private val count: ThreadLocal<Int> = ThreadLocal.withInitial { 0 }
 
     override fun beforeQuery(query: String) {
       if (!transacter.isCheckEnabled(Check.FULL_SCATTER)) return
@@ -402,7 +405,7 @@ class VitessScaleSafetyChecks(
     }
   }
 
-  inner class TableScanDetector : ExtendedQueryExectionListener() {
+  inner class TableScanDetector : ExtendedQueryExecutionListener() {
     private val mysqlTimeBeforeQuery: ThreadLocal<Timestamp?> =
         ThreadLocal.withInitial { null }
 
@@ -473,18 +476,17 @@ class VitessScaleSafetyChecks(
     }
   }
 
-  inner class CowriteDetector : ExtendedQueryExectionListener() {
-    private val keyspaceIdsThisTransaction: ThreadLocal<LinkedHashSet<String>> =
-        ThreadLocal.withInitial { LinkedHashSet<String>() }
+  inner class CowriteDetector : ExtendedQueryExecutionListener() {
+
+    private val transactionDeque: ThreadLocal<ArrayDeque<LinkedHashSet<String>>> =
+        ThreadLocal.withInitial { ArrayDeque<LinkedHashSet<String>>() }
 
     override fun beforeStartTransaction() {
       // Connect before the query because connecting spits out a bunch of crap in the general_log
       // that makes it harder for us to get to the thread id
       connect()
 
-      check(keyspaceIdsThisTransaction.get().isEmpty()) {
-        "Transaction state has not been cleaned up, beforeEndTransaction was never executed"
-      }
+      transactionDeque.get().push(LinkedHashSet())
     }
 
     override fun afterQuery(query: String) {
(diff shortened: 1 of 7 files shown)


tirsen:
Very nice! Do you want to add a test?

Syeberman:
I will

misk-hibernate-testing/src/main/kotlin/misk/jdbc/VitessScaleSafetyChecks.kt
```diff
@@ -46,7 +47,7 @@ internal data class Variables(
   val QueriesProcessed: Map<String, Int>
 )
 
-open class ExtendedQueryExectionListener : QueryExecutionListener, MethodExecutionListener {
+open class ExtendedQueryExecutionListener : QueryExecutionListener, MethodExecutionListener {
```
Syeberman: Fixed a typo

misk-hibernate-testing/src/main/kotlin/misk/jdbc/VitessScaleSafetyChecks.kt
```diff
@@ -169,7 +170,7 @@ class VitessScaleSafetyChecks(
 
   private val fullScatterDetector = FullScatterDetector()
   private val crossEntityGroupTransactionDetector = CowriteDetector()
-  val enabled = ThreadLocal.withInitial { true }
+  val enabled: ThreadLocal<Boolean> = ThreadLocal.withInitial { true }
```
Syeberman: IntelliJ recommends explicit types when the implicit type comes from Java.

misk-hibernate-testing/src/main/kotlin/misk/jdbc/VitessScaleSafetyChecks.kt
```diff
@@ -356,30 +357,3 @@ private inline fun <T, R> ThreadLocal<T>.withValue(value: T, body: () -> R): R {
     set(prev)
   }
 }
-
-/**
- * Returns a list containing elements after the last element that satisfied the given [predicate]
- * or the empty list if no elements satisfy the predicate.
- */
-inline fun <T> List<T>.sublistAfterMatch(predicate: (T) -> Boolean): List<T> {
```
Syeberman: Unused.

misk-hibernate-testing/src/main/kotlin/misk/jdbc/VitessScaleSafetyChecks.kt
```diff
@@ -356,30 +357,3 @@ private inline fun <T, R> ThreadLocal<T>.withValue(value: T, body: () -> R): R {
...
-      return ArrayList<T>(expectedSize).apply {
-        while (iterator.hasNext())
-          add(iterator.next())
-      }
-    }
-  }
-  return emptyList()
-}
-
-fun String.quoteSql(escape: String): String =
```
Syeberman: Unused.

misk-hibernate-testing/src/test/kotlin/misk/jdbc/ParseQueryPlansTest.kt
```diff
@@ -10,7 +10,7 @@ import kotlin.test.assertEquals
 import kotlin.test.assertFalse
 import kotlin.test.assertTrue
 
-class VitessScaleSafetyChecksTest {
+class ParseQueryPlansTest {
```
Syeberman: I renamed the file as this class is only testing the query plan parser.

misk-hibernate/src/main/kotlin/misk/hibernate/DbUnsharded.kt
```diff
@@ -3,4 +3,4 @@ package misk.hibernate
 /**
  * Marker interface for persistent entities that have a bounded growth and do not require sharding.
  */
-interface DbUnsharded<T : DbEntity<T>> : DbEntity<T>
\ No newline at end of file
+interface DbUnsharded<T : DbUnsharded<T>> : DbEntity<T>
```
Syeberman: This more closely mirrors how `DbRoot`, `DbSharded`, etc are defined.

misk-hibernate/src/main/kotlin/misk/hibernate/RealTransacter.kt
```diff
@@ -57,7 +57,7 @@ internal class RealTransacter private constructor(
 
   override fun isCheckEnabled(check : Check): Boolean {
     val session = threadLocalSession.get()
-    return session != null && !session.disabledChecks.contains(check)
+    return session == null || !session.disabledChecks.contains(check)
```
Syeberman: This will enable all checks for raw Hibernate sessions, which was the behaviour previously. If we have a need to disable these checks for Hibernate sessions, we can tackle that in the future. But it's easier to disable a check on something that expects it, than to enable a check on something that doesn't.

tirsen: I agree, I think it makes sense.

misk-hibernate-testing/src/main/kotlin/misk/jdbc/VitessScaleSafetyChecks.kt
tirsen: sure

misk-hibernate/src/main/kotlin/misk/hibernate/DbUnsharded.kt
tirsen: yes!

misk-hibernate-testing/src/main/kotlin/misk/jdbc/VitessScaleSafetyChecks.kt
```diff
@@ -29,6 +29,7 @@ import java.sql.SQLException
 import java.sql.SQLSyntaxErrorException
 import java.sql.Timestamp
 import java.util.Locale
+import java.util.Stack
```
tirsen: `java.util.Stack` is legacy and no longer recommended. You don't need thread safety here so you should use `java.util.ArrayDeque` instead.

https://stackoverflow.com/questions/47933459/when-should-i-use-java-util-stack-vs-my-own-implementation

misk-hibernate-testing/src/main/kotlin/misk/jdbc/VitessScaleSafetyChecks.kt
```diff
@@ -473,18 +474,18 @@ class VitessScaleSafetyChecks(
     }
   }
 
-  inner class CowriteDetector : ExtendedQueryExectionListener() {
-    private val keyspaceIdsThisTransaction: ThreadLocal<LinkedHashSet<String>> =
-        ThreadLocal.withInitial { LinkedHashSet<String>() }
+  inner class CowriteDetector : ExtendedQueryExecutionListener() {
+
+    private val vtgateKeyspaceIdRegex = "vtgate:: keyspace_id:([^ ]+)".toRegex()
```
tirsen: I think this should be in the companion object so we don't have to parse the regexp every time we start a transaction.

misk-hibernate-testing/src/main/kotlin/misk/jdbc/VitessScaleSafetyChecks.kt
Syeberman: Done

misk-hibernate-testing/src/main/kotlin/misk/jdbc/VitessScaleSafetyChecks.kt
Syeberman: Done

PR #902, opened 2019-04-04
Initialize the test module like the real one

yoavamit:
Otherwise, the KeyManager isn't initialized properly


ryanhall07:
is there a test we can write to catch this problem/verify?

PR #903, opened 2019-04-04
enable gradle build cache on circleci

mightyguava:
Totally copied from https://github.com/square/okhttp/pull/4866


mightyguava:
Pretty fast if no tests end up actually running :) 
image

PR #905, opened 2019-04-05
Support multiple keys per cipher

yoavamit:
In order to support apps being deployed in multiple regions.

Each `Cipher` object can have multiple `KeysetHandle`s associated with it.
When encrypting, the first `KeysetHandle` is used. Tink also includes the key ID used in the payload.
When decrypting, iterate through the `KeysetHandle`s and return the result of the first successful decryption. 
Tink will fail to decrypt and throw a `GeneralSecurityException` if the encrypted data doesn't match any of the key IDs listed in a `KeysetHandle` and non of them managed to decrypt.
In that case, we'll just try the next `KeysetHandle` until one of them succeeds.

If no `KeysetHandle` was able to decrypt the data, throw an exception.


misk-crypto/src/main/kotlin/misk/crypto/Cipher.kt
```diff
@@ -31,23 +33,23 @@ interface Cipher {
...
 }
 
 /**
  * Real implementation of the [Cipher] interface.
  * This class uses key [KeysetHandle] that's been decrypted by the KMS when the module was loaded.
  */
 class RealCipher internal constructor(
-  private val keysetHandle: KeysetHandle,
-  private val masterKey: Aead
+  private val keys: List<Pair<KeysetHandle, Aead>>
```
jjestrel: Maybe this should be `List<SomeDataClass>` instead

yoavamit: Agree, I usually don't like using `Pair`, but Kotlin is really accommodating with these objects (w/ stuff like `{ (left, right) -> ... }`) and I got tempted.

I'll switch it to something more readable.

misk-crypto/src/main/kotlin/misk/crypto/Cipher.kt
```diff
@@ -23,6 +24,7 @@ interface Cipher {
    * Decrypt the given data.
    * If anything goes wrong, including failing to decipher the data,
    * throws a [java.security.GeneralSecurityException]
+   * and [NullPointerException] if no suitable key was found to decrypt the data.
```
ryanhall07: prefer to throw a more meaningful exception than NPE.

misk-crypto/src/main/kotlin/misk/crypto/CryptoModule.kt
```diff
@@ -21,32 +21,40 @@ class CryptoModule(
...
 
-    private fun readKey(keyConfig: Secret<String>, masterKey: Aead): Cipher {
-      val keysetHandle = KeysetHandle.read(JsonKeysetReader.withString(keyConfig.value), masterKey)
-      return RealCipher(keysetHandle, masterKey)
+    private fun readKey(keyConfig: Secret<String>, masterKey: Aead): KeysetHandle {
+      return KeysetHandle.read(JsonKeysetReader.withString(keyConfig.value), masterKey)
+    }
+
+    private fun getMasterKeyUri(keyDetails: EncryptedKey): String {
+      check(!(keyDetails.gcp_key_uri != null && keyDetails.aws_kms_key_alias != null))
```
wesleyk: can we leave a message here noting that exactly one has to be set?

I think we can simplify both of these checks to this as well, if you don't find it confusing.

`check((keyDetails.gcp_key_uri == null) != (keyDetails.aws_kms_key_alias == null))`

PR #921, opened 2019-04-11
Changed crypto configuration

yoavamit:
Now there's only one place to specify the KMS URI.
This makes the module environment agnostic.

Users now must specify the full KMS URI to use with their encrypted keys.

### DIFF
--- a/misk-crypto/README.md
+++ b/misk-crypto/README.md
@@ -34,6 +34,7 @@ tinkey create-keyset --key-template AES256-GCM --master-key-uri aws-kms://arn:km
 Then, to specify a new `Cipher` key called "myKey", add the following in your app's configuration file:
 ```$yaml
 crypto:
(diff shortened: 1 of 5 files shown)


misk-crypto/README.md
```diff
@@ -34,6 +34,7 @@ tinkey create-keyset --key-template AES256-GCM --master-key-uri aws-kms://arn:km
 Then, to specify a new `Cipher` key called "myKey", add the following in your app's configuration file:
 ```$yaml
 crypto:
+  kms_uri: "aws-kms://arn:kms:<region>:<account-id>:key/<key-id>"
```
mightyguava: nit: `key_uri` instead? kms is specifically Amazon KMS isn't it?

yoavamit: KMS is acronym for Key Management System, both Google and Amazon use it.

wesleyk:
Nice!

PR #922, opened 2019-04-11
Web Actions Tab: Rewritten with data submitted to endpoint

adrw:
Apr 15, 2019 Web Actions Typed Forms


wesleyk:
Nice! Could we increase the indent a bit on nested fields? From the gif it looks a bit hard to tell which child falls under which parent.

PR #926, opened 2019-04-16
Removed eager initialization

yoavamit:
When running in STAGING or PRODUCTION environemnts, this object needs to
communicate with a live KMS, which makes it very difficult to test -
specifically the InjectorTest which loads each env's complete
configuration and makes sure Guice can initialize it.
Removing the `.asEagerSingleton()` part means that when the `Cipher`
object is constructed it won't need the `KmsClient` until first use.

There are other ways to exercise all the relevant code paths in `Cipher`
for testing without using a real KMS.

### DIFF
--- a/misk-crypto/src/main/kotlin/misk/crypto/CryptoModule.kt
+++ b/misk-crypto/src/main/kotlin/misk/crypto/CryptoModule.kt
@@ -27,7 +27,6 @@ class CryptoModule(
       bind<Cipher>()
           .annotatedWith(Names.named(key.key_name))
           .toProvider(CipherProvider(config.kms_uri, key))
-          .asEagerSingleton()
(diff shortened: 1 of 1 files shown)


mightyguava:
Is there an initialization error? If there's initialization required for a binding we usually make it a Service, which doesn't get run on injector tests. Or is this in tink and outside your control?

dhanji:
You might want to add `in(Singleton.class)` to preserve the scope. As is this is turning it from a singleton into a non-singleton (unless overridden in the implementation with `@Singleton`)

yoavamit:
Closing this PR because there are other changes I'd like to introduce (https://github.com/square/misk/pull/927) and I don't see a reason to merge this by itself

PR #927, opened 2019-04-16
Added support for using more Tink primitives (Mac)

yoavamit:
Key changes here:
- Added support for the `Mac` Tink primitive `CryptoModule` will now be able to initialize and inject
- Got rid of the `Cipher` object. It started as a basic wrapper for Tink's `Aead` implementation, but we figured it's probably best to stick with the interface Google provides here.
- Split the `KeyManager` object to `MacKeyManager` and `AeadKeyManager` - it makes it easier to reason and handle the object the `get` method returns here.
- Removed the `asEagerSingleton()` when initializing these objects. When Tink constructs these objects, it makes a request to resolve an encrypted key from AWS KMS, and we don't want that.


misk-crypto/src/main/kotlin/misk/crypto/CryptoConfig.kt
```diff
@@ -13,5 +13,11 @@ data class CryptoConfig(
...
 data class Key(
   val key_name: String,
+  val key_type: KeyType,
   val encrypted_key: Secret<String>
-) : Config
\ No newline at end of file
+) : Config
+
+enum class KeyType {
+  ENCRYPTION,
```
wesleyk: worth adding docs explaining when folks would use each type?

misk-crypto/src/main/kotlin/misk/crypto/CryptoModule.kt
```diff
@@ -22,28 +29,77 @@ class CryptoModule(
   override fun configure() {
     requireBinding(KmsClient::class.java)
     AeadConfig.register()
+    MacConfig.register()
 
+    check(config.keys?.map { it.key_name }?.distinct()?.size == config.keys?.size) {
+      "Found duplicate key name"
```
wesleyk: add the key name?

misk-crypto/src/main/kotlin/misk/crypto/KeyManager.kt
```diff
@@ -1,21 +1,56 @@
...
-  private var keys: HashMap<String, Cipher> = LinkedHashMap()
+  internal operator fun set(name: String, aead: Aead) {
+    aeads[name] = aead
+  }
+
+  operator fun get(name: String): Aead? {
+    var aead = aeads[name]
+    if (aead == null) {
+      try {
+        aead = injector.getInstance(Key.get(Aead::class.java, Names.named(name)))
```
wesleyk: why lazily create these instead of doing it all on startup?

misk-crypto/src/test/kotlin/misk/crypto/CryptoModuleTest.kt
```diff
@@ -22,17 +23,55 @@ class CryptoModuleTest {
...
+  private fun getInjector(keyMap: List<Pair<String, KeysetHandle>>): Injector{
+    val keys = keyMap.map {
+      var keyType = KeyType.ENCRYPTION
+      if (it.second.keysetInfo.getKeyInfo(0).typeUrl.endsWith("HmacKey")) {
+        keyType = KeyType.MAC
+      }
+      Key(it.first, keyType, generateEncryptedKey(it.second))
+    }
+    val config = CryptoConfig(keys, "test_master_key")
+    return  Guice.createInjector(CryptoTestModule(), CryptoModule(config))
```
wesleyk: /nit extra space

misk-crypto/src/main/kotlin/misk/crypto/KeyManager.kt
yoavamit: This is kind of a weird situation...
These are actually being `set()` in the `KeyManager` by their respective providers in `CryptoModule`. For example:
```
return MacFactory.getPrimitive(keysetHandle)
          .also { keyManager[key.key_name] = it }
```
And changing providers binding to `asEagerSingleton` will work and we won't need this lazy initialization.
But, it would also screw up the `InjectorTest` in every app, because when an `Aead` or `Mac` object is initialized, it needs the AWS KMS client to decrypt its key - making the `InjectorTest` test fail in every environment that's not development/testing.

By removing the `asEagerSingleton` in the module class, it means that `Aead` and `Mac` object are only initialized on first use - which is a good thing - but as a side effect also means that they won't be set in the `KeyManager` until first use *via injection*.
Which could lead to keys being missing from the `KeyManager` if they're being fetched from it on the first time.

mightyguava: You could probably also multi-bind the primitives to avoid using the injector in your key managers

misk-crypto/src/main/kotlin/misk/crypto/CryptoConfig.kt
```diff
@@ -8,10 +8,42 @@ import misk.config.Secret
  */
```
mightyguava: thanks for the documentation!

PR #929, opened 2019-04-16
Explicit propagation of span to e.g. sub threads

tirsen:


jjestrel:
`ActionScopedExecutorService` do what you want already?

tirsen:
I ended up just using `TracedExecutorService` instead

PR #930, opened 2019-04-17
Log the root cause of startup failures up front

dhanji:
otherwise it's buried under a mountain of:

```java
java.lang.IllegalStateException: Expected to be healthy after starting. The following services are not running: {FAILED=[SchemaValidatorService [FAILED]]}

	at com.google.common.util.concurrent.ServiceManager$ServiceManagerState.checkHealthy(ServiceManager.java:741)
	at com.google.common.util.concurrent.ServiceManager$ServiceManagerState.awaitHealthy(ServiceManager.java:568)
	at com.google.common.util.concurrent.ServiceManager.awaitHealthy(ServiceManager.java:329)
```


tirsen:
YAY!

PR #931, opened 2019-04-17
Allow specifying number of retries in Transacter

surrahman:
Quick change to allow us to specify the number of retries in `Transacter`.

### DIFF
--- a/misk-hibernate/src/main/kotlin/misk/hibernate/RealTransacter.kt
+++ b/misk-hibernate/src/main/kotlin/misk/hibernate/RealTransacter.kt
@@ -135,7 +135,7 @@ internal class RealTransacter private constructor(
     }
   }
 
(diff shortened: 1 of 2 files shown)


misk-hibernate/src/main/kotlin/misk/hibernate/RealTransacter.kt
```diff
@@ -135,7 +135,7 @@ internal class RealTransacter private constructor(
     }
   }
 
-  override fun retries(): Transacter = withOptions(options.copy(maxAttempts = 2))
+  override fun retries(maxAttempts: Int): Transacter = withOptions(options.copy(maxAttempts = maxAttempts))
```
wesleyk: how about we set the default to 2 to maintain compatibility?

surrahman: Sure why not

PR #935, opened 2019-04-18
Add update_version to events

qelper:


misk-events/src/main/kotlin/misk/events/Event.kt
```diff
@@ -19,6 +19,13 @@ data class Event(
   /** a global unique id for the event */
   val id: ByteString,
 
+  /**
+   * Events often represent a stream of state changes or entity updates; the update_version
+   * can be used to indicate the version of the updated entity or state machine at the time
+   * the event was generated. Not all events will have this populated, therefore optional.
```
ryanhall07: we should make it clear that not setting this field is considered deprecated and all new events should have this field. eventually it will be required.

qelper: update the comment to reflect that

PR #936, opened 2019-04-18
Web Actions: Return empty body for forms without input

adrw:


adrw:
Going to combine with follow up PR

PR #940, opened 2019-04-20
[DNM] Debug swallowed 500s

jjestrel:


adrw:
Bug was found elsewhere in the stack.

PR #941, opened 2019-04-23
Add Session.loadOrNull

shellderp:
Motivation: I need to load by id, but handle missing entities. The id comes from user input.

`session.load()` can already return null from java code, but it's not marked `T?` in kotlin. Thanks kotlin.

Ideally we'd make `load()` return `T?` but that would break a lot of code.

### DIFF
--- a/misk-hibernate/src/main/kotlin/misk/hibernate/RealTransacter.kt
+++ b/misk-hibernate/src/main/kotlin/misk/hibernate/RealTransacter.kt
@@ -247,6 +247,10 @@ internal class RealTransacter private constructor(
       return session.get(type.java, gid)
     }
 
+    override fun <T : DbEntity<T>> loadOrNull(id: Id<T>, type: KClass<T>): T? {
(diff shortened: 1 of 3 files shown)


swankjesse:
I don’t think this will work. I think Hibernate will give you a lazy instance that crashes when you try to use it.

swankjesse:
Adding a test case will show you the truth!

shellderp:
> Adding a test case will show you the truth!

I tested in my service but I'll add a misk test!
The docs for hibernate Session.get():
```
        /**
	 * Return the persistent instance of the given entity class with the given identifier,
	 * or null if there is no such persistent instance. (If the instance is already associated
	 * with the session, return that instance. This method never returns an uninitialized instance.)
	 *
	 * @param entityType The entity type
	 * @param id an identifier
	 *
	 * @return a persistent instance or null
	 */
	<T> T get(Class<T> entityType, Serializable id);
```

PR #943, opened 2019-04-24
Fix potential NPE in test log statement

dhanji:


misk-testing/src/main/kotlin/misk/testing/MiskTestExtension.kt
```diff
@@ -82,7 +82,7 @@ internal class MiskTestExtension : BeforeEachCallback, AfterEachCallback {
           serviceManager.startAsync().awaitHealthy(60, TimeUnit.SECONDS)
         } catch (e: IllegalStateException) {
           // Unearth the root cause so we can see it in logs easily.
-          log.error("Startup failure -- ${e.cause!!.message}")
```
ryanhall07: lol kotlin definitely removes NPEs

mightyguava: the docs made me laugh:

> The third option is for NPE-lovers: the not-null assertion operator (!!) converts any value to a non-null type and throws an exception if the value is null.

PR #944, opened 2019-04-24
Added support for digital signatures

yoavamit:
`misk.crypto` now also has support for providing Tink primitives for signing and verifying signed data.

### DIFF
--- a/misk-crypto/src/main/kotlin/misk/crypto/KeyManager.kt
+++ b/misk-crypto/src/main/kotlin/misk/crypto/KeyManager.kt
@@ -119,6 +119,7 @@ class DigitalSignatureKeyManager @Inject internal constructor(
     }
     return verifier
   }
-  data class DigitalSignature(val signer: PublicKeySign, val verifier: PublicKeyVerify)
 }
+
+internal data class DigitalSignature(val signer: PublicKeySign, val verifier: PublicKeyVerify)
 
--- a/misk-crypto/src/main/kotlin/misk/crypto/KeyManager.kt
+++ b/misk-crypto/src/main/kotlin/misk/crypto/KeyManager.kt
@@ -2,6 +2,8 @@ package misk.crypto
 
 import com.google.crypto.tink.Aead
 import com.google.crypto.tink.Mac
+import com.google.crypto.tink.PublicKeySign
+import com.google.crypto.tink.PublicKeyVerify
 import com.google.inject.ConfigurationException
 import com.google.inject.Inject
 import com.google.inject.Injector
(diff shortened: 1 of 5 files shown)


misk-crypto/src/main/kotlin/misk/crypto/KeyManager.kt
```diff
@@ -67,4 +69,56 @@ class MacKeyManager @Inject internal constructor(
...
+  }
+
+  fun getSigner(name: String): PublicKeySign? {
+    var signer = signers[name]?.signer
+    if (signer == null) {
+      try {
+        signer = injector.getInstance(Key.get(PublicKeySign::class.java, Names.named(name)))
+        val verifier = injector.getInstance(Key.get(PublicKeyVerify::class.java, Names.named(name)))
+        this[name] = DigitalSignature(signer, verifier)
+      } catch (e: ConfigurationException) {
```
wesleyk: should we be logging or exploding more loudly here? When would this happen?

mightyguava:
I was thinking about this quite differently:

The private key lives with the service, encrypted using the master key, and can be used to create both the Signer and Verifier.

The public key is in plaintext, not sure where it lives yet, but can be used to create the Verifier only.

yoavamit:
> I was thinking about this quite differently:
> 
> The private key lives with the service, encrypted using the master key, and can be used to create both the Signer and Verifier.
> 
> The public key is in plaintext, not sure where it lives yet, but can be used to create the Verifier only.
> 
> At least, this is the approach I took in gap

Yeah, I think making the public key somehow exportable could really be useful here.
Especially if you want to use it outside the service (in a client library for example).
OTOH, how is key rotation and syncing being handled then?
FWIW, it's always possible to get the `KeysetHandle` for the public key and write it to somewhere, so I don't think that should be a blocker

misk-crypto/src/main/kotlin/misk/crypto/KeyManager.kt
yoavamit: It shouldn't, but any call to `Injector:getInstance` might produce a `ConfigurationException` and should be wrapped with a `try/catch`.

I know Kotlin has some nicer ways/tricks to deal with checked exceptions - is there anything you think I should change here?

wesleyk: hmm, I don't think `ConfigurationException` is a checked exception, can you try removing this?

mightyguava:
> Especially if you want to use it outside the service (in a client library for example).

That's exactly how we'll need to use the public key for sessions. Sessions service keeps the private key for signing, but clients like Franklin will be using the public key for verification.

The use-case we should try to address in `misk-crypto` first is loading a public key for signature verification, but this change only supports loading a private key.

> OTOH, how is key rotation and syncing being handled then? FWIW, it's always possible to get the KeysetHandle for the public key and write it to somewhere, so I don't think that should be a blocker

I think we should come up with a strategy for propagating public keys and keeping them in sync before pushing this forward. Some options that I've briefly thought about:

* Since we haven't actually gotten to key rotation support yet, maybe it's OK to just keep the public keys checked in on the client side?
* We'll have an `sqm` command soon for generating keys, maybe that command can put them somewhere the service can load on startup?
* Maybe for sessions it's ok for the client to call sessions and ask for the public key?

PR #947, opened 2019-04-26
Add a FakeJobQueue for testing

jmuia:


misk-jobqueue-testing/src/main/kotlin/misk/jobqueue/FakeJobHandlerModule.kt
```diff
@@ -0,0 +1,37 @@
+package misk.jobqueue
+
+import misk.inject.KAbstractModule
+import kotlin.reflect.KClass
+
+class FakeJobHandlerModule<T : JobHandler> private constructor(
```
ryanhall07: i'd rather not have separate modules for testing. seems reasonable that a module with an API also includes the proper fake for testing

misk-jobqueue-testing/src/main/kotlin/misk/jobqueue/FakeJobQueue.kt
```diff
@@ -0,0 +1,58 @@
...
+
+import misk.tokens.TokenGenerator
+import java.time.Duration
+import java.util.concurrent.ConcurrentHashMap
+import java.util.concurrent.ConcurrentLinkedDeque
+import javax.inject.Inject
+import javax.inject.Singleton
+
+@Singleton
+class FakeJobQueue @Inject constructor(
```
ryanhall07: I think it would be nice if library fakes had tests. Even better if we had a test that verifies `FakeJobQueue` behaves just like the real job queue using in the in memory SQS

misk-jobqueue-testing/src/main/kotlin/misk/jobqueue/FakeJobHandlerModule.kt
jmuia: that is reasonable, though it means you always include testing code. Misk's a bit conflicted in terms of having separated testing modules right now.

misk-jobqueue-testing/src/main/kotlin/misk/jobqueue/FakeJobQueue.kt
jmuia: 👍 I had 'em in the jobqueue exemplar, but I'll add here.

misk-jobqueue-testing/src/main/kotlin/misk/jobqueue/FakeJobHandlerModule.kt
ryanhall07: i think including testing code is the right idea. why wouldn't you want this fake for your tests if you're using a job queue?

jmuia: you'd potentially not want it in a prod artifact. but for testing, definitely.

ryanhall07: that's a fair point. personally I don't think it's worth it, but we should probably poll the misk audience to get a consensus for consistency

misk-jobqueue-testing/src/main/kotlin/misk/jobqueue/FakeJobQueue.kt
```diff
@@ -0,0 +1,58 @@
+package misk.jobqueue
```
mightyguava: It's not clear to me how to use this package. Could you add docs?

misk-jobqueue-testing/src/main/kotlin/misk/jobqueue/FakeJobQueue.kt
```diff
@@ -0,0 +1,58 @@
...
+    val jobs = jobQueues[queueName]
+    return jobs?.toList() ?: listOf()
+  }
+
+  fun handleJobs(queueName: QueueName) {
+    val jobHandler = jobHandlers[queueName]!!
+    val jobs = jobQueues[queueName] ?: return
+
+    while (true) {
+      val job = jobs.poll() ?: break
```
mightyguava: This should probably support a timeout or your tests might hang when there are bugs.

jmuia: @mightyguava added a timeout

misk-jobqueue-testing/src/main/kotlin/misk/jobqueue/FakeJobQueue.kt
jmuia: done!

misk-jobqueue-testing/src/main/kotlin/misk/jobqueue/FakeJobQueue.kt
jmuia: @ryanhall07 added some tests

misk-jobqueue-testing/src/main/kotlin/misk/jobqueue/FakeJobQueue.kt
mightyguava: I actually meant you should pass the timeout into the `poll()`. That's the thing that might block indefinitely if you try to consume jobs that never got queued. Or I might be misunderstanding how this package is to be used.

misk-jobqueue-testing/src/test/kotlin/misk/jobqueue/FakeJobQueueTest.kt
```diff
@@ -0,0 +1,210 @@
...
+  @Test
+  fun failsWithTimeout() {
+    assertThat(fakeJobQueue.peekJobs(GREEN_QUEUE)).isEmpty()
+
+    exampleJobEnqueuer.enqueueGreen("sleep-10s")
+
+    val jobs = fakeJobQueue.peekJobs(GREEN_QUEUE)
+    assertThat(jobs).hasSize(1)
+
+    val timeout = Duration.ofSeconds(2)
```
mightyguava: Does this test take 2 seconds to run...? Could we lower that to like, 10ms?

misk-jobqueue-testing/src/main/kotlin/misk/jobqueue/FakeJobQueue.kt
jmuia: ahhh, I thought you might be worried about a self-enqueuing job or similar. `poll()` is non-blocking and returns null if the queue is empty.

mightyguava: Oh, I totally misunderstood. That makes sense. Thanks

jmuia: I'm just going to get rid of this. As-is It's not that useful since the jobs are running in the same thread. If it is problematic we can start handling the fake jobs in a separate thread and interrupt it after a timeout.

PR #948, opened 2019-04-27
Prefer more modern implementation/testImplementation tasks over compile/testCompile

wesleyk:


wesleyk:
eh looks like Misk actually depends on some transitive dependencies moreso than our polyrepo codebases. Will tease out what we actually need as compile.

PR #954, opened 2019-05-01
Use wire Grpc client instead of misks

Synesso:


mightyguava:
Yay grpc! Are there docs for this client? Is this the code? https://github.com/square/wire/tree/master/wire-grpc-client. How does it relate to https://github.com/grpc/grpc-java?

Synesso:
> Yay grpc! Are there docs for this client? Is this the code? https://github.com/square/wire/tree/master/wire-grpc-client. 

That looks like the one. Any docs available @oldergod ?

> How does it relate to https://github.com/grpc/grpc-java?

 AFAIK (which isn't much, admittedly), there's no connection.

mightyguava:
Hmm it doesn't look like a full grpc client impl with all the goodies: https://github.com/grpc/grpc/tree/master/doc. Are we rebuilding a subset of grpc features in this client?

Synesso:
@mightyguava I guess that's another question for @oldergod / wire-grpc-client.

misk-grpc-tests/src/main/java/routeguide/RouteGuide.kt
```diff
@@ -0,0 +1,41 @@
...
+package routeguide
+
+import com.squareup.wire.Service
+import com.squareup.wire.WireRpc
+import kotlin.Pair
+import kotlinx.coroutines.Deferred
+import kotlinx.coroutines.channels.ReceiveChannel
+import kotlinx.coroutines.channels.SendChannel
+
+// TODO(jwilson): generate this and remove it from git.
```
oldergod: This is already generated code though?

misk-grpc-tests/src/main/kotlin/misk/grpc/miskclient/MiskGrpcClientModule.kt
```diff
@@ -43,5 +44,14 @@ class MiskGrpcClientModule : KAbstractModule() {
   fun provideGrpcClient(
     client: OkHttpClient,
     @Named("grpc server") url: HttpUrl
-  ) = GrpcClient(client, url)
+  ): GrpcClient {
+    return GrpcClient.Builder()
+        .client(client)
+        .baseUrl(url.toString()) // TODO(jwilson): this should also take a URL. Come on guys.
```
oldergod: https://github.com/square/wire/issues/945

mightyguava:
> We intend to generate code for services from a server point of view; what else do you need?

I can't claim to know what we _need_ in a grpc client. Haven't personally used it in a high traffic setting like square. But grpc has more stuff than just the wire protocol.

From the FAQ, it has:
* interaction with flow-control at the application layer
* cascading call-cancellation (this is super cool)
* client-side load balancing & failover (this is kinda annoying actually, but they make it out to be something that becomes more useful at larger scale?)

It should also be really stable now since it's been available for several years. We probably don't actually need all of these features, but I'd love to understand why we want our own implementation of the client, versus just using the official implementation.

I really like that the Kotlin/wire for protos and grpc is much cleaner with `wire-grpc-client`, but that seems like something we could build as an adapter, rather than a standalone re-implementation.

swankjesse:
My plan is to do gRPC on the same rails as our web APIs. Same TLS, tracing, metrics, auth, config, load balancing, routing, ... everything. The only thing different between web and gRPC is that gRPC specifies schema & action names in .proto.

If we use Google’s gRPC stack we can’t share much with the web.

misk-grpc-tests/src/main/java/routeguide/RouteGuide.kt
swankjesse: Wasn't getting emitted for us. Maybe fixed beyond alpha01?

PR #957, opened 2019-05-01
Catch Throwable for tasks scheduled via RepeatedTaskQueue#schedule

nbroyles:


misk/src/main/kotlin/misk/tasks/RepeatedTaskQueue.kt
```diff
@@ -76,16 +76,30 @@ class RepeatedTaskQueue @VisibleForTesting internal constructor(
...
-   * next execution time
+   * next execution time. Provide an optional retryDelayOnFailure parameter to determine when
+   * the job should be retried in the case of an unhandled exception by the client
    */
-  fun schedule(delay: Duration, task: () -> Result) {
+  fun schedule(delay: Duration, retryDelayOnFailure: Duration? = null, task: () -> Result) {
+    val wrappedTask: () -> Result = {
+      try {
+        task()
+      } catch (th: Throwable) {
```
mightyguava: I think you want `Exception`? `Errors` aren't usually meant to be caught

ryanhall07: I think you want `Throwable`. if some library accidentally throws an `AssertionError` that will kill the scheduling thread.

misk/src/main/kotlin/misk/tasks/RepeatedTaskQueue.kt
```diff
@@ -100,7 +114,7 @@ class RepeatedTaskQueue @VisibleForTesting internal constructor(
     failureBackoff: Backoff = ExponentialBackoff(timeBetweenRuns, defaultMaxDelay, defaultJitter),
     task: () -> Status
   ) {
-    val wrappedTask: () -> Result = {
+    schedule(delay = initialDelay) {
```
ryanhall07: this is kind of strange since it results in double wrapping the task. can this just call enqueue?

nbroyles: https://github.com/square/misk/pull/960

PR #959, opened 2019-05-01
Allow multiple zookeeper bindings

ryanhall07:
In addition, split a ZookeeperModule out of the ZkLeaseModule, so it can
be reused by other Zookeeper installs later.

The motivation is to add a connection to a global zk for cluster
weights.


misk-zookeeper/src/main/kotlin/misk/clustering/zookeeper/ZkLeaseModule.kt
```diff
@@ -9,20 +9,20 @@ import misk.inject.KAbstractModule
...
+ */
+class ZkLeaseModule(private val config: ZookeeperConfig) : KAbstractModule() {
   override fun configure() {
-    bind<ZookeeperConfig>().toInstance(config)
-    multibind<Service>().to<ZkService>()
     multibind<Service>().to<ZkLeaseManager>()
     multibind<Service>().to(RepeatedTaskQueue::class.toKey(ForZkLease::class)).asSingleton()
     bind<LeaseManager>().to<ZkLeaseManager>()
-    bind<CuratorFramework>().toProvider(CuratorFrameworkProvider::class.java).asSingleton()
+    install(ZookeeperModule(config, ForZkLease::class))
```
wesleyk: nice

misk-zookeeper/src/test/kotlin/misk/clustering/zookeeper/ZkTestModule.kt
```diff
@@ -27,9 +29,10 @@ internal class ZkTestModule : KAbstractModule() {
     val truststrorePath = this::class.java.getResource("/zookeeper/truststore.jks").path
     bind<String>().annotatedWith<AppName>().toInstance("my-app")
 
+
```
wesleyk: revert

misk-zookeeper/src/main/kotlin/misk/zookeeper/ZookeeperModule.kt
```diff
@@ -0,0 +1,44 @@
...
+import org.apache.curator.framework.CuratorFramework
+import javax.inject.Provider
+import kotlin.reflect.KClass
+
+/**
+ * Binds a [CuratorFramework] for an application to use. A [Service] is also installed to manage
+ * the lifecycle of the [CuratorFramework].
+ *
+ * If an application needs to connect to multiple Zookeepers, an optional qualifier can be passed
+ * resulting in an annotated [CuratorFramework] binding.
```
wesleyk: worth pointing to leases as an example?

wesleyk: eh the example you have below is already plenty descriptive

misk-zookeeper/src/main/kotlin/misk/clustering/zookeeper/ZkService.kt
```diff
@@ -5,16 +5,13 @@ import com.google.inject.Key
...
 
-@Singleton
-internal class ZkService @Inject internal constructor(
+internal class ZkService internal constructor(
   private val curatorFramework: CuratorFramework
 ) : AbstractIdleService(), DependentService {
 
   override val consumedKeys: Set<Key<*>> = setOf()
-  override val producedKeys: Set<Key<*>> = setOf(ZookeeperModule.serviceKey)
+  override val producedKeys: Set<Key<*>> = setOf(ZkLeaseModule.serviceKey)
```
mightyguava: won't this end up in multiple instances of ZkService producing the same key? Might be useful to have a test where you start multiple zookeepers

ryanhall07: fixed this up in a new commit 93d007f

I actually went with reusing the same zookeeper instances in tests. Not sure there is a lot of value in actually having separate instances running

PR #963, opened 2019-05-02
Enforce envelope encryption at AEAD provider

yan:


misk-crypto/src/main/kotlin/misk/crypto/CryptoModule.kt
```diff
@@ -72,16 +75,24 @@ class CryptoModule(
...
-      return AeadFactory.getPrimitive(keysetHandle)
-          .also { keyManager[key.key_name] = it }
+      val kek = AeadFactory.getPrimitive(keysetHandle)
+      val envelopeKey = KmsEnvelopeAead(DEK_TEMPLATE, kek)
+
+      return envelopeKey.also { keyManager[key.key_name] = it }
     }
 
+    companion object {
+      val DEK_TEMPLATE : KeyTemplate = AeadKeyTemplates.AES128_GCM
```
mightyguava: You are not using this outside this file. Do you mean to use it in tests?

misk-crypto/src/main/kotlin/misk/crypto/CryptoModule.kt
```diff
@@ -72,16 +75,24 @@ class CryptoModule(
     }
   }
 
-  private class AeadProvider(val keyUri: String, val key: Key) : Provider<Aead> {
+  /**
+   * We do not support creating bare AEAD keys, only via envelope encryption.
```
mightyguava: nit: Probably better to word it as "we only support Aead keys via envelope encryption"

misk-crypto/src/main/kotlin/misk/crypto/CryptoModule.kt
yan: Just wanted to keep the DEK template visually separate from its use, if we want to adjust it later. Don't mind inlining it however.

misk-crypto/src/main/kotlin/misk/crypto/CryptoModule.kt
```diff
@@ -49,7 +52,7 @@ class CryptoModule(
         KeyType.AEAD -> {
           bind<Aead>()
               .annotatedWith(Names.named(key.key_name))
-              .toProvider(AeadProvider(config.kms_uri, key))
+              .toProvider(AeadEnvelopeProvider(config.kms_uri, key))
```
yoavamit: Might be worth adding documentation about this in the README file on in the `AeadKeyManager` javadoc

yan: 👍 updated the docstring to expand a bit on operating mode.

Issue #41, opened 2017-12-15
Add tests for web dispatch


mmihic:
Pretty much the entirety of the web dispatch layer is untested. This needs tests.

mmihic:
Addressed in https://github.com/square/misk/pull/48

PR #964, opened 2019-05-03
Add ExecutorServiceModule

### LINKED ISSUE #284: ExecutorServiceModule
I want very regular access to executor services. Something like this:

```
  install(ExecutorServiceModule(Jetty::class, thread_count = 5))
```

```
   @Inject @Jetty lateinit var executorService: ExecutorService
```

monkey-mas:
Fixes #284 
- Allow us to add metrics in the future.
- Makes it easier to get ExecutorServices instead of using `@Providers`

### DIFF
--- a/misk/src/main/kotlin/misk/concurrent/ExecutorServiceModule.kt
+++ b/misk/src/main/kotlin/misk/concurrent/ExecutorServiceModule.kt
@@ -1,0 +1,26 @@
+package misk.concurrent
+
+import com.google.common.util.concurrent.ThreadFactoryBuilder
+import misk.inject.KAbstractModule
+import java.util.concurrent.ExecutorService
+import java.util.concurrent.Executors
+import kotlin.reflect.KClass
+
+class ExecutorServiceModule(
+  private val annotation: KClass<out Annotation>,
+  private val executorService: ExecutorService
+) : KAbstractModule() {
+  override fun configure() {
+    bind<ExecutorService>().annotatedWith(annotation.java).toInstance(executorService)
+  }
+
+  companion object {
+    fun withCachedThreadPool(
+      annotation: KClass<out Annotation>,
+      nameFormat: String
+    ): ExecutorServiceModule {
+      val threadFactory = ThreadFactoryBuilder().setNameFormat(nameFormat).build()
+      return ExecutorServiceModule(annotation, Executors.newCachedThreadPool(threadFactory))
+    }
+  }
+}
--- a/misk-aws/src/main/kotlin/misk/jobqueue/sqs/AwsSqsJobQueueModule.kt
+++ b/misk-aws/src/main/kotlin/misk/jobqueue/sqs/AwsSqsJobQueueModule.kt
@@ -4,18 +4,16 @@ import com.amazonaws.auth.AWSCredentialsProvider
 import com.amazonaws.services.sqs.AmazonSQS
 import com.amazonaws.services.sqs.AmazonSQSClientBuilder
 import com.google.common.util.concurrent.Service
-import com.google.common.util.concurrent.ThreadFactoryBuilder
 import com.google.inject.Provider
 import com.google.inject.Provides
 import com.google.inject.Singleton
 import misk.cloud.aws.AwsRegion
+import misk.concurrent.ExecutorServiceModule
 import misk.inject.KAbstractModule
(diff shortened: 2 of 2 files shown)


misk/src/main/kotlin/misk/concurrent/ExecutorServiceModule.kt
```diff
@@ -0,0 +1,34 @@
...
+    bind<ExecutorService>().annotatedWith(annotation.java).toInstance(executorService)
+  }
+
+  companion object {
+    fun of(annotation: KClass<out Annotation>, nameFormat: String): ExecutorServiceModule {
+      val threadFactory = ThreadFactoryBuilder().setNameFormat(nameFormat).build()
+      return ExecutorServiceModule(annotation, Executors.newCachedThreadPool(threadFactory))
+    }
+
+    fun of(annotation: KClass<out Annotation>, nameFormat: String, nThreads: Int)
```
jjestrel: We should follow the kotlin style guide
```
fun of(annotation: KClass<out Annotation>, nameFormat: String, nThreads: Int)
: ExecutorServiceModule {
```

should be
```

fun of(
    annotation: KClass<out Annotation>, 
    nameFormat: String, 
    nThreads: Int
): ExecutorServiceModule {
```

misk/src/main/kotlin/misk/concurrent/ExecutorServiceModule.kt
```diff
@@ -0,0 +1,34 @@
...
+class ExecutorServiceModule(
+  private val annotation: KClass<out Annotation>,
+  private val executorService: ExecutorService
+) : KAbstractModule() {
+  override fun configure() {
+    bind<ExecutorService>().annotatedWith(annotation.java).toInstance(executorService)
+  }
+
+  companion object {
+    fun of(annotation: KClass<out Annotation>, nameFormat: String): ExecutorServiceModule {
```
mightyguava: Per Effective Java #41, can we name these according to the `Executors` methods? This could be `withCachedThreadPool()` and the other one `withFixedThreadPool()`

misk/src/main/kotlin/misk/concurrent/ExecutorServiceModule.kt
```diff
@@ -0,0 +1,34 @@
+package misk.concurrent
+
+import com.google.common.util.concurrent.ThreadFactoryBuilder
+import misk.inject.KAbstractModule
+import java.util.concurrent.ExecutorService
+import java.util.concurrent.Executors
+import kotlin.reflect.KClass
+
+/**
```
mightyguava: remove empty comment, though would be nice to have docs for the methods

misk/src/main/kotlin/misk/concurrent/ExecutorServiceModule.kt
monkey-mas: done :)

misk/src/main/kotlin/misk/concurrent/ExecutorServiceModule.kt
monkey-mas: done :)

misk/src/main/kotlin/misk/concurrent/ExecutorServiceModule.kt
monkey-mas: I removed this this time. :)

misk/src/main/kotlin/misk/concurrent/ExecutorServiceModule.kt
```diff
@@ -0,0 +1,26 @@
...
+class ExecutorServiceModule(
+  private val annotation: KClass<out Annotation>,
+  private val executorService: ExecutorService
+) : KAbstractModule() {
+  override fun configure() {
+    bind<ExecutorService>().annotatedWith(annotation.java).toInstance(executorService)
+  }
+
+  companion object {
+    fun withCachedThreadPool(
```
swankjesse: this is awesome for everything EXCEPT cached thread pools are bad and you should never use them

monkey-mas: Yeah, got it! Will make a PR to use `Executors.newFixedThreadPool(...)` instead. :)

PR #966, opened 2019-05-03
Use FixedThreadPool instead of CashedThreadPool in ExecutorServiceModule

monkey-mas:
- Since it's not really good to use CashedThreadPool, we use FixedThreadPool.[1]
- Also, we add a new line at the EOF.

[1] feedback: https://github.com/square/misk/pull/964#discussion_r280903015

issue #284

### DIFF
--- a/misk-aws/src/main/kotlin/misk/jobqueue/sqs/AwsSqsJobQueueModule.kt
+++ b/misk-aws/src/main/kotlin/misk/jobqueue/sqs/AwsSqsJobQueueModule.kt
@@ -31,7 +31,7 @@ class AwsSqsJobQueueModule(
     bind<TransactionalJobQueue>().to<SqsTransactionalJobQueue>()
     multibind<Service>().to<SqsJobConsumer>()
 
-    install(ExecutorServiceModule.withCachedThreadPool(ForSqsConsumer::class, "sqs-consumer-%d"))
+    install(ExecutorServiceModule.withFixedThreadPool(ForSqsConsumer::class, "sqs-consumer-%d", 4))
 
     // Bind a map of AmazonSQS clients for each external region that we need to contact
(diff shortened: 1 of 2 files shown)


monkey-mas:
Hmm, `failsIfServerIsUntrusted() - misk.web.ssl.PemSslClientServerTest` failed on CI, but the test passes on my local env though.

jjestrel:
> Hmm, `failsIfServerIsUntrusted() - misk.web.ssl.PemSslClientServerTest` failed on CI, but the test passes on my local env though.

it's flaky I think. I re-ran

misk-aws/src/main/kotlin/misk/jobqueue/sqs/AwsSqsJobQueueModule.kt
```diff
@@ -31,7 +31,7 @@ class AwsSqsJobQueueModule(
     bind<TransactionalJobQueue>().to<SqsTransactionalJobQueue>()
     multibind<Service>().to<SqsJobConsumer>()
 
-    install(ExecutorServiceModule.withCachedThreadPool(ForSqsConsumer::class, "sqs-consumer-%d"))
+    install(ExecutorServiceModule.withFixedThreadPool(ForSqsConsumer::class, "sqs-consumer-%d", 4))
```
mightyguava: This seems like it should be configurable

wesleyk: Maybe as part of `AwsSqsJobQueueConfig`?

jjestrel: If it was unbounded before I think making it configurable is a separate PR.

monkey-mas:
Thanks for your review! :) will make another PR to make it configurable!

PR #968, opened 2019-05-05
Allow thread size to be configurable for AwsSqsJobQueue

monkey-mas:
It'd be nice to have a config for the number of threads instead of hardcoding it.

Feedback: https://github.com/square/misk/pull/966#discussion_r280936506
Issue ref: https://github.com/square/misk/issues/284

### DIFF
--- a/misk-aws/src/main/kotlin/misk/jobqueue/sqs/AwsSqsJobQueueConfig.kt
+++ b/misk-aws/src/main/kotlin/misk/jobqueue/sqs/AwsSqsJobQueueConfig.kt
@@ -17,6 +17,11 @@ class AwsSqsJobQueueConfig(
    * External queues is a set of externally owned SQS queues accessed by this service, mapping
    * an internal queue name to the (account ID, region, name) of the queue in the external account
    */
-  val external_queues: Map<String, AwsSqsQueueConfig> = mapOf()
+  val external_queues: Map<String, AwsSqsQueueConfig> = mapOf(),
+
+  /**
+   * Number of jobs that can be processed concurrently.
+   */
+  val consumer_thread_pool_size: Int = 4
 ) : Config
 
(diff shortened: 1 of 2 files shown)


misk-aws/src/main/kotlin/misk/jobqueue/sqs/AwsSqsJobQueueConfig.kt
```diff
@@ -17,6 +17,12 @@ class AwsSqsJobQueueConfig(
...
    * an internal queue name to the (account ID, region, name) of the queue in the external account
    */
-  val external_queues: Map<String, AwsSqsQueueConfig> = mapOf()
+  val external_queues: Map<String, AwsSqsQueueConfig> = mapOf(),
+
+  /**
+   * Number of consumer threads to run. This parameter controls the number of jobs handled by
+   * consumer concurrently.
+   */
+  val consumer_thread_pool_size: Int = 4
```
monkey-mas: Any suggestions for the comment and var name? :)

wesleyk: the naming seems reasonable!

The second part of the comment doesn't seem quite accurate: "This parameter controls the number of jobs handled by consumer concurrently." Maybe like, "Number of threads to run for the job queue consumer"?

ryanhall07: number of jobs that can be processed concurrently

monkey-mas: Thanks for the feedback! I'll pick @ryanhall07 's idea `Number of jobs that can be processed concurrently.` if it's okay! :)

monkey-mas:
Thanks for the review! :)

PR #969, opened 2019-05-05
Use ExecutorServiceModule for ZookeeperModule

monkey-mas:
issue ref: https://github.com/square/misk/issues/284

### DIFF
--- a/misk-zookeeper/src/main/kotlin/misk/clustering/zookeeper/ZookeeperModule.kt
+++ b/misk-zookeeper/src/main/kotlin/misk/clustering/zookeeper/ZookeeperModule.kt
@@ -1,10 +1,10 @@
 package misk.clustering.zookeeper
 
 import com.google.common.util.concurrent.Service
-import com.google.common.util.concurrent.ThreadFactoryBuilder
 import com.google.inject.Key
 import com.google.inject.Provides
 import misk.clustering.lease.LeaseManager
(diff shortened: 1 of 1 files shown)


misk-zookeeper/src/main/kotlin/misk/clustering/zookeeper/ZookeeperModule.kt
```diff
@@ -23,6 +23,7 @@ class ZookeeperModule(private val config: ZookeeperConfig) : KAbstractModule() {
     multibind<Service>().to(RepeatedTaskQueue::class.toKey(ForZkLease::class)).asSingleton()
     bind<LeaseManager>().to<ZkLeaseManager>()
     bind<CuratorFramework>().toProvider(CuratorFrameworkProvider::class.java).asSingleton()
+    install(ExecutorServiceModule.withFixedThreadPool(ForZkLease::class, "zk-lease-poller", 1))
```
monkey-mas: Can we make the `nThreads` configurable in `ZookeeperConfig`? 🤔

wesleyk: I think clients won't really care about configuring this pool, since the client's usage shouldn't affect how many threads it needs for the zk lease poller task.

PR #971, opened 2019-05-06
Include id when shard not found

ryanhall07:


ryanhall07:
> this is just for debugging?

yes

PR #973, opened 2019-05-07
Add clients for calling other members in an app's cluster

alecholmes:
An app running in a cluster may want to call specific peers in the cluster group. This builds on top of the `Cluster` abstraction to create both untyped and retrofit-typed clients, using mTLS to validate that the peer's OU is the same as the calling app's OU.

Example using this to have a specific cluster member handle all `foo`-related requests:
```
@Inject lateinit var cluster: Cluster
@Inject lateinit var clientFactory: TypedPeerClientFactory<ReturnADinosaur>

@Test fun demo() {
  val client: ReturnADinosaur = clientFactory.client(cluster.snapshot.resourceMapper["foo"])
}
```

### DIFF
--- a/misk/src/main/kotlin/misk/client/PeerClientFactory.kt
+++ b/misk/src/main/kotlin/misk/client/PeerClientFactory.kt
@@ -1,0 +1,113 @@
+package misk.client
+
+import com.google.common.cache.CacheBuilder
+import com.google.common.cache.CacheLoader
+import com.google.inject.Provides
+import misk.clustering.Cluster
+import misk.config.AppName
+import misk.inject.KAbstractModule
+import misk.security.cert.X500Name
+import misk.web.WebConfig
+import misk.web.jetty.JettyService
+import okhttp3.OkHttpClient
+import java.security.cert.X509Certificate
+import java.util.concurrent.TimeUnit
+import javax.inject.Singleton
+
+/**
+ * Binds a [PeerClientFactory] that calls peers on the HTTPS port of this process's server,
+ * as determined by the SSL port in the [WebConfig].
+ */
+class PeerClientModule : KAbstractModule() {
+  @Provides @Singleton fun peerClientFactory(
+    @AppName appName: String,
+    httpClientsConfig: HttpClientsConfig,
+    httpClientFactory: HttpClientFactory,
+    webConfig: WebConfig
+  ): PeerClientFactory {
+    check(webConfig.ssl?.port ?: 0 > 0) { "server must have static HTTPS port" }
+
+    return PeerClientFactory(
+        appName = appName,
+        httpClientsConfig = httpClientsConfig,
+        httpClientFactory = httpClientFactory,
+        httpsPort = webConfig.ssl!!.port
+    )
+  }
+}
+
+/**
+ * For testing.
+ *
+ * Binds a [PeerClientFactory] that calls peers on the HTTPS port of this process's server,
+ * as determined by the Jetty server's port.
+ */
+class JettyPortPeerClientModule : KAbstractModule() {
+  @Provides @Singleton fun peerClientFactory(
+    @AppName appName: String,
+    httpClientsConfig: HttpClientsConfig,
+    httpClientFactory: HttpClientFactory,
+    jetty: JettyService
+  ): PeerClientFactory {
+    return PeerClientFactory(
+        appName = appName,
+        httpClientsConfig = httpClientsConfig,
+        httpClientFactory = httpClientFactory,
+        httpsPort = jetty.httpsServerUrl!!.port()
+    )
+  }
+}
+
+/**
+ * Factory that creates [OkHttpClient]s for connecting to another instance of the same application
+ * running in the same cluster.
+ *
+ * An [OkHttpClient] is cached for each peer.
+ */
+class PeerClientFactory(
+  private val appName: String,
+  private val httpClientsConfig: HttpClientsConfig,
+  private val httpClientFactory: HttpClientFactory,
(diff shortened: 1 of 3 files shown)


misk/src/main/kotlin/misk/client/TypedHttpClientModule.kt
```diff
@@ -101,3 +103,96 @@ class TypedHttpClientModule<T : Any>(
...
+    bind(key).toProvider(PeerTypedClientProvider(kclass, name))
+  }
+
+  companion object {
+    inline fun <reified T : Any> create(name: String): TypedPeerHttpClientModule<T> {
+      return TypedPeerHttpClientModule(T::class, name)
+    }
+  }
+
+  private class PeerTypedClientProvider<T : Any>(
```
alecholmes: Adapted (err, cargo culted) from `TypedHttpClientModule`

misk/src/main/kotlin/misk/client/PeerClientFactory.kt
```diff
@@ -0,0 +1,56 @@
...
+import javax.inject.Singleton
+
+/**
+ * Factory that creates [OkHttpClient]s for connecting to another instance of the same application
+ * running in the same cluster.
+ */
+@Singleton
+class PeerClientFactory @Inject constructor(
+  @AppName private val appName: String,
+  private val jetty: JettyService,
```
alecholmes: Using Jetty feels a bit weird and heavy. Injecting the web server config doesn't work in tests, though, since ports are dynamically assigned.

ryanhall07: lol need dat PortPicker

misk/src/main/kotlin/misk/client/PeerClientFactory.kt
```diff
@@ -0,0 +1,56 @@
...
+  private val httpClientsConfig: HttpClientsConfig,
+  private val httpClientFactory: HttpClientFactory
+) {
+
+  init {
+    require(jetty.httpsServerUrl != null) { "SSL must be configured to use peer clients" }
+
+    // There must be web client config. The URL and Envoy config are ultimately ignored.
+    httpClientsConfig[appName] // This throws if config is missing
```
ryanhall07: does it throw with a useful message?

misk/src/main/kotlin/misk/client/TypedHttpClientModule.kt
ryanhall07: yea this feels copied. can we do better and share? these configs/bindings are especially tricky to get right. e.g that stupid provider cycle detector note below

misk/src/main/kotlin/misk/client/PeerClientFactory.kt
mightyguava: wouldn't be terrible to statically assign ports in tests. We already expect in tests 2181 for ZK and 27001-4 for Vitess, etc. Though maybe the argument should be that those services should be using dynamically assigned ports in tests...

mightyguava: also does something need to depend on jetty service's key in the producer/consumer graph for the port to be assigned, or does that happen earlier?

misk/src/main/kotlin/misk/client/PeerClientFactory.kt
```diff
@@ -0,0 +1,56 @@
...
+  fun baseUrl(peer: Cluster.Member): String {
+    return "https://${peer.ipAddress}:${jetty.httpsServerUrl!!.port()}"
+  }
+
+  /**
+   * Get a client to call the given peer cluster member.
+   * This client will fail when calling different services, as determined by the OU in the certificate
+   * returned by the called service.
+   */
+  fun client(peer: Cluster.Member): OkHttpClient {
```
mightyguava: what level does connection pooling happen at? Should the clients per peer be singletons too?

alecholmes: Good question. `OkHttpClient` pools connections so it would make sense to for `PeerClientFactory` to cache them.

@swankjesse, I noticed the existing `HttpClientModule` doesn't scope clients as singletons. Any particular reason why to keep in mind?

misk/src/main/kotlin/misk/client/PeerClientFactory.kt
alecholmes: Yes

alecholmes:
This is ready for review. Overview of changes:
* `OkHttpClient`s are now cached per peer, to take advantage of connection pooling.
* Jetty dependency is now only used in tests.
* More code reuse in `TypedHttpClientModule.kt`

misk/src/main/kotlin/misk/client/PeerClientFactory.kt
```diff
@@ -0,0 +1,113 @@
...
+ *
+ * Binds a [PeerClientFactory] that calls peers on the HTTPS port of this process's server,
+ * as determined by the Jetty server's port.
+ */
+class JettyPortPeerClientModule : KAbstractModule() {
+  @Provides @Singleton fun peerClientFactory(
+    @AppName appName: String,
+    httpClientsConfig: HttpClientsConfig,
+    httpClientFactory: HttpClientFactory,
+    jetty: JettyService
```
mmihic: I wonder if we want to have a `Provider<ServerPorts>` that returns a `data class ServerPorts(val http: Int, val https: Int)` object with the actual HTTP/HTTPS ports, derived from the config in normal mode and the `JettyService` for dynamic port assignmint in tests

PR #974, opened 2019-05-08
Refactor key providers

yan:
This change unifies test and non-test versions of the various key providers to make them usable in both environments. After some discussion, we settled on using actual keys for testing and development, removing the need for diverging code paths and hard-coded keys.

This might supersede #972

### DIFF
--- a/misk-crypto/src/main/kotlin/misk/crypto/CryptoTestModule.kt
+++ b/misk-crypto/src/main/kotlin/misk/crypto/CryptoTestModule.kt
@@ -48,9 +48,6 @@ class CryptoTestModule(
               .asEagerSingleton()
         }
         KeyType.DIGITAL_SIGNATURE -> {
-          // val keysetHandle = KeysetHandle.generateNew(SignatureKeyTemplates.ED25519)
-          // val signer = PublicKeySignFactory.getPrimitive(keysetHandle)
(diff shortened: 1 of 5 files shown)


misk-crypto/src/main/kotlin/misk/crypto/CryptoTestModule.kt
```diff
@@ -35,68 +26,42 @@ class CryptoTestModule(
...
+              .toProvider(MacProvider(key, null))
               .asEagerSingleton()
         }
         KeyType.DIGITAL_SIGNATURE -> {
-          val keysetHandle = KeysetHandle.generateNew(SignatureKeyTemplates.ED25519)
-          val signer = PublicKeySignFactory.getPrimitive(keysetHandle)
-          val verifier = PublicKeyVerifyFactory.getPrimitive(keysetHandle.publicKeysetHandle)
+          // val keysetHandle = KeysetHandle.generateNew(SignatureKeyTemplates.ED25519)
+          // val signer = PublicKeySignFactory.getPrimitive(keysetHandle)
+          // val verifier = PublicKeyVerifyFactory.getPrimitive(keysetHandle.publicKeysetHandle)
```
mightyguava: remove?

PR #975, opened 2019-05-08
Connect to Zookeeper using EnsembleProvider instead of connection string

nbroyles:
Not sure if this is the _best_ way to wire this up, but it works. Skim will have a `SkimZkLeaseModule` that calls `ZkLeaseModule` and also binds up the `EnsembleProvider`

This change is part of the fix to resolve these errors:
```
Invalid config event received: {server.2=localhost:2881:3881:participant, server.1=localhost:2880:3880:participant, server.5=localhost:2884:3884:participant, server.4=localhost:2883:3883:participant, server.3=localhost:2882:3882:participant, version=0}
```
A separate change will allow these events to be received and, once that's done, this change will properly disregard them so that our connection string doesn't get updated with localhost values. See ZookeeperEnsembleProvider for more details

### DIFF
--- a/misk-zookeeper/src/main/kotlin/misk/clustering/zookeeper/ZkLeaseCommonModule.kt
+++ b/misk-zookeeper/src/main/kotlin/misk/clustering/zookeeper/ZkLeaseCommonModule.kt
@@ -1,0 +1,48 @@
+package misk.clustering.zookeeper
+
+import com.google.common.util.concurrent.Service
+import com.google.inject.Key
+import com.google.inject.Provides
+import misk.clustering.lease.LeaseManager
+import misk.concurrent.ExecutorServiceModule
+import misk.inject.KAbstractModule
+import misk.inject.asSingleton
+import misk.inject.toKey
+import misk.tasks.RepeatedTaskQueue
+import misk.zookeeper.CuratorFrameworkModule
+import java.time.Clock
+import java.util.concurrent.ExecutorService
+import javax.inject.Singleton
+
+
+class ZkLeaseCommonModule(private val config: ZookeeperConfig) : KAbstractModule() {
+  override fun configure() {
+    install(ZkLeaseManagerModule())
+    install(ExecutorServiceModule.withFixedThreadPool(ForZkLease::class, "zk-lease-poller", 1))
+    install(CuratorFrameworkModule(config, ForZkLease::class))
+  }
+
+  companion object {
+    /** @property Key<*> the Key of the lease manager service */
+    val leaseManagerKey: Key<*> = Key.get(ZkLeaseManager::class.java)
+  }
+
+  @Provides @ForZkLease @Singleton
+  fun provideTaskQueue(
+    clock: Clock,
+    @ForZkLease executorService: ExecutorService
+  ): RepeatedTaskQueue {
+    return RepeatedTaskQueue("zk-lease-poller", clock, executorService)
+  }
+}
+
+/**
+ * Common bindings between [ZkLeaseCommonModule] and [ZkLeaseTestModule].
+ */
+internal class ZkLeaseManagerModule : KAbstractModule() {
+  override fun configure() {
+    multibind<Service>().to<ZkLeaseManager>()
+    multibind<Service>().to(RepeatedTaskQueue::class.toKey(ForZkLease::class)).asSingleton()
+    bind<LeaseManager>().to<ZkLeaseManager>()
+  }
+}
--- a/misk-zookeeper/src/main/kotlin/misk/clustering/zookeeper/ZkLeaseModule.kt
+++ b/misk-zookeeper/src/main/kotlin/misk/clustering/zookeeper/ZkLeaseModule.kt
@@ -1,50 +1,14 @@
 package misk.clustering.zookeeper
 
-import com.google.common.util.concurrent.Service
-import com.google.inject.Key
-import com.google.inject.Provides
-import misk.clustering.lease.LeaseManager
-import misk.concurrent.ExecutorServiceModule
 import misk.inject.KAbstractModule
-import misk.inject.asSingleton
-import misk.inject.toKey
-import misk.tasks.RepeatedTaskQueue
-import misk.zookeeper.ZookeeperModule
-import java.time.Clock
-import java.util.concurrent.ExecutorService
-import javax.inject.Singleton
+import misk.zookeeper.FixedEnsembleProviderModule
 
 /**
  * Binds a [LeaseManager] that uses Zookeeper.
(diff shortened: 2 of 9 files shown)


misk-zookeeper/src/main/kotlin/misk/zookeeper/ZookeeperEnsembleModule.kt
```diff
@@ -0,0 +1,22 @@
...
+
+import misk.clustering.zookeeper.ZookeeperConfig
+import misk.inject.KAbstractModule
+import misk.inject.keyOf
+import org.apache.curator.ensemble.EnsembleProvider
+import org.apache.curator.ensemble.fixed.FixedEnsembleProvider
+import javax.inject.Provider
+import kotlin.reflect.KClass
+
+class ZookeeperEnsembleModule(
```
ryanhall07: this should be called FixedEnsembleModule or something

misk-zookeeper/src/test/kotlin/misk/zookeeper/ZkTestModule.kt
```diff
@@ -22,12 +22,13 @@ class ZkTestModule(
...
-
-    install(ZookeeperModule(ZookeeperConfig(
+    val config = ZookeeperConfig(
         zk_connect = "127.0.0.1:$zkPortKey",
         cert_store = CertStoreConfig(keystorePath, "changeit", SslLoader.FORMAT_JKS),
-        trust_store = TrustStoreConfig(truststorePath, "changeit", SslLoader.FORMAT_JKS)),
-        qualifier))
+        trust_store = TrustStoreConfig(truststorePath, "changeit", SslLoader.FORMAT_JKS))
+
+    install(ZookeeperCommonModule(config, qualifier))
```
ryanhall07: it seems like these 2 modules should be packaged up into `ZookeeperModule` for you.

misk-zookeeper/src/main/kotlin/misk/zookeeper/ZookeeperEnsembleProvider.kt
```diff
@@ -0,0 +1,48 @@
+package misk.zookeeper
+
+import org.apache.curator.ensemble.EnsembleProvider
+import org.slf4j.LoggerFactory
+
+/**
+ * ZookeeperEnsembleProvider is modeled after [FixedEnsembleProvider] but does not allow the
+ * connection string to be changed (a la [ExhibitorEnsembleProvider]). By default, Curator attempts
+ * to handle load balancing connections across ZooKeeper servers. However, the connection string
```
wesleyk: thinking out loud, is this a misk layer concept or a skim one?

nbroyles: Ah, yeah, this comment here is a bit Square specific, but the Provider itself could be useful to anyone who wants to handle load balancing themselves. I'll clean that up

misk-zookeeper/src/main/kotlin/misk/zookeeper/ZookeeperEnsembleModule.kt
nbroyles: Yeah, I'll rename this to something better

misk-zookeeper/src/main/kotlin/misk/zookeeper/ZookeeperEnsembleProvider.kt
nbroyles: Eh, on second thought. Just going to move it back to skim

misk-zookeeper/src/main/kotlin/misk/clustering/zookeeper/ZkLeaseCommonModule.kt
```diff
@@ -0,0 +1,48 @@
...
+import misk.inject.asSingleton
+import misk.inject.toKey
+import misk.tasks.RepeatedTaskQueue
+import misk.zookeeper.ZookeeperCommonModule
+import java.time.Clock
+import java.util.concurrent.ExecutorService
+import javax.inject.Singleton
+
+
+class ZkLeaseCommonModule(private val config: ZookeeperConfig) : KAbstractModule() {
```
nbroyles: This was `ZkLeaseModule` but renamed.

misk-zookeeper/src/main/kotlin/misk/clustering/zookeeper/ZkLeaseCommonModule.kt
```diff
@@ -0,0 +1,48 @@
...
+    @ForZkLease executorService: ExecutorService
+  ): RepeatedTaskQueue {
+    return RepeatedTaskQueue("zk-lease-poller", clock, executorService)
+  }
+}
+
+/**
+ * Common bindings between [ZkLeaseCommonModule] and [ZkLeaseTestModule].
+ */
+internal class ZkLeaseManagerModule : KAbstractModule() {
```
nbroyles: Not in love with this name, but `ZkLeaseCommonModule` was a much better name for the outer module given that it will be used in Skim. This was the best I could think of :-/

misk-zookeeper/src/main/kotlin/misk/clustering/zookeeper/ZkLeaseModule.kt
```diff
@@ -1,50 +1,15 @@
...
-import java.time.Clock
-import java.util.concurrent.ExecutorService
-import javax.inject.Singleton
+import misk.zookeeper.FixedEnsembleProviderModule
+import misk.zookeeper.ZookeeperCommonModule
 
 /**
  * Binds a [LeaseManager] that uses Zookeeper.
  */
 class ZkLeaseModule(private val config: ZookeeperConfig) : KAbstractModule() {
```
nbroyles: And, we're now using `ZkLeaseModule` to bind the common lease code plus the ensemble provider. Skim will do the same but used a different ensemble provider

misk-zookeeper/src/main/kotlin/misk/zookeeper/ZookeeperCommonModule.kt
```diff
@@ -0,0 +1,45 @@
...
+ * ```
+ * @Qualifier
+ * annotation class MyZk
+ *
+ * install(ZookeeperCommonModule(config, MyZk::class)
+ *
+ * @Inject @MyZk CuratorFramework
+ * ```
+ */
+class ZookeeperCommonModule(
```
ryanhall07: nit: could probably just call this the `CuratorModule`, since that's the intention

PR #979, opened 2019-05-10
Support DAEAD encryption

yan:
This PR updates `SecretColumn` to:
 1. Encrypt its values with Deterministic AEAD by default. This sacrifices semantic security for improved usability. Using DAEAD implies that multiple identical plaintexts will encrypt to identical ciphertexts. This makes encrypted rows queryable, but leaks some information about ciphertexts. 
 2. Accept a parameter to the attribute (`indexable`) that a user can set to `false` to enable use of non-deterministic AEAD. This attribute is defaulted to `true` for usability.

### DIFF
--- a/misk-crypto/src/main/kotlin/misk/crypto/CryptoModule.kt
+++ b/misk-crypto/src/main/kotlin/misk/crypto/CryptoModule.kt
@@ -1,6 +1,7 @@
 package misk.crypto
 
 import com.google.crypto.tink.Aead
+import com.google.crypto.tink.DeterministicAead
 import com.google.crypto.tink.KmsClient
 import com.google.crypto.tink.aead.AeadConfig
 import com.google.crypto.tink.mac.MacConfig
@@ -42,6 +43,12 @@ class CryptoModule(
           bind<Aead>()
               .annotatedWith(Names.named(key.key_name))
               .toProvider(AeadEnvelopeProvider(key, config.kms_uri))
+              .`in`(Singleton::class.java)
+        }
+        KeyType.DAEAD -> {
+          bind<DeterministicAead>()
+              .annotatedWith(Names.named(key.key_name))
+              .toProvider(DeterministicAeadProvider(key, config.kms_uri))
               .`in`(Singleton::class.java)
         }
         KeyType.MAC -> {
--- a/misk-crypto/src/main/kotlin/misk/crypto/CryptoModule.kt
+++ b/misk-crypto/src/main/kotlin/misk/crypto/CryptoModule.kt
@@ -10,6 +10,7 @@ import misk.inject.KAbstractModule
 import com.google.crypto.tink.Mac
 import com.google.crypto.tink.PublicKeySign
 import com.google.crypto.tink.PublicKeyVerify
+import com.google.crypto.tink.daead.DeterministicAeadConfig
 import com.google.inject.Singleton
 import com.google.inject.name.Names
 import okio.ByteString
@@ -28,6 +29,7 @@ class CryptoModule(
     config.keys ?: return
     requireBinding(KmsClient::class.java)
     AeadConfig.register()
+    DeterministicAeadConfig.register()
     MacConfig.register()
     SignatureConfig.register()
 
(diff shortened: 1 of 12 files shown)


misk-crypto/src/main/kotlin/misk/crypto/CryptoModule.kt
```diff
@@ -44,6 +47,12 @@ class CryptoModule(
               .toProvider(AeadEnvelopeProvider(key, config.kms_uri))
               .`in`(Singleton::class.java)
         }
+        KeyType.DAEAD -> {
```
yoavamit: I'm really happy with now easy it seems to add another primitive to `misk-crypto`

misk-hibernate/src/main/kotlin/misk/hibernate/SecretColumn.kt
```diff
@@ -1,9 +1,15 @@
 package misk.hibernate
 
 /**
- * [SecretColumn] is an annotation used to get Hibernate to encrypt a field before writing it
- * to the database.
+ * [SecretColumn] is an annotation used to get Hibernate to encrypt a field before writing it  to the database.
```
yoavamit: super nit: extra space at the end of the sentence

misk-hibernate/src/main/kotlin/misk/hibernate/SecretColumn.kt
```diff
@@ -25,13 +31,15 @@ package misk.hibernate
...
  * ```
- * - It cannot be a part of an index
  * - It cannot be annotates with any other custom column annotations like [ProtoColumn] or [JsonColumn].
  *
- * *Note*: the resulting ciphertext that is persisted in the database may be much larger in size
- * than the original plaintext because it also contains some metadata.
- * Please make sure to allocate enough space when defining the column using `VARBINARY()`.
+ * *Note*:
+ *
+ *  1. the resulting ciphertext that is persisted in the database may be much larger in size than
```
yoavamit: nit: remove the `.1`

misk-hibernate/src/main/kotlin/misk/hibernate/SecretColumnType.kt
```diff
@@ -34,14 +36,15 @@ internal class SecretColumnType : UserType, ParameterizedType, TypeConfiguration
...
-    } catch (ex: KeyNotFoundException) {
-      throw HibernateException("Cannot set field, key $keyName not found")
+    val indexable = parameters.getProperty(FIELD_ENCRYPTION_INDEXABLE)!!.toBoolean()
+
+    encryptionAdapter = if (indexable) {
+      DeterministicAeadAdapter(_typeConfiguration, keyName)
+    } else {
+      AeadAdapter(_typeConfiguration, keyName)
     }
+
```
yoavamit: nit: stray empty line

misk-hibernate/src/main/kotlin/misk/hibernate/SecretColumnType.kt
```diff
@@ -90,4 +93,59 @@ internal class SecretColumnType : UserType, ParameterizedType, TypeConfiguration
...
+    val keyManager = typeConfig.metadataBuildingContext.bootstrapContext.serviceRegistry.injector
+            .getInstance(DeterministicAeadKeyManager::class.java)
+    try {
+      daead = keyManager[keyName]
+    } catch (ex: KeyNotFoundException) {
+      throw HibernateException("Cannot set field, key $keyName not found")
+    }
+  }
+
+  // DeterministicAEAD throws if associatedData is null, so we pass an empty array if it is.
```
yoavamit: This comment should be inside the function body

misk-hibernate/src/test/kotlin/misk/hibernate/SecretColumnTest.kt
```diff
@@ -178,6 +197,32 @@ class SecretColumnTest {
     }
   }
 
+  @Test
+  fun testEncryptedIndxedQueryMultiple() {
+    val reviewer = "Some Reviewer".toByteArray()
+    transacter.transaction { session ->
+        session.save(DbJerryGarciaSong("title 1", 123, "album".toByteArray(), reviewer))
```
yoavamit: Maybe add real songs? 😁

misk-hibernate/src/main/kotlin/misk/hibernate/SecretColumnType.kt
yan: It was initially but it covers both encrypt and decrypt. I can move it in though.

PR #981, opened 2019-05-14
A few minor improvements related to misk-crypto

yoavamit:
Added extension functions in `misk-crypto`.

Modified the `disassemble` and `assemble` functions in `SecretColumnType` so data is encrypted when cached by Hibernate.

### DIFF
--- a/misk-crypto/src/main/kotlin/misk/crypto/CryptoModule.kt
+++ b/misk-crypto/src/main/kotlin/misk/crypto/CryptoModule.kt
@@ -15,6 +15,7 @@ import com.google.inject.Singleton
 import com.google.inject.name.Names
 import okio.ByteString
 import okio.ByteString.Companion.toByteString
+import java.util.Base64
 
 /**
  * Configures and registers the keys listed in the configuration file.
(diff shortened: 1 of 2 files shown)


misk-crypto/src/main/kotlin/misk/crypto/CryptoModule.kt
```diff
@@ -94,4 +95,44 @@ fun Aead.decrypt(ciphertext: ByteString, aad: ByteArray? = null): ByteString {
   val decrypted = decryptedBytes.toByteString()
   decryptedBytes.fill(0)
   return decrypted
+}
+
+/**
+ * Extension function for convenient encryption of [ByteString]s.
+ * This function also makes sure that no extra copies of the plaintext data are kept in memory.
+ */
+fun DeterministicAead.encryptDeterministically(plaintext: ByteString, aad: ByteArray? = null
```
wesleyk: is `Deterministically` redundant given it's a `DeterministicAead`?

misk-hibernate/src/main/kotlin/misk/hibernate/SecretColumnType.kt
```diff
@@ -57,9 +56,17 @@ internal class SecretColumnType : UserType, ParameterizedType, TypeConfiguration
...
 
-  override fun assemble(cached: Serializable?, owner: Any?) = cached
+  override fun assemble(cached: Serializable?, owner: Any?): ByteArray {
+    return encryptionAdapter.decrypt(cached as ByteArray, null)
+  }
 
-  override fun disassemble(value: Any?) = (value as ByteArray).copyOf() as Serializable
+  /**
+   * This method is used by Hibernate when caching values, see [org.hibernate.type.Type.disassemble].
+   * This implementation makes sure that data is stored encrypted even when being cached in memory.
```
wesleyk: woah, nice!

misk-crypto/src/main/kotlin/misk/crypto/CryptoModule.kt
yoavamit: I wanted this to match the method name provided by the `DeterministicAead` interface for compatibility.
This way, it's more like method overloading rather then an extension

yan: This is an extension for Tink's `DeterministicAead` which uses the methods `encryptDeterministically`/`decryptDeterministically`, and this just wraps them. I believe Tink chose the naming scheme to ensure AEAD and DeterministicAead were not swappable.

yan: (woops, submitted at the same time)

PR #984, opened 2019-05-15
ServiceGraphBuilder

keeferrourke:
This PR adds two classes:
 - _`CoordinatedService2`_ which is intended to replace the existing `CoordinatedService`
 - _`ServiceGraphBuilder`_ which is a builder class for Guava `ServiceManager`s that manage `CoordinatedService2`s

CoordinatedService2 adds the concept of "enhancements" in addition to the traditional downstream dependencies. Hopefully the documentation is clear as this was tricky to reason about at first!

### DIFF
--- a/misk/src/main/kotlin/misk/CoordinatedService2.kt
+++ b/misk/src/main/kotlin/misk/CoordinatedService2.kt
@@ -4,54 +4,26 @@ import com.google.common.util.concurrent.AbstractService
 import com.google.common.util.concurrent.MoreExecutors
 import com.google.common.util.concurrent.Service
 import com.google.common.util.concurrent.Service.*
-import misk.CoordinatedService2.Companion.CycleValidity
 
+/**
+ * CoordinatedService2 wraps a Service to defer its start up and shut down until its dependent
+ * services are ready.
+ *
+ * This Service will stall in the `STARTING` state until all upstream services are `RUNNING`.
+ * Symmetrically it stalls in the `STOPPING` state until all dependent services are `TERMINATED`.
+ *
+ * This Service may be "enhanced" by other services. That is to say that if there exists a
+ * complementary service that is dependent on this service, it will be started after this service
+ * starts, and stopped after this service stops. Information about this service's enhancements is
+ * hidden from its dependent services.
+ *
+ * @param service The Service to wrap.
+ */
 internal class CoordinatedService2(val service: Service) : AbstractService() {
-  val upstream = mutableSetOf<CoordinatedService2>()      // upstream services dependent on me
-  val downstream = mutableSetOf<CoordinatedService2>()    // downstream dependencies
-  val enhancements = mutableSetOf<CoordinatedService2>()  // services that enhance me (depend on me)
-  var target: CoordinatedService2? = null                 // service I enhance
-
-  fun getServicesThatINeed(): Set<CoordinatedService2> {
-    // target
-    // upstream providers
-    val neededServices = mutableSetOf<CoordinatedService2>()
-    if (target != null) {
-      neededServices.add(target!!)
-    }
-    for (provider in upstream) {
-      neededServices.add(provider)
-      neededServices.addAll(provider.getTransitiveEnhancements())
-    }
-    return neededServices
-  }
-
-  // obtains all dependencies and enhancements (and their dependencies and enhancements)
-  // for this service
-  fun getServicesThatNeedMe(): Set<CoordinatedService2> {
-    // my enhancements
-    // my target's dependencies
-    // my target's target's dependencies
-    // etc.
-
-    val needsMe = mutableSetOf<CoordinatedService2>()
-    needsMe.addAll(enhancements)
-    var t: CoordinatedService2? = this
-    while (t != null) {
-      needsMe.addAll(t.downstream)
-      t = t.target
-    }
-    return needsMe
-  }
-
-  private fun getTransitiveEnhancements() : Set<CoordinatedService2> {
-    val list = mutableSetOf<CoordinatedService2>()
-    list.addAll(enhancements)
-    for (enhancement in enhancements) {
-      list.addAll(enhancement.getTransitiveEnhancements())
-    }
-    return list
-  }
+  private val upstream = mutableSetOf<CoordinatedService2>()      // upstream services dependent on me
+  private val downstream = mutableSetOf<CoordinatedService2>()    // downstream dependencies
+  private val enhancements = mutableSetOf<CoordinatedService2>()  // services that enhance me (depend on me)
+  private var target: CoordinatedService2? = null                 // service I enhance
 
   init {
     service.addListener(object : Listener() {
@@ -59,14 +31,14 @@ internal class CoordinatedService2(val service: Service) : AbstractService() {
         synchronized(this) {
           notifyStarted()
         }
-        getServicesThatNeedMe().forEach { it.startIfReady() }
+        getReliantServices().forEach { it.startIfReady() }
       }
 
       override fun terminated(from: State?) {
         synchronized(this) {
           notifyStopped()
         }
-        getServicesThatINeed().forEach { it.stopIfReady() }
+        getRequiredServices().forEach { it.stopIfReady() }
       }
 
       override fun failed(from: State, failure: Throwable) {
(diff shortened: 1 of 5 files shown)


mightyguava:
cool stuff! What problem does it solve?

misk/src/main/kotlin/misk/ServiceGraphBuilder.kt
```diff
@@ -0,0 +1,133 @@
...
+
+    for ((_, service) in serviceMap) {
+      val cycle = service.findCycle(validityMap)
+      if (cycle != null) {
+        throw IllegalStateException("Detected cycle: ${cycle.joinToString(" -> ")}")
+      }
+    }
+  }
+
+  // Checks that each service registered with this builder has its dependencies registered.
```
swankjesse: Use `/** */` KDoc comments

misk/src/main/kotlin/misk/ServiceGraphBuilder.kt
```diff
@@ -0,0 +1,133 @@
...
+        for (dependent in dependents) {
+          stringBuilder.append("${serviceMap[dependent]}")
+        }
+        throw IllegalStateException("${stringBuilder.toString()} requires $big but no such service "
+            + "was registered with the builder")
+      }
+    }
+  }
+
+  // Checks that no enhancement has been applied more than once.
```
swankjesse: same

misk/src/main/kotlin/misk/ServiceGraphBuilder.kt
```diff
@@ -0,0 +1,133 @@
...
+  private fun validateEnhancementMap() {
+    val enhancementList = mutableListOf<Key<*>>()
+    for ((_, set) in enhancementMap.asMap()) {
```
swankjesse: `for (set in enhancementMap.asMap().values())`

swankjesse: or maybe just
```
enhancementList.addAll(enhancementMap.values())
```
?

misk/src/main/kotlin/misk/ServiceGraphBuilder.kt
```diff
@@ -0,0 +1,133 @@
...
+
+/**
+ * Builds a graph of `CoordinatedService2`s which defer start up and shut down until their dependent
+ * services are ready.
+ */
+class ServiceGraphBuilder {
+
+  private var serviceMap = mutableMapOf<Key<*>, CoordinatedService2>() // map of all services
+  private val dependencyMap = LinkedHashMultimap.create<Key<*>, Key<*>>()
+  private val enhancementMap = LinkedHashMultimap.create<Key<*>, Key<*>>()
```
swankjesse: should this be a regular map? No duplicate keys right?

misk/src/main/kotlin/misk/ServiceGraphBuilder.kt
```diff
@@ -0,0 +1,133 @@
...
+  }
+
+  /**
+   * Adds the [dependency] service as a dependency to the [service] provider service.
+   *
+   * @param dependency The identifier for the dependent service.
+   * @param service The identifier for the service that provides for `dependency`.
+   */
+  fun addDependency(service: Key<*>, dependency: Key<*>) {
+    val dependencySet = dependencyMap[service]
```
swankjesse: dependencyMap.put(service, dependency)

misk/src/main/kotlin/misk/ServiceGraphBuilder.kt
swankjesse: (and if you get multiple puts on the same key, you can fail on the 2nd put, not later in a validation step)

misk/src/main/kotlin/misk/ServiceGraphBuilder.kt
```diff
@@ -0,0 +1,133 @@
...
+   */
+  fun build(): ServiceManager {
+    validateDependencyMap()
+    validateEnhancementMap()
+    linkDependencies()
+    checkCycles()
+    return ServiceManager(serviceMap.values)
+  }
+
+  // Builds CoordinatedService2s from the instructions provided in the dependency and enhancement
```
swankjesse: `/** */`

misk/src/main/kotlin/misk/ServiceGraphBuilder.kt
```diff
@@ -0,0 +1,133 @@
...
+    }
+  }
+
+  // Checks that no service in this builder has specified a dependency cycle.
+  private fun checkCycles() {
+    val validityMap = mutableMapOf<CoordinatedService2, CycleValidity>()
+
+    for ((_, service) in serviceMap) {
+      val cycle = service.findCycle(validityMap)
+      if (cycle != null) {
```
swankjesse: replace `if` with `check`

misk/src/main/kotlin/misk/ServiceGraphBuilder.kt
```diff
@@ -0,0 +1,133 @@
...
+      val dependencies = dependencyMap[key]?.map { serviceMap[it]!! } ?: listOf()
+      service.addDependencies(dependencies)
+    }
+  }
+
+  // Checks that no service in this builder has specified a dependency cycle.
+  private fun checkCycles() {
+    val validityMap = mutableMapOf<CoordinatedService2, CycleValidity>()
+
+    for ((_, service) in serviceMap) {
```
swankjesse: iterate on `serviceMap.values()`

misk/src/main/kotlin/misk/ServiceGraphBuilder.kt
```diff
@@ -0,0 +1,133 @@
...
+      if (cycle != null) {
+        throw IllegalStateException("Detected cycle: ${cycle.joinToString(" -> ")}")
+      }
+    }
+  }
+
+  // Checks that each service registered with this builder has its dependencies registered.
+  // (i.e. no one service requires a dependency or enhancement that doesn't exist.)
+  private fun validateDependencyMap() {
+    for ((big, dependents) in dependencyMap.asMap()) {
```
swankjesse: `big` –> `service`

misk/src/main/kotlin/misk/ServiceGraphBuilder.kt
```diff
@@ -0,0 +1,133 @@
...
+      if (serviceMap[big] == null) {
```
swankjesse: replace if with check?

misk/src/main/kotlin/misk/ServiceGraphBuilder.kt
keeferrourke: I could do this if I swap the mapping.

Right now it's _service_ to _set of enhancements_, and the validation works by making sure no enhancement is in multiple sets.

keeferrourke: Alright this has been updated in commit 5b34926. Note that _`ServiceGraphBuilder.addEnhancement()`_ now throws from a check if there is an attempt to apply an enhancement to more than one service.

I also modified the _CoordinatedService2_ API to take varargs for `addDependencies()` and `addEnhancements()` respectively, since as a consequence of this change we are no longer always adding a list of services.

misk/src/main/kotlin/misk/ServiceGraphBuilder.kt
keeferrourke: done in b6b460d

misk/src/main/kotlin/misk/ServiceGraphBuilder.kt
keeferrourke: done in b6b460d

misk/src/main/kotlin/misk/ServiceGraphBuilder.kt
keeferrourke: done in b6b460d

misk/src/main/kotlin/misk/ServiceGraphBuilder.kt
keeferrourke: done in b6b460d

misk/src/main/kotlin/misk/ServiceGraphBuilder.kt
keeferrourke: done in b6b460d

misk/src/main/kotlin/misk/ServiceGraphBuilder.kt
keeferrourke: done in b6b460d

misk/src/main/kotlin/misk/ServiceGraphBuilder.kt
keeferrourke: done in b6b460d. went with `+=` instead of `addAll()`

misk/src/main/kotlin/misk/ServiceGraphBuilder.kt
keeferrourke: done in b6b460d

misk/src/main/kotlin/misk/ServiceGraphBuilder.kt
keeferrourke: done in b6b460d

keeferrourke:
> @mightyguava: cool stuff! What problem does it solve?

Jesse and I added an example of a use case in the docstring for `CoordinatedService2`! Hopefully its a helpful illustration of the problem :)

misk/src/main/kotlin/misk/CoordinatedService2.kt
```diff
@@ -0,0 +1,238 @@
...
+    while (t != null) {
+      result += t.directDependencies
+      t = t.enhancementTarget
+    }
+    result
+  }
+
+  /** Adds [services] as dependents downstream. */
+  fun addDependencies(vararg services: CoordinatedService2) {
+    directDependencies += services
```
mmihic: This seems dangerous. I don't think we want to allow dynamically added dependencies at runtime - dependencies should be built during graph initialization and the made immutable prior to service start. Ideally the service graph construction would occur outside the service itself such that by the time the service is created it is immutable; failing that we should be checking invariants here and failing if you attempt to add a dependency after the service is started.

misk/src/main/kotlin/misk/CoordinatedService2.kt
```diff
@@ -0,0 +1,238 @@
...
+ *
+ * ### How It Works
+ *
+ * CoordinatedService2 wraps the actual Service to defer its start up and shut down until its
+ * dependent services are ready.
+ *
+ * This Service will stall in the `STARTING` state until all upstream services are `RUNNING`.
+ * Symmetrically it stalls in the `STOPPING` state until all dependent services are `TERMINATED`.
+ */
+internal class CoordinatedService2(val service: Service) : AbstractService() {
```
mmihic: Is there going to be fast-follow PR replacing of all uses of the existing ``CoordinatedService``?

keeferrourke: I'm still new to the project and this PR was made with a lot of guidance from @swankjesse... But I believe that's the idea, yes.

misk/src/main/kotlin/misk/CoordinatedService2.kt
keeferrourke: @mmihic due to the self-relationships in the class, it doesn't seem possible to add enhancements/dependencies at construction time, but commit 10e38a8 adds runtime checks (and tests) that:
 - no _`CoordinatedService2`_ can wrap an already running Service
 - no _`CoordinatedService2`_ can have a dependency/enhancement added if it or the services to be added to it are not all `NEW`

Hopefully this addresses your concerns!

misk/src/main/kotlin/misk/CoordinatedService2.kt
swankjesse: Yep. Kill CoordinatedService, then take its name.

misk/src/main/kotlin/misk/ServiceGraphBuilder.kt
```diff
@@ -0,0 +1,122 @@
...
+  private var serviceMap = mutableMapOf<Key<*>, CoordinatedService2>()
+  private val dependencyMap = LinkedHashMultimap.create<Key<*>, Key<*>>()
+  private val enhancementMap = mutableMapOf<Key<*>, Key<*>>()
+  /**
+   * Registers a [service] with this [ServiceGraphBuilder]
+   *
+   * A service should be added before dependencies or enhancements are specified.
+   * Keys must be unique. If a key is reused, then the original key-service pair will be replaced.
+   *
+   * @param key A Guice Key used to identify the service.
```
swankjesse: these docs aren’t useful, they are implied by the parameter types

PR #986, opened 2019-05-15
Added VerifiedColumn annotation

yoavamit:
When used, this annotation will verify the authenticity of the data
stored in the corresponding column and will warn and thrown an exception
if the data failed verification.

### DIFF
--- a/misk-hibernate/src/main/kotlin/misk/hibernate/SessionFactoryService.kt
+++ b/misk-hibernate/src/main/kotlin/misk/hibernate/SessionFactoryService.kt
@@ -149,6 +149,7 @@ internal class SessionFactoryService(
     if (value !is SimpleValue) return
 
     val field = field(persistentClass, property)
+
     if (field.isAnnotationPresent(JsonColumn::class.java)) {
       value.typeName = JsonColumnType::class.java.name
       if (value.typeParameters == null) {
@@ -157,14 +158,12 @@ internal class SessionFactoryService(
       value.typeParameters.setField("jsonColumnField", field)
     } else if (field.isAnnotationPresent(ProtoColumn::class.java)) {
       value.typeName = ProtoColumnType::class.java.name
-
       if (value.typeParameters == null) {
         value.typeParameters = Properties()
       }
       value.typeParameters.setField("protoColumnField", field)
     } else if (field.isAnnotationPresent(SecretColumn::class.java)) {
       value.typeName = SecretColumnType::class.java.name
-
       if (value.typeParameters == null) {
         value.typeParameters = Properties()
       }
@@ -172,6 +171,13 @@ internal class SessionFactoryService(
           field.getAnnotation(SecretColumn::class.java).keyName)
       value.typeParameters.setProperty(SecretColumnType.FIELD_ENCRYPTION_INDEXABLE,
           field.getAnnotation(SecretColumn::class.java).indexable.toString())
+    } else if (field.isAnnotationPresent(VerifiedColumn::class.java)) {
+      value.typeName = VerifiedColumnType::class.java.name
+      if (value.typeParameters == null) {
+        value.typeParameters = Properties()
+      }
+      value.typeParameters.setProperty(VerifiedColumnType.FIELD_HMAC_KEY_NAME,
+          field.getAnnotation(VerifiedColumn::class.java).keyName)
     }
   }
 
--- a/misk-hibernate/src/main/kotlin/misk/hibernate/VerifiedColumnType.kt
+++ b/misk-hibernate/src/main/kotlin/misk/hibernate/VerifiedColumnType.kt
@@ -5,6 +5,7 @@ import misk.crypto.MacKeyManager
 import misk.logging.getLogger
 import org.hibernate.HibernateException
 import org.hibernate.engine.spi.SharedSessionContractImplementor
+import org.hibernate.id.UUIDGenerator
 import org.hibernate.type.spi.TypeConfiguration
 import org.hibernate.type.spi.TypeConfigurationAware
 import org.hibernate.usertype.ParameterizedType
@@ -27,6 +28,9 @@ internal class VerifiedColumnType : UserType, ParameterizedType, TypeConfigurati
   private lateinit var mac: Mac
   private lateinit var keyName: String
   private lateinit var _typeConfiguration: TypeConfiguration
+  private val uuidGenerator: UUIDGenerator =
+      UUIDGenerator.buildSessionFactoryUniqueIdentifierGenerator()
+
   override fun getTypeConfiguration(): TypeConfiguration = _typeConfiguration
 
   override fun setTypeConfiguration(typeConfiguration: TypeConfiguration) {
@@ -60,10 +64,12 @@ internal class VerifiedColumnType : UserType, ParameterizedType, TypeConfigurati
     index: Int,
     session: SharedSessionContractImplementor?
   ) {
+    val salt = uuidGenerator.generate(session, value).toString().toByteArray()
     (value as String?)?.let {
       st.setString(index, it)
-      st.setBytes(index + 1, mac.computeMac(it.toByteArray()))
-    } ?: st.setNull(index, Types.VARBINARY)
+      st.setBytes(index + 1, salt)
+      st.setBytes(index + 2, mac.computeMac(it.toByteArray() + salt))
+    } ?: st.setNull(index, Types.VARCHAR)
   }
 
   override fun nullSafeGet(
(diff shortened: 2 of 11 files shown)


misk-hibernate/src/main/kotlin/misk/hibernate/VerifiedColumnType.kt
```diff
@@ -0,0 +1,98 @@
...
+import java.io.Serializable
+import java.security.GeneralSecurityException
+import java.sql.PreparedStatement
+import java.sql.ResultSet
+import java.sql.Types
+import java.util.Base64
+import java.util.Objects
+import java.util.Properties
+
+internal class VerifiedColumnType : UserType, ParameterizedType, TypeConfigurationAware {
```
swankjesse: This is definitely the most natural way to implement this, but I’d like to discuss whether it’s the most useful. As-is an attacker can copy a value + HMAC from another row to bypass the check

misk-hibernate/src/main/kotlin/misk/hibernate/SessionFactoryService.kt
```diff
@@ -157,21 +158,26 @@ internal class SessionFactoryService(
...
       value.typeName = SecretColumnType::class.java.name
-
       if (value.typeParameters == null) {
         value.typeParameters = Properties()
       }
       value.typeParameters.setProperty(SecretColumnType.FIELD_ENCRYPTION_KEY_NAME,
           field.getAnnotation(SecretColumn::class.java).keyName)
       value.typeParameters.setProperty(SecretColumnType.FIELD_ENCRYPTION_INDEXABLE,
           field.getAnnotation(SecretColumn::class.java).indexable.toString())
+    } else if (field.isAnnotationPresent(VerifiedColumn::class.java)) {
```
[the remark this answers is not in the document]
yoavamit: That is a tricky question!
We've thought about it before, and it makes it does make sense and could be a nice feature to have.
On the other hand, it has 2 major disadvantages:
- It makes it hand to reason what should be the underlying datatype in the database. For example, if you want to have an encrypted/verified json, should it be defined as a `BLOB`? `VARBINARY`? `VARCHAR`?
- There's also the problem of the order of operations, how can we make sure for example that serialization happens before encryption/hashing, also in what form should it be cached?

In order to both avoid these issues and also provide explicitness to user defined entities, I've decided to not try and support that, *for now*.
If one wants to define a table that contains an encrypted or verified proto/json, one should take care of the serialization itself and provide a `String` to be hashed and persisted by the column annotated with `@VerifiedColumn`/`@SecretColumn`.

misk-hibernate/src/main/kotlin/misk/hibernate/VerifiedColumnType.kt
yoavamit: That's correct, as I wrote the documentation for the `VerifiedColumn` annotation, this doesn't provide any secrecy. An attacker could definitely steal the data and try to use it. The thing this provide is the guarantee that existing data cannot be modified out of band.
The main use case we had in mind while writing this is, suppose you have a ledger of transactions, and an attacker tried to modify an existing transaction or add a new one.
Given an attacker gains write access to the database, it cannot do either because it lacks access to the key used when doing the hmac, and verification will fail.

There are cases where we'd like to provide both secrecy and authenticity of the data, but those are less common, and we don't plan to deal with for for now (you could always define an `@SecretColumn` and do the hashing yourself).

AlexSzlavik: To Jesse's point though: Let's say you have two rows in the table and they have different verifiable data. An adversary could still change content of either row to be the value of the other and everything would check out. So while an attacker can't change the data to arbitrary content, they could change it to another value already present in the table.
Are we missing something here? What about deletes? Could an adversary just delete the column and the hmac and everything would be ok?

misk-hibernate/src/main/kotlin/misk/hibernate/SessionFactoryService.kt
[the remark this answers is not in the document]
ddz: I can totally see the reasoning on why it could be desirable to make these column attributes composable, but it's not a great fit with cryptographic best practices. In general, the best practice for modern encryption is to do authenticated encryption using an algorithm such as AES-GCM or ChaCha20Poly1305. So for our "Secret" columns, they are really already "Secret+Verified", but the 2nd part is implicit. I am pretty sure that we don't want to offer an option of making a Secret-but-not-verified column.

This might be something that we can make clearer with our attribute naming to clarify.

misk-hibernate/src/main/kotlin/misk/hibernate/VerifiedColumnType.kt
ddz: I think the crux of @swankjesse's point is that our encryption schemes for both `SecretColumn` and `VerifiedColumn` do not guarantee row-level integrity and it is an excellent point.

This could be addressed by incorporating the primary key of the row in the additional authenticated data used by the AEAD primitive for `SecretColumn` and HMAC for the `VerifiedColumn`. Both of these would cryptographically prevent copying ciphertext/MAC'd data from one row into another row.

ddz: Forgot to address @AlexSzlavik's question about dropping a column. A whole-row integrity check value (ICV) column could be added to prevent dropping any single column. We may need to provide some table-level annotation for stronger guarantees such as whole-row integrity and whole table integrity (e.g. preventing malicious deletion of a row).

yoavamit: These are great points!
We've already started planning a solution for some of them, but that would probably be done in another PR.
For `SecretColumn` - I'd like to introduce a `AssociatedAuthenticationDataProvider` that will be able to produce consistent `aad`s for encrypted values.
A similar thing could be added for `VerifiedColumn` as well.

About the whole-row integrity issue, that is a bit more tricky and I'm not familiar enough with Hibernate yet to know if there's an easy solution here.

swankjesse: The challenge with whole-row is that the primary key is typically generated after the cell values are all known. This necessitates a 2nd write for the initial insert of the row.

I think we should figure out which thread vectors we’re defending against, and then build solutions specific to them. This particular feature is well built, but I have trouble anticipating an attack it defends against.

I’m anxious about adding this as-is because it suggests to its user that it adds verification, but there are too many caveats for it to be used effectively.

yoavamit: I've been doing a little bit of researching yesterday about the whole-row integrity, and I think it's not going to be easily feasible.
Even those Hibernate provides an interceptor interface for `INSERT` and `UPDATE` statements, it doesn't provide a way to intercept `SELECT` queries in a standard way, so verifying the row's integrity is going to be tough.
Also, there's a whole new world of problems when performing a query for only some of the columns in the table.

I totally understand why adding this as-is is going to be a problem, calling a `VerifiedColumn` is going to be a big overstatement and misleading.
For that end, I think the best course of action here is to:
- For the `VerifiedColumn`, add a salt. This can be another column in the table. It doesn't have to be secret or anything, but must have a uniqueness constraint
- For the `SecretColumn` start using the `aad` when encrypting/decrypting. It can also be implemented as another column - not secret but unique across the table.

wdyt?

swankjesse:
Great idea! Proceed with the salt!

swankjesse:
Hmmmm.... It's weakness will be swapping will still work, and we don't enforce uniqueness across shards.

misk-hibernate/src/main/kotlin/misk/hibernate/VerifiedColumnType.kt
yoavamit: OK, the think the only major caveat that still remains here is the uniqueness across shards, which I'm still trying to solve.
Is there a way to get some Id generator that can guarantee this constraint?

swankjesse: I think that's right. We don't currently have an ID generator that runs before the insert, but we'll have lots of reasons to want one and this is such a case.

yoavamit: Got it.

Another thing I was thinking about, (not trying to be lazy here, but...) if someone has access to multiple shards doesn't it mean we have bigger things to worry about?
AFAIK, shards are meant to be this "non-sharable" thing, where each shard can act as a standalone instance of the database and is completely independent of the rest of the shards.
If someone can copy authenticated/encrypted data from one shard to another, that someone basically own the entire database and can do a lot worse.
What I'm saying is, is this something we really want to have in scope right now?

swankjesse: Good point! So what attack vectors we are attempting to address with this. Naughty internal actors? SQL injection?

chris-ryan-square:

chris-ryan-square:

PR #990, opened 2019-05-16
Make UserType function parameters nullable in BoxedStringType

### LINKED ISSUE #989: Handle a nullable BoxedStringType in misk-hibernate
For e.g., if you attempt to save a nullable `BoxedStringType`:

```
data class OptionalToken(val token: String)

@Entity
class DbEntity() ... {
  ...

  @Column(nullable = true)
  var optional_token: OptionalToken? = null

  ...
}
```

You get the error:
```
java.lang.IllegalArgumentException: Parameter specified as non-null is null: method misk.hibernate.BoxedStringType.deepCopy, parameter value

	at misk.hibernate.BoxedStringType.deepCopy(BoxedStringType.kt)
```

This is because the `BoxedStringType` overrides some functions from a hibernate class and [assumes non-null parameters](https://github.com/square/misk/blob/master/misk-hibernate/src/main/kotlin/misk/hibernate/BoxedStringType.kt#L41).

The other `UserType` fields all handle this by making the [parameters to the overridden functions nullable](https://github.com/square/misk/blob/master/misk-hibernate/src/main/kotlin/misk/hibernate/ByteStringType.kt#L15), we should do the same for `BoxedStringType`.

hamdanjaveed:
[CLOSES #989]

### DIFF
--- a/misk-hibernate/src/test/kotlin/misk/hibernate/BoxedStringColumnTest.kt
+++ b/misk-hibernate/src/test/kotlin/misk/hibernate/BoxedStringColumnTest.kt
@@ -29,10 +29,13 @@ class BoxedStringColumnTest {
   @Test
   fun happyPath() {
     val abcToken = GoodLuckToken("abc")
+    val abcOptionalToken = GoodLuckToken("abc optional")
+
     val defToken = GoodLuckToken("def")
+    val defOptionalToken = GoodLuckToken("def optional")
     transacter.transaction { session ->
-      session.save(DbTextToken("abc", abcToken))
-      session.save(DbTextToken("def", defToken))
+      session.save(DbTextToken("abc", abcToken, abcOptionalToken))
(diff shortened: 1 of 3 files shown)


misk-hibernate/src/test/kotlin/misk/hibernate/BoxedStringColumnTest.kt
```diff
@@ -29,17 +29,37 @@ class BoxedStringColumnTest {
...
           .token(abcToken)
           .uniqueResult(session)!!
       assertThat(textHash.text).isEqualTo("abc")
       assertThat(textHash.token).isEqualTo(abcToken)
+      assertThat(textHash.optional_token).isEqualTo(abcOptionalToken)
+    }
+  }
+
+  @Test
+  fun nullOptionalToken() {
```
wesleyk: nice test!

PR #994, opened 2019-05-17
Don't error on warnings

oldergod:


jjestrel:
why do we want this?

swankjesse:
This setting during development is punitive. Unused variable? No test execution for you.

oldergod:
My 2 cents would be: if we want an error behavior on warnings; then let's make it so they are actually errors.

PR #996, opened 2019-05-20
Log error when service fails

tirsen:


misk-testing/src/main/kotlin/misk/testing/MiskTestExtension.kt
```diff
@@ -78,13 +78,7 @@ internal class MiskTestExtension : BeforeEachCallback, AfterEachCallback {
 
     override fun beforeEach(context: ExtensionContext) {
       if (context.startService()) {
-        try {
-          serviceManager.startAsync().awaitHealthy(60, TimeUnit.SECONDS)
-        } catch (e: IllegalStateException) {
-          // Unearth the root cause so we can see it in logs easily.
-          log.error("Startup failure -- ${e.cause?.message}")
```
tirsen: This doesn't work it just prints this:
```
    06:52:24.358 [Test worker] ERROR misk.testing.MiskTestExtension - Startup failure -- null
```

PR #1000, opened 2019-05-21
Fail initialization in the wrong environment

yan:
Adds checks that make sure unencrypted keys can not be used in a non-testing/non-development environment.

### DIFF
--- a/misk-crypto/src/main/kotlin/misk/crypto/KeyProviders.kt
+++ b/misk-crypto/src/main/kotlin/misk/crypto/KeyProviders.kt
@@ -19,89 +19,13 @@ import com.google.crypto.tink.aead.KmsEnvelopeAead
 import com.google.crypto.tink.daead.DeterministicAeadFactory
 import com.google.inject.Inject
 import com.google.inject.Provider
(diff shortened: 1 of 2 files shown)


misk-crypto/src/main/kotlin/misk/crypto/KeyProviders.kt
```diff
@@ -19,14 +19,37 @@ import com.google.crypto.tink.aead.KmsEnvelopeAead
...
 
 val logger by lazy { getLogger<CryptoModule>() }
 
+open class KeyReader {
+  @Inject lateinit var env: Environment
+
+  fun readKey(key: Key, kmsUri: String?, kmsClient: KmsClient): KeysetHandle {
+    val keyJson = JsonKeysetReader.withString(key.encrypted_key.value)
+
+    return if (kmsUri != null) {
```
yoavamit: I'd break this big return statement into smaller chunks.
Something like
```
if (kmsUri != null) {
  return ...
}

if (env != TESTING && env != DEV) {
throw ...
}

return ...
```

PR #1002, opened 2019-05-24
Add max concurrent streams setting

jjestrel:


mightyguava:
What does this do for you? Force clients to create more h2 connections because they can't just infinitely multiplex a single one?

misk/src/test/kotlin/misk/web/AbstractRebalancingTest.kt
```diff
@@ -39,7 +42,18 @@ abstract class AbstractRebalancingTest(
           port = 0,
           idle_timeout = 500000,
           host = "127.0.0.1",
-          close_connection_percent = percent
+          close_connection_percent = percent,
+          ssl = WebSslConfig(0,
```
jmuia: nit: add the param name

jjestrel:
@mightyguava  yeah, jetty was picking 128 by default, and I believe we can force okhttp with http2 to create more sockets by picking a sufficiently low value

misk/src/main/kotlin/misk/web/WebConfig.kt
```diff
@@ -15,6 +15,7 @@ data class WebConfig(
   val queue_size: Int? = null,
   val jetty_max_thread_pool_size: Int? = null,
   val action_exception_log_level: ActionExceptionLogLevelConfig = ActionExceptionLogLevelConfig(),
+  val jetty_max_concurrent_streams: Int? = null,
```
swankjesse: tempted to rename this to `max_concurrent_streams_per_connection` since it’s not the max streams overall

swankjesse: (is it?)

jjestrel: I'll clean up this and other comment in a follow up PR

PR #1003, opened 2019-05-24
Deprecate CoordinatedService and clarify SGBuilder API

keeferrourke:
`CoordinatedService` usages have been fully removed and replaced by `ServiceGraphBuilder`. It is now deprecated, and associated tests have been disabled.

The `provideServiceManager` provider method follows the keys specified by `DependentService`s and has optional support for usage of the `install(ServiceModule<...>())` API.

The API and documentation for `ServiceGraphBuilder` has been clarfied because I found myself tripping over it before. Hopefully it is easier to use now.

### DIFF
--- a/misk/src/main/kotlin/misk/CoordinatedService.kt
+++ b/misk/src/main/kotlin/misk/CoordinatedService.kt
@@ -1,3 +1,5 @@
+@file:Suppress("DEPRECATION")
+
 package misk
 
 import com.google.common.util.concurrent.AbstractService
@@ -14,6 +16,7 @@ import com.google.inject.Key
  * starts up. Symmetrically it stalls in `STOPPING` until all downstream services are `TERMINATED`,
  * then it actually shuts down.
  */
+@Deprecated("Replaced by CoordinatedService2.")
 internal class CoordinatedService(val service: Service) : AbstractService(), DependentService {
   override val producedKeys = (service as? DependentService)?.producedKeys ?: setOf()
   override val consumedKeys = (service as? DependentService)?.consumedKeys ?: setOf()
--- a/misk/src/main/kotlin/misk/MiskServiceModule.kt
+++ b/misk/src/main/kotlin/misk/MiskServiceModule.kt
@@ -131,7 +131,7 @@ class MiskCommonServiceModule : KAbstractModule() {
       builder.addService(entry.key, service)
     }
     for (edge in dependencies) {
-      builder.addDependency(dependent = edge.service, dependsOn = edge.dependsOn)
+      builder.addDependency(dependent = edge.dependent, dependsOn = edge.dependsOn)
     }
     for (edge in enhancements) {
       builder.enhanceService(toBeEnhanced = edge.toBeEnhanced, enhancement = edge.enhancement)
@@ -179,23 +179,9 @@ class MiskCommonServiceModule : KAbstractModule() {
     val serviceSetKey = Key.get(Types.setOf(Service::class.java)) as Key<Set<Service>>
     private val log = KotlinLogging.logger {}
   }
-
 }
 
-/**
- * Normal dependency edge in the service dependency graph.
- *
- * @param service Identifier for the dependent service.
- * @param dependsOn Identifier for the service that [service] depends on.
- */
-internal data class DependencyEdge(val service: Key<*>, val dependsOn: Key<*>)
-
-/**
- * Enhancement edge in the service dependency graph.
- *
- * @param toBeEnhanced The key of the service to be enhanced.
- * @param enhancement The key of the service that enhances [toBeEnhanced].
- */
+internal data class DependencyEdge(val dependent: Key<*>, val dependsOn: Key<*>)
 internal data class EnhancementEdge(val toBeEnhanced: Key<*>, val enhancement: Key<*>)
 internal data class ServiceEntry(val key: Key<out Service>)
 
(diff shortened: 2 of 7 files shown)


misk/src/main/kotlin/misk/CoordinatedService.kt
```diff
@@ -1,3 +1,5 @@
+@file:Suppress("DEPRECATION")
```
swankjesse: Yuck. Errors as warnings mean we just suppress them which means we don’t even get to see them in IntelliJ. Yuck.

misk/src/main/kotlin/misk/MiskServiceModule.kt
```diff
@@ -184,8 +182,21 @@ class MiskCommonServiceModule : KAbstractModule() {
...
+ * @param service Identifier for the dependent service.
+ * @param dependsOn Identifier for the service that [service] depends on.
+ */
+internal data class DependencyEdge(val service: Key<*>, val dependsOn: Key<*>)
+
+/**
+ * Enhancement edge in the service dependency graph.
+ *
+ * @param toBeEnhanced The key of the service to be enhanced.
+ * @param enhancement The key of the service that enhances [toBeEnhanced].
```
keeferrourke: The parameter docs here might be a bit redundant, but I think its necessary to be very explicit on this, considering I wrote this API and then was subsequently confused by it 😓

misk/src/main/kotlin/misk/MiskServiceModule.kt
```diff
@@ -131,17 +107,39 @@ class MiskCommonServiceModule : KAbstractModule() {
...
+      var key : Key<*>
+      when (service) {
+        is DependentService -> {
+          key = service.producedKeys.firstOrNull() ?: service::class.toKey()
+          for (consumedKey in service.consumedKeys) {
+            builder.addDependency(dependent = key, dependsOn = consumedKey)
+          }
+        }
+        else -> {
+          key = service::class.toKey()
```
swankjesse: Can you change addService() to explode if the same key is added multiple times? I’m worried about name collisions here but that should protect against it.

misk/src/main/kotlin/misk/MiskServiceModule.kt
```diff
@@ -184,8 +182,21 @@ class MiskCommonServiceModule : KAbstractModule() {
 
 }
 
-internal data class DependencyEdge(val service: Key<*>, val dependency: Key<*>)
-internal data class EnhancementEdge(val service: Key<*>, val enhancement: Key<*>)
+/**
+ * Normal dependency edge in the service dependency graph.
+ *
+ * @param service Identifier for the dependent service.
+ * @param dependsOn Identifier for the service that [service] depends on.
```
swankjesse: redundant Javadoc!

misk/src/main/kotlin/misk/MiskServiceModule.kt
```diff
@@ -184,8 +182,21 @@ class MiskCommonServiceModule : KAbstractModule() {
...
+ */
+internal data class DependencyEdge(val service: Key<*>, val dependsOn: Key<*>)
```
swankjesse: Be consistent! This should be `dependent`

misk/src/main/kotlin/misk/MiskServiceModule.kt
```diff
@@ -184,8 +182,21 @@ class MiskCommonServiceModule : KAbstractModule() {
...
+
+/**
+ * Enhancement edge in the service dependency graph.
+ *
+ * @param toBeEnhanced The key of the service to be enhanced.
```
swankjesse: More redundant docs!

misk/src/main/kotlin/misk/MiskServiceModule.kt
```diff
@@ -85,38 +85,14 @@ class MiskCommonServiceModule : KAbstractModule() {
...
+  internal fun provideServiceManager(
     injector: Injector,
     services: List<Service>,
-    listeners: List<ServiceManager.Listener>
-  ): ServiceManager {
-    // NB(mmihic): We get the binding for the Set<Service> because this uses a multibinder,
-    // which allows us to retrieve the bindings for the elements
-    val serviceListBinding = injector.getBinding(serviceSetKey)
-    val invalidServices = serviceListBinding
-        .acceptTargetVisitor(CheckServicesVisitor())
```
swankjesse: We still have this check somewhere else?

misk/src/main/kotlin/misk/CoordinatedService.kt
keeferrourke: Yeah I'm not happy about this, but it was the only way to get the code to compile without modifying the Gradle script 😦

misk/src/main/kotlin/misk/MiskServiceModule.kt
keeferrourke: Line 99 in the updated MiskServiceModule. The diff is not very smart because it was duplicate code that I removed.

misk/src/main/kotlin/misk/MiskServiceModule.kt
keeferrourke: Nevermind 🤐

misk/src/main/kotlin/misk/MiskServiceModule.kt
keeferrourke: Done!

misk/src/main/kotlin/misk/MiskServiceModule.kt
keeferrourke: removed 😅

misk/src/main/kotlin/misk/MiskServiceModule.kt
keeferrourke: Done

misk/src/main/kotlin/misk/MiskServiceModule.kt
keeferrourke: Removed 😅

PR #1004, opened 2019-05-24
Use ServiceGraphBuilder in Misk AWS and Web modules

keeferrourke:


misk-aws/src/main/kotlin/misk/jobqueue/sqs/AwsSqsJobHandlerModule.kt
```diff
@@ -20,7 +18,8 @@ class AwsSqsJobHandlerModule<T : JobHandler> private constructor(
 ) : KAbstractModule() {
   override fun configure() {
     newMapBinder<QueueName, JobHandler>().addBinding(queueName).to(handler.java)
-    multibind<Service>().to<AwsSqsJobHandlerSubscriptionService>()
+    install(ServiceModule<AwsSqsJobHandlerSubscriptionService>()
+        .dependsOn<SqsJobConsumer>()) // Installed in AwsSqsJobQueueModule.
```
swankjesse: remove the comment!

misk/src/main/kotlin/misk/MiskServiceModule.kt
```diff
@@ -108,7 +108,27 @@ class MiskCommonServiceModule : KAbstractModule() {
 
     val builder = ServiceGraphBuilder()
 
+    // Avoid adding services to the builder twice, while we migrate the API.
```
swankjesse: Why? I think it should explode if it’s added 2x

misk/src/main/kotlin/misk/MiskServiceModule.kt
```diff
@@ -108,7 +108,27 @@ class MiskCommonServiceModule : KAbstractModule() {
...
+    for (edge in dependencies) {
+      builder.addDependency(dependent = edge.dependent, dependsOn = edge.dependsOn)
+    }
+    for (edge in enhancements) {
+      builder.enhanceService(toBeEnhanced = edge.toBeEnhanced, enhancement = edge.enhancement)
+    }
+
     // Support the deprecated DependantService interface.
+    //
+    // Since this loop operates on the set *all* services bound to the injector, we might already
```
swankjesse: It shouldn’t. It should only see services installed like this:

```
multibind<Service>() ...

```

misk/src/main/kotlin/misk/MiskServiceModule.kt
keeferrourke: Side effect of supporting both the `DependentService` and `ServiceGraphBuilder` APIs at the same time! `SGBuilder` uses `ServiceEntry` which grabs the `Service` instance from the injector, but the `DependentService` API grabs a set of all `Services` bound ever! Since `install(ServiceModule<>())` creates a binding, you'll get a service twice :(

keeferrourke: I had to add this change because `JettyService` would cause everything to explode even though it *is* only registered once.

misk/src/main/kotlin/misk/MiskServiceModule.kt
keeferrourke: `install(ServiceModule<>())` creates such a binding 😦

misk-aws/src/main/kotlin/misk/jobqueue/sqs/AwsSqsJobHandlerModule.kt
keeferrourke: done!

misk/src/main/kotlin/misk/MiskServiceModule.kt
swankjesse: change it to not?

PR #1009, opened 2019-05-27
Deprecated `misk.service.ServiceTestingModule`

keeferrourke:
`ServiceTestingModule` appears to have existed to accomodate for
deficiencies in service-dependency declaration at testing sites.
This module no longer needs to exist, as `misk.ServiceModule` provides a
Guice-compatible API which can be used to express full service
dependency graphs.


jjestrel:
A lot of projects have warnings as errors, so this will end up being a breaking change

keeferrourke:
Also holding off on merging this for now.

PR #1011, opened 2019-05-27
Provision Services as requested by the service graph

keeferrourke:
`CoordinatedService2` now takes a `Provider<Service>` instead of a `Service` in its constructor. 

This allows for the services to be provisioned as requested by the coordinated boot-up order in the service graph.

### DIFF
--- a/misk/src/main/kotlin/misk/CoordinatedService2.kt
+++ b/misk/src/main/kotlin/misk/CoordinatedService2.kt
@@ -5,6 +5,7 @@ import com.google.common.util.concurrent.MoreExecutors
 import com.google.common.util.concurrent.Service
 import com.google.common.util.concurrent.Service.Listener
 import com.google.common.util.concurrent.Service.State
+import javax.inject.Provider
 
 /**
  * Services in Misk can depend on other services.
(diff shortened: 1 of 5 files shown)


misk/src/main/kotlin/misk/CoordinatedService2.kt
```diff
@@ -55,7 +56,14 @@ import com.google.common.util.concurrent.Service.State
  * This Service will stall in the `STARTING` state until all upstream services are `RUNNING`.
  * Symmetrically it stalls in the `STOPPING` state until all dependent services are `TERMINATED`.
  */
-internal class CoordinatedService2(val service: Service) : AbstractService() {
+internal class CoordinatedService2(
+    private val serviceProvider : Provider<out Service>
+) : AbstractService() {
+
+  val service : Service by lazy {
```
swankjesse: COOL

misk/src/main/kotlin/misk/MiskServiceModule.kt
```diff
@@ -183,7 +182,7 @@ class MiskCommonServiceModule : KAbstractModule() {
 
 internal data class DependencyEdge(val dependent: Key<*>, val dependsOn: Key<*>)
 internal data class EnhancementEdge(val toBeEnhanced: Key<*>, val enhancement: Key<*>)
-internal data class ServiceEntry(val key: Key<out Service>)
+internal data class ServiceEntry(val key: Key<out Service>, val provider: Provider<out Service>)
```
swankjesse: omit the `Provider` field and compute it on-demand?

swankjesse: (that’ll be `injector.getProvider(key)` on line 133)

keeferrourke: fixed

PR #1019, opened 2019-05-29
Simplify service relationships in HibernateModule

keeferrourke:
Changes the dependency graph declaration (but not the graph itself!) to have `TransactorService` enhanced by `SchemaValidatorService` and `SchemaMigratorService`. Since `SessionFactoryService` is an internal class, I've introduced the public interface `TransactorService` to act as a level of indirection. Services that need a database connection can depend on `TransactorService` and an instance of `SessionFactoryService` will be started for them.

This will allow us to avoid the semantic annoyance of having services directly depend on `SchemaMigratorService` when all they care about is acquiring a connection to the database.

e.g., in tests, `DropTableService` is an enhancement of `TransactorService`.

The following are equivalent:

```kotlin
// Old World
install(ServiceModule<MoviesService>()
    .dependsOn<SchemaMigratorService>(Movies::class))
```

```kotlin
// New World:
// We really want to depend on a SessionFactoryService and have SchemaMigratorService
// start for us. So now we depend on TransacterService instead.
install(ServiceModule<MoviesService>()
    .dependsOn<TransacterService>(Movies::class))
```

With these changes, nothing but internal classes need to care about `SchemaMigratorService` or `SchemaValidatorService`, so both can be demoted to internal classes at a later point.

### DIFF
--- a/misk-hibernate/src/main/kotlin/misk/hibernate/HibernateModule.kt
+++ b/misk-hibernate/src/main/kotlin/misk/hibernate/HibernateModule.kt
@@ -65,7 +65,7 @@ class HibernateModule(
       }
     }).asSingleton()
 
-    // Bind PingDataBaseService.
+    // Bind PingDatabaseService.
     bind(keyOf<PingDatabaseService>(qualifier)).toProvider(Provider<PingDatabaseService> {
       PingDatabaseService(config, environmentProvider.get())
     }).asSingleton()
@@ -91,24 +91,6 @@ class HibernateModule(
     }).asSingleton()
     install(ServiceModule<DataSourceService>(qualifier)
         .dependsOn<PingDatabaseService>(qualifier))
-
-    // Bind SessionFactoryService.
-    val entitiesProvider = getProvider(setOfType(HibernateEntity::class).toKey(qualifier))
-    val eventListenersProvider =
-        getProvider(setOfType(ListenerRegistration::class).toKey(qualifier))
-    val hibernateInjectorAccessProvider = getProvider(HibernateInjectorAccess::class.java)
-    val dataSourceProvider = getProvider(keyOf<DataSource>(qualifier))
-
-    bind(keyOf<SessionFactory>(qualifier))
-        .toProvider(keyOf<SessionFactoryService>(qualifier))
-        .asSingleton()
-    bind(keyOf<SessionFactoryService>(qualifier)).toProvider(Provider<SessionFactoryService> {
-      SessionFactoryService(qualifier, config, dataSourceProvider,
-          hibernateInjectorAccessProvider.get(),
-          entitiesProvider.get(), eventListenersProvider.get())
-    }).asSingleton()
-    install(ServiceModule<SessionFactoryService>(qualifier)
-        .dependsOn<DataSourceService>(qualifier))
 
     // Bind SchemaMigratorService.
     val transacterKey = Transacter::class.toKey(qualifier)
@@ -142,8 +124,7 @@ class HibernateModule(
           config
       )
     }).asSingleton()
-    install(ServiceModule<SchemaMigratorService>(qualifier)
-        .dependsOn<SessionFactoryService>(qualifier))
+    install(ServiceModule<SchemaMigratorService>(qualifier))
 
     // Bind SchemaValidatorService.
     val sessionFactoryServiceProvider = getProvider(keyOf<SessionFactoryService>(qualifier))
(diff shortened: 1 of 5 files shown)


misk-hibernate/src/main/kotlin/misk/hibernate/HibernateModule.kt
```diff
@@ -157,8 +138,28 @@ class HibernateModule(
           )
         }).asSingleton()
     install(ServiceModule<SchemaValidatorService>(qualifier)
-        .dependsOn<SchemaMigratorService>(qualifier)
-        .dependsOn<SessionFactoryService>(qualifier))
+        .dependsOn<SchemaMigratorService>(qualifier))
+
+    // Bind SessionFactoryService as implementation of TransacterService.
```
keeferrourke: Note that I moved this block (verbatim) from above to keep the bindings in order of usage. Since `SchemaValidatorService` and `SchemaMigratorService` no longer depend on `SessionFactoryService`, this has been moved to ensure this configuration block continues to read as linearly as possible.

misk-hibernate/src/test/kotlin/misk/hibernate/SchemaValidatorTest.kt
```diff
@@ -46,31 +46,25 @@ class SchemaValidatorTest {
...
     override fun configure() {
       install(MiskTestingServiceModule())
+
       val qualifier = ValidationDb::class
 
       val dataSourceService =
           DataSourceService(qualifier, config.data_source, Environment.TESTING, emptySet())
 
       val injectorServiceProvider = getProvider(HibernateInjectorAccess::class.java)
-      val sessionFactoryServiceKey = SessionFactoryService::class.toKey(qualifier)
```
keeferrourke: I just inlined a bunch of keys here to clean up this configure block a bit.

misk-hibernate/src/test/kotlin/misk/hibernate/SchemaValidatorTest.kt
```diff
@@ -116,22 +111,23 @@ class SchemaValidatorTest {
...
             dataSourceService,
             injectorServiceProvider.get(),
             entitiesProvider.get())
       }).asSingleton()
-      install(ServiceModule<SessionFactoryService>(qualifier)
+      install(ServiceModule<TransacterService>(qualifier)
+          .enhancedBy<SchemaMigratorService>(qualifier)
           .dependsOn<DataSourceService>(qualifier))
-
-      install(ServiceModule<SchemaMigratorService>(qualifier)
```
keeferrourke: Similarly, reordered these declarations to ensure this reads linearly.

misk-hibernate/src/main/kotlin/misk/hibernate/HibernateModule.kt
swankjesse: FYI @tirsen this might collide with you?

misk-hibernate/src/test/kotlin/misk/hibernate/SchemaValidatorTest.kt
```diff
@@ -46,31 +46,25 @@ class SchemaValidatorTest {
...
-      val sessionFactoryServiceKey = SessionFactoryService::class.toKey(qualifier)
-
-      sessionFactoryService = getProvider(sessionFactoryServiceKey)
-
-      val entitiesKey = setOfType(HibernateEntity::class).toKey(qualifier)
-      val entitiesProvider = getProvider(entitiesKey)
 
-      val sessionFactoryKey = SessionFactory::class.toKey(qualifier)
-      val sessionFactoryProvider = getProvider(sessionFactoryKey)
+      sessionFactoryService = getProvider(keyOf<SessionFactoryService>(qualifier))
```
swankjesse: FWIW, we should suffix all Provider instances with the Provider suffix!

misk-hibernate/src/test/kotlin/misk/hibernate/SchemaValidatorTest.kt
```diff
@@ -46,31 +46,25 @@ class SchemaValidatorTest {
...
-      val transacterKey = Transacter::class.toKey(qualifier)
-      val transacterProvider = getProvider(transacterKey)
+      val entitiesProvider = getProvider(setOfType(HibernateEntity::class).toKey(qualifier))
 
-      val schemaMigratorKey = SchemaMigrator::class.toKey(qualifier)
-      val schemaMigratorProvider = getProvider(schemaMigratorKey)
+      val sessionFactoryProvider = getProvider(keyOf<SessionFactory>(qualifier))
+      bind<SessionFactory>()
+          .annotatedWith<ValidationDb>()
+          .toProvider(keyOf<SessionFactoryService>(qualifier))
```
swankjesse: nice improvement

PR #1022, opened 2019-05-30
Ignore resource extension when loading strings and byte arrays

yan:


misk/src/main/kotlin/misk/config/MiskConfig.kt
```diff
@@ -214,6 +214,13 @@ object MiskConfig {
           source
         }
         else -> {
+          // Ignore extension if we're requesting a string or a bytearray
+          if (type.rawClass == source.javaClass) {
```
acostama: Instead of comparing to `source.javaClass` it might better to explicitly compare to `String::class.java`. The test would break if source ever changes type.

PR #1023, opened 2019-05-30
Actually use connection_timeout/connection_max_lifetime in hikariconfig

wesleyk:


misk-hibernate/src/main/kotlin/misk/jdbc/DataSourceConfig.kt
```diff
@@ -33,7 +33,6 @@ data class DataSourceConfig(
   val database: String? = null,
   val username: String? = null,
   val password: String? = null,
-  val connection_properties: Map<String, String> = mapOf(),
```
wesleyk: was unused and not useful at this time so leaving it out

PR #1025, opened 2019-05-31
Promoted importance of ServiceModule

keeferrourke:
# What happening

 - `CoordinatedService` is gone!
- `ServiceModule` is more important than ever!

## `CoordinatedService` is gone?!

`CoordinatedService` is dead! Long live `CoordinatedService`!

`CoordinatedService` has been fully deprecated and isn't used *anywhere*.  It (and associated tests) has been deleted.

There will be a swift follow-up PR that renames `CoordinatedService2` back to `CoordinatedService`. Both are internal classes, so this will be a purely aesthetic change.

## `ServiceModule` changes

`ServiceModule` has been promoted to its own file, and documentation has been reorganized to make sure it is easy to use. Useful explanations were hidden on an internal class before, but that is no longer the case.

Additionally, provisioning a `ServiceManager` will now a log a warning if a service is multibound in the application, as they really should be installed via `ServiceModule`s.


misk/src/main/kotlin/misk/MiskServiceModule.kt
```diff
@@ -120,6 +119,14 @@ class MiskCommonServiceModule : KAbstractModule() {
     }
 
     // Support the deprecated DependantService interface.
+    if (services.isNotEmpty()) {
```
swankjesse: love this

PR #1030, opened 2019-06-04
Prepare to combine parameter binding with marshalling.

swankjesse:
This will make it possible to do streaming requests and responses
pretty seamlessly in Misk. In particular, I'm working towards passing
a response body stream in as a parameter.


swankjesse:
Yeah, using it will show us what we forgot and what we don’t need!

PR #1035, opened 2019-06-05
Change the way network interceptors work.

swankjesse:
Today these are supposed to return a Response<ResponseBody> which is
written to the servlet's response stream at the end of the call. This
works for basic calls, but doesn't permit the action to access the
stream directly, such as via a `@ResponseBody` parameter.

This is step 2 or 3, the next step is a mechanical rename of the
following:
  - Request becomes HttpCall
  - FakeRequest becomes FakeHttpCall

Once these are done we can adopt the WebActionBinding API to properly
combine all binding features under a single system.


swankjesse:
Test fails until I rebase on this:
https://github.com/square/misk/pull/1037/files

misk/src/main/kotlin/misk/web/Request.kt
```diff
@@ -1,29 +1,83 @@
...
+
+  fun setResponseHeader(name: String, value: String)
+  fun addResponseHeaders(headers: Headers)
+
+  /**
+   * Claim ownership of the request body stream. Returns null if the stream has already been
+   * claimed. Callers should read the HTTP request body or call [withRequestBody] to create a new
+   * chain with a request body that is unclaimed.
+   */
+  fun takeRequestBody(): BufferedSource?
```
oldergod: With feature bindings guaranteeing one and only one handler is taking of the request body, why do we have to provide a nullable type and one claims something already claimed?

misk/src/main/kotlin/misk/web/ServletHttpCall.kt
```diff
@@ -0,0 +1,95 @@
+package misk.web
```
oldergod: copyright

misk/src/main/kotlin/misk/web/interceptors/ResponseBodyMarshallerFactory.kt
```diff
@@ -0,0 +1,40 @@
+package misk.web.interceptors
```
oldergod: copyright

oldergod: I see some existing files don't have copyright either, we don't care?

misk/src/main/kotlin/misk/web/interceptors/ResponseBodyMarshallerFactory.kt
```diff
@@ -0,0 +1,40 @@
...
+    val contentTypeMarshaller = marshallerFactories.mapNotNull {
+      it.create(responseMediaType, action.returnType)
+    }.firstOrNull()
+
+    return contentTypeMarshaller
+        ?: genericMarshallerFor(responseMediaType, action.returnType)
+  }
+
+  private fun genericMarshallerFor(mediaType: MediaType?, type: KType): Marshaller<Any> {
+    return GenericMarshallers.from(mediaType, type) ?: throw IllegalArgumentException(
```
oldergod: nit, I would break before elvis

misk/src/main/kotlin/misk/web/jetty/WebActionsServlet.kt
```diff
@@ -107,33 +114,88 @@ internal class WebActionsServlet @Inject constructor(
...
-  return Request(
-      urlString()!!,
-      dispatchMechanism(),
-      headers(),
-      inputStream.source().buffer()
-  )
+internal fun HttpServletRequest.httpUrl(): HttpUrl {
+  return if (queryString == null)
+    HttpUrl.get("$requestURL") else
+    HttpUrl.get("$requestURL?$queryString")
```
oldergod: I'm not use to this kind of indention for if/else statements..

misk/src/test/kotlin/misk/web/interceptors/TracingInterceptorTest.kt
```diff
@@ -42,13 +40,11 @@ class TracingInterceptorTest {
   @Test
   fun initiatesTrace() {
     val tracingInterceptor = tracingInterceptorFactory.create(TracingTestAction::call.asAction())!!
-    val request = Request(
-        HttpUrl.parse("http://foo.bar/")!!,
-        DispatchMechanism.GET,
-        body = Buffer()
+    val request = FakeRequest(
```
oldergod: nit: probably fit on one line

misk/src/main/kotlin/misk/web/interceptors/ResponseBodyMarshallerFactory.kt
swankjesse: We have it in the root of the project!

misk/src/main/kotlin/misk/web/Request.kt
swankjesse: It’s so we can respond "internal error" if the application code throws, but only if it hasn’t already starting responding

misk/src/main/kotlin/misk/web/interceptors/ResponseBodyMarshallerFactory.kt
swankjesse: Done.

misk/src/main/kotlin/misk/web/jetty/WebActionsServlet.kt
swankjesse: Fixed.

misk/src/test/kotlin/misk/web/interceptors/TracingInterceptorTest.kt
swankjesse: Fixed.

PR #1040, opened 2019-06-06
Split FormValueParameterExtractorFactory into 3 parts.

swankjesse:
FormValueParameterExtractorFactory: glue to move form data into Action parameters
FormData: a model of form data
FormAdapter: glue to turn form data into an instance of a Kotlin class

Previously these were all in FormValueParameterExtractorFactory and that felt awkward.

### DIFF
--- a/misk/src/main/kotlin/misk/web/extractors/FormAdapter.kt
+++ b/misk/src/main/kotlin/misk/web/extractors/FormAdapter.kt
@@ -1,0 +1,77 @@
+package misk.web.extractors
+
+import misk.exceptions.requireRequest
+import misk.web.FormField
+import kotlin.reflect.KClass
+import kotlin.reflect.KFunction
+import kotlin.reflect.KParameter
+import kotlin.reflect.full.findAnnotation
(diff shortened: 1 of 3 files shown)


misk/src/main/kotlin/misk/web/extractors/FormAdapter.kt
```diff
@@ -0,0 +1,77 @@
...
+      return FormAdapter(constructor, fields)
+    }
+
+    private fun KParameter.toField(): Field {
+      val annotation = findAnnotation<FormField>()
+      val name = annotation?.name?.toLowerCase()
+          ?: name?.toLowerCase()
+          ?: throw IllegalStateException("cannot introspect parameter name")
+
+      // TODO(jwilson): explode if no converter is found. As is we just replace values with nulls.
```
adrw: Should this be addressed in this PR?

swankjesse: Nah, gotta follow rule #1. Never combine refactorings with behavior changes.

PR #1041, opened 2019-06-06
Add HTTP trailers to our HttpCall abstraction

swankjesse:
This makes it so that gRPC has one fewer weird thing to downcast.

### DIFF
--- a/misk/src/main/kotlin/misk/web/HttpCall.kt
+++ b/misk/src/main/kotlin/misk/web/HttpCall.kt
@@ -27,30 +27,61 @@ interface HttpCall {
   fun addResponseHeaders(headers: Headers)
 
   /**
+   * Call this before the response body is written to make sure it is encoded in a way that'll
+   * permit trailers to be sent. This will do chunked encoding for HTTP/1. For HTTP/2 trailers are
+   * always permitted. It is an error to call this for web socket calls.
+   */
+  fun requireTrailers()
+
+  /**
+   * Add a trailer. This requires that [requireTrailers] was called before the response body is
+   * written.
+   */
+  fun setResponseTrailer(name: String, value: String)
+
+  /**
    * Claim ownership of the request body stream. Returns null if the stream has already been
-   * claimed. Callers should read the HTTP request body or call [withRequestBody] to create a new
+   * claimed. Callers should read the HTTP request body or call [putRequestBody] to create a new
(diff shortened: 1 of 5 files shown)


misk/src/main/kotlin/misk/web/HttpCall.kt
```diff
@@ -26,31 +26,62 @@ interface HttpCall {
...
+   * permit trailers to be sent. This will do chunked encoding for HTTP/1. For HTTP/2 trailers are
+   * always permitted. It is an error to call this for web socket calls.
+   */
+  fun requireTrailers()
+
+  /**
+   * Add a trailer. This requires that [requireTrailers] was called before the response body is
+   * written.
+   */
+  fun setResponseTrailer(name: String, value: String)
```
oldergod: Let's rename it to `addResponseTrailer`? Otherwise feels like it's overriding

misk/src/main/kotlin/misk/web/ServletHttpCall.kt
```diff
@@ -36,36 +36,55 @@ internal data class ServletHttpCall(
...
+  }
 
   /** Adapts the underlying servlet call to send data to the client. */
   interface UpstreamResponse {
     var statusCode: Int
     val headers: Headers
     fun setHeader(name: String, value: String)
     fun addHeaders(headers: Headers)
+    fun requireTrailers()
+    fun setTrailer(name: String, value: String)
```
oldergod: again, `addTrailer`? The response can have many

misk/src/main/kotlin/misk/web/HttpCall.kt
swankjesse: Yes! Lemme fix in follow-up

misk/src/main/kotlin/misk/web/ServletHttpCall.kt
swankjesse: Yes!

swankjesse: Actually, studying this, the implementation does a set and not an add. (You can have multiple values for a headers otherwise).

oldergod: hmmm, ok. Still confusing to me. `Map` uses `put` for that; it's more or less the same I guess

PR #1048, opened 2019-06-10
Release leases if the cluster weight is 0

ryanhall07:
This allows applications to run in an Active-Passive mode. If the
current weight for a cluster is 0, leases are not acquired or released
if already acquired. This allows an application to only run leased
background tasks in the Active region.

This iteration requires misk users to provide their own implementation
for ClusterWeightProvider. SQ services will use a global ZK impl, which
we could consider contributing as well in the future.

### DIFF
--- a/misk/src/main/kotlin/misk/clustering/weights/ActiveClusterWeight.kt
+++ b/misk/src/main/kotlin/misk/clustering/weights/ActiveClusterWeight.kt
@@ -1,0 +1,22 @@
+package misk.clustering.weights
+
+import misk.inject.KAbstractModule
+
+/**
+ * A static [ClusterWeightProvider] that always returns 100
+ */
+class ActiveClusterWeight : ClusterWeightProvider {
+
+  override fun get(): Int {
+    return 100
+  }
+}
+
+/**
+ * Provides an [ActiveClusterWeight]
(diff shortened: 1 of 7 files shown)


misk/src/main/kotlin/misk/clustering/weights/ActiveClusterWeight.kt
```diff
@@ -0,0 +1,22 @@
+package misk.clustering.weights
+
+import misk.inject.KAbstractModule
+
+/**
+ * A static [ClusterWeightProvider] that always returns 100
+ */
+class ActiveClusterWeight : ClusterWeightProvider {
+
+  override fun get(): Int {
```
swankjesse: don’t love Integers for this, though I also don’t love floats?

ryanhall07: i'm open to suggestions, but not really a fan of floats. is the float value 0.801 or 80.1? at least an Int is clear in that regard. Loss of precision didn't seem important for something like cluster weights.

I briefly considered a boolean, since in practice it's 0 or 100. However, it's reasonable in the future we might support a canary cluster running some percentage of the background tasks.

PR #1052, opened 2019-06-14
Allow declaring dependent services for a JobHandler

frojasg:
We may need dependant service before we start consuming events in a JobHandler.

### DIFF
--- a/misk-aws/src/main/kotlin/misk/jobqueue/sqs/AwsSqsJobHandlerModule.kt
+++ b/misk-aws/src/main/kotlin/misk/jobqueue/sqs/AwsSqsJobHandlerModule.kt
@@ -1,8 +1,11 @@
 package misk.jobqueue.sqs
 
 import com.google.common.util.concurrent.AbstractIdleService
+import com.google.common.util.concurrent.Service
+import com.google.inject.Key
 import misk.ServiceModule
 import misk.inject.KAbstractModule
+import misk.inject.toKey
 import misk.jobqueue.JobHandler
 import misk.jobqueue.QueueName
 import javax.inject.Inject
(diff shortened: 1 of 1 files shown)


misk-aws/src/main/kotlin/misk/jobqueue/sqs/AwsSqsJobHandlerModule.kt
```diff
@@ -14,17 +17,21 @@ import kotlin.reflect.KClass
...
-  private val handler: KClass<T>
+  private val handler: KClass<T>,
+  private val dependsOn: List<Key<out Service>> = listOf()
 ) : KAbstractModule() {
   override fun configure() {
     newMapBinder<QueueName, JobHandler>().addBinding(queueName).to(handler.java)
-    install(ServiceModule<AwsSqsJobHandlerSubscriptionService>()
-        .dependsOn<SqsJobConsumer>())
+    install(ServiceModule(AwsSqsJobHandlerSubscriptionService::class.toKey(),
+        dependsOn + listOf(SqsJobConsumer::class.toKey())))
```
swankjesse: FYI @keeferrourke

swankjesse: I’d love to give you a mechanism to do this elsewhere

swankjesse:
... I’m trying to think of what the obvious test looks like, and I think it probably asserts something about the ServiceGraph. @keeferrourke is that easy?

My recommendation is you merge as is, and I borrow some of @keeferrourke’s time to add the test. Cause testing this might not be as easy as it should be.

keeferrourke:
> ... I’m trying to think of what the obvious test looks like, and I think it probably asserts something about the ServiceGraph. @keeferrourke is that easy?
> 
> My recommendation is you merge as is, and I borrow some of @keeferrourke’s time to add the test. Cause testing this might not be as easy as it should be.

That sounds good to me!

ryanhall07:
> ... I’m trying to think of what the obvious test looks like, and I think it probably asserts something about the ServiceGraph. 

My initial thought was a test specific to the jobqueue. You create a fake dependent service that flips some boolean and in the job handler you verify that boolean has been flipped.

misk-aws/src/main/kotlin/misk/jobqueue/sqs/AwsSqsJobHandlerModule.kt
keeferrourke: #1055

swankjesse:
I think this can be closed. We fixed it so that all services are running before any jobs run, which makes the contract between jobs and their handlers very straightforward.

frojasg:
closing in favor of #1055

PR #1055, opened 2019-06-14
Ensure that SQS job handlers do not run until healthy

keeferrourke:
`JobHandlers` can be totally arbitrary and might depend on services in a misk application (e.g. database, etc.). This patch ensures `SqsJobConsumer` doesn't start consuming jobs until all services are up and running in a healthy state.

### DIFF
--- a/misk-aws/src/main/kotlin/misk/jobqueue/sqs/SqsJobConsumer.kt
+++ b/misk-aws/src/main/kotlin/misk/jobqueue/sqs/SqsJobConsumer.kt
@@ -2,6 +2,7 @@ package misk.jobqueue.sqs
 
 import com.amazonaws.services.sqs.model.ReceiveMessageRequest
 import com.google.common.util.concurrent.AbstractIdleService
+import com.google.common.util.concurrent.ServiceManager
 import io.opentracing.Tracer
 import io.opentracing.tag.StringTag
 import io.opentracing.tag.Tags
@@ -15,6 +16,7 @@ import java.util.concurrent.ConcurrentHashMap
 import java.util.concurrent.ExecutorService
 import java.util.concurrent.atomic.AtomicBoolean
 import javax.inject.Inject
+import javax.inject.Provider
 import javax.inject.Singleton
 import kotlin.concurrent.thread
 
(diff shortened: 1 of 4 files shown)


misk-aws/src/main/kotlin/misk/jobqueue/sqs/SqsJobConsumer.kt
```diff
@@ -62,6 +70,10 @@ internal class SqsJobConsumer @Inject internal constructor(
     private val running = AtomicBoolean(true)
 
     override fun run() {
+      // Don't call handlers until all services are ready, otherwise handlers will crash because the
+      // services they might need (databases, etc.) won't be ready.
+      serviceManagerProvider.get().awaitHealthy()
```
ryanhall07: why this approach as opposed to the original proposal in https://github.com/cashapp/misk/pull/1052/files ?

the original proposal seemed fine and it was more explicit about declaring what you depend on.

ryanhall07: nit: blocking here is going to screw with the timing metrics for the job.

swankjesse: I think it’s a mistake ’cause it means we’re running Jobs before the service is officially booted up. That probably means every job needs to know every service it depends on, which is weird.

(We wouldn’t want HTTP actions to be received when the webserver is running but before the database is ready)

swankjesse: Timing metrics for the job happen below. This is just the runloop for the entire job processing thread.

misk-aws/src/test/kotlin/misk/jobqueue/sqs/SqsJobQueueServiceTest.kt
```diff
@@ -0,0 +1,92 @@
...
+  @AfterEach
+  internal fun tearDown() {
+    serviceManager.stopAsync()
+    serviceManager.awaitStopped(20, TimeUnit.SECONDS)
+  }
+
+  @Test fun jobsNotHandledUntilAllServicesAreRunning() {
+    val log = LinkedBlockingDeque<String>()
+
+    sleep(100)
```
wesleyk: can we avoid having all these actual sleeps in here? It adds unnecessary test runtime. It's also not clear why we need them all, since the job handler shouldn't run until manual start service is started if I understand correctly.

keeferrourke: Removed the sleeps.

swankjesse: Put the sleeps back! It’s the simplest way to make sure that if we regress here we’ll catch it. Using latches or other concurrency mechanisms is possible but it adds complexity.

keeferrourke: Put them back.

jjestrel: What does this catch? If the job was running before the services were ready then the job would process the log.put's?

swankjesse: Then the log.puts will be out-of-order and we know we have a problem.

wesleyk: Why do the sleeps help here though? For example, nothing happens before this first sleep.

PR #1057, opened 2019-06-17
Fix ReadinessCheckAction

### LINKED ISSUE #1056: Don't return 200s until all services are running
The readiness check should confirm that all services are running, and if not return a non-200 code.

keeferrourke:
`ReadinessCheckAction` was incorrectly looking for a list of `Service`
provided by multibindings. It now looks for a service manager and gets
the services from there.

Also fixes `LivenessCheckAction` and `StatusAction`.

---

Closes #1056

### DIFF
--- a/misk/src/main/kotlin/misk/web/actions/ReadinessCheckAction.kt
+++ b/misk/src/main/kotlin/misk/web/actions/ReadinessCheckAction.kt
@@ -1,6 +1,6 @@
 package misk.web.actions
 
-import com.google.common.util.concurrent.Service
+import com.google.common.util.concurrent.ServiceManager
 import misk.healthchecks.HealthCheck
 import misk.logging.getLogger
 import misk.security.authz.Unauthenticated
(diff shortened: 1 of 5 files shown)


misk/src/main/kotlin/misk/web/actions/ReadinessCheckAction.kt
```diff
@@ -9,21 +9,25 @@ import misk.web.Response
 import misk.web.ResponseContentType
 import misk.web.mediatype.MediaTypes
 import javax.inject.Inject
+import javax.inject.Provider
 import javax.inject.Singleton
 
 private val logger = getLogger<ReadinessCheckAction>()
 
 @Singleton
 class ReadinessCheckAction @Inject internal constructor(
```
swankjesse: Also fix LivenessAction, StatusAction

keeferrourke: Fixed!

swankjesse:
```
BUILD SUCCESSFUL in 6m 10s
107 actionable tasks: 68 executed, 39 up-to-date
```

PR #1058, opened 2019-06-17
Fail the HibernateHealthCheck if the service isn't running

swankjesse:
Previously we'd attempt to get a SessionFactory and that would crash in injection


swankjesse:
```
BUILD SUCCESSFUL in 4m 18s
130 actionable tasks: 129 executed, 1 up-to-date
```

PR #1062, opened 2019-06-18
Query zk connection directly for ZkLeaseManager connection state

ryanhall07:
There is no need to track the state locally in the manager, since the
curator client is already tracking it locally. In addition, there is a
chance for them to be out of sync, resulting in nasty race conditions.
If the manager does not think the connection is established, no leases
are acquired!

Previously this was relying on the connection listener to update the
local manager state. However, there was a race between the manager
starting and the connection being established. If the connection was
established before the manager started, the listener to update the local
state was never dispatched.


nbroyles:
Nice find!

PR #1065, opened 2019-06-18
Add support for explicit envelope encryption

yan:
Data generated by pre-PR services will not be decryptable.


misk-crypto/src/main/kotlin/misk/crypto/KeyProviders.kt
```diff
@@ -27,25 +27,36 @@ import java.security.GeneralSecurityException
...
+      KeysetHandle.read(reader, kek)
+    } catch (ex: GeneralSecurityException) {
+      logger.warn { "using single-encrypted KEKs" }
+      val reader = JsonKeysetReader.withString(key.encrypted_key.value)
+      KeysetHandle.read(reader, masterKey)
+    }
   }
 
   fun readKey(key: Key, kmsUri: String?, kmsClient: KmsClient): KeysetHandle {
-    val keyJson = JsonKeysetReader.withString(key.encrypted_key.value)
```
yoavamit: nit: Why were the method signatures for `readEncryptedKey` and `readCleartextKey` changed instead of reading the key here?

misk-crypto/src/test/kotlin/misk/crypto/CryptoModuleTest.kt
```diff
@@ -42,14 +42,6 @@ class CryptoModuleTest {
     assertThat(testKey).isNotNull()
   }
 
-  @Test
-  fun testAeadIsEnvelopeKey() {
```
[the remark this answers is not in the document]
yan: This was removed since we swapped the wrapping of keys; the actual AEAD returned from a keymanager will be a plain `Aead`, not the explicitly created `KmsEnvelopeAead`.

misk-crypto/src/main/kotlin/misk/crypto/KeyProviders.kt
yan: Because in the `readEncryptedKey` case, when the new format of wrapping fails, we'll try to re-read the key using the old format and we'd have to recreate the reader since first attempt would have drained the key reader.

PR #1067, opened 2019-06-19
Ensure services added via ServiceModule are singleton

keeferrourke:
For feature parity with the legacy explicit service multibindings.

### DIFF
--- a/misk/src/main/kotlin/misk/MiskServiceModule.kt
+++ b/misk/src/main/kotlin/misk/MiskServiceModule.kt
@@ -97,18 +97,17 @@ class MiskCommonServiceModule : KAbstractModule() {
     val serviceListBinding = injector.getBinding(serviceSetKey)
     val invalidServices = serviceListBinding
         .acceptTargetVisitor(CheckServicesVisitor())
-        .sorted()
-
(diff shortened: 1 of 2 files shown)


misk/src/main/kotlin/misk/MiskServiceModule.kt
```diff
@@ -97,18 +97,17 @@ class MiskCommonServiceModule : KAbstractModule() {
...
-    check(invalidServices.isEmpty()) {
-      "the following services are not marked as @Singleton: ${invalidServices.joinToString(", ")}"
-    }
+        .sorted().toMutableList()
 
     val builder = ServiceGraphBuilder()
 
     // Support the new ServiceModule API.
     for (entry in serviceEntries) {
+      // Check that the bound ServiceEntry is Singleton. This does the same thing as the visitor for
```
swankjesse: This comment probably doesn’t make much sense to future readers of this code?

keeferrourke: It refers to the stuff that is happening on lines 95-100.

When we eventually remove the multibinder for `Service` (this is planned, right?), this comment can be removed with the code on 95-100 and the `CheckServicesVisitor` class.

PR #1070, opened 2019-06-20
Add Service.Listener to RepeatedTaskQueue

ryanhall07:
Helpful if the background thread ever terminates for some unknown reason


misk/src/main/kotlin/misk/tasks/RepeatedTaskQueue.kt
```diff
@@ -45,7 +45,28 @@ class RepeatedTaskQueue @VisibleForTesting internal constructor(
   override fun startUp() {
     if (!running.compareAndSet(false, true)) return
 
-    log.info { "starting repeated task queue $name" }
```
ryanhall07: this is now covered by the running msg

PR #1071, opened 2019-06-21
Use a RepeatedTaskQueue in SqsJobConsumer

### LINKED ISSUE #1069: misk.jobqueue.sqs.QueueReceiver needs to handle errors in client call
[If the client call](https://github.com/cashapp/misk/blob/f7f2f432ec99ddf7c82a2e0b3685e4513c3c0335/misk-aws/src/main/kotlin/misk/jobqueue/sqs/SqsJobConsumer.kt#L67-L72) throws an exception it will kill the thread consuming the queue events.

ryanhall07:
This fixes #1069, which was originally a bug in RepeatedTaskQueue back
in the day. We should prefer using RepeatedTaskQueue instead of hand
rolling threads to prevent repeating the same mistakes. In addition, job
consumers will backoff if there is no work or run into errors.

Now every parallel job consumer runs as a RepeatedTask and each Task
consumes N messages and dispatches them to separate executor service for
running.

### DIFF
--- a/misk-aws/src/main/kotlin/misk/jobqueue/sqs/AwsSqsJobQueueConfig.kt
+++ b/misk-aws/src/main/kotlin/misk/jobqueue/sqs/AwsSqsJobQueueConfig.kt
@@ -1,6 +1,7 @@
 package misk.jobqueue.sqs
 
 import misk.config.Config
+import misk.tasks.RepeatedTaskQueueConfig
 
 /**
  * [AwsSqsJobQueueConfig] is the configuration for job queueing backed by Amazon's
@@ -22,6 +23,8 @@ class AwsSqsJobQueueConfig(
   /**
    * Number of jobs that can be processed concurrently.
    */
-  val consumer_thread_pool_size: Int = 4
+  val consumer_thread_pool_size: Int = 4,
+
+  val task_queue: RepeatedTaskQueueConfig = RepeatedTaskQueueConfig()
 ) : Config
 
--- a/misk-aws/src/main/kotlin/misk/jobqueue/sqs/ForSqsConsumer.kt
+++ b/misk-aws/src/main/kotlin/misk/jobqueue/sqs/ForSqsConsumer.kt
@@ -3,4 +3,5 @@ package misk.jobqueue.sqs
 import javax.inject.Qualifier
 
 @Qualifier
+@Target(AnnotationTarget.FIELD, AnnotationTarget.FUNCTION, AnnotationTarget.VALUE_PARAMETER)
 internal annotation class ForSqsConsumer
--- a/misk-aws/src/main/kotlin/misk/jobqueue/sqs/SqsJobConsumer.kt
+++ b/misk-aws/src/main/kotlin/misk/jobqueue/sqs/SqsJobConsumer.kt
@@ -15,8 +15,8 @@ import misk.time.timed
 import misk.tracing.traceWithSpan
 import java.time.Duration
 import java.util.concurrent.ConcurrentHashMap
+import java.util.concurrent.ExecutionException
 import java.util.concurrent.ExecutorService
-import java.util.concurrent.atomic.AtomicBoolean
 import javax.inject.Inject
 import javax.inject.Provider
 import javax.inject.Singleton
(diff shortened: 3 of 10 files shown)


misk-aws/src/main/kotlin/misk/jobqueue/sqs/AwsSqsJobQueueConfig.kt
```diff
@@ -22,6 +22,11 @@ class AwsSqsJobQueueConfig(
   /**
    * Number of jobs that can be processed concurrently.
    */
-  val consumer_thread_pool_size: Int = 4
+  val consumer_thread_pool_size: Int = 4,
+
+  /**
+   * The amount of jitter to use when scheduling backoff times.
+   */
+  val backoff_jitter_ms: Long = 50
```
mightyguava: is this just for testing?

misk-aws/src/main/kotlin/misk/jobqueue/sqs/SqsJobConsumer.kt
```diff
@@ -53,70 +49,65 @@ internal class SqsJobConsumer @Inject internal constructor(
...
       }
-    }
 
-    override fun close() {
-      if (!subscriptions.remove(queueName, this)) return
-
-      log.info { "closing subscription to queue ${queueName.value}" }
-      running.set(false)
+      return when {
+        failure.get() -> Status.FAILED
```
mightyguava: don't you need to wait for the dispatched tasks to complete before checking this?

misk/src/main/kotlin/misk/tasks/RepeatedTaskQueueConfig.kt
```diff
@@ -0,0 +1,33 @@
...
+import misk.backoff.ExponentialBackoff
+import java.time.Duration
+
+data class RepeatedTaskQueueConfig(
+  /**
+   * The default amount of jitter to use when scheduling backoffs.
+   *
+   * Can be overridden when scheduling a tasks.
+   */
+  val default_jitter_ms: Long = 50,
```
mightyguava: is this just for tests?

ryanhall07: well i'm using it for tests, but it seems reasonable to make it configurable

misk-aws/src/main/kotlin/misk/jobqueue/sqs/SqsJobConsumer.kt
ryanhall07: umm yea. should probably have a test :(

ryanhall07: fixed in 9cdece6a298d316f06d8cbd6a2d8920ce4526c44

misk-aws/src/main/kotlin/misk/jobqueue/sqs/SqsJobConsumer.kt
```diff
@@ -94,20 +96,24 @@ internal class SqsJobConsumer @Inject internal constructor(
...
       }
 
-      return when {
-        failure.get() -> Status.FAILED
-        messages.size == 0 -> Status.NO_WORK
-        else -> Status.OK
+      for (future in futures) {
+        try {
+          future.get()
+        } catch (e: Exception) {
```
mightyguava: not `Throwable`?

ryanhall07: it doesn't really matter here, you're not trying to prevent an Exception from being leaked (RepeatedTaskQueue handles that). Just trying to catch the signal that the dispatched handler threw an exception.

misk-aws/src/main/kotlin/misk/jobqueue/sqs/SqsJobConsumer.kt
ryanhall07: added a test too

misk-aws/src/main/kotlin/misk/jobqueue/sqs/ForSqsConsumer.kt
```diff
@@ -3,4 +3,5 @@ package misk.jobqueue.sqs
 import javax.inject.Qualifier
 
 @Qualifier
-internal annotation class ForSqsConsumer
\ No newline at end of file
+@Target(AnnotationTarget.FIELD, AnnotationTarget.FUNCTION, AnnotationTarget.VALUE_PARAMETER)
```
mmihic: hm, what necessitated this change (it's fine by me, just curious)

misk-aws/src/main/kotlin/misk/jobqueue/sqs/SqsJobConsumer.kt
```diff
@@ -53,70 +49,76 @@ internal class SqsJobConsumer @Inject internal constructor(
...
+
+      val futures = messages.map { SqsJob(queueName, queues, metrics, it) }.map { message ->
+        dispatchThreadPool.submit {
+          metrics.jobsReceived.labels(queueName.value).inc()
+
+          tracer.traceWithSpan("handle-job-${queueName.value}") { span ->
+            // If the incoming job has an original trace id, set that as a tag on the new span.
+            // We don't turn that into the parent of the current span because that would
+            // incorrectly include the execution time of the job in the execution time of the
+            // action that triggered the job        q
```
mmihic: trailing `    q`

misk-aws/src/main/kotlin/misk/jobqueue/sqs/ForSqsConsumer.kt
ryanhall07: so you could field inject in the test

```@Inject @ForSqsConsumer RepeatedTaskQueue```

ryanhall07: without having to do `@field:ForSqsConsumer`

PR #1075, opened 2019-06-24
Add HibernateExceptionMappers.kt

jjestrel:
map a bunch of Hibernate conflict exceptions to 409s

### DIFF
--- a/misk-hibernate/src/main/kotlin/misk/hibernate/HibernateExceptionMappers.kt
+++ b/misk-hibernate/src/main/kotlin/misk/hibernate/HibernateExceptionMappers.kt
@@ -1,0 +1,64 @@
+package misk.hibernate
+
+import com.google.common.base.Throwables
+import misk.exceptions.ConflictException
+import misk.web.Response
+import misk.web.ResponseBody
+import misk.web.exceptions.ExceptionMapper
+import misk.web.mediatype.MediaTypes
+import misk.web.toResponseBody
+import okhttp3.Headers
+import org.hibernate.StaleObjectStateException
+import org.hibernate.exception.ConstraintViolationException
+import org.slf4j.event.Level
+import java.sql.SQLIntegrityConstraintViolationException
+import javax.inject.Inject
+import javax.persistence.OptimisticLockException
+
+internal class RetryTransactionExceptionMapper @Inject internal constructor(
+) : ExceptionMapper<RetryTransactionException> {
+
+  override fun toResponse(th: RetryTransactionException): Response<ResponseBody>
+    = ConflictExceptionResponder.toResponse()
+
+  override fun canHandle(th: Throwable): Boolean {
+    val rootCause = Throwables.getRootCause(th)
+
+    return rootCause is SQLIntegrityConstraintViolationException ||
+      rootCause is StaleObjectStateException
+  }
+
+  override fun loggingLevel(th: RetryTransactionException) = Level.WARN
+}
+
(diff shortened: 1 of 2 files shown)


misk-hibernate/src/main/kotlin/misk/hibernate/HibernateExceptionMappers.kt
```diff
@@ -0,0 +1,67 @@
...
+import misk.web.toResponseBody
+import okhttp3.Headers
+import org.hibernate.StaleObjectStateException
+import org.hibernate.exception.ConstraintViolationException
+import java.sql.SQLIntegrityConstraintViolationException
+import javax.inject.Inject
+import javax.persistence.OptimisticLockException
+
+internal class RetryTransactionExceptionMapper @Inject internal constructor(
+  val config: ActionExceptionLogLevelConfig
```
jjestrel: We copy this from `ActionExceptionMapper`. maybe this should be renamed?

misk-hibernate/src/main/kotlin/misk/hibernate/HibernateExceptionMappers.kt
```diff
@@ -0,0 +1,64 @@
...
+}
+
+internal class OptimisticLockExceptionMapper @Inject internal constructor(
+) : ExceptionMapper<OptimisticLockException> {
+  override fun toResponse(th: OptimisticLockException): Response<ResponseBody>
+    = ConflictExceptionResponder.toResponse()
+
+  override fun canHandle(th: Throwable): Boolean = th is OptimisticLockException
+
+  override fun loggingLevel(th: OptimisticLockException) = Level.WARN
```
jjestrel: This matches `ActionExceptionMapper`'s default

misk-hibernate/src/main/kotlin/misk/hibernate/HibernateExceptionMappers.kt
```diff
@@ -0,0 +1,64 @@
...
+}
+
+object ConflictExceptionResponder {
```
swankjesse: internal?

jjestrel: changed

misk-hibernate/src/main/kotlin/misk/hibernate/HibernateExceptionMappers.kt
```diff
@@ -0,0 +1,64 @@
...
+import misk.web.toResponseBody
+import okhttp3.Headers
+import org.hibernate.StaleObjectStateException
+import org.hibernate.exception.ConstraintViolationException
+import org.slf4j.event.Level
+import java.sql.SQLIntegrityConstraintViolationException
+import javax.inject.Inject
+import javax.persistence.OptimisticLockException
+
+internal class RetryTransactionExceptionMapper @Inject internal constructor(
```
ryanhall07: this does not seem like a general mapper for `RetryTransactionException`, but very specific to the actual cause of the root exception. probably needs a better name to make that clear

jjestrel: agreed it's not general. when we have another case I'll rewrite this to get the root cause and check if there's a mapper for it. Or maybe `Throwables.getCasualChains()` and keep going down until there's an exception mapper (or up?)

Issue #1078, opened 2019-06-25
Replace @Authenticated.roles with capabilities


ryanhall07:
Using the term `roles` is extremely confusing to our cashapp developers since they assume it maps to the Roles in our internal Registry service. However, they actually map to capabilities in our internal Registry service. 

In addition, capabilities is a common way to manage permissions in an RBAC system. An API endpoint requires a capability (e.g view financial records) and that capability can be assigned to a role (e.g VeryImportantPerson). Many humans can then be added to that role.

The rollout process

1. Add `capabilities` to `@Authenticated` and friends. All readers should use a new `allCapabilities` property which is `capabilities + roles`.
2. Update all services to set `capabilities` instead of roles.
3. Remove `roles`

PR #1079, opened 2019-06-25
Add capabilities to @Authenticated

ryanhall07:
This is the first part of https://github.com/cashapp/misk/issues/1078 to
add a new capabilities field to @Authenticated that will eventually
replace the roles field.

I did not Deprecate the roles field, because many repos have -WError set
which will fail the build if there is a Deprecated Warning.

### DIFF
--- a/misk/src/main/kotlin/misk/MiskCaller.kt
+++ b/misk/src/main/kotlin/misk/MiskCaller.kt
@@ -9,7 +9,11 @@ data class MiskCaller(
   val user: String? = null,
 
   /** Set of roles to which the human user belongs, typically provided by the SSO infrastructure */
-  val roles: Set<String> = setOf()
+  // TODO(rhall): Deprecate this https://github.com/cashapp/misk/issues/1078
+  val roles: Set<String> = setOf(),
+
+  /** Set of capabilities given to a human user, typically provided by the SSO infrastructure */
+  val capabilities: Set<String> = setOf()
 ) {
   init {
     require(service != null || user != null) { "one of service or user is required" }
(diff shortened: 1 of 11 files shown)


misk/src/main/kotlin/misk/MiskCaller.kt
```diff
@@ -9,7 +9,11 @@ data class MiskCaller(
   val user: String? = null,
 
   /** Set of roles to which the human user belongs, typically provided by the SSO infrastructure */
-  val roles: Set<String> = setOf()
+  // TODO(rhall): Deprecate this https://github.com/cashapp/misk/issues/1078
```
ryanhall07: it's kind of lame that you can add `@Deprecated` with some kind of INFO level that won't break builds that have `-WError` set

misk/src/main/kotlin/misk/web/WebTab.kt
```diff
@@ -26,6 +26,8 @@ class DashboardTab(
   url_path_prefix: String,
   val name: String,
   val category: String = "Container Admin",
+    // TODO(rhall): Deprecate this https://github.com/cashapp/misk/issues/1078
   roles: Set<String> = setOf(),
```
ryanhall07: are these roles providing any value? couldn't we just show a 403 if one of the backend calls failed with Unauthorized? Is it just to gate showing the Tab? Seems not worth it to me at the expense of maintaining multiple ACL sets.

misk/src/main/kotlin/misk/MiskCaller.kt
swankjesse: I think `-WError` sucks

misk/src/main/kotlin/misk/web/WebTab.kt
swankjesse: Yeah, that might be simpler.

PR #1082, opened 2019-06-26
Add and apply spotless + ktlint

adrw:


build.gradle
```diff
@@ -74,6 +74,13 @@ subprojects {
     main.java.srcDirs += 'src/main/kotlin/'
     test.java.srcDirs += 'src/test/kotlin/'
   }
+  spotless {
+    kotlin {
+      target "**/*.kt"
+      ktlint(dep.ktlintVersion).userData(['indent_size': '2', 'continuation_indent_size' : '2'])
+    }
+  }
+  compileKotlin.dependsOn 'spotlessKotlinApply'
```
wesleyk: do we want to incur this with every compile?

adrw: It seems to be a trivial addition (< 1 sec) even on a project as large as Misk. If we want it to be moved to another stage of the lifecycle, it should exist at minimum as a pre-commit hook so that committed code has been linted and fixed.

wesleyk: just want to make sure it's not too much of a surprise to developers. How does it look in intellij?

adrw: For the most part it only appears as an extra line in the tasks blob in Gradle output. If it gets stuck on any fix, the error makes clear that it is a linting error and clearly states the problem and solution.

PR #1086, opened 2019-06-28
Use envelope master keys

yan:
This PR enforces an envelope encryption scheme that makes working with KMS's more secure.

### DIFF
--- a/misk-crypto/src/main/kotlin/misk/crypto/KeyProviders.kt
+++ b/misk-crypto/src/main/kotlin/misk/crypto/KeyProviders.kt
@@ -13,7 +13,6 @@ import com.google.crypto.tink.signature.PublicKeySignFactory
 import com.google.crypto.tink.signature.PublicKeyVerifyFactory
 import com.google.crypto.tink.JsonKeysetReader
 import com.google.crypto.tink.KeysetHandle
-import com.google.crypto.tink.KeysetReader
 import com.google.crypto.tink.aead.AeadFactory
 import com.google.crypto.tink.aead.AeadKeyTemplates
 import com.google.crypto.tink.aead.KmsEnvelopeAead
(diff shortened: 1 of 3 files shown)


misk-crypto/src/test/kotlin/misk/crypto/CryptoModuleTest.kt
```diff
@@ -116,6 +109,37 @@ class CryptoModuleTest {
...
+    val client = FakeKmsClient()
+    val (out, _) = captureOutput {
+      kr.readKey(key, "aws-kms://some-uri", client)
+    }
+    assertThat(out).contains("WARN")
+    assertThat(out).contains("using obsolete key format")
+  }
+
+  // Run a block, capturing stdout and stderr; return as strings
+  private fun captureOutput(f: () -> Any?): Pair<String, String> {
```
yoavamit: I'm not entirely sure about it, but I think Kotlin/Misk might have some tools for capturing/analyzing logs and output in tests

misk-crypto/src/main/kotlin/misk/crypto/KeyProviders.kt
```diff
@@ -13,37 +13,48 @@ import com.google.crypto.tink.signature.PublicKeySignFactory
...
 import com.google.inject.Inject
 import com.google.inject.Provider
 import misk.logging.getLogger
+import java.security.GeneralSecurityException
 
 val logger by lazy { getLogger<CryptoModule>() }
 
 open class KeyReader {
-  private fun readCleartextKey(reader: KeysetReader): KeysetHandle {
+  companion object {
```
yoavamit: nit: should this be private?

misk-crypto/src/test/kotlin/misk/crypto/CryptoModuleTest.kt
yoavamit: Yeah, take a look at `misk.logging.LogCollector`

PR #1092, opened 2019-07-05
Clarify Transacter documentation

keeferrourke:


misk-hibernate/src/main/kotlin/misk/hibernate/Transacter.kt
```diff
@@ -50,6 +50,28 @@ fun Transacter.shards() = transaction { it.shards() }
...
+ *
+ * For example if we were to save a new movie to a movie database, and update the gross revenue of
+ * existing movie:
+ * ```
+ * transaction.transaction { session ->
+ *   // Saving a new entity to the database needs an explicit call.
+ *   session.save(starWars)
+ *
+ *   // Updating a movie from the database is done by modifying the object.
+ *   // Changes are saved implciitly.
```
jjestrel: Implicitly

keeferrourke: Fixed

misk-hibernate/src/main/kotlin/misk/hibernate/Transacter.kt
```diff
@@ -50,6 +50,28 @@ fun Transacter.shards() = transaction { it.shards() }
...
+ *
+ * New objects must be persisted with an explicit call to [Session.save].
+ * Updates are performed implicitly by modifying objects returned from a query.
+ *
+ * For example if we were to save a new movie to a movie database, and update the gross revenue of
+ * existing movie:
+ * ```
+ * transaction.transaction { session ->
+ *   // Saving a new entity to the database needs an explicit call.
+ *   session.save(starWars)
```
swankjesse: This example needs a `new` call. Otherwise the saved instance comes from outside the transaction, which we forbid

misk-hibernate/src/main/kotlin/misk/hibernate/Transacter.kt
```diff
@@ -50,6 +50,28 @@ fun Transacter.shards() = transaction { it.shards() }
...
+ *
+ *   // Updating a movie from the database is done by modifying the object.
+ *   // Changes are saved implicitly.
+ *   val movie: DbMovie? = queryFactory.newQuery<MovieQuery>().id(id).uniqueResult(session)
```
swankjesse: Add the !! at the end of this line and change the type to `DbMovie`.

misk-hibernate/src/main/kotlin/misk/hibernate/Transacter.kt
keeferrourke: Fixed

misk-hibernate/src/main/kotlin/misk/hibernate/Transacter.kt
keeferrourke: Fixed

PR #1093, opened 2019-07-08
Add pingInterval support for HttpClientsConfig

jjestrel:


misk/src/main/kotlin/misk/client/HttpClientFactory.kt
```diff
@@ -26,6 +26,9 @@ class HttpClientFactory @Inject constructor(
         TimeUnit.MILLISECONDS) }
     config.writeTimeout?.let { builder.writeTimeout(it.toMillis(),
         TimeUnit.MILLISECONDS) }
+    config.pingInterval?.let {
```
adrw: Nice use of `let` 👍

PR #1096, opened 2019-07-11
Add TooManyRequestsException

jjestrel:


misk/src/main/kotlin/misk/exceptions/Exceptions.kt
```diff
@@ -48,6 +48,9 @@ open class BadRequestException(message: String = "", cause: Throwable? = null) :
 open class ConflictException(message: String = "", cause: Throwable? = null) :
     ActionException(StatusCode.CONFLICT, message, cause)
 
+open class TooManyRequestsException(message: String = "", cause: Throwable? = null) :
```
adrw: Is this triggered in the case of number of requests passing some threshold?

PR #1104, opened 2019-07-15
Reduce web-actions tab typing latency by 98%

adrw:
* Latency typing performance improved from >140ms for `keypress` to <2ms
* Use `onBlur` calls for Redux update functions instead of `onChange` which means the expensive Redux update only happens when the text field is exited, instead of on every keypress
* `defaultValue` instead of `value` is used to support existing functionality with the above `onBlur` change


mightyguava:
Consider debouncing your change handler instead to keep the field responsive while you type?

adrw:
Thanks @mightyguava ! I'm continuing research into a method to debounce the change handler (possibly https://github.com/ryanseddon/redux-debounced) and will issue a PR once that is fully tested. This PR makes the web-actions tab usable until that fix is landed.

PR #1106, opened 2019-07-19
Expose hibernate session.delete in Session

### LINKED ISSUE #1105: Expose hibernate's session.delete in misk's Session class
Right now we can only delete using `Query`. We should expose hibernate's `session.delete(entity)`

sfavron:
Fixes #1105

### DIFF
--- a/misk-hibernate/src/main/kotlin/misk/hibernate/RealTransacter.kt
+++ b/misk-hibernate/src/main/kotlin/misk/hibernate/RealTransacter.kt
@@ -255,6 +255,13 @@ internal class RealTransacter private constructor(
       } as Id<T>
     }
 
+    override fun <T : DbEntity<T>> delete(entity: T) {
+      check(!readOnly) {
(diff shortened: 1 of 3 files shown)


misk-hibernate/src/main/kotlin/misk/hibernate/RealTransacter.kt
```diff
@@ -255,6 +255,13 @@ internal class RealTransacter private constructor(
       } as Id<T>
     }
 
+    override fun <T : DbEntity<T>> delete(entity: T) {
+      if (readOnly) {
```
jjestrel: nit:
```
check(readOnly) {
    "Deleting isn't permitted in a read only session."
}
```

swankjesse: check(!read-only) { "Deleting isn't permitted in a read only session." }

PR #1108, opened 2019-07-23
Upgrade to Gradle 5.5

swankjesse:


jjestrel:
should we stop checking in the `gradle/wrapper` folder?

PR #1109, opened 2019-07-23
Fix BulkShardMigrator same shard migration logs showing %s instead of value

jjestrel:


misk-hibernate/src/main/kotlin/misk/hibernate/migrate/BulkShardMigrator.kt
```diff
@@ -182,8 +182,8 @@ class BulkShardMigrator<R : DbRoot<R>, C : DbChild<R, C>> private constructor(
           if (sourceRecords.isEmpty()) {
             0
           } else {
-            logger.info("Bulk migrating (same shard) %s entities for table %s",
-                sourceRecords.size, tableName)
+            logger.info("Bulk migrating (same shard) ${sourceRecords.size} entities for table" +
```
adrw: Good use of Kotlin string templating 👍

PR #1110, opened 2019-07-23
Add misk-guice module containing former :misk:misk misk.inject package

jjestrel:
part of the work to pull out `misk-client`


misk-aws/src/main/kotlin/misk/cloud/aws/AwsEnvironmentModule.kt
```diff
@@ -3,7 +3,7 @@ package misk.cloud.aws
 import com.google.inject.Provides
 import misk.environment.EnvVarLoader
 import misk.environment.ForEnvVars
-import misk.inject.KAbstractModule
+import misk.guice.KAbstractModule
```
swankjesse: I like the new module, but renaming the package seems like needless churn. Is this so we can have both concurrently?

misk-guice/build.gradle
```diff
@@ -0,0 +1,14 @@
+buildscript {
+  dependencies {
+    classpath dep.kotlinNoArgPlugin
+  }
+}
+
+apply plugin: 'kotlin-jpa'
```
swankjesse: Necessary?

misk-aws/src/main/kotlin/misk/cloud/aws/AwsEnvironmentModule.kt
jjestrel: changed back

misk-guice/build.gradle
jjestrel: removed

jjestrel:
The format changes are auto intellij changes

PR #1111, opened 2019-07-23
Use Wire's gRPC types in Misk

swankjesse:


build.gradle
```diff
@@ -1,5 +1,8 @@
 buildscript {
   repositories {
+    maven {
+      url "https://artifactory.global.square/artifactory/releases"
```
swankjesse: # Do not merge

This needs a Wire release first!

PR #1113, opened 2019-07-23
[misk.client] Records metrics for timeout requests

frojasg:


misk/src/test/kotlin/misk/client/ClientMetricsInterceptorTest.kt
```diff
@@ -1,123 +1,117 @@
...
-            "Content-type: " + MediaTypes.APPLICATION_JSON)
-    fun ping(@Body request: AppRequest): Call<AppResponse>
-  }
+  @Test
+  fun timeouts() {
+    assertThatExceptionOfType(ExecutionException::class.java).isThrownBy {
+      client.ping(AppRequest(200)).execute().code()
+    }.withCauseInstanceOf(SocketTimeoutException::class.java)
 
-  class PingAction @Inject constructor() : WebAction {
```
frojasg: It was easier to simulate timeouts by using  okhttp `MockWebServer`.

misk/src/test/kotlin/misk/client/ClientMetricsInterceptorTest.kt
```diff
@@ -1,123 +1,117 @@
...
     @Provides
     @Singleton
-    fun provideHttpClientConfig(): HttpClientsConfig {
+    fun provideHttpClientConfig(server: MockWebServer): HttpClientsConfig {
+      val url = server.url("/")
       return HttpClientsConfig(
-              endpoints = mapOf("pinger" to HttpClientEndpointConfig(jetty.httpServerUrl.toString())))
+          endpoints = mapOf("pinger" to HttpClientEndpointConfig(
+              url = url.toString(),
+              readTimeout = Duration.ofMillis(10)
```
frojasg: short timeout so we don't slow down the tests

misk/src/main/kotlin/misk/client/ClientMetricsInterceptor.kt
```diff
@@ -13,14 +16,20 @@ class ClientMetricsInterceptor internal constructor(
...
+      val result = chain.proceed(chain.request)
+      val elapsedMillis = stopwatch.stop().elapsed(TimeUnit.MILLISECONDS).toDouble()
+      requestDuration.record(elapsedMillis, actionName, "all")
+      requestDuration.record(elapsedMillis, actionName, "${result.code / 100}xx")
+      requestDuration.record(elapsedMillis, actionName, "${result.code}")
+      return result
+    } catch (e: SocketTimeoutException) {
+      val elapsedMillis = stopwatch.stop().elapsed(TimeUnit.MILLISECONDS).toDouble()
+      requestDuration.record(elapsedMillis, actionName, "all")
+      requestDuration.record(elapsedMillis, actionName, "timeout")
```
wesleyk: is this that useful? It's essentially what the client timeout is set to, no?

misk/src/test/kotlin/misk/client/ClientMetricsInterceptorTest.kt
```diff
@@ -1,123 +1,117 @@
...
-    @POST("/ping")
-    @Headers(
-            "Accept: " + MediaTypes.APPLICATION_JSON,
-            "Content-type: " + MediaTypes.APPLICATION_JSON)
-    fun ping(@Body request: AppRequest): Call<AppResponse>
-  }
+  @Test
+  fun timeouts() {
+    assertThatExceptionOfType(ExecutionException::class.java).isThrownBy {
+      client.ping(AppRequest(200)).execute().code()
```
wesleyk: interesting, where do we setup the mock server to timeout?

misk/src/main/kotlin/misk/client/ClientMetricsInterceptor.kt
frojasg: Yeah, the histogram metric is not very interesting other than validate you're timing out on the expected time. I was more interested to know how many times outs the client is seeing. Also, it consistent with the other metrics. Let me know if you want to use other metric.

misk/src/test/kotlin/misk/client/ClientMetricsInterceptorTest.kt
frojasg: You don't. The mock server only returns the enqueued requests. If you don't queue anything, it won't return, causing a timeout.

misk/src/main/kotlin/misk/client/ClientMetricsInterceptor.kt
swankjesse: I think it’s useful in that if it spikes it tells us something

PR #1117, opened 2019-07-25
Log redacted caller in AccessInterceptor

keeferrourke:
This can assist in debugging configuration of capabilities for users/webactions.

### DIFF
--- a/misk/src/main/kotlin/misk/MiskCaller.kt
+++ b/misk/src/main/kotlin/misk/MiskCaller.kt
@@ -18,4 +18,13 @@ data class MiskCaller(
 
   /** The identity of the calling principal, regardless of whether they are a service or a user */
   val principal: String get() = service ?: user!!
+
+  /** We don't like to log usernames. */
+  override fun toString(): String {
+    return if (user != null) {
+      "user=${user.firstOrNull() ?: ""}███████, capabilities=$capabilities"
+    } else {
+      "service=$service"
+    }
+  }
 }
--- a/misk/src/main/kotlin/misk/security/authz/AccessInterceptor.kt
+++ b/misk/src/main/kotlin/misk/security/authz/AccessInterceptor.kt
@@ -6,6 +6,7 @@ import misk.Chain
 import misk.MiskCaller
 import misk.exceptions.UnauthenticatedException
 import misk.exceptions.UnauthorizedException
+import misk.logging.getLogger
 import misk.scope.ActionScoped
 import javax.inject.Inject
 import kotlin.reflect.KClass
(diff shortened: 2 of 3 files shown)


misk/src/main/kotlin/misk/MiskCaller.kt
```diff
@@ -18,4 +18,13 @@ data class MiskCaller(
 
   /** The identity of the calling principal, regardless of whether they are a service or a user */
   val principal: String get() = service ?: user!!
+
+  /** We don't like to log usernames. */
+  override fun toString(): String {
+    return if (user != null) {
+      "user=${user.firstOrNull() ?: ""}███████, capabilities=$capabilities"
```
adrw: This looks like something that could be confirmed in a test, want to add one?

misk/src/main/kotlin/misk/security/authz/AccessInterceptor.kt
```diff
@@ -19,6 +20,7 @@ class AccessInterceptor private constructor(
   override fun intercept(chain: Chain): Any {
     val caller = caller.get() ?: throw UnauthenticatedException()
     if (!isAllowed(caller)) {
+      log.info { "$caller is not allow to access ${chain.action}" }
```
adrw: * `not allowed`

misk/src/main/kotlin/misk/security/authz/AccessInterceptor.kt
```diff
@@ -96,4 +98,8 @@ class AccessInterceptor private constructor(
     private inline fun <reified T : Annotation> Action.hasAnnotation() =
         function.annotations.any { it.annotationClass == T::class }
   }
+
+  companion object {
+    val log = getLogger<AccessInterceptor>()
```
adrw: We tend to use `logger` as the val instead of `log`.

misk/src/main/kotlin/misk/MiskCaller.kt
keeferrourke: Added

misk/src/main/kotlin/misk/security/authz/AccessInterceptor.kt
keeferrourke: fixed

misk/src/main/kotlin/misk/security/authz/AccessInterceptor.kt
keeferrourke: fixed

misk/src/test/kotlin/misk/MiskCallerTest.kt
```diff
@@ -0,0 +1,40 @@
...
+import javax.inject.Qualifier
+
+@MiskTest(startService = false)
+internal class MiskCallerTest {
+  @MiskTestModule val module = MiskCallerTestModule()
+  @Inject @TestUser lateinit var testUser: MiskCaller
+  @Inject @TestService lateinit var testService: MiskCaller
+
+  @Test fun userNameIsRedactedFromToString() {
+    assertThat("$testUser").doesNotContain(testUser.user)
```
adrw: 🥇

PR #1120, opened 2019-07-26
Track whether the connection changed on retries

swankjesse:
We've had bugs where MySQL appears to be returning stale data. The
goal here is to see if this problem is isolated to reused connections.


misk-hibernate/src/main/kotlin/misk/hibernate/RealTransacter.kt
```diff
@@ -81,22 +81,30 @@ internal class RealTransacter private constructor(
     while (true) {
       try {
         attempt++
-        return transactionInternal(block)
+        val result = transactionInternal(block)
+
+        if (attempt > 0) {
```
acostama: Seems like this is always true. Maybe `attempt > 1`?

swankjesse: Good catch!

PR #1121, opened 2019-07-26
Support Retrofit Converter factories for typed HTTP clients

alecholmes:
Example use:

```
    install(TypedPeerHttpClientModule(
        kclass = FooClient.Api::class,
        name = "foo",
        converterFactories = listOf(WireConverterFactory.create())))
```

### DIFF
--- a/misk/src/main/kotlin/misk/client/TypedHttpClientModule.kt
+++ b/misk/src/main/kotlin/misk/client/TypedHttpClientModule.kt
@@ -15,11 +15,17 @@ import retrofit2.Retrofit
 import java.lang.reflect.Proxy
 import kotlin.reflect.KClass
 
-/** Creates a retrofit-backed typed client given an API interface and an HTTP configuration */
+/**
+ * Creates a retrofit-backed typed client given an API interface and an HTTP configuration.
+ *
+ * @param retrofitBuilderProvider Optional provider of a [Retrofit.Builder]. This provider should
+ * not return a singleton since the builder it returns will be mutated.
+ */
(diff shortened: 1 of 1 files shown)


swankjesse:
Should we just have a built-in set that does the right thing?

alecholmes:
> Should we just have a built-in set that does the right thing?

I don't think there's one set that can cover everything. In my case, my ultimate goal is to add a Wire adapter. An alternative would be to allow passing an optional Guice `Key` for a `Retrofit.Builder` provider/factory to the `TypedPeerHttpClientModule` constructor. This is much more flexible although I wonder if it makes the binding harder to reason about.

alecholmes:
@swankjesse On second thought, passing a `Provider<Retrofit.Builder>` might be better in that it won't usually be used but allows the flexibility when needed.

```
class TypedPeerHttpClientModule<T : Any>(
  private val kclass: KClass<T>,
  private val name: String,
  private val retrofitBuilderProvider: Provider<Retrofit.Builder>? = null // <-- new
) { ... }
```
Usage:
```
class SomeAppModule : KAbstractModule() {
  override fun configure() {
    install(TypedPeerHttpClientModule(
        kclass = Hello.Api::class,
        name = "hello",
        retrofitBuilderProvider = getProvider(
            Key.get(Retrofit.Builder::class.java, Names.named("wire")))))
  }

  @Provides @Named("wire") fun wireRetrofitBuilder(): Retrofit.Builder {
    return Retrofit.Builder()
        .addConverterFactory(WireConverterFactory.create())
  }
```

PR #1122, opened 2019-07-27
Exclude junit4 so it stops getting accidentally imported

jjestrel:
I think using `implementation` also fixes this but that's a bigger change.

People have been accidentally importing the wrong `@Test` which leads to injection stuff not getting set up which is a very confusing error message

### DIFF
--- a/dependencies.gradle
+++ b/dependencies.gradle
@@ -39,6 +39,7 @@ ext.dep = [
   "jettyWebsocketServer": "org.eclipse.jetty.websocket:websocket-server:9.4.18.v20190429",
   "jettyWebsocketServlet": "org.eclipse.jetty.websocket:websocket-servlet:9.4.18.v20190429",
   "jnrUnixsocket": "com.github.jnr:jnr-unixsocket:0.22",
+  "junit4": "junit:junit:4.12",
   "junitApi": "org.junit.jupiter:junit-jupiter-api:5.3.1",
   "junitEngine": "org.junit.jupiter:junit-jupiter-engine:5.3.1",
   "junitGradlePlugin": "org.junit.platform:junit-platform-gradle-plugin:1.0.2",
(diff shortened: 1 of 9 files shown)


dependencies.gradle
```diff
@@ -39,6 +39,7 @@ ext.dep = [
   "jettyWebsocketServer": "org.eclipse.jetty.websocket:websocket-server:9.4.18.v20190429",
   "jettyWebsocketServlet": "org.eclipse.jetty.websocket:websocket-servlet:9.4.18.v20190429",
   "jnrUnixsocket": "com.github.jnr:jnr-unixsocket:0.22",
+  "junit4": "junit:junit:4.12",
```
jjestrel: `misk:misk` needs this because okhttp mock webserver needs it

misk-crypto/src/test/kotlin/misk/crypto/CryptoModuleTest.kt
```diff
@@ -17,16 +17,20 @@ import misk.environment.EnvironmentModule
...
+import org.junit.jupiter.api.Disabled
+import org.junit.jupiter.api.Test
 import java.io.ByteArrayOutputStream
 import java.security.GeneralSecurityException
 
 @MiskTest(startService = true)
 class CryptoModuleTest {
+  @Suppress("unused")
+  @MiskTestModule
+  val module = CryptoTestModule()
```
jjestrel: This test was broken for injection the whole time

misk-testing/src/test/kotlin/org/assertj/core/api/AssertExtensionsTest.kt
```diff
@@ -37,7 +37,12 @@ internal class AssertExtensionsTest {
 }
 """)
     }).hasMessage(
-        """expected:<"{ "my_structure[]" : [ "this" , 45, "...> but was:<"{ "my_structure[2]" : [ "this" , 45, "...>""")
```
jjestrel: These messages are different in junit5

misk/build.gradle
```diff
@@ -54,4 +54,5 @@ dependencies {
   testCompile dep.kotlinxCoroutines
   testCompile dep.mockitoCore
   testCompile project(':misk-testing')
+  testImplementation dep.junit4
```
jjestrel: `implementation` so we don't leak this dependency transitively

PR #1124, opened 2019-07-30
Support graceful shutdown for real this time

jjestrel:
```

     * The {@link StatisticsHandler} must be configured so that open connections can
     * be tracked for a graceful shutdown.
```

### DIFF
--- a/misk/src/main/kotlin/misk/web/jetty/JettyService.kt
+++ b/misk/src/main/kotlin/misk/web/jetty/JettyService.kt
@@ -16,6 +16,7 @@ import org.eclipse.jetty.server.HttpConnectionFactory
 import org.eclipse.jetty.server.NetworkConnector
 import org.eclipse.jetty.server.SecureRequestCustomizer
 import org.eclipse.jetty.server.Server
+import org.eclipse.jetty.server.ServerConnectionStatistics
 import org.eclipse.jetty.server.ServerConnector
 import org.eclipse.jetty.server.SslConnectionFactory
 import org.eclipse.jetty.server.handler.ContextHandler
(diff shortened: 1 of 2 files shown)


misk/src/main/kotlin/misk/web/jetty/JettyService.kt
```diff
@@ -117,12 +119,22 @@ class JettyService @Inject internal constructor(
...
-    server.addManaged(servletContextHandler)
-    server.handler = servletContextHandler
+    handlers.addHandler(servletContextHandler)
+
+    val statisticsHandler = StatisticsHandler()
+    statisticsHandler.server = server
+    handlers.addHandler(statisticsHandler)
 
     server.stopAtShutdown = true
+    server.stopTimeout = 20_000
```
jjestrel: 20 seconds is arbitrary but good since kubernetes gives us 30 second

swankjesse: Promote this to a code comment?

swankjesse: Is this line the only behavior change?

jjestrel:
@wesleyk yes, I tried to fix this before but forgot to add `StatisticsHandler`

misk/src/main/kotlin/misk/web/jetty/JettyService.kt
```diff
@@ -117,12 +120,21 @@ class JettyService @Inject internal constructor(
     }
 
     // TODO(mmihic): Force security handler?
+
```
alecholmes: Extra lines snuck in

jjestrel: fixed

jjestrel:
@swankjesse the behaviour change is that jetty ignores all the graceful shutdown stuff if you don't install `StatisticsHandler` so it should be _on_ now

PR #1125, opened 2019-07-31
Reflection utilities so we can handle annotations on supertypes.

swankjesse:
I want to do @WireRpc on supertypes but that PR isn't ready yet.

### DIFF
--- a/misk/src/main/kotlin/misk/web/actions/reflect.kt
+++ b/misk/src/main/kotlin/misk/web/actions/reflect.kt
@@ -1,0 +1,63 @@
+package misk.web.actions
+
+import java.lang.reflect.Method
+import java.util.ArrayDeque
+import kotlin.reflect.KFunction
+import kotlin.reflect.jvm.javaMethod
+
+/** Returns an instance of [T] annotating this method or a method it overrides. */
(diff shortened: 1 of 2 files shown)


misk/src/main/kotlin/misk/web/actions/reflect.kt
```diff
@@ -0,0 +1,60 @@
...
+import java.util.ArrayDeque
+import kotlin.reflect.KFunction
+import kotlin.reflect.jvm.javaMethod
+
+/** Returns an instance of [T] annotating this method or a method it overrides. */
+internal inline fun <reified T : Annotation> KFunction<*>.findAnnotationWithOverrides(): T? {
+  return javaMethod!!.findAnnotationWithOverrides(T::class.java)
+}
+
+/** Returns an instance of [T] annotating this method or a method it overrides. */
```
acostama: Maybe mention that if the method overrides a method from multiple superclasses all with the same annotation, it returns one of these "arbitrarily".

swankjesse: Thanks! Fixed.

PR #1128, opened 2019-08-01
Support customizing okhttp dispatcher max request limits

alecholmes:
This adds support for a passing through okhttp dispatcher request concurrency maximums from config.


jjestrel:
We should give a high default

alecholmes:
@jjestrel Updated this to have default Misk values higher than the OkHttp defaults.

PR #1130, opened 2019-08-02
RFC: Request deadlines

alecholmes:
This is not meant to be a final PR. Instead, the goal is to drive discussion on interfaces and behavior.

Applications often have upper bounds for how long a request can take before giving up or retrying. This upper bound is a deadline: a wall-clock instant by which work is completed or considered failed. Since work for a request may span multiple systems, it's useful to have a uniform way to propagate and adhere to deadlines. This PR starts to explore what deadline propagation could look like in Misk.

**Setting and passing deadlines**
Deadlines are passed through request side channels as timeouts in milliseconds. They are not passed as timestamps due to clock skew.

This PR uses a non-standard HTTP header, `X-Request-Timeout-Ms`, to pass the timeout in milliseconds. This does not use the standard HTTP header `Request-Timeout` as that uses _seconds_ as a time unit. A Misk server receiving a request will then set its own deadline of `clock.now().plusMillis(timeoutMs)`. The absence of this header indicates the request has no deadline.

GRPC has a standard sidechannel for passing timeouts as nanoseconds. Future GRPC-Misk integration would use that side channel.

**Deadlines in code**
Misk will be responsible for exposing timeouts from a request and scoping them to an action in a way that is abstracted from the transport. This PR has an `ActionDeadline` type that is scoped to an action using `ActionScoped<ActionDeadline>`. This is available even if there is no deadline for the action. Using `ActionScoped` fits with existing Misk patterns for request metadata.

If a client-determined deadline passes, Misk will not necessarily return an error to the caller and abort code. However, some Misk-wired components, such as HTTP clients, may choose to check deadlines and try and abort a request early by raising `ActionException`s. See `ClientDeadlineAppInterceptor` in this PR as an example.

**Overriding deadlines**
Actions often interact with other systems or databases and they may want to set tighter deadlines for some of those calls. For example, an action may to make two client calls in parallel, each with a different short deadline.

This PR allows overriding a deadline for a block of code. Importantly, this overridden deadline does not propagate to other threads. This supports an action doing work on concurrent threads with different deadlines, at the expense of framework code _possibly_ having to propagate deadlines to threads they spawm.

```
// This line is subject to the action's deadline
actionDeadline.get().overriding(Duration.ofMillis(100)) {
  // This block now has a 100ms deadline
  actionDeadline.get().overriding(Duration.ofMillis(5)) {
    // This block has a 5 ms deadline
  }
  // This is subject to the earlier 100ms deadline

  Thread(Runnable {
    // This thread is subject to the action's deadline (not the 100ms deadline!)
  }).start()
}
```

**Propagating deadlines**
This PR illustrates how deadlines can be propagated outwards in HTTP client calls. Future work could include:
* Having typed HTTP clients not only propagate timeouts but actually timeout on the calling side.
* Using timeouts in Transacter-created DB sessions.
* Passing timeouts to GRPC calls.

**Misc**

Deadlines and latency. Calls to the systems have varying latency overhead that eats into the deadline. E.g. if a client has a 100ms deadline and network RTT is 25ms, the server handling the request only has ~75ms to handle the request; longer than that and the client may time out. This PR does not have an opinion on managing this. It would naively pass alog a 100ms timeout to the server.

### DIFF
--- a/misk/src/main/kotlin/misk/Deadline.kt
+++ b/misk/src/main/kotlin/misk/Deadline.kt
@@ -9,19 +9,66 @@ import java.time.Duration
 import java.time.Instant
 import javax.inject.Inject
 
-class Deadline(
+/**
+ * Deadline for an action. A deadline can be overridden, with the caveat that the overridden deadline
+ * does not propagate to other threads.
+ *
+ * All actions have [ActionDeadline]s available even if there is initially no deadline.
+ */
+class ActionDeadline(
   private val clock: Clock,
-  val deadline: Instant
+  deadline: Instant?
 ) {
 
-  constructor(clock: Clock, duration: Duration) :
-      this(clock, clock.instant().plusNanos(duration.toNanos()))
+  constructor(clock: Clock, duration: Duration) : this(clock, clock.instant().plus(duration))
 
-  fun expired() = !clock.instant().isBefore(deadline)
+  private val threadDeadline: ThreadLocal<Instant?> = ThreadLocal.withInitial { deadline }
 
-  fun remaining(): Duration {
+  /**
+   * Get the current deadline, or null if there is none.
+   *
+   * If the deadline is currently being overridden in the same thread, that value is returned.
+   */
+  fun current(): Instant? = threadDeadline.get()
+
+  /** @return If the deadline is expired. If there is no deadline this returns false */
+  fun expired(): Boolean = current()?.let { it < clock.instant() } ?: false
+
+  /** @return Get the duration until the deadline is hit, or null if there is no deadline */
+  fun remaining(): Duration? = current()?.let { deadline ->
     val left = Duration.between(clock.instant(), deadline)
-    return if (left.isNegative) Duration.ZERO else left
+    if (left.isNegative) Duration.ZERO else left
+  }
+
+  /**
+   * Override the deadline for the given block. The overridden deadline only applies to the current
+   * thread.
+   *
+   * The lesser of the given deadline and current deadline is used.
+   */
+  fun <T> overriding(newDeadline: Instant, fn: () -> T): T {
+    val current = threadDeadline.get()
+    if (current != null && newDeadline >= current) {
+      return fn()
+    }
+
+    threadDeadline.set(newDeadline)
+    return try {
+      fn()
+    } finally {
+      threadDeadline.set(current)
+    }
+  }
+
+  /**
+   * Override the deadline for the given block. The overridden deadline only applies to the current
+   * thread.
+   *
+   * The lesser of the given deadline and current deadline is used.
+   */
+  fun <T> overriding(duration: Duration, fn: () -> T): T {
+    check(!duration.isZero && !duration.isNegative) { "duration must be positive" }
+    return overriding(clock.instant().plus(duration), fn)
   }
 }
 
@@ -29,21 +76,25 @@ class Deadline(
  * Provides an optional deadline scoped to the current action, determined by the optional
  * "X-Request-Timeout-Ms" HTTP header.
  */
-class DeadlineProvider @Inject constructor(
+class ActionDeadlineProvider @Inject constructor(
   private val clock: Clock,
   var currentRequest: ActionScoped<HttpCall>
-) : ActionScopedProvider<Deadline?> {
-  override fun get(): Deadline? = currentRequest.get().requestHeaders[HTTP_HEADER]?.let {
-    try {
-      val duration = Duration.ofMillis(it.toLong())
-      if (duration.isNegative || duration.isZero) {
-        throw BadRequestException(
-            "Invalid header value for $HTTP_HEADER; deadline must be positive")
+) : ActionScopedProvider<ActionDeadline> {
+  override fun get(): ActionDeadline {
+    val deadline = currentRequest.get().requestHeaders[HTTP_HEADER]?.let {
+      try {
+        val timeoutMs = it.toLong()
+        if (timeoutMs <= 0) {
+          throw BadRequestException(
+              "Invalid header value for $HTTP_HEADER; deadline must be positive")
+        }
+        clock.instant().plusMillis(timeoutMs)
+      } catch (ex: NumberFormatException) {
+        throw BadRequestException("Invalid header value for $HTTP_HEADER")
       }
-      Deadline(clock, duration)
-    } catch (ex: NumberFormatException) {
-      throw BadRequestException("Invalid header value for $HTTP_HEADER")
     }
+
+    return ActionDeadline(clock, deadline)
   }
 
   companion object {
--- a/misk/src/main/kotlin/misk/Deadline.kt
+++ b/misk/src/main/kotlin/misk/Deadline.kt
@@ -1,0 +1,52 @@
+package misk
+
+import misk.exceptions.BadRequestException
+import misk.scope.ActionScoped
+import misk.scope.ActionScopedProvider
+import misk.web.HttpCall
+import java.time.Clock
+import java.time.Duration
+import java.time.Instant
+import javax.inject.Inject
+
+class Deadline(
+  private val clock: Clock,
+  val deadline: Instant
+) {
+
+  constructor(clock: Clock, duration: Duration) :
+      this(clock, clock.instant().plusNanos(duration.toNanos()))
+
+  fun expired() = !clock.instant().isBefore(deadline)
+
+  fun remaining(): Duration {
+    val left = Duration.between(clock.instant(), deadline)
+    return if (left.isNegative) Duration.ZERO else left
+  }
+}
+
+/**
+ * Provides an optional deadline scoped to the current action, determined by the optional
+ * "X-Request-Timeout-Ms" HTTP header.
+ */
+class DeadlineProvider @Inject constructor(
+  private val clock: Clock,
+  var currentRequest: ActionScoped<HttpCall>
+) : ActionScopedProvider<Deadline?> {
+  override fun get(): Deadline? = currentRequest.get().requestHeaders[HTTP_HEADER]?.let {
+    try {
+      val duration = Duration.ofMillis(it.toLong())
+      if (duration.isNegative || duration.isZero) {
+        throw BadRequestException(
+            "Invalid header value for $HTTP_HEADER; deadline must be positive")
+      }
+      Deadline(clock, duration)
+    } catch (ex: NumberFormatException) {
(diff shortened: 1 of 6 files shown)


misk/src/main/kotlin/misk/Deadline.kt
```diff
@@ -0,0 +1,103 @@
...
+}
+
+/**
+ * Provides an optional deadline scoped to the current action, determined by the optional
+ * "X-Request-Timeout-Ms" HTTP header.
+ */
+class ActionDeadlineProvider @Inject constructor(
+  private val clock: Clock,
+  var currentRequest: ActionScoped<HttpCall>
+) : ActionScopedProvider<ActionDeadline> {
```
frojasg: I'm not sure `ActionScopedProvider` works. Is the `get` method executed when the request is received a memorized somehow?  Reading the code it gives me the impression we're executing the `get`  method lazily in the client interceptor.

alecholmes:
@frojasg Thanks for reviewing and for the feedback. The deadline propagation is something I haven't found a satisfying answer to and your comments on coroutines add a level of complexity to it that I hadn't originally considered. The Go convention of apps and libraries passing a `Context` around all the time would be helpful here, but alas.

I think the ultimate question is this: should a request/job deadline be an immutable piece of data, like the calling user or request body, or should a deadline be more of a scope that code runs in, similar to how tracing metadata and logging MDC.

If the choice is immutability, then I think the outcome is client and DB interfaces taking timeouts as inputs but it ultimately being the responsibility of the app developer to propagate any deadlines. However, we could still provide some framework code that can check and use the _request_ deadline if the caller hasn't provided an override. The upside to this approach is avoiding the impedance mismatch between thread locals and coroutines.

Alternatively, if we deadlines are scoped to the Guice request/job `Scope` then we can have a more unified way of setting deadlines across different framework components. I don't think it's feasibly to do this without thread locals since we aren't in a fully coroutine world, so this would mean using some coroutine `ThreadContextElement` boilerplate.

misk/src/main/kotlin/misk/Deadline.kt
alecholmes: `ActionScopedProvider` is Misk-specific but it builds on Guice scopes. Each action executes within a Guice scope, which allows Guice to provide instances, like `MiskCaller`, that are relevant only to that scope and shouldn't be injected more widely.

I think that the value provided by `ActionScopedProvider` is treated as a Guice Singleton but I'll confirm that just in case it's not.

`ActionDeadline` being `ActionScoped` means that the same deadline can be injected at multiple points during an action without it being exposed to unrelated actions executing at the same time. It also means that any thread locals for the `ActionDeadline` are cleaned up when the scope ends, avoiding them leaking to future actions executing on the same thread.

swankjesse:
I love it. The propagation is great. The overrides are great. Well thought out, and simple but very powerful! I also think the tradeoff of assuming zero network latency is good enough. Canceling is best-effort, and the extra 2ms of time budget doesn’t hurt.

Other miscellaneous notes:

1. Pass nanoseconds-on-the-wire instead of milliseconds for easier interop with gRPC? Especially sticky because truncation to zero is weird.

2. OkHttp calls permit a full-operation timeout to be configured. OkHttp will cancel the call when it’s reached. Might be a handy building block.
https://square.github.io/okhttp/4.x/okhttp/okhttp3/-call/timeout/

3. Okio’s [AsyncTimeout](https://square.github.io/okio/2.x/okio/okio/-async-timeout/) fires a callback if a configured deadline is exceeded. Useful if you want to call Thread.interrupt() or something. Or it could call [Session.cancelQuery](https://docs.jboss.org/hibernate/orm/5.2/javadocs/org/hibernate/Session.html#cancelQuery--) on a DB query.

4. When a deadline is reached many systems may fail simultaneously, generating multiple exceptions with the same root cause. This might be noisy in our logs. Correlating them would be good.

5. We should also be careful about how we blame exceeded deadlines. If you ask a clientsync to do 100 ms worth of work with a 30 ms budget, it’ll time out. We don’t want to waste engineering effort trying to fix that bug because it isn’t one. That’ll be difficult.

6. Please consider to treat ThreadLocals as a tool that you can use. If/when we put coroutines into Misk we’ll swap out ThreadLocal with something we’ll build like CoroutineLocal. The procedural behavior will be the same.

misk/src/main/kotlin/misk/Deadline.kt
```diff
@@ -0,0 +1,103 @@
...
+import java.time.Clock
+import java.time.Duration
+import java.time.Instant
+import javax.inject.Inject
+
+/**
+ * Deadline for an action. A deadline can be overridden, with the caveat that the overridden deadline
+ * does not propagate to other threads.
+ *
+ * All actions have [ActionDeadline]s available even if there is initially no deadline.
```
swankjesse: As a follow up I’d love our default deadline to be say, 10 seconds. You can specify a longer deadline, but the default is not infinite.

misk/src/main/kotlin/misk/Deadline.kt
```diff
@@ -0,0 +1,103 @@
...
+  }
+
+  /**
+   * Override the deadline for the given block. The overridden deadline only applies to the current
+   * thread.
+   *
+   * The lesser of the given deadline and current deadline is used.
+   */
+  fun <T> overriding(duration: Duration, fn: () -> T): T {
+    check(!duration.isZero && !duration.isNegative) { "duration must be positive" }
```
swankjesse: Should this throw a DeadlineExceededException ?

misk/src/main/kotlin/misk/Deadline.kt
```diff
@@ -0,0 +1,103 @@
...
+
+/**
+ * Provides an optional deadline scoped to the current action, determined by the optional
+ * "X-Request-Timeout-Ms" HTTP header.
+ */
+class ActionDeadlineProvider @Inject constructor(
+  private val clock: Clock,
+  var currentRequest: ActionScoped<HttpCall>
+) : ActionScopedProvider<ActionDeadline> {
+  override fun get(): ActionDeadline {
```
swankjesse: I think you should move this into the web package, which should call `overriding`. That way actions created by other means (jobs!) still participate.

misk/src/main/kotlin/misk/Deadline.kt
```diff
@@ -0,0 +1,103 @@
...
+    val deadline = currentRequest.get().requestHeaders[HTTP_HEADER]?.let {
+      try {
+        val timeoutMs = it.toLong()
+        if (timeoutMs <= 0) {
```
swankjesse:
```
requireRequest(timeoutMs > 0)
```

misk/src/main/kotlin/misk/Deadline.kt
```diff
@@ -0,0 +1,103 @@
...
+ * "X-Request-Timeout-Ms" HTTP header.
+ */
+class ActionDeadlineProvider @Inject constructor(
+  private val clock: Clock,
+  var currentRequest: ActionScoped<HttpCall>
+) : ActionScopedProvider<ActionDeadline> {
+  override fun get(): ActionDeadline {
+    val deadline = currentRequest.get().requestHeaders[HTTP_HEADER]?.let {
+      try {
+        val timeoutMs = it.toLong()
```
swankjesse: I prefer nanos here because calls could in under a millisecond. If you can’t express 0.9 ms, that sucks!

misk/src/main/kotlin/misk/client/ClientDeadlineAppInterceptor.kt
```diff
@@ -0,0 +1,81 @@
...
+  override fun interceptBeginCall(chain: BeginClientCallChain): Call<Any> {
+    val deadline = deadlineProvider.get().get()
+
+    val remaining = deadline.remaining()
+    if (remaining == null) {
+      return chain.proceed(chain.args)
+    }
+    if (remaining.isZero) {
+      // TODO: Different status code?
+      throw ActionException(StatusCode.SERVICE_UNAVAILABLE, "deadline exceeded; skipping call")
```
swankjesse: create DeadlineExceededException so we can better identify these in logs?

misk/src/main/kotlin/misk/client/ClientDeadlineAppInterceptor.kt
```diff
@@ -0,0 +1,81 @@
...
+      return chain.proceed(chain.request)
+    }
+
+    val remaining = deadlineProvider.get().get().remaining()
+    if (remaining == null) {
+      return chain.proceed(chain.request)
+    }
+
+    // TODO(alec): Hacky for the time being. Deadline may have passed between early check.
+    // This could return a 503 here, although that feels slightly weird.
```
swankjesse: Just do it here exclusively? That way you’re only sampling the clock once.

If you want OkHttp to cancel the call for you automatically:

```
chain.call().timeout().deadline(5, TimeUnit.MILLISECONDS)
```

But you’ll need to expose the OkHttp Chain to ClientNetworkChain in NetworkInterceptorWrapper. Or give ClientNetworkChain a timeout.

alecholmes: I like the idea of setting the timeout on the `OkHttp` call but can't figure out a clean way to expose the OkHttp `Call` to Misk network interceptors, since that seems like a hidden implementation detail of the Retrofit `Call`. Any suggestions, or am I missing something?

swankjesse: Might be able to do with the `ClientNetworkInterceptor`? That’s the one with access to the network bits.

misk/src/main/kotlin/misk/Deadline.kt
```diff
@@ -0,0 +1,103 @@
...
+      } catch (ex: NumberFormatException) {
+        throw BadRequestException("Invalid header value for $HTTP_HEADER")
+      }
+    }
+
+    return ActionDeadline(clock, deadline)
+  }
+
+  companion object {
+    const val HTTP_HEADER = "X-Request-Timeout-Ms"
```
swankjesse: FYI: https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_filters/router_filter#config-http-filters-router-x-envoy-expected-rq-timeout-ms

misk/src/main/kotlin/misk/client/ClientDeadlineAppInterceptor.kt
ryanhall07: https://github.com/cashapp/misk/pull/1519

chris-ryan-square:

PR #1123, opened 2019-07-29
Allow misk tests to use new database between runs

keeferrourke:
This change introduces a `DatabasePool` which is used by `DataSourceService` to obtain (and under test, allocate) database names.

In production, this changes nothing; we will always use the database that was specified by the config.

In test, each `MiskTest` annotation test will operate on a special database called `database_yyyyMMdd_s`, where `database` is the name specified by the `DataSourceConfig.database` property, `yyyyMMdd` is the current date, and `s` is a sequence number.

e.g. If two tests for a cinema service are running in parallel and they request access to the same database, then the first test will use a database called `movies_20190731_1`, and the second test will use a database called `movies_20190731_2`. These databases are released for usage by other tests that run afterwards, should they require them.

This feature can be opted into by configuring the `HibernateModule` as follows in a testing module:

```kotlin
install(HibernateModule(Movies::class, config, SHARED_TEST_DATABASE_POOL))
```

PR #1134, opened 2019-08-06
Propagate config changes from DataSourceService

keeferrourke:
`DataSourceService` can now modify the config to specify which database to use under test (as of #1123 and #1131), however these changes were not propagated to other services that were concerned with the current database under test (e.g. SchemaMigratorService, TruncateTablesService, SessionFactoryService, etc).

This change makes it so that the database name assigned by DataSourceService is actually used by other services in testing.

### DIFF
--- a/misk-hibernate-testing/src/main/kotlin/misk/jdbc/TruncateTablesService.kt
+++ b/misk-hibernate-testing/src/main/kotlin/misk/jdbc/TruncateTablesService.kt
@@ -23,7 +23,7 @@ private val logger = getLogger<TruncateTablesService>()
  */
 class TruncateTablesService(
   private val qualifier: KClass<out Annotation>,
-  private val config: DataSourceConfig,
+  private val connector: DataSourceConnector,
   private val transacterProvider: Provider<Transacter>,
   private val startUpStatements: List<String> = listOf(),
   private val shutDownStatements: List<String> = listOf()
@@ -45,6 +45,7 @@ class TruncateTablesService(
     val truncatedTableNames = transacterProvider.get().shards().flatMap { shard ->
       transacterProvider.get().transaction(shard) { session ->
         session.withoutChecks {
+          val config = connector.config()
           val tableNamesQuery = when (config.type) {
             DataSourceType.MYSQL -> {
               "SELECT table_name FROM information_schema.tables where table_schema='${config.database}'"
--- a/misk-hibernate-testing/src/test/kotlin/misk/jdbc/TruncateTablesServiceTest.kt
+++ b/misk-hibernate-testing/src/test/kotlin/misk/jdbc/TruncateTablesServiceTest.kt
@@ -24,9 +24,9 @@ internal class TruncateTablesServiceTest {
   @MiskTestModule
   val module = TestModule()
 
-  @Inject @TestDatasource lateinit var config: DataSourceConfig
   @Inject @TestDatasource lateinit var sessionFactory: SessionFactory
   @Inject @TestDatasource lateinit var transacter: Transacter
+  @Inject @TestDatasource lateinit var connector: DataSourceConnector
 
   @BeforeEach
   internal fun setUp() {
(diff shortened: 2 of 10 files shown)


misk-hibernate-testing/src/main/kotlin/misk/jdbc/TestDatabasePool.kt
```diff
@@ -32,7 +33,7 @@ val SHARED_TEST_DATABASE_POOL = TestDatabasePool(
  *
  * Thread-safe.
  */
-class TestDatabasePool(
+class TestDatabasePool @Inject constructor(
```
swankjesse: I don’t think we wanted this created by Guice?

misk-hibernate-testing/src/main/kotlin/misk/jdbc/TruncateTablesService.kt
```diff
@@ -23,13 +23,15 @@ private val logger = getLogger<TruncateTablesService>()
...
   private val qualifier: KClass<out Annotation>,
-  private val config: DataSourceConfig,
+  internal val connector: DataSourceConnector,
   private val transacterProvider: Provider<Transacter>,
   private val startUpStatements: List<String> = listOf(),
   private val shutDownStatements: List<String> = listOf()
 ) : AbstractIdleService() {
   private val persistentTables = setOf("schema_version")
 
+  private val config get() = connector.config()
```
swankjesse: I’d avoid this because it doesn’t act like a field. (you get a different value if you access it before startup, right?)

misk-hibernate-testing/src/test/kotlin/misk/jdbc/TruncateTablesServiceTest.kt
```diff
@@ -109,12 +110,9 @@ internal class TruncateTablesServiceTest {
...
       val config = MiskConfig.load<TestConfig>("test_truncatetables_app", Environment.TESTING)
-      install(HibernateModule(
-          TestDatasource::class, config.data_source))
-      install(object : HibernateEntityModule(
-          TestDatasource::class) {
-        override fun configureHibernate() {
-        }
+      install(HibernateModule(TestDatasource::class, config.data_source))
+      install(object : HibernateEntityModule(TestDatasource::class) {
+        override fun configureHibernate() {}
```
swankjesse: hmm. . . . I thought our Spotless forced `{}` to be on two lines. (Special cases are bad)

misk-hibernate/src/main/kotlin/misk/hibernate/HibernateModule.kt
```diff
@@ -99,6 +101,9 @@ class HibernateModule(
           metrics = metrics
       )
     }).asSingleton()
+    val dataSourceServiceProvider = getProvider(keyOf<DataSourceService>(qualifier))
+    bind(keyOf<DataSourceConnector>(qualifier)).toProvider(dataSourceServiceProvider).asSingleton()
```
swankjesse: neat

swankjesse: I’d omit `.asSingleton()` since it already inherits singleton-ness from the DataSourceService

misk-hibernate/src/main/kotlin/misk/hibernate/SchemaMigratorService.kt
```diff
@@ -5,21 +5,28 @@ import com.google.common.util.concurrent.Service
...
 import misk.healthchecks.HealthStatus
+import misk.jdbc.DataSourceConnector
 import misk.jdbc.DataSourceType
+import javax.inject.Provider
 import kotlin.reflect.KClass
 
 class SchemaMigratorService internal constructor(
   private val qualifier: KClass<out Annotation>,
   private val environment: Environment,
-  private val schemaMigratorProvider: javax.inject.Provider<SchemaMigrator>, // Lazy!
```
swankjesse: thanks!

misk-hibernate-testing/src/main/kotlin/misk/jdbc/TestDatabasePool.kt
keeferrourke: Reverted

misk-hibernate-testing/src/test/kotlin/misk/jdbc/TruncateTablesServiceTest.kt
keeferrourke: I'll fix it. Didn't realize that was against this was against the style guide.

misk-hibernate-testing/src/main/kotlin/misk/jdbc/TruncateTablesService.kt
keeferrourke: I suppose that `private val config = connection.config()` is preferred then? Will fix.

PR #1135, opened 2019-08-07
Cast ids to misk.hibernate.Id instead of Long in BulkShardMigrator

surrahman:
We've been seeing `ClassCastException`s in our logs for the `BulkShardMigrator`: `class  misk.hibernate.Id cannot be cast to class java.lang.Long (misk.hibernate.Id is in unnamed module of loader 'app'; java.lang.Long is in module java.base of loader 'bootstrap')`. This was happening because we were attempting to cast `misk.hibernate.Id` to a `Long`. This change corrects that by first casting to a `misk.hibernate.Id` and then calling `.id` to get the required `Long` id.


misk-hibernate/src/main/kotlin/misk/hibernate/migrate/BulkShardMigrator.kt
```diff
@@ -206,10 +206,9 @@ class BulkShardMigrator<R : DbRoot<R>, C : DbChild<R, C>> private constructor(
             "SELECT id FROM $tableName WHERE $rootColumnName = :target AND id IN (:ids)"
         ).setParameter("target", targetRoot!!.id)
           .setParameterList("ids", sourceRecords.keys)
-          .list()
-          .stream()
-          .map { it as Long }
-          .collect(Collectors.toSet<Long>())
+            .resultList
```
adrw: Is `resultList` short for `.list().stream()`?

misk-hibernate/src/main/kotlin/misk/hibernate/migrate/BulkShardMigrator.kt
```diff
@@ -206,10 +206,9 @@ class BulkShardMigrator<R : DbRoot<R>, C : DbChild<R, C>> private constructor(
...
+            .map { (it as Id<*>).id }
```
adrw: Good fix 👍

misk-hibernate/src/main/kotlin/misk/hibernate/migrate/BulkShardMigrator.kt
surrahman: I believe so .. feels more _kotlinish_ than the clumsy java-style `.list().stream()`.

wesleyk: can you fix/align indenting here?

wesleyk:
any test you can add/update corresponding to this fix?

surrahman:
We're not sure what's causing this failure .. it happens sometimes and there's nothing in the database or logs to suggest a difference between inputs for successful and failed merges, other than the attached [stacktrace](https://github.com/cashapp/misk/files/3478907/stacktrace.txt). We're going to deploy this and see if it fixes the error. If not, we'll revert and continue investigating.

PR #1137, opened 2019-08-08
Allow to set vitess docker image via environment variable when starti…

kalfonso:
Allow to set the Vitess docker image via environment variable when starting Vitess services

This helps testing some vitess changes via Misk services


misk-hibernate/src/main/kotlin/misk/vitess/StartVitessService.kt
```diff
@@ -207,7 +207,9 @@ class DockerVitessCluster(
     if (containerId == null) {
       StartVitessService.logger.info(
           "Starting Vitess cluster with command: ${cmd.joinToString(" ")}")
-      containerId = docker.createContainerCmd(VITESS_IMAGE)
+      val vitessImage: String = System.getenv("VITESS_IMAGE") ?: VITESS_IMAGE
```
tirsen: You also have to change the "docker pull" part above

kalfonso:
I'll just change the hardcoded version in Misk to my local one momentarily for testing

PR #1139, opened 2019-08-09
exclude log4j from transitive dependencies of curator

mightyguava:
Don't know if this does what I want. I want misk to not include
`slf4j-log4j12`, which comes from curator:

```
 |    +--- org.apache.curator:curator-framework:4.0.1
|    |    \--- org.apache.curator:curator-client:4.0.1
|    |         +--- org.apache.zookeeper:zookeeper:3.5.3-beta -> 3.5.4-beta
|    |         |    +--- org.slf4j:slf4j-api:1.7.25
|    |         |    +--- org.slf4j:slf4j-log4j12:1.7.25
```

I also want any apps that depend on both Misk and Curator, to
automatically exclude `slf4j-log4j12`, without needing a similar clause
to what I'm adding here in the app's build.gradle.

Relevant Zookeeper ticket that seems to be abandoned https://issues.apache.org/jira/browse/ZOOKEEPER-1371

### DIFF
--- a/misk-zookeeper/build.gradle
+++ b/misk-zookeeper/build.gradle
@@ -7,7 +7,9 @@ buildscript {
 apply plugin: 'kotlin-jpa'
 
 dependencies {
-  compile dep.curatorFramework
+  compile(dep.curatorFramework) {
+    exclude(group: 'org.slf4j', module: 'slf4j-log4j12')
+  }
   compile project(':misk')
   compile dep.zookeeper
 


misk-zookeeper/build.gradle
```diff
@@ -7,7 +7,9 @@ buildscript {
 apply plugin: 'kotlin-jpa'
 
 dependencies {
-  compile dep.curatorFramework
+  compile(dep.curatorFramework) {
```
keeferrourke: These `compile` declarations should be `api` (same thing, but newer).

I believe this does what you want: `dep.curatorFramework` should be made available to misk apps through this (transitive) dependency. Misk apps that use _shouldn't_ need to directly depend on `dep.curatorFramework` -- instead they can use the version supplied by Misk.

keeferrourke: If misk app _do_ depend on curatorFramework explicitly, they'll need to do the same exclusion. Explicitly depending on curatorFramework should be discouraged though.

mightyguava: > If misk app do depend on curatorFramework explicitly, they'll need to do the same exclusion.

Ideally they would not need this exclusion. But I guess we can live with this.

keeferrourke: They can avoid depending on curatorFramework entirely though!

```gradle
// ZookeeperApp
dependencies {
  api dep.misk // <- has api.curatorFramework, so curatorFramework is available to this app for usage
  ...
}
```

PR #1141, opened 2019-08-09
exclude log4j from transitive dependencies of curator (try again)

mightyguava:
The previous attempt didn't prune everything https://github.com/cashapp/misk/pull/1139

### DIFF
--- a/misk-zookeeper/build.gradle
+++ b/misk-zookeeper/build.gradle
@@ -9,13 +9,20 @@ apply plugin: 'kotlin-jpa'
 dependencies {
   compile(dep.curatorFramework) {
     exclude(group: 'org.slf4j', module: 'slf4j-log4j12')
+    exclude(group: 'log4j', module: 'log4j')
   }
   compile project(':misk')
-  compile dep.zookeeper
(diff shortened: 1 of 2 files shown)


misk-zookeeper/build.gradle
```diff
@@ -9,13 +9,20 @@ apply plugin: 'kotlin-jpa'
...
-  compile dep.zookeeper
+  compile(dep.zookeeper) {
+    exclude(group: 'org.slf4j', module: 'slf4j-log4j12')
+    exclude(group: 'log4j', module: 'log4j')
+  }
 
   testCompile dep.docker
   testCompile dep.junitApi
-  testCompile dep.zookeeper
+  testCompile(dep.zookeeper) {
```
keeferrourke: `testCompile` includes the configurations from `compile` does it not?

mightyguava: the excludes don't propagate

mightyguava: so, I'm still not totally sure this'll work, but hopeful

PR #1142, opened 2019-08-09
[CLOSES #1129] DashboardTabProvider

### LINKED ISSUE #1129: Set authentication in AdminDashboardTab bindings for Config and Web-Actions tab from the bound AdminDashboardAccess

adrw:
* New pattern of multibinding tabs to a DashboardTabProvider along with an AccessAnnotation
* Allows for Misk tabs to be set with the authentication from the multibound `AdminDashboardAccess` access annotation
* Also allows for service tabs to be easily set with service specific Access Annotations

* Fixes small bug in Loader tab where dashboard doesn't load when no tabs are bound


adrw:
Thanks to @mmihic for help with Guice Provider pattern!

misk/src/main/kotlin/misk/web/WebTab.kt
```diff
@@ -10,22 +10,11 @@ abstract class WebTab(
...
-  fun isAuthenticated(caller: MiskCaller?): Boolean {
-    return when {
-      capabilities.isEmpty() && services.isEmpty() -> true // no capabilities/service requirement => unauthenticated requests allowed (including when caller is null)
-      caller == null -> false // capability/service requirement present but caller null => assume authentication broken
-      capabilities.any { caller.capabilities.contains(it) } -> true // matching capability
-      services.any { caller.service == it } -> true // matching service
-      else -> false
-    }
+  fun isAuthenticated(caller: MiskCaller?): Boolean = when {
+    capabilities.isEmpty() && services.isEmpty() -> true // no capabilities/service requirement => unauthenticated requests allowed (including when caller is null)
```
swankjesse: nit for next time: if ya gonna reformat, wrap to 100 please

misk/src/main/kotlin/misk/web/metadata/AdminDashboardModule.kt
```diff
@@ -30,63 +31,66 @@ class AdminDashboardModule(val environment: Environment) : KAbstractModule() {
...
+      slug = "loader",
+      web_proxy_url = "http://localhost:3100/"
     ))
     install(WebTabResourceModule(
-        environment = environment,
-        slug = "loader",
-        web_proxy_url = "http://localhost:3100/",
-        url_path_prefix = "/_admin/",
-        resourcePath = "classpath:/web/_tab/loader/"
+      environment = environment,
```
swankjesse: this reformatting is wrong

misk/src/test/kotlin/misk/web/actions/AdminDashboardTabActionTest.kt
```diff
@@ -0,0 +1,74 @@
...
+  val module = AdminDashboardActionTestingModule()
+
+  @Inject private lateinit var jetty: JettyService
+  @Inject private lateinit var httpClientFactory: HttpClientFactory
+  @Inject private lateinit var adminDashboardTabAction: AdminDashboardTabAction
+
+  val path = "/api/admindashboardtabs"
+  val empty = "{\"adminDashboardTabs\":[]}"
+  val tabs =
+    "{\"adminDashboardTabs\":[{\"slug\":\"config\",\"url_path_prefix\":\"/_admin/config/\",\"name\":\"Config\",\"category\":\"Container Admin\",\"capabilities\":[\"admin_access\"],\"services\":[]},{\"slug\":\"web-actions\",\"url_path_prefix\":\"/_admin/web-actions/\",\"name\":\"Web Actions\",\"category\":\"Container Admin\",\"capabilities\":[\"admin_access\"],\"services\":[]}]}"
```
swankjesse: pretty print this JSON and use a multiline string?

misk/src/test/kotlin/misk/web/actions/DashboardTabTest.kt
```diff
@@ -54,20 +26,23 @@ internal class DashboardTabTest {
 
   @Test
   internal fun tabGoodCategory() {
-    DashboardTab(slug = "slug", url_path_prefix = "/a/path/", name = "Name", category = "@tea-pot_418")
+    DashboardTab(slug = "slug", url_path_prefix = "/a/path/", name = "Name",
+      category = "@tea-pot_418")
```
swankjesse: wrap at +4!

misk/src/test/kotlin/misk/web/actions/AdminDashboardTabActionTest.kt
swankjesse: and then you can assert on the object, not on its string representation?
```
val moshi = Moshi.Builder().build()
val responseAdapter = moshi.adapter<AdminDashboardTabAction.Response>()
return responseAdapter.fromJson(response.body!!.source())
```

misk/src/main/kotlin/misk/web/metadata/AdminDashboardModule.kt
adrw: Our KtLint configuration changes the `continuation_indent_size` from the Kotlin default 4 to 2:

```
ktlint(dep.ktlintVersion).userData(['indent_size': '2', 'continuation_indent_size' : '2'])
```

misk/src/test/kotlin/misk/web/actions/AdminDashboardTabActionTest.kt
adrw: Perfect! I wanted to do this but I didn't know how 👍

misk/src/main/kotlin/misk/web/WebTab.kt
```diff
@@ -10,22 +10,19 @@ abstract class WebTab(
...
 ) : ValidWebEntry(slug = slug, url_path_prefix = url_path_prefix) {
-  fun isAuthenticated(caller: MiskCaller?): Boolean {
-    return when {
-      capabilities.isEmpty() && services.isEmpty() -> true // no capabilities/service requirement => unauthenticated requests allowed (including when caller is null)
-      caller == null -> false // capability/service requirement present but caller null => assume authentication broken
-      capabilities.any { caller.capabilities.contains(it) } -> true // matching capability
-      services.any { caller.service == it } -> true // matching service
-      else -> false
-    }
+  fun isAuthenticated(caller: MiskCaller?): Boolean = when {
```
wesleyk: any particular reason we have to re-implement our authentication logic for web tabs vs web actions? Would be good to have a single source of truth for what is pretty critical security logic

misk/src/main/kotlin/misk/web/DashboardTab.kt
```diff
@@ -0,0 +1,54 @@
...
+}
+
+inline fun <reified A : Annotation> DashboardTabProvider(
+  slug: String,
+  url_path_prefix: String,
+  name: String,
+  category: String = "Admin"
+) = DashboardTabProvider(
+  slug = slug,
+  url_path_prefix = url_path_prefix,
```
wesleyk: camelcase?

misk/src/main/kotlin/misk/web/DashboardTab.kt
```diff
@@ -0,0 +1,54 @@
...
+import javax.inject.Inject
+import javax.inject.Provider
+import kotlin.reflect.KClass
+
+class DashboardTab(
+  slug: String,
+  url_path_prefix: String,
+  val name: String,
+  val category: String = "Admin",
+  capabilities: Set<String> = setOf(),
```
wesleyk: do we actually want defaults for this?

misk/src/main/kotlin/misk/web/WebTab.kt
adrw: Good point, we can reused the `isAllowed` function from `AccessInterceptor`. Will do so in a follow up PR.

misk/src/main/kotlin/misk/web/DashboardTab.kt
adrw: Snakecase was recommended in early work on tabs to signify that the variable was exposed as part of the API. I'm open to changing it to camelcase if we decide to set that recommendation aside. @swankjesse

misk/src/main/kotlin/misk/web/DashboardTab.kt
adrw: I think so, though I'd be open to removing the default and forcing developers to deliberately choose authentication vs. being open by default. Is that what you'd recommend?

misk/src/main/kotlin/misk/web/DashboardTab.kt
wesleyk: mm gotcha. I guess in my mind that makes sense for the API object itself, which I don't think this provider is

misk/src/main/kotlin/misk/web/DashboardTab.kt
wesleyk: > deliberately choose authentication vs. being open by default

I think being open by default seems like a dangerous default.

PR #1144, opened 2019-08-12
Access Vitess using the MySQL driver

tirsen:
Adds a new VITESS_MYSQL data source type which accesses Vitess using the
mysql protocol and the MySQL JDBC driver

### DIFF
--- a/misk-hibernate/src/main/kotlin/misk/hibernate/SchemaMigratorService.kt
+++ b/misk-hibernate/src/main/kotlin/misk/hibernate/SchemaMigratorService.kt
@@ -22,7 +22,8 @@ class SchemaMigratorService internal constructor(
     val schemaMigrator = schemaMigratorProvider.get()
     val connector = connectorProvider.get()
     if (environment == Environment.TESTING || environment == Environment.DEVELOPMENT) {
-      if (connector.config().type != DataSourceType.VITESS) {
+      val type = connector.config().type
+      if (type != DataSourceType.VITESS && type != DataSourceType.VITESS_MYSQL) {
         val appliedMigrations = schemaMigrator.initialize()
         migrationState = schemaMigrator.applyAll("SchemaMigratorService", appliedMigrations)
       } else {
--- a/misk-hibernate/src/main/kotlin/misk/vitess/StartVitessService.kt
+++ b/misk-hibernate/src/main/kotlin/misk/vitess/StartVitessService.kt
@@ -58,6 +58,7 @@ class VitessCluster(
   val moshi: Moshi = Moshi.Builder().build()
 ) {
   val schemaDir: Path
+  val configDir: Path
 
   init {
     if (config.vitess_schema_dir != null) {
@@ -85,6 +86,15 @@ class VitessCluster(
         schemaDir.toFile().deleteRecursively()
       })
     }
+
+    // Copy out all the resources from the current package
+    // We use the my.cnf configuration file to configure MySQL (e.g. the default time zone)
+    configDir = Paths.get("/tmp/vitess_conf_${System.currentTimeMillis()}")
+    Files.createDirectories(configDir)
+    resourceLoader.copyTo("classpath:/misk/vitess", configDir)
+    Runtime.getRuntime().addShutdownHook(thread(start = false) {
+      configDir.toFile().deleteRecursively()
+    })
   }
 
   val keyspaceAdapter = moshi.adapter<Keyspace>()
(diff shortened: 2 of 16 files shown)


misk-hibernate/src/main/kotlin/misk/hibernate/SchemaMigratorService.kt
```diff
@@ -22,7 +22,7 @@ class SchemaMigratorService internal constructor(
     val schemaMigrator = schemaMigratorProvider.get()
     val connector = connectorProvider.get()
     if (environment == Environment.TESTING || environment == Environment.DEVELOPMENT) {
-      if (connector.config().type != DataSourceType.VITESS) {
+      if (connector.config().type != DataSourceType.VITESS && connector.config().type != DataSourceType.VITESS_MYSQL) {
```
keeferrourke: Extract `connector.config()` to a variable?

misk-hibernate/src/main/kotlin/misk/vitess/StartVitessService.kt
```diff
@@ -58,6 +58,7 @@ class VitessCluster(
   val moshi: Moshi = Moshi.Builder().build()
 ) {
   val schemaDir: Path
+  val confDir: Path
```
keeferrourke: Small nit: I think we use `config` more than we use `conf`?

mightyguava:
Oh is this a new Vitess feature? How well has it been tested?

misk-hibernate/src/test/resources/moviestestmodule-common.yaml
```diff
@@ -1,8 +1,13 @@
-data_source:
+vitess_data_source:
   type: VITESS
   username: vt_app
   password: whatever
   vitess_schema_resource_root: classpath:/misk/hibernate/moviestestmodule-schema
+vitess_mysql_data_source:
+  type: VITESS_MYSQL
```
mightyguava: our configs are auto-provisioned, how were you envisioning this to work? Generate another copy of each db for VITESS_MYSQL?

tirsen:
@mightyguava Nope it's not new at all. It's been around for years. They actually initially built it because we (Square) asked for it. Ironically *we* never ended up using it but pretty much everyone else is using it. Slack uses the mysql protocol for all their production usage. (It's the best way to connect to Vitess via PHP.)

misk-hibernate/src/test/resources/moviestestmodule-common.yaml
tirsen: Yeah it will be a gradual roll out. I imagine we add an option to the hcl which makes it generate `VITESS_MYSQL` config instead of `VITESS` and then we roll it one app the at the time and then eventually make it the default. It should be completely transparent to the app. You shouldn't even need to recompile (except to include this version of misk of course).

But first of all I want to make some scale/perf testing on this using Arnhem. It's still somewhat experimental at this stage I should say.

misk-hibernate/src/test/kotlin/misk/hibernate/MoviesTestModule.kt
```diff
@@ -33,14 +34,17 @@ class MoviesTestModule(
...
   private fun selectDataSourceConfig(config: MoviesConfig): DataSourceConfig {
-    return if (useVitess)
-      config.data_source
-    else
-      config.mysql_data_source
+    return when (type) {
+      DataSourceType.VITESS -> config.vitess_data_source
+      DataSourceType.VITESS_MYSQL -> config.vitess_mysql_data_source
+      DataSourceType.MYSQL -> config.mysql_data_source
+      DataSourceType.HSQLDB -> throw RuntimeException("Not supported (yet?)")
```
swankjesse: We should remove it

misk-hibernate/src/test/kotlin/misk/hibernate/TimestampListenerTest.kt
```diff
@@ -67,11 +67,29 @@ class TimestampListenerTest {
...
+                .isEqualTo(clock.instant().toEpochMilli())
           }
         }
       }
     }
   }
 }
+
+@MiskTest(startService = true)
+class MySQLTimestampListenerTest : TimestampListenerTest() {
```
swankjesse: supernit: MySql: https://google.github.io/styleguide/javaguide.html#s5.3-camel-case

tirsen: oh right I keep forgetting :-)

misk-hibernate/src/test/kotlin/misk/hibernate/MoviesTestModule.kt
tirsen: yeah probably

PR #1146, opened 2019-08-12
Remove duplicate authentication logic in WebTab

adrw:
https://github.com/cashapp/misk/pull/1142#discussion_r312961702


misk/src/main/kotlin/misk/security/authz/AccessInterceptor.kt
```diff
@@ -101,5 +90,16 @@ class AccessInterceptor private constructor(
 
   companion object {
     val logger = getLogger<AccessInterceptor>()
+
+    internal fun isAllowed(caller: MiskCaller, allowedCapabilities: Set<String>, allowedServices: Set<String>): Boolean {
```
wesleyk: how do you feel about this being a method on MiskCaller?

adrw: I think that'd make sense! 👍

Issue #1149, opened 2019-08-13
Avoid SHOW WARNINGS Hibernate misfeature


swankjesse:
Apparently Hibernate’s default is to do an extra DB roundtrip to look for potential warnings. We should turn this off because it doesn’t actually show useful warnings in practice.

https://stackoverflow.com/questions/44211525/how-do-i-disable-show-warnings-from-hibernate

PR #1151, opened 2019-08-13
Log org.hibernate at ERROR

maniksurtani:
Re: #1149 

@swankjesse I presume this would work as a sensible default unless overridden in app-specific configs?

### DIFF
--- a/misk/src/main/resources/log4j2.yaml
+++ b/misk/src/main/resources/log4j2.yaml
@@ -10,7 +10,9 @@ Configuration:
       -
         name: misk
         level: debug
-
+      -
+        name: org.hibernate
+        level: error
     Root:
       level: info
       AppenderRef:


maniksurtani:
slf4j is a log API, not a logger. You'd still need a logging engine behind it. Log4j2 is the newest, with all the benefits of Logback plus some other niceties including [minimal GC impact](http://home.apache.org/~rpopma/log4j/2.6/garbagefree.html). There's almost no point in even talking about the original log4j, which is so outdated.

Hibernate uses JBoss Logging - another log bridge - that plays nice with most major logging engines, including Log4j2.

swankjesse:
@maniksurtani so if I’m reading this right, we should:

1. Migrate from logback to log4j2 as a backend
2. Continue to use slf4j as a frontend

maniksurtani:
Yep.

mightyguava:
It doesn't make a difference to me, as long as we don't install 2 logging backends at the same time (that's what I removed recently), and as long as we end up logging JSON.

swankjesse:
https://github.com/cashapp/misk/issues/1153

PR #1152, opened 2019-08-14
disable SchemaValidatorService

mightyguava:
`SchemaValidatorService` lacks forwards compatibility. Adding a column to a table causes the service to fail to start. This caused a production outage when a service was mid-migration with no way to roll forward or back.

Let's disable temporarily until a fix has been made.


swankjesse:
Can you paste the schema validation error? I was under the impression we’d tested forward-compatibility cases but obviously missed some.

mightyguava:
@swankjesse circleci is failing on `annotationWithOverrides$misk() - misk.web.actions.ReflectTest`, it does not reproduce locally. You added it 2 weeks ago from the git history.

I put the code back in `if (false)`. I tend to delete code that we want to disable temporarily (who knows how temporary things really are), since version control will keep everything anyways!

mightyguava:
Changing to not installing the service rather than disabling its execution path, per chat with @swankjesse

PR #1154, opened 2019-08-15
Add MkDocs Material + Dokka powered docs site

adrw:
* Inspired liberally by existing MkDocs sites in OkHttp and Barber
* MkDocs really chugs generating the site for this many Dokka generated docs so it takes some patience if you want to build the docs site locally (2-5 mins to build instead of near instantly with other repos)
* There are numerous warnings from MkDocs of dead links in the Dokka generated content. This PR makes no changes to the existing docs in Misk projects. Future PRs should.

Screen Shot 2019-08-15 at 00 09 35
Screen Shot 2019-08-15 at 00 15 07


misk-gcp/build.gradle
```diff
@@ -14,3 +14,10 @@ dependencies {
   testCompile project(':misk-testing')
   testCompile project(':misk-gcp-testing')
 }
+
+afterEvaluate { project ->
```
wesleyk: can we avoid having to set this in every sub project?

adrw: I tried numerous ways to move it to the root `build.gradle` without success, though I agree that'd be best. OkHttp and Barber also seem to have the same stanza copy-pasta in all of the subprojects.

PR #1156, opened 2019-08-16
Allow disabling maxRows check in query

keeferrourke:
Sometimes you do want to have an unbounded query, and logging a warning is too noisy.

This change makes it possible to disable the max row count check on per-query basis.


swankjesse:
https://github.com/cashapp/misk/issues/1157

Issue #1159, opened 2019-08-17
Tests: ReflectTest.annotationWithOverrides() is flaky


swankjesse:
https://circleci.com/gh/cashapp/misk/2339?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

```
org.junit.ComparisonFailure: expected:<"[square]"> but was:<"[territory]">
	at java.base/jdk.internal.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
	at java.base/jdk.internal.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
	at misk.web.actions.ReflectTest.annotationWithOverrides$misk(ReflectTest.kt:53)
```

PR #1160, opened 2019-08-17
Prefer non-synthetic methods in findAnnotationWithOverrides

swankjesse:
Closes: https://github.com/cashapp/misk/issues/1159

### DIFF
--- a/misk/src/main/kotlin/misk/web/actions/reflect.kt
+++ b/misk/src/main/kotlin/misk/web/actions/reflect.kt
@@ -35,7 +35,8 @@ internal fun Method.overrides(): Set<Method> {
 internal fun Class<*>.getOverriddenMethod(override: Method): Method? {
   return try {
     check(this.isAssignableFrom(override.declaringClass))
-    getDeclaredMethod(override.name, *override.parameterTypes)
+    val overridden = getDeclaredMethod(override.name, *override.parameterTypes)
+    return overridden.preferNonSynthetic()
   } catch (_: NoSuchMethodException) {
     null
   }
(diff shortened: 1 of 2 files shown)


misk/src/main/kotlin/misk/web/actions/reflect.kt
```diff
@@ -61,3 +62,45 @@ internal fun Class<*>.superclasses(): Set<Class<*>> {
...
+ * public Point next()
+ * public synthetic bridge Object next()
+ * ```
+ *
+ * We prefer non-synthetic methods for reflection because that's where annotations will be.
+ */
+internal fun Method.preferNonSynthetic(): Method {
+  if (!isSynthetic) return this
+
+  for (peer in declaringClass.methods) {
```
adrw: Is there a case where there are multiple non-synthetic methods returned? Or is there only ever two choices: synthetic + non-synthetic?

misk/src/main/kotlin/misk/web/actions/reflect.kt
```diff
@@ -61,3 +62,45 @@ internal fun Class<*>.superclasses(): Set<Class<*>> {
   }
   return result
 }
+
+/**
+ * Attempts to find a non-synthetic method with the same declaring class, name, and parameters. If
```
adrw: Thanks for the docs, helps to understand the purpose (and flakiness without) this code!

misk/src/main/kotlin/misk/web/actions/reflect.kt
swankjesse: there can be multiple synthetic methods!

misk/src/main/kotlin/misk/web/actions/reflect.kt
```diff
@@ -61,3 +62,45 @@ internal fun Class<*>.superclasses(): Set<Class<*>> {
...
+ *
+ * ```
+ * public boolean hasNext()
+ * public Point next()
+ * public synthetic bridge Object next()
+ * ```
+ *
+ * We prefer non-synthetic methods for reflection because that's where annotations will be.
+ */
+internal fun Method.preferNonSynthetic(): Method {
```
tomjadams: nit: This name reads to me like it should have a boolean return type, would a better name be `nonSyntheticIfAvailable`? I don't think that's much better though TBH :)

PR #1163, opened 2019-08-19
Suggest endpoints in NotFoundAction response

### LINKED ISSUE #1162: Provide more context when the caller’s Content-Type is wrong
It’s super frustrating to call endpoints with the wrong Content-Type and get a 404. We should provide more context like so:

```
No such resource.

Received:
  POST /foo/bar
  Content-Type: none specified

Alternative:
  GET /foo/bar

Alternative:
  POST /foo/bar
  Content-Type: application/json
```

keeferrourke:
Closes #1162

### DIFF
--- a/misk/src/main/kotlin/misk/web/BoundAction.kt
+++ b/misk/src/main/kotlin/misk/web/BoundAction.kt
@@ -57,11 +57,22 @@ internal class BoundAction<A : WebAction>(
     val requestCharsetMatch = requestContentTypeMatch?.matchesCharset ?: false
 
     return BoundActionMatch(
-        this,
-        pathMatcher,
-        acceptedMediaRange,
-        requestCharsetMatch,
-        action.responseContentType ?: MediaTypes.ALL_MEDIA_TYPE
+        action = this,
+        pathMatcher = pathMatcher,
+        acceptedMediaRange = acceptedMediaRange,
+        requestCharsetMatch = requestCharsetMatch,
+        responseContentType = action.responseContentType ?: MediaTypes.ALL_MEDIA_TYPE
+    )
+  }
+
+  fun matchByUrl(url: HttpUrl): BoundActionMatch? {
+    val patchMather = pathPattern.matcher(url) ?: return null
+    return BoundActionMatch(
+        action = this,
+        pathMatcher = patchMather,
+        acceptedMediaRange = MediaRange.ALL_MEDIA,
+        requestCharsetMatch = false,
+        responseContentType = MediaTypes.ALL_MEDIA_TYPE
     )
   }
 
@@ -108,7 +119,8 @@ internal class BoundAction<A : WebAction>(
         dispatchMechanism = action.dispatchMechanism,
         allowedServices = fetchAllowedCallers(
             applicationInterceptors, AccessInterceptor::allowedServices),
-        allowedCapabilities = fetchAllowedCallers(applicationInterceptors, AccessInterceptor::allowedCapabilities)
+        allowedCapabilities = fetchAllowedCallers(applicationInterceptors,
+            AccessInterceptor::allowedCapabilities)
     )
   }
 
--- a/misk/src/main/kotlin/misk/web/actions/NotFoundAction.kt
+++ b/misk/src/main/kotlin/misk/web/actions/NotFoundAction.kt
@@ -1,29 +1,41 @@
 package misk.web.actions
 
 import misk.logging.getLogger
+import misk.scope.ActionScoped
 import misk.security.authz.Unauthenticated
 import misk.web.Get
+import misk.web.HttpCall
 import misk.web.PathParam
 import misk.web.Post
 import misk.web.RequestContentType
 import misk.web.Response
 import misk.web.ResponseBody
 import misk.web.ResponseContentType
+import misk.web.jetty.WebActionsServlet
 import misk.web.mediatype.MediaTypes
 import misk.web.toResponseBody
 import okhttp3.Headers.Companion.headersOf
 import javax.inject.Inject
+import javax.inject.Provider
 import javax.inject.Singleton
 
 @Singleton
-class NotFoundAction @Inject constructor() : WebAction {
+class NotFoundAction @Inject constructor(
+  @JvmSuppressWildcards private val clientHttpCall: ActionScoped<HttpCall>
+) : WebAction {
+  @Inject internal lateinit var servletProvider: Provider<WebActionsServlet>
+
   @Get("/{path:.*}")
   @Post("/{path:.*}")
   @RequestContentType(MediaTypes.ALL)
   @ResponseContentType(MediaTypes.ALL)
   @Unauthenticated
   fun notFound(@PathParam path: String): Response<ResponseBody> {
-    return response(path)
+    return Response(
+        body = notFoundMessage(path, clientHttpCall.get()).toResponseBody(),
+        headers = headersOf("Content-Type", MediaTypes.TEXT_PLAIN_UTF8),
+        statusCode = 404
+    )
   }
 
   companion object {
@@ -36,6 +48,37 @@ class NotFoundAction @Inject constructor() : WebAction {
           headers = headersOf("Content-Type", MediaTypes.TEXT_PLAIN_UTF8),
           statusCode = 404
       )
+    }
+  }
+
+  /** Generates a message for a client that hints what may have been wrong with their request. */
+  private fun notFoundMessage(path: String, httpCall: HttpCall) = buildString {
+    val boundActions = servletProvider.get().boundActions
+    val alternativeActions = boundActions.mapNotNull {
+      it.matchByUrl(httpCall.url)
+    }.filterNot { it.action.pathPattern.matchesWildcardPath }
+
+    append("Nothing found at /$path.\n\n")
+    append("Received:\n")
+    append("${httpCall.dispatchMechanism.method} /$path\n")
+    for (accept in httpCall.accepts()) {
+      append("Accept: $accept\n")
+    }
+    val contentType = httpCall.contentType()
+    if (contentType != null) {
+      append("Content-Type: ${httpCall.contentType()}\n")
+    }
+
+    for (i in 0 until alternativeActions.size) {
+      append("\n")
+      val alternative = alternativeActions[i]
+      val metadata = alternative.action.metadata
+      append("Alternative:\n")
+      append("${metadata.dispatchMechanism.method} ${metadata.pathPattern}\n")
+      append("Accept: ${metadata.responseMediaType}\n")
+      for (requestContentType in metadata.requestMediaTypes) {
+        append("Content-Type: $requestContentType\n")
+      }
     }
   }
 }
--- a/misk/src/test/kotlin/misk/web/actions/NotFoundActionTest.kt
+++ b/misk/src/test/kotlin/misk/web/actions/NotFoundActionTest.kt
@@ -4,6 +4,10 @@ import com.squareup.moshi.Moshi
 import misk.inject.KAbstractModule
 import misk.testing.MiskTest
 import misk.testing.MiskTestModule
+import misk.web.Post
+import misk.web.RequestBody
+import misk.web.RequestContentType
+import misk.web.ResponseContentType
 import misk.web.WebActionModule
 import misk.web.WebTestingModule
 import misk.web.jetty.JettyService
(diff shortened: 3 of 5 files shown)


misk/src/main/kotlin/misk/web/BoundAction.kt
```diff
@@ -65,6 +65,17 @@ internal class BoundAction<A : WebAction>(
     )
   }
 
+  fun matchByUrl(url: HttpUrl): BoundActionMatch? {
+    val patchMather = pathPattern.matcher(url) ?: return null
+    return BoundActionMatch(
+        this,
```
adrw: Consider using Kotlin named paramters so it's more clear at call site.

misk/src/main/kotlin/misk/web/actions/NotFoundAction.kt
```diff
@@ -1,29 +1,41 @@
...
   @Post("/{path:.*}")
   @RequestContentType(MediaTypes.ALL)
   @ResponseContentType(MediaTypes.ALL)
   @Unauthenticated
   fun notFound(@PathParam path: String): Response<ResponseBody> {
-    return response(path)
+    return Response(
+        body = notFoundMessage(path, clientHttpCall.get()).toResponseBody(),
+        headers = headersOf("Content-Type", MediaTypes.TEXT_PLAIN_UTF8),
+        statusCode = 404
```
adrw: Was NotFoundAction not returning a `404` before?

misk/src/main/kotlin/misk/web/actions/NotFoundAction.kt
```diff
@@ -38,4 +50,40 @@ class NotFoundAction @Inject constructor() : WebAction {
       )
     }
   }
+
+  /** Generates a message for a client that hints what may have been wrong with their request. */
+  private fun notFoundMessage(path: String, httpCall: HttpCall) = buildString {
```
adrw: Nice use of `buildString`

misk/src/main/kotlin/misk/web/actions/NotFoundAction.kt
```diff
@@ -38,4 +50,40 @@ class NotFoundAction @Inject constructor() : WebAction {
...
+    val boundActions = servletProvider.get().boundActions
+    val alternativeActions = boundActions.mapNotNull {
+      it.matchByUrl(httpCall.url)
+    }.filterNot { it.action.pathPattern.matchesWildcardPath }
+
+    append("Nothing found at /$path.\n\n")
+    append("Received:\n")
+    append("  ${httpCall.dispatchMechanism.method} /$path\n")
+    for (accept in httpCall.accepts()) {
+      append("  Accept: $accept\n")
```
acostama: nit: have the `Accept` line follow the `Content-Type` as in the alternative (or the other way around)

misk/src/main/kotlin/misk/web/actions/NotFoundAction.kt
```diff
@@ -38,4 +50,40 @@ class NotFoundAction @Inject constructor() : WebAction {
...
+  /** Generates a message for a client that hints what may have been wrong with their request. */
+  private fun notFoundMessage(path: String, httpCall: HttpCall) = buildString {
+    val boundActions = servletProvider.get().boundActions
+    val alternativeActions = boundActions.mapNotNull {
+      it.matchByUrl(httpCall.url)
+    }.filterNot { it.action.pathPattern.matchesWildcardPath }
+
+    append("Nothing found at /$path.\n\n")
+    append("Received:\n")
+    append("  ${httpCall.dispatchMechanism.method} /$path\n")
```
adrw: Consider an explicit indentation character like `\t` or the following which only need to be set once vs. manually adding spaces.

- Increase indent: `⇥`
- Decrease indent: `⇤`

misk/src/test/kotlin/misk/web/actions/NotFoundActionTest.kt
```diff
@@ -85,6 +89,61 @@ class NotFoundActionTest {
...
+      |  Accept: */*
+      |  Content-Type: application/weird; charset=utf-8
+      |
+      |Alternative:
+      |  POST /echo
+      |  Content-Type: text/plain;charset=utf-8
+      |  Accept: text/plain;charset=utf-8
+      |""".trimMargin())
+  }
+  @Test fun responseMessageSuggestsAcceptType() {
```
adrw: nit: add new line above.

misk/src/test/kotlin/misk/web/actions/NotFoundActionTest.kt
```diff
@@ -85,6 +89,61 @@ class NotFoundActionTest {
     assertThat(response.code).isEqualTo(404)
   }
 
+  @Test fun responseMessageSuggestsAlternativeMethod() {
+    val wrongMethod = get("/echo", plainTextMediaType)
+    val response = httpClient.newCall(wrongMethod).execute()
+    assertThat(response.code).isEqualTo(404)
+    assertThat(response.body!!.source().readUtf8()).isEqualTo("""
```
adrw: Nice tests

misk/src/test/kotlin/misk/web/actions/NotFoundActionTest.kt
```diff
@@ -110,14 +169,35 @@ class NotFoundActionTest {
     return Request.Builder()
         .post(content.toRequestBody(contentType))
         .url(jettyService.httpServerUrl.newBuilder().encodedPath(path).build())
-        .header("Accept", acceptedMediaType.toString())
+        .apply {
+          if (acceptedMediaType != null)
```
adrw: Add braces for if statement

misk/src/test/kotlin/misk/web/actions/NotFoundActionTest.kt
```diff
@@ -110,14 +169,35 @@ class NotFoundActionTest {
...
+            header("Accept", acceptedMediaType.toString())
```
adrw: Does the request blow up if Accept header isn't passed in?

misk/src/test/kotlin/misk/web/proxy/WebProxyActionTest.kt
```diff
@@ -161,7 +160,13 @@ class WebProxyActionTest {
   internal fun getNotForwardedPathNoMatch() {
     val request = get("/local/notredirectedprefix/tacos", weirdMediaType)
     val response = optionalBinder.proxyClient.newCall(request).execute().toMisk()
-    assertThat(response.readUtf8()).isEqualTo("Nothing found at /local/notredirectedprefix/tacos")
+    assertThat(response.readUtf8()).isEqualTo("""
```
adrw: Thanks for updating these tests

misk/src/test/kotlin/misk/web/proxy/WebProxyActionTest.kt
```diff
@@ -224,7 +229,16 @@ class WebProxyActionTest {
...
+      |No such resource.
+      |
+      |Received:
+      |  POST /local/notredirectedprefix/tacos
+      |  Accept: application/weird
+      |  Content-Type: application/weird; charset=utf-8
+      |""".trimMargin())
+
+        // "Nothing found at /local/notredirectedprefix/tacos")
```
adrw: rm comment

misk/src/main/kotlin/misk/web/actions/NotFoundAction.kt
swankjesse: nah, it was on the old line 37

misk/src/test/kotlin/misk/web/actions/NotFoundActionTest.kt
keeferrourke: Heh, yeah it sends the string "null" as the accept header value, which throws
```
java.lang.IllegalArgumentException: null is not a valid media range
	at misk.web.mediatype.MediaRange$Companion.parse(MediaRange.kt:87)
	at misk.web.mediatype.MediaRange$Companion.parseRanges(MediaRange.kt:81)
	at misk.web.HttpCall$DefaultImpls.accepts(HttpCall.kt:101)
	at misk.web.ServletHttpCall.accepts(ServletHttpCall.kt:13)
	at misk.web.jetty.WebActionsServlet.handleCall(WebActionsServlet.kt:65)
	at misk.web.jetty.WebActionsServlet.doPost(WebActionsServlet.kt:51)
...
```

misk/src/main/kotlin/misk/web/actions/NotFoundAction.kt
```diff
@@ -38,4 +50,40 @@ class NotFoundAction @Inject constructor() : WebAction {
       )
     }
   }
+
+  /** Generates a message for a client that hints what may have been wrong with their request. */
+  private fun notFoundMessage(path: String, httpCall: HttpCall) = buildString {
+    val boundActions = servletProvider.get().boundActions
+    val alternativeActions = boundActions.mapNotNull {
+      it.matchByUrl(httpCall.url)
+    }.filterNot { it.action.pathPattern.matchesWildcardPath }
```
swankjesse: does this prevent it from matching useful things?

misk/src/main/kotlin/misk/web/actions/NotFoundAction.kt
swankjesse: nooo tabs are bad!

misk/src/main/kotlin/misk/web/actions/NotFoundAction.kt
keeferrourke: It was, but it was doing so with the `response()` method of its Companion, which isn't capable of providing the contextful response.

misk/src/main/kotlin/misk/web/actions/NotFoundAction.kt
```diff
@@ -38,4 +50,40 @@ class NotFoundAction @Inject constructor() : WebAction {
...
+    for (i in 0 until alternativeActions.size) {
+      val alternative = alternativeActions[i]
+      val metadata = alternative.action.metadata
+      append("Alternative:\n")
+      append("  ${metadata.dispatchMechanism.method} ${metadata.pathPattern}\n")
+      for (requestContentType in metadata.requestMediaTypes) {
+        append("  Content-Type: $requestContentType\n")
+      }
+      append("  Accept: ${metadata.responseMediaType}\n")
+      if (i < alternativeActions.size - 1) {
```
swankjesse: you could make your life easier by moving this newline to the beginning of the loop and removing the check on line 71

misk/src/main/kotlin/misk/web/actions/NotFoundAction.kt
adrw: Agreed 💯 . No indentation for the following lines would also be totally valid too!

misk/src/main/kotlin/misk/web/actions/NotFoundAction.kt
keeferrourke: Fixed

misk/src/main/kotlin/misk/web/BoundAction.kt
keeferrourke: Fixed

misk/src/main/kotlin/misk/web/actions/NotFoundAction.kt
keeferrourke: Ah I just updated this to use tabs! What if I did `val INDENT = "  "` and used that everywhere? Would that be an okay solution?

misk/src/test/kotlin/misk/web/actions/NotFoundActionTest.kt
keeferrourke: Fixed

misk/src/test/kotlin/misk/web/actions/NotFoundActionTest.kt
keeferrourke: thanks!

misk/src/test/kotlin/misk/web/actions/NotFoundActionTest.kt
keeferrourke: Fixed

misk/src/test/kotlin/misk/web/proxy/WebProxyActionTest.kt
keeferrourke: Fixed, sorry about that!

misk/src/main/kotlin/misk/web/actions/NotFoundAction.kt
keeferrourke: It might. 

I did this for two reasons:
1. Prevent from matching on `/{path: .*}` routes, which aren't useful suggestions.
2. Concerns about leaking too much information.

As @adrw mentioned:
> Curious about the security implications around giving more information on endpoints through the error messages. Otherwise looks good 👍

I don't want to suggest too much &mdash; just things that look like the caller is definitely aware of. If a client is making a call to weird wildcard path, they might not be provided any suggestions, but they _should_ know about what the endpoint they are trying to call requires... but they might not realize they forgot to set the right content-type, so context is still helpful.

misk/src/main/kotlin/misk/web/actions/NotFoundAction.kt
adrw: I'd lean towards just not having an indent at all

misk/src/main/kotlin/misk/web/actions/NotFoundAction.kt
adrw: Would there be a way to also only recommend paths that they have sufficient authentication to see?

misk/src/main/kotlin/misk/web/actions/NotFoundAction.kt
keeferrourke: Removed the indentation. No tabs, no spaces, just info :)

misk/src/main/kotlin/misk/web/actions/NotFoundAction.kt
keeferrourke: Fixed!

misk/src/main/kotlin/misk/web/actions/NotFoundAction.kt
keeferrourke: I'd expect so. Exploring that as an option.

keeferrourke: The information leakage is very high. Not going to delve into this. Any info that would be leaked by this is already discoverable. 

Someone who really cares about this could deploy an interceptor or an endpoint that supercedes the behaviour of the NotFoundAction.

misk/src/main/kotlin/misk/web/actions/NotFoundAction.kt
```diff
@@ -38,4 +50,35 @@ class NotFoundAction @Inject constructor() : WebAction {
       )
     }
   }
-}
+
+  /** Generates a message for a client that hints what may have been wrong with their request. */
+  private fun notFoundMessage(path: String, httpCall: HttpCall) = buildString {
+    val boundActions = servletProvider.get().boundActions
```
jjestrel: Is it easy to get routes that match your own level of authentication

keeferrourke: It's not :(

PR #1164, opened 2019-08-20
Replica reads

tirsen:


misk-hibernate/src/main/kotlin/misk/hibernate/Destination.kt
```diff
@@ -0,0 +1,61 @@
...
+      check(keyspace == shard.keyspace)
+    }
+  }
+
+  constructor(shard: Shard) : this(shard.keyspace, shard, null)
+
+  constructor(tabletType: TabletType) : this(null, null, tabletType)
+
+  override fun toString(): String {
+    val tabletType = if (tabletType != null) "@$tabletType" else ""
```
kalfonso: nit: should check if tabletType already starts with '@'?

misk-hibernate/src/main/kotlin/misk/hibernate/Destination.kt
```diff
@@ -0,0 +1,61 @@
...
+    if (shard != null) {
+      return "$shard$tabletType"
+    }
+    if (keyspace != null) {
+      return "$shard$tabletType"
```
kalfonso: nit: shard is null at this point. Did you mean "keyspace"?

misk-hibernate/src/main/kotlin/misk/hibernate/Destination.kt
tirsen: I don't think it can. It's an enum! :-)

misk-hibernate/src/main/kotlin/misk/hibernate/Destination.kt
tirsen: Good point. Lemme write some tests for the `toString` method too!

tirsen: there!

misk-hibernate/src/main/kotlin/misk/hibernate/RealTransacter.kt
```diff
@@ -313,41 +329,45 @@ internal class RealTransacter private constructor(
...
-            // TODO we need to parse out the tablet type (replica or master) from the current target and keep that when we target the new shard
-            // We should only change the shard we're targeting, not the tablet type
-            val previousTarget =
-                connection.createStatement().use { statement ->
-                  statement.executeQuery("SHOW VITESS_TARGET").uniqueString()
-                }
-            connection.createStatement().use { statement ->
-              statement.execute("USE `$shard`")
-            }
+      return if (isVitess()) {
```
kalfonso: can this be replaced with ```target(Destination(shard), function)```. "target" body seems to exercise same logic

misk-hibernate/src/main/kotlin/misk/hibernate/Destination.kt
kalfonso: Makes sense! :-)

misk-hibernate/src/main/kotlin/misk/hibernate/RealTransacter.kt
tirsen: Yes it can but I want to expose `target(shard)` and `replicaRead()` to the end user as I believe those are the two main use cases I want app developers to use. So it's a way to guide them down the right path.

PR #1167, opened 2019-08-21
Don't merge in state on (UN)SET_DIRTY_INPUT_FIELD

jaredwindover:


misk/web/tabs/web-actions/src/ducks/webActions.ts
```diff
@@ -937,8 +937,6 @@ export const WebActionsReducer = (
 ) => {
   switch (action.type) {
     case WEBACTIONS.ADD_REPEATED_FIELD:
-    case WEBACTIONS.SET_DIRTY_INPUT_FIELD:
```
adrw: Amazing what 2 lines can do!

Context: There was a recursively expanding state as a result of these two Redux actions unnecessarily being merged into state on each user interaction. Removing them from merging into the state shrinks the state size and eliminates the recursive performance error.

Issue #230, opened 2018-06-05
MySQL query checker


swankjesse:
It should be possible to install a thing that detects and warns on unindexed queries.

PR #1168, opened 2019-08-21
Add QuerySniper and TransactionRunner

keeferrourke:
This change addresses issue (#230) and modifies the `Transacter` to:
 1. Run transactions with configurable time limits (entire transaction)
 2. Run individual queries with configurable time limits

Some new internal classes have been introduced to accomplish this:

 - `TransactionRunner`
 - `QuerySniper`

## TransactionRunner

Every transacter gets one. As its name suggests, it runs the transaction block. It also schedules events to warn about or cancel its transaction if the Timeouts are exceeded.

## QuerySniper

Every session gets one. This is like `TransactionRunner`, but it listens on individual hibernate queries.

## How do these work?

`Timeouts` are specified in the `DataSourceConfig`. The `HibernateModule` will bind a `TimeoutsConfig` object based on the values configured there.

There are three kinds of timeouts:
 - transaction timeouts: the time allotted for a transaction as a whole
 - query timeouts: the time allotted for a regular query
 - slow query timeouts: the time allotted for markedly slow queries

By default, these timeouts are all disabled.

When the timeouts are active, the runner attached to a transacter will automatically kill long running transactions, and each session's sniper will automatically kill long running queries.

### Transaction timeout runtime configuration

You can opt out of a transaction's timeouts with

    transacter.noTimeouts().transaction { ... }

or specify custom transaction timeouts with

    transacter.withTimeouts(Timeouts.create(...)).transaction { ... }

Note that these timeouts do _NOT_ affect query timeouts, which must be configured separately. It is possible to never have a transaction timeout, but still have query timeouts, or vice versa.

### Query timeout runtime configuration

If you know that a particular query is slow running, you can tell it to use the slow query timeout configuration by

      session.runSlowQuery { ... }
    }

The `runSlowQuery` lambda uses the session as a reciever.

If you want to change the query or slow query timeouts at runtime, you can do so on the transacter:

    transacter.withQueryTimeouts(Timeouts.create(...))
         .withSlowQueryTimeouts(Timeouts.create(...))
         .transaction { ... }

That's it!

### DIFF
--- a/misk-hibernate/src/main/kotlin/misk/hibernate/RealTransacter.kt
+++ b/misk-hibernate/src/main/kotlin/misk/hibernate/RealTransacter.kt
@@ -18,6 +18,7 @@ import java.sql.SQLRecoverableException
 import java.sql.SQLSyntaxErrorException
 import java.time.Duration
 import java.util.EnumSet
+import java.util.concurrent.atomic.AtomicBoolean
 import javax.inject.Provider
 import javax.persistence.OptimisticLockException
 import kotlin.reflect.KClass
@@ -28,23 +29,31 @@ internal class RealTransacter private constructor(
   private val qualifier: KClass<out Annotation>,
   private val sessionFactoryProvider: Provider<SessionFactory>,
   private val threadLatestSession: ThreadLocal<RealSession>,
-  private val options: TransacterOptions,
+  internal val options: TransacterOptions,
   private val queryTracingListener: QueryTracingListener,
-  private val tracer: Tracer?
+  private val tracer: Tracer?,
+  private val transactionRunnerFactory: TransactionRunner.Factory,
+  private val querySniperFactory: QuerySniper.Factory,
+  private val timeoutsConfig: TimeoutsConfig
 ) : Transacter {
-
   constructor(
     qualifier: KClass<out Annotation>,
     sessionFactoryProvider: Provider<SessionFactory>,
     queryTracingListener: QueryTracingListener,
-    tracer: Tracer?
+    tracer: Tracer?,
+    transactionRunnerFactory: TransactionRunner.Factory,
+    querySniperFactory: QuerySniper.Factory,
+    timeoutsConfig: TimeoutsConfig
   ) : this(
-      qualifier,
-      sessionFactoryProvider,
-      ThreadLocal(),
-      TransacterOptions(),
-      queryTracingListener,
-      tracer
+      qualifier = qualifier,
+      sessionFactoryProvider = sessionFactoryProvider,
+      threadLatestSession = ThreadLocal(),
+      options = TransacterOptions(),
+      queryTracingListener = queryTracingListener,
+      tracer = tracer,
+      transactionRunnerFactory = transactionRunnerFactory,
+      querySniperFactory = querySniperFactory,
+      timeoutsConfig = timeoutsConfig
   )
 
   private val sessionFactory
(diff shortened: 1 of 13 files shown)


tirsen:
Is this for MySQL only? Vitess has an entirely different way of enforcing and specifying query and transaction timeouts.

misk-hibernate/src/main/kotlin/misk/hibernate/RealTransacter.kt
```diff
@@ -195,39 +215,75 @@ internal class RealTransacter private constructor(
...
+
+  override fun withQueryTimeouts(timeouts: Timeouts): Transacter {
+    return withTimeouts(timeoutsConfig.copy(query = timeouts))
+  }
+
+  override fun withSlowQueryTimeouts(timeouts: Timeouts): Transacter {
+    return withTimeouts(timeoutsConfig.copy(slowQuery = timeouts))
+  }
+
+  override fun noTimeouts(): Transacter = withOptions(options.copy(disableTimeouts = true))
```
tirsen: This won't work on Vitess. Vitess always enforces query and transaction timeouts.

misk-hibernate/src/main/kotlin/misk/hibernate/Session.kt
```diff
@@ -50,6 +50,20 @@ interface Session {
...
   fun <T : DbEntity<T>> delete(entity: T)
+
+  /** Retrieve the current timeouts used by this. */
+  fun timeouts(): Timeouts
+
+  /**
+   * Marks a query as slow. If a [QuerySniper] is watching this session, then it will not warn
+   * or kill it.
+   */
+  fun <T> runSlowQuery(query: Session.() -> T): T
```
tirsen: This won't work on Vitess. Vitess enforces timeouts at the vtgate and has a different way of changing those timeouts.

misk-hibernate/src/main/kotlin/misk/hibernate/Transacter.kt
```diff
@@ -17,12 +17,16 @@ interface Transacter {
...
-   * If the block raises an exception the transaction will be rolled back instead of committed.
+   * Starts a transaction, blocks on the current thread until [block] has finished execution, and
+   * commits the transaction. If the block raises an exception the transaction will be rolled back
+   * instead of committed.
    *
    * If retries are permitted (the default), a failed but recoverable transaction will be
    * reattempted after rolling back.
    *
+   * Unless the timeout is disabled, this transaction will be cancelled and rolled back if it
+   * exceeds the duration specified by [misk.jdbc.DataSourceConfig.transaction_timeout].
```
tirsen: Again, this is not true for Vitess. If the Vitess configured timeout is shorter than the datasource timeout then that will be enforced.

tirsen:
I think we might need to rethink this for Vitess. Vitess enforces various timeout at the vtgate level. The advantages with that is that the timeouts are enforced even if the application crashes where as these timeouts will only cancel the query or transaction if the application is still running.

tirsen:
I think either we need to make it clear that these timeouts apply to MySQL only or we need to actually pass down the changed timeouts to Vitess. We are transitioning to the MySQL protocol so we need to ensure the timeouts can be passed to the vtgate through the MySQL protocol.

keeferrourke:
> @tirsen: I think either we need to make it clear that these timeouts apply to MySQL only or we need to actually pass down the changed timeouts to Vitess. We are transitioning to the MySQL protocol so we need to ensure the timeouts can be passed to the vtgate through the MySQL protocol.

Is it sufficient to add checks on the driver type and update the docs to clarify these features only work for plain MySQL? What is the implication of transitioning our Vitess implementation to the MySQL protocol? Will that break any driver checks that I might implement?

I could look into propogating the timeouts to the vtgate, but I'm not really that familiar with Vitess.

tirsen:
@keeferrourke I'm not sure we should build extensive features for MySQL only. Are you building this specifically for an application that will run only on MySQL? Usually these type of features are required for high scalability apps and if you have high scale you should generally use Vitess.

I think what we might need for Vitess is the slow query support. I think for that we need to propagate the timeout either using a `SET` statement, a query hint or a JDBC flag. You might like to coordinate with @systay to figure this out.

tirsen:
I think what we should do is work with the standard JDBC methods for as much of this as possible. For example, there is `java.sql.Statement#setQueryTimeout`. We should make sure that is propagated down to Vitess. I think current versions of MySQL supports query timeouts and I would assume also respects this setting but that should obviously be verified. In that case we would not need a QuerySniper.

Transaction timeout is enforced on Vitess as mentioned but I'm not sure there is an equivalent transaction timeout at the JDBC level that we can use to propagate an extended timeout to Vitess. I'm thinking we can use some kind of `SET vitess_transaction_timeout = ...` session setting on Vitess for that.

What do you think @systay ?

tirsen:
As discussed in Slack we need to solve this in a different way in the cloud. I'm going to close this. Nice code though!

PR #1169, opened 2019-08-22
Web-Actions: setDirty onChange for easier tab nav

adrw:
* Form performance issues allow for reverting `onBlur` changes in below commit
* git revert b4bf5af8a33f67d903d2e83c9329372f4f17c4af


jaredwindover:
Is performance still okay with this?

adrw:
Initial testing seems to indicate it is, seems like the primary bottleneck was the bug you fixed.

PR #1183, opened 2019-09-04
Make it possible to turn off HTTP/2 support

swankjesse:
We're running into problems with HTTP/2 and rebalancing

### DIFF
--- a/misk/src/main/kotlin/misk/web/WebConfig.kt
+++ b/misk/src/main/kotlin/misk/web/WebConfig.kt
@@ -6,21 +6,44 @@ import misk.security.ssl.TrustStoreConfig
 import misk.web.exceptions.ActionExceptionLogLevelConfig
 
 data class WebConfig(
+  /** HTTP port to listen on, or 0 for any available port. */
   val port: Int,
+
+  /** If a connection is unused for this many milliseconds, it is closed. */
   val idle_timeout: Long,
+
+  /** The network interface to bind to. Null or 0.0.0.0 to bind to all interfaces. */
   val host: String? = null,
+
   val ssl: WebSslConfig? = null,
+
+  /** HTTP/2 support is currently opt-in because we can't load balance it dynamically. */
+  val http2: Boolean = true,
+
+  /** Number of NIO selector threads. */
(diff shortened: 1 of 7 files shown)


misk/src/main/kotlin/misk/web/jetty/JettyService.kt
```diff
@@ -96,16 +109,34 @@ class JettyService @Inject internal constructor(
...
+        }
+        httpsConnectionFactories += http2
       }
+
       val http1 = HttpConnectionFactory(httpsConfig)
-      val httpsConnector = ServerConnector(server, null, null, null,
-          webConfig.acceptors ?: -1, webConfig.selectors ?: -1, ssl, alpn, http2, http1)
+      httpsConnectionFactories += http1
+
+      val httpsConnector = ServerConnector(
```
zhxnlai: This `startUp` method is long. I would move this line to a `createHttpsConnector` method (same for `createHttpConnector`)

misk/src/main/kotlin/misk/web/WebConfig.kt
```diff
@@ -6,21 +6,44 @@ import misk.security.ssl.TrustStoreConfig
...
+
+  /** If a connection is unused for this many milliseconds, it is closed. */
   val idle_timeout: Long,
+
+  /** The network interface to bind to. Null or 0.0.0.0 to bind to all interfaces. */
   val host: String? = null,
+
   val ssl: WebSslConfig? = null,
+
+  /** HTTP/2 support is currently opt-in because we can't load balance it dynamically. */
```
alecholmes: `true` being the default makes HTTP/2 opt-out.

swankjesse: Yep. I’m gonna flip in follow-up

misk/src/main/kotlin/misk/web/jetty/JettyService.kt
swankjesse: Lemme consider for follow-up. I’m not bothered by long plumbing methods.

PR #1184, opened 2019-09-04
add misk-datadog

mightyguava:
very basic datadog integration (just binding the tracer to misk). Apps still need to use the datadog java agent for this to work, but that's at a different layer.


settings.gradle
```diff
@@ -21,4 +21,5 @@ include ':samples:exemplar'
 include ':samples:exemplarchat'
 include ':misk-metrics-digester'
 include ':misk-crypto'
-include ':misk-inject'
\ No newline at end of file
+include ':misk-inject'
+include ':misk-datadog'
```
adrw: Alphabetical ordering?

PR #1188, opened 2019-09-06
datadog: inject trace/span ids into MDC

mightyguava:


wesleyk:
interesting, should we request upstream for them to give us this for free?

wesleyk:
> interesting, should we request upstream for them to give us this for free?

err actually saw your other PR where we disable automated injection? Is this what we're doing instead?>

mightyguava:
> err actually saw your other PR where we disable automated injection? Is this what we're doing instead?

Yeah, they hard-code `dd.span_id` and `dd.trace_id`. We don't want the `dd` prefix. I think they have it to enable directly linking for tracing to logs, but we don't use datadog logs so it doesn't matter.

Could probably ask them to turn it into an option, but it's not really worth the effort to me.

mightyguava:
Also this is based on their impl: https://github.com/DataDog/dd-trace-java/blob/master/dd-java-agent/agent-tooling/src/main/java/datadog/trace/agent/tooling/log/LogContextScopeListener.java

PR #1189, opened 2019-09-10
Fall back to replica to fix the spare region health check issue

lyonlai:
This PR address the health check connectivity problem on the spare region, where vitess cluster doesn't have a master tablet inside. Proposed way to solve it is to fall back the connection/transaction to point to replicas when fails to connect to master tablet.

Major changes happens in the following class.
1. `PingDatabaseService`
2. `DataSourceService`
3. `Transacter` having a `failSafeReal` method for `SchemaMigratorService` to use

### DIFF
--- a/misk-hibernate/src/main/kotlin/misk/hibernate/RealTransacter.kt
+++ b/misk-hibernate/src/main/kotlin/misk/hibernate/RealTransacter.kt
@@ -1,6 +1,5 @@
 package misk.hibernate
 
-import com.google.common.base.Supplier
 import com.google.common.base.Suppliers
 import com.google.common.util.concurrent.ThreadFactoryBuilder
 import io.opentracing.Tracer
@@ -13,7 +12,6 @@ import misk.jdbc.map
 import misk.jdbc.uniqueString
 import misk.logging.getLogger
 import misk.tracing.traceWithSpan
-import okhttp3.internal.threadFactory
 import org.hibernate.FlushMode
 import org.hibernate.SessionFactory
 import org.hibernate.StaleObjectStateException
@@ -26,7 +24,6 @@ import java.util.EnumSet
 import java.util.concurrent.Callable
 import java.util.concurrent.CompletableFuture
 import java.util.concurrent.Executors
-import java.util.concurrent.Future
 import java.util.concurrent.TimeUnit
 import javax.inject.Provider
 import javax.persistence.OptimisticLockException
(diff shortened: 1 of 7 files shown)


misk-hibernate/src/main/kotlin/misk/hibernate/RealTransacter.kt
```diff
@@ -133,6 +130,45 @@ internal class RealTransacter private constructor(
...
+    check(config.type.isVitess) {
+      "Only applicable with vitess connection"
+    }
+
+    return withSession { session ->
+      try {
+        session.target(Destination(TabletType.MASTER)) {
+          block(session)
+        }
+      } catch (e: Throwable) {
```
tirsen: Never catch Throwable: https://rules.sonarsource.com/java/tag/error-handling/RSPEC-1181

Catch SQLException instead. Can we find out exactly SQLException is being thrown? Is it an SQLRecoverableException for example? We should be as specific as possible so we don't hide other errors.

misk-hibernate/src/main/kotlin/misk/hibernate/RealTransacter.kt
```diff
@@ -133,6 +130,45 @@ internal class RealTransacter private constructor(
...
+        try {
+          session.target(Destination(TabletType.REPLICA)) {
+            block(session)
+          }
+        } catch (e: Throwable) {
+          throw e
```
tirsen: You don't need to catch this if you're just rethrowing it

misk-hibernate/src/main/kotlin/misk/hibernate/Transacter.kt
```diff
@@ -49,6 +49,12 @@ interface Transacter {
    */
   fun <T> replicaRead(block: (session: Session) -> T): T
 
+  /**
+   * Runs a read on master first then tries it on replicas on failure. This method is here only for
+   * health check purpose for standby regions.
+   */
+  fun <T> failSafeRead(block: (session: Session) -> T): T
```
tirsen: I think this can be made an extension method as well right? You should be able to build it purely on top of the existing Session interface.

misk-hibernate/src/main/kotlin/misk/jdbc/DataSourceConfig.kt
```diff
@@ -219,6 +219,35 @@ data class DataSourceConfig(
 
     return pathToUse
   }
+
+  fun getReplicaDatabaseConfig(): DataSourceConfig {
```
tirsen: `asReplica`

PR #1195, opened 2019-09-12
Use the Guice annotation for optional=false

swankjesse:


swankjesse:
Tracking upstream: https://youtrack.jetbrains.com/issue/KT-33822

PR #1197, opened 2019-09-12
Datadog span context in queries

systay:


systay:
Will re-do using opentracing TextMap

PR #1200, opened 2019-09-16
Add span context to all queries when using VITESS over MySQL

systay:


systay:
This time I'm only using datadog for testing.

misk-hibernate/build.gradle
```diff
@@ -25,6 +25,7 @@ dependencies {
   compile dep.openTracingUtil
   compile dep.openTracingJdbc
   compile dep.tracingJaeger
+  compile dep.openTracingDatadog
```
mightyguava: This should be a `testCompile`

PR #1201, opened 2019-09-16
Bump Misk-Web 0.1.19 and remove deprecated APIs

adrw:
* Rename `loader` tab to `admin-dashboard`
* Update ducks file imports to use unified `simpleRedux`
* Use `simpleSelectorGet` everywhere instead of `simpleSelect`
* Remove `/_tab/*` React routes


misk/src/main/kotlin/misk/web/metadata/AdminDashboardModule.kt
```diff
@@ -27,29 +27,29 @@ class AdminDashboardModule(val environment: Environment) : KAbstractModule() {
...
-      resourcePath = "classpath:/web/_tab/loader/"
+      resourcePath = "classpath:/web/_tab/admin-dashboard/"
     ))
 
     // @misk packages
     install(WebTabResourceModule(
       environment = environment,
       slug = "@misk",
-      web_proxy_url = "http://localhost:9100/",
+      web_proxy_url = "http://localhost:3100/",
```
aashnisshah: Any particular reason for this change?

misk/web/tabs/config/src/components/ConfigComponent.tsx
```diff
@@ -10,9 +10,11 @@ export interface IConfigResource {
 }
 
 export interface IConfigProps {
-  resources: string
+  resources?: string
 }
 
+const cssH3 = css({ fontFamily: "Fira Code, Menlo" })
```
aashnisshah: can we pull changes like this out to a standard/main stylesheet unless there's a specific reason to only have it here

aashnisshah: If we want it specifically here, perhaps we can make use of `styled-components` instead?

misk/web/tabs/config/src/components/ConfigComponent.tsx
```diff
@@ -26,14 +28,32 @@ export default class ConfigComponent extends React.PureComponent<IConfigProps> {
...
+      )
+    } else {
+      return (
+        <div>
+          <H1>Config</H1>
+          <br />
+          <H3 css={cssH3}>{"Effective Config"}</H3>
+          <CodePreContainer>{<Spinner />}</CodePreContainer>
+          <br />
+          <H3 css={cssH3}>{"classpath://common.yaml"}</H3>
```
aashnisshah: are these paths correct?

misk/web/tabs/web-actions/src/containers/SendRequestCollapseContainer.tsx
```diff
@@ -74,66 +73,88 @@ const SendRequestCollapseContainer = (
...
-  const response = simpleSelect(props.simpleNetwork, `${tag}::Response`)
-  const responseData = response.data
-  const whichFormData = simpleSelect(
-    props.simpleForm,
-    `${tag}::RequestBodyFormInputType`,
-    "data",
-    simpleType.boolean
+
+  // Response with fallback to error
+  const response = simpleSelectorGet(
```
aashnisshah: This is confusing. What's happening here and why?

misk/web/tabs/web-actions/src/ducks/webActions.ts
```diff
@@ -901,8 +901,12 @@ export const processMetadata = (webActionMetadata: IWebActionAPI[]): any =>
     .value()
 
 function* handleMetadata() {
+  const useTestData = false
```
aashnisshah: what's the `useTestData` used for?

misk/web/tabs/config/src/components/ConfigComponent.tsx
adrw: This is the loading screen when data isn't present yet

misk/web/tabs/web-actions/src/ducks/webActions.ts
adrw: Testing the web-actions tab without a live running service

misk/web/tabs/web-actions/src/containers/SendRequestCollapseContainer.tsx
adrw: `simpleSelectorGet` follows Lodash Get syntax (https://lodash.com/docs#get) so the third parameter is the fallback value. In this case, if `Response.data` hasn't been set it looks for to see if an error has been set instead.

misk/web/tabs/config/src/components/ConfigComponent.tsx
adrw: That's something I'll consider in a follow up. I don't think we gain/lose much by having occasional one-off css-in-JS styling. Notably, this syntax is still css-in-js using `@emotion/core` instead of `styled-components` because of better performance and testing story.

misk/src/main/kotlin/misk/web/metadata/AdminDashboardModule.kt
adrw: I found this bug and fixed it! There isn't a separate `@misk/` tab, Below you can see that the `resourcePath` is part of the `admin-dashboard` tab, so the ports should match, and now they do!

adrw:
Promises and mapper functions coming in a follow up Misk-Web PR. The `simpleSelectorGet` and `simpleMergeData` functions are also documented in the Misk-Web code base. Discoverability of documentation on the Misk-Web docs site will eventually help a lot with providing context for these larger API changes.

misk/web/tabs/config/src/components/ConfigComponent.tsx
aashnisshah: Ahh awesome! I'll read up more about `@emotion/core` since it sounds like a better option :)

PR #1205, opened 2019-09-20
[CLOSES #1138] Graceful failure of Web Actions tab for large protos

### LINKED ISSUE #1138: Web Actions tab falling over for actions with large protos
Should add a check to bail out of web actions that have large protos so that it shows just the empty textbox and doesn't fail the entire tab.

adrw:


wesleyk:
What size of protobuf considered too large? Do we understand what the actual parser failure is (is it timing out, does it run out of memory, etc)?

adrw:
@wesleyk I don’t have a concrete idea of what is too large but I can look at the large endpoints of ClientSync and get an idea, it is likely on the order of many dozens of fields. The failure is from ImmutableJS running out of call stack while it is building up the representation of all of the field nodes in an array based tree. Moving away from Immutable in the long term could improve performance a bit though their primitives are helpful in preserving ordering and handling merges.

misk/web/tabs/web-actions/src/ducks/webActions.ts
```diff
@@ -791,48 +792,58 @@ const generateFieldTypesMetadata = (
...
-          ["0", "idChildren"],
-          typesMetadata.getIn(["0", "idChildren"]).add(id)
-        )
+          typesMetadata = typesMetadata.setIn(
+            ["0", "idChildren"],
+            typesMetadata.getIn(["0", "idChildren"]).add(id)
+          )
+        }
       }
+    } catch (e) {
```
wesleyk: should we be catching _any_ exception? Wouldn't this mask bugs?

adrw: Good catch! I'll try and see if we can narrow it down to only catch the call stack errors!

wesleyk: "Good catch" 😏

PR #1206, opened 2019-09-23
RequestLoggingInterceptor logs response time and code

mightyguava:
This implementation isn't ideal. It splits the interceptor into `RequestLoggingInterceptor` (NetworkInterceptor) and `RequestBodyLoggingInterceptor` (ApplicationInterceptor).

Request body is not available to a NetworkInterceptor since the body hasn't been decoded yet. Response code is not available to an ApplicationInterceptor since ApplicationInterceptors are actually implemented as the last NetworkInterceptor the request chain (the response chain is reversed), and `ExceptionHandlingInterceptor` hasn't executed yet.

Splitting means that if request/responses are sampled and `includeBody` is on, the `RequestLoggingInterceptor` and `RequestBodyLoggingInterceptor` do not log on the same set of requests. While not ideal, it should be fine since they are effectively independent.

### DIFF
--- a/misk/src/main/kotlin/misk/web/MiskWebModule.kt
+++ b/misk/src/main/kotlin/misk/web/MiskWebModule.kt
@@ -35,6 +35,7 @@ import misk.web.interceptors.InternalErrorInterceptorFactory
 import misk.web.interceptors.MetricsInterceptor
 import misk.web.interceptors.RebalancingInterceptor
 import misk.web.interceptors.RequestLogContextInterceptor
+import misk.web.interceptors.RequestBodyLoggingInterceptor
 import misk.web.interceptors.RequestLoggingInterceptor
 import misk.web.interceptors.TracingInterceptor
 import misk.web.jetty.JettyConnectionMetricsCollector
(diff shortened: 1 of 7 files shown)


misk/src/main/kotlin/misk/web/MiskWebModule.kt
```diff
@@ -113,10 +114,14 @@ class MiskWebModule(private val config: WebConfig) : KAbstractModule() {
     multibind<NetworkInterceptor.Factory>(MiskDefault::class)
         .to<ExceptionHandlingInterceptor.Factory>()
 
-    // Optionally log request and response details
-    multibind<ApplicationInterceptor.Factory>(MiskDefault::class)
+    // Log request and response details
```
wesleyk: isn't this optional as well based on sampling?

mightyguava:
> LG, but I'm not really seeing how the new interceptors are logging on different sets of requests. Mind elaborating?

The interceptors share the same sampling rate, but independently generate random numbers to determine what is sampled.

wesleyk:
> The interceptors share the same sampling rate, but independently generate random numbers to determine what is sampled.

mm gotcha, so there could be overlap. Thanks for clarifying!

misk/src/main/kotlin/misk/web/MiskWebModule.kt
mightyguava: I mistakenly thought this was opt-out, but it's opt-in. 👍

PR #1207, opened 2019-09-24
Log uncaught exceptions at servlet level

mightyguava:
Guice exceptions in `BoundAction` (and maybe other exceptions), can leak up. `HttpChannel` will end up logging at WARN, but we want it to be error.

### DIFF
--- a/misk/src/main/kotlin/misk/web/jetty/WebActionsServlet.kt
+++ b/misk/src/main/kotlin/misk/web/jetty/WebActionsServlet.kt
@@ -1,6 +1,7 @@
 package misk.web.jetty
 
 import misk.exceptions.StatusCode
+import misk.logging.getLogger
 import misk.web.BoundAction
 import misk.web.DispatchMechanism
 import misk.web.ServletHttpCall
@@ -32,6 +33,10 @@ internal class WebActionsServlet @Inject constructor(
   webActionFactory: WebActionFactory,
   webActionEntries: List<WebActionEntry>
 ) : WebSocketServlet() {
+
+  companion object {
+    val log = getLogger<WebActionsServlet>()
+  }
 
   internal val boundActions: MutableSet<BoundAction<out WebAction>> = mutableSetOf()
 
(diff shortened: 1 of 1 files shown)


misk/src/main/kotlin/misk/web/jetty/WebActionsServlet.kt
```diff
@@ -75,6 +80,9 @@ internal class WebActionsServlet @Inject constructor(
       }
     } catch (_: ProtocolException) {
       // Probably an unexpected HTTP method. Send a 404 below.
+    } catch (e: Throwable) {
+      log.error { "Uncaught exception on ${request.dispatchMechanism()} ${request.httpUrl()}" }
+      throw e
```
ryanhall07: should this just gracefully return a 500 below? otherwise you'll get duplicate log messages right? one ERROR, one WARN

misk/src/main/kotlin/misk/web/jetty/WebActionsServlet.kt
```diff
@@ -75,6 +80,9 @@ internal class WebActionsServlet @Inject constructor(
       }
     } catch (_: ProtocolException) {
       // Probably an unexpected HTTP method. Send a 404 below.
+    } catch (e: Throwable) {
+      log.error { "Uncaught exception on ${request.dispatchMechanism()} ${request.httpUrl()}" }
```
alanpaulin: Should this also log the exception itself?

mightyguava: Yeah good catch

misk/src/main/kotlin/misk/web/jetty/WebActionsServlet.kt
mightyguava: I'm not totally sure of semantics if this crashes after something already started writing a response. I guess we can't fail any worse if we are already here so 👍

PR #1208, opened 2019-09-25
Replica reads for MySQL

tirsen:


misk-hibernate/src/main/kotlin/misk/hibernate/RealTransacter.kt
```diff
@@ -44,19 +52,23 @@ internal class RealTransacter private constructor(
...
-      qualifier,
-      sessionFactoryProvider,
-      config,
-      ThreadLocal(),
-      TransacterOptions(),
-      queryTracingListener,
-      tracer
+      qualifier = qualifier,
+      sessionFactoryProvider = sessionFactoryProvider,
+      readerSessionFactoryProvider = readerSessionFactoryProvider,
```
kalfonso: nit: why not just invoke ```this(...)``` with ```readerSessionFactoryProvider``` added?

misk-hibernate/src/test/kotlin/misk/hibernate/TransacterTest.kt
```diff
@@ -244,6 +246,11 @@ class TransacterTest {
 
   @Test
   fun `can run consecutive replica reads and transactions`() {
+    // This test only makes sense with Vitess
+    if (!transacter.config().type.isVitess) {
```
kalfonso: why does it only make sense with Vitess? We should be able to run replica reads in MySQL as well

tirsen: yeah you're right, there's some vitess specific assertions in there but I can make those conditional

PR #1209, opened 2019-09-25
Prevent multiple Actions with identical routing annotations

ryanhall07:
This causes a server start up failure, which should result in catching
the error in a unit test before committing the bug.

2 actions with identical routing annotations is most likely a bug
introduced by bad copy/paste. One of the Actions will never be
dispatched, so it provides no value.

### DIFF
--- a/misk/src/main/kotlin/misk/web/jetty/WebActionsServlet.kt
+++ b/misk/src/main/kotlin/misk/web/jetty/WebActionsServlet.kt
@@ -46,6 +46,16 @@ internal class WebActionsServlet @Inject constructor(
     for (entry in webActionEntries) {
       boundActions += webActionFactory.newBoundAction(entry.actionClass, entry.url_path_prefix)
     }
+    // Verify no two Actions have identical routing annotations, which is most likely a bad
+    // copy/paste error and results in unexpected results for a developer. This fails the service
+    // startup, which should be caught with a unit test.
+    for (action in boundActions) {
+      for (other in boundActions) {
+        check(action === other || !action.hasIdenticalRouting(other)) {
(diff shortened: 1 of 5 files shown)


misk/src/test/kotlin/misk/web/actions/NotFoundActionTest.kt
```diff
@@ -180,7 +180,6 @@ class NotFoundActionTest {
   class TestModule : KAbstractModule() {
     override fun configure() {
       install(WebTestingModule())
-      install(WebActionModule.create<NotFoundAction>())
```
ryanhall07: no need to explicitly bind these, they are automatically bound for a Service in `MiskWebModule`. now these are resulting in dupe errors :)

alanpaulin:
Thanks @ryanhall07 !

misk/src/main/kotlin/misk/web/jetty/WebActionsServlet.kt
```diff
@@ -46,6 +46,16 @@ internal class WebActionsServlet @Inject constructor(
     for (entry in webActionEntries) {
       boundActions += webActionFactory.newBoundAction(entry.actionClass, entry.url_path_prefix)
     }
+    // Verify no two Actions have identical routing annotations, which is most likely a bad
+    // copy/paste error and results in unexpected results for a developer. This fails the service
+    // startup, which should be caught with a unit test.
+    for (action in boundActions) {
+      for (other in boundActions) {
+        check(action === other || !action.hasIdenticalRouting(other)) {
+          "Actions [${action.action.name}, ${other.action.name}] have identical routing annotations."
```
adrw: Nice 👍

PR #1210, opened 2019-09-27
Wait for longer for Vitess to start

tirsen:
This can sometimes take quite a long time on Kochiku it seems. Fail instantly if Vschema is busted

### DIFF
--- a/misk-hibernate/src/main/kotlin/misk/vitess/StartVitessService.kt
+++ b/misk-hibernate/src/main/kotlin/misk/vitess/StartVitessService.kt
@@ -152,14 +152,27 @@ class DockerVitessCluster(
 
   private var isRunning = false
   private var stopContainerOnExit = true
+  private var startupFailure: Exception? = null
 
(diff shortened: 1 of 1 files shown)


misk-hibernate/src/main/kotlin/misk/vitess/StartVitessService.kt
```diff
@@ -262,24 +276,30 @@ class DockerVitessCluster(
...
+            } else {
+              throw e
+            }
           }
         }
       }
+    } catch (e: DontRetryException) {
+      throw Exception(e.message)
+    } catch (e: Exception) {
+      throw Exception("Vitess cluster failed to start up in time", e)
```
systay: Why this rethrowing?

tirsen: We still want to see this error message in the test logs. We just need to enrich it with a better message.

PR #1212, opened 2019-10-01
LaunchDarkly modules

mightyguava:
Porting this over from skim. Adds the following modules:

`misk-feature` -- feature service interfaces
`misk-feature-testing` -- testing impl of said interfaces
`misk-launchdarkly-core` -- LaunchDarkly SDK wrapper that doesn't depend on `misk`, so we can use it in our monorepo
`misk-launchdarkly` -- Misk guice bindings for launchdarkly


misk-feature-testing/src/main/kotlin/misk/feature/testing/TestFeatureFlags.kt
```diff
@@ -0,0 +1,68 @@
...
+import misk.feature.FeatureFlags
+import misk.feature.FeatureService
+import javax.inject.Inject
+import javax.inject.Singleton
+
+/**
+ * In-memory test implementation of [FeatureFlags] that allows flags to be overridden.
+ */
+@Singleton
+class TestFeatureFlags @Inject constructor() : AbstractIdleService(),
```
swankjesse: Can you rename this to `FakeFeatureFlags` ? Or perhaps `SimpleFeatureFlags` ?

misk-feature-testing/src/main/kotlin/misk/feature/testing/TestFeatureFlags.kt
```diff
@@ -0,0 +1,68 @@
...
+import com.google.common.util.concurrent.AbstractIdleService
+import misk.feature.Attrs
+import misk.feature.Feature
+import misk.feature.FeatureFlags
+import misk.feature.FeatureService
+import javax.inject.Inject
+import javax.inject.Singleton
+
+/**
+ * In-memory test implementation of [FeatureFlags] that allows flags to be overridden.
```
swankjesse: should this be thread-safe?

misk-feature-testing/src/main/kotlin/misk/feature/testing/TestFeatureFlags.kt
```diff
@@ -0,0 +1,68 @@
...
+@Singleton
+class TestFeatureFlags @Inject constructor() : AbstractIdleService(),
+    FeatureFlags,
+    FeatureService {
+  override fun startUp() {}
+  override fun shutDown() {}
+
+  private val overrides = mutableMapOf<Feature, Any>()
+
+  override fun getBool(feature: Feature, token: String, attrs: Attrs?): Boolean {
```
swankjesse: Bool –> Boolean

misk-feature-testing/src/main/kotlin/misk/feature/testing/TestFeatureFlags.kt
```diff
@@ -0,0 +1,68 @@
+package misk.feature.testing
+
+import com.google.common.util.concurrent.AbstractIdleService
+import misk.feature.Attrs
```
swankjesse: Please rename this

swankjesse: Convention in Java land is to not abbreviate things. Motivation is that abbreviations are hard to guess and hard to grep

misk-feature-testing/src/main/kotlin/misk/feature/testing/TestFeatureFlagsModule.kt
```diff
@@ -0,0 +1,41 @@
...
+import misk.inject.toKey
+import kotlin.reflect.KClass
+
+/**
+ * Binds a [TestFeatureFlags] that allows tests to override values.
+ */
+class TestFeatureFlagsModule(
+  private val qualifier: KClass<out Annotation>? = null
+) : KAbstractModule() {
+  private val testFeatureFlags = TestFeatureFlags()
```
swankjesse: This class has an injectable constructor, you shouldn’t instantiate it manually

misk-feature-testing/src/main/kotlin/misk/feature/testing/TestFeatureFlagsModule.kt
```diff
@@ -0,0 +1,41 @@
...
+
+  override fun configure() {
+    val key = TestFeatureFlags::class.toKey(qualifier)
+    bind(key).toInstance(testFeatureFlags)
```
swankjesse: you can say `bind(key);` and that’ll bind it to the injectable constructor

misk-feature-testing/src/main/kotlin/misk/feature/testing/TestFeatureFlagsModule.kt
```diff
@@ -0,0 +1,41 @@
...
+   * instead of within individual test classes.
+   *
+   * Usage:
+   * ```
+   * install(TestFeatureFlagsModule().withOverrides { flags ->
+   *   flags.overrideBool(Feature("foo"), true)
+   * })
+   * ```
+   */
+  fun withOverrides(lambda: (TestFeatureFlags) -> Unit): TestFeatureFlagsModule {
```
swankjesse: If you want a fancier API:

```
   install(TestFeatureFlagsModule().withOverrides {
     overrideBool(Feature("foo"), true)
   })
```

You do this:

```
  fun withOverrides(lambda: TestFeatureFlags.() -> Unit): TestFeatureFlagsModule {
```

misk-feature-testing/src/test/kotlin/misk/feature/testing/TestFeatureFlagsTest.kt
```diff
@@ -0,0 +1,50 @@
...
+
+  @Test
+  fun getEnum() {
+    // Defaults to first enum.
+    assertThat(subject.getEnum<Dinosaur>(FEATURE, TOKEN))
+        .isEqualTo(Dinosaur.PTERODACTYL)
+
+    // Can be overridden
+    subject.override(FEATURE, Dinosaur.TYRANNOSAURUS)
+    assertThat(subject.getEnum<Dinosaur>(FEATURE, TOKEN))
```
swankjesse: Nice API!

misk-feature/src/main/kotlin/misk/feature/FeatureFlags.kt
```diff
@@ -0,0 +1,74 @@
...
+
+/**
+ * Typed feature string.
+ */
+data class Feature(val name: String)
+
+/**
+ * Extra attributes to be used for evaluating features.
+ */
+data class Attrs(
```
swankjesse: Could you make an instance of Attrs that’s emtpy, and make _that_ the default? As-is I can specify both empty attrs and null.

misk-feature/src/main/kotlin/misk/feature/FeatureFlags.kt
```diff
@@ -0,0 +1,74 @@
...
+  val stringAttrs: Map<String, String> = mapOf(),
```
swankjesse: Do we want to pass this through exactly as-is? Do we have a use case? This seems like unnecessary complexity

mightyguava: This is how you'd pass user properties through for segmenting them, like by country, platform (ios/android)

misk-feature-testing/src/main/kotlin/misk/feature/testing/TestFeatureFlags.kt
mightyguava: Yeah. Should be fine to just make the underlying map a ConcurrentHashMap right?

misk-feature-testing/src/main/kotlin/misk/feature/testing/TestFeatureFlags.kt
mightyguava: Should I make it `GetInteger` then?

misk-feature-testing/src/main/kotlin/misk/feature/testing/TestFeatureFlags.kt
mightyguava: Standard abbreviations are fine no? I don't think anyone would be confused what Attr means. We use HTTP and RPC all over the place. RPC is more alien than Attr imo

misk-feature-testing/src/main/kotlin/misk/feature/testing/TestFeatureFlagsModule.kt
mightyguava: I instantiate manually to enable the `withOverrides` thing.

misk-feature-testing/src/main/kotlin/misk/feature/testing/TestFeatureFlagsModule.kt
mightyguava: Cool!

misk-feature-testing/src/main/kotlin/misk/feature/testing/TestFeatureFlags.kt
swankjesse: Yep!

misk-feature-testing/src/main/kotlin/misk/feature/testing/TestFeatureFlags.kt
swankjesse: No, a good rule is to never invent synonyms. Kotlin gives us Int and Boolean and those are the words we should use.

misk-feature/src/main/kotlin/misk/feature/FeatureFlags.kt
swankjesse: Gotcha. Can you add that example? Should we use types for this?

mightyguava: I'm not attached to this. I think almost all usages will be passing in a string map.

Not sure what you mean by using types. Can you expand on that?

misk-feature-testing/src/main/kotlin/misk/feature/testing/TestFeatureFlags.kt
swankjesse: I'm recalling Effective Java Item 68. This is how we’ve named most of the things in Misk and I think it’s also good for us to be consistent!

swankjesse: (Abbreviations and Acronyms are different things, and the rules of acronyms is that they’re written like words: XmlHttpRequest instead of XMLHTTPRequest.)

misk-feature/src/main/kotlin/misk/feature/FeatureFlags.kt
swankjesse: In my experience if you give developers a `Map<String, String>` then some of them will use `state/prov: CALIFORNIA` and others will use `STATE: ca`. Giving them a type instead makes it easier to be consistent, which increases the value of the data gathered.

Here’s a hypothetical sketch:

```
/** Subtypes are data classes whose fields are captured with feature flag lookups. */
interface Attributes {
}

data class PlasmaAttributes {
  val state: UspsState,
  val statusLevel: StatusLevel,
  val userAgent: UserAgent
} : Attributes
```

used like this:

```
  override fun getBoolean(feature: Feature, token: String, attributes: Attributes) Boolean {
    ...
  }
```

That way every developer says “what is an Attributes?” and they find their project’s implementation and that makes things typesafe.

Lets land this as-is before adding that though.

misk-feature-testing/src/main/kotlin/misk/feature/testing/FakeFeatureFlags.kt
```diff
@@ -0,0 +1,69 @@
...
+  private val overrides = ConcurrentHashMap<Feature, Any>()
+
+  override fun getBoolean(feature: Feature, token: String, attributes: Attributes): Boolean {
+    return overrides.getOrDefault(feature, false) as Boolean
+  }
+
+  override fun getInt(feature: Feature, token: String, attributes: Attributes): Int {
+    checkNotNull(
+        overrides[feature]) { "Int flag $feature must be overridden with override() before use" }
+    return overrides[feature] as Int
```
swankjesse: if you want, you can combine the cast with the nullcheck AND get a free typecheck

```
return overrides[feature] as? Int
  ?: throw IllegalArgumentException("Int flag $feature must be overridden with override() before use")
```

misk-launchdarkly-core/src/main/kotlin/misk/feature/launchdarkly/LaunchDarklyFeatureFlags.kt
```diff
@@ -0,0 +1,116 @@
...
+    throw IllegalStateException("Feature flag $feature is off but no off variation is specified")
+  }
+
+  private fun buildUser(token: String, attributes: Attributes): LDUser {
+    val builder = LDUser.Builder(token)
+    attributes.text.forEach { (k, v) ->
+      when (k) {
+        // LaunchDarkly has some built-in keys that have to be initialized with their named
+        // methods.
+        "secondary" -> builder.secondary(k)
```
swankjesse: with respect to types, it’d be better if this was part of the API instead of part of the implementation.

misk-launchdarkly-core/src/main/kotlin/misk/feature/launchdarkly/LaunchDarklyFeatureFlags.kt
```diff
@@ -0,0 +1,116 @@
...
+        "ip" -> builder.ip(k)
+        "email" -> builder.email(k)
```
swankjesse: We probably never want to send them email addresses right?

mightyguava: Probably not, I'm just transcribing the set of "special" keys

misk-feature/src/main/kotlin/misk/feature/FeatureFlags.kt
mightyguava: We want to have consistency across services. I'd like to explore deeper how to do that in a type safe manner. My initial thought is that we may get most of what we want by exposing constants for common keys rather than adding types

PR #1214, opened 2019-10-02
Add Service.name() to misk module

jjestrel:


misk/src/main/kotlin/misk/Service.kt
```diff
@@ -0,0 +1,8 @@
+package misk
+
+import com.google.common.util.concurrent.Service
+
+fun Service.name(): String = when (this) {
```
jjestrel: I'm open to putting this in a different package. This way we have misk expose a way to get a "name" then callers don't need to care (at the call site) about dealing with CoordinatedService vs plain service

swankjesse: I think this wants to be CoordinatedService’s toString().

adrw: You could move this and all related Service files (CoordinatorService, MiskServiceModule, ServiceGraphBuilder, ServiceModule...)  into a `misk.service` package.

jjestrel: I think the toString change is a different unit of work than this PR. It would be nice if Service just had name() in the interface

swankjesse: Unfortunately that’s not how extension methods work. In particular, custom implementations of `Service` cannot override this method to provide their own name.

jjestrel:
We did something else

PR #1216, opened 2019-10-03
split out misk-service module for misk's ServiceManager

mightyguava:
This makes `misk.ServiceModule` usable in `misk-feature-testing`. `ServiceModule` is needed by basically every misk module, so splitting it out will make it easier for us to use it internally in dc.


misk/src/main/kotlin/misk/web/actions/StatusAction.kt
```diff
@@ -30,7 +30,7 @@ class StatusAction @Inject internal constructor(
     val services = serviceManagerProvider.get().servicesByState().values().asList()
     val serviceStatus = services.map {
       when (it) {
-        is CoordinatedService -> it.service.javaClass.simpleName to it.state()
+        is DelegatingService -> it.service.javaClass.simpleName to it.state()
```
mightyguava: @jjestrel fyi I added a new interface to work around `CoordinatedService` being internal

jjestrel: @swankjesse and I spent a lot of time talking about this in a different PR. I think `DelegatingService` is better than what I tried in my other PR. LGTM

PR #1217, opened 2019-10-04
Update to wire3.0.0-rc03

oldergod:
- [x] Use Wire3.0.0-rc03 which doesn't depend on okio 2.5.0-SNAPSHOT.


misk/src/main/kotlin/misk/grpc/GrpcMessageSink.kt
```diff
@@ -38,5 +42,16 @@ internal class GrpcMessageSink<T : Any> constructor(
...
+  override fun close() {
+    if (closed) return
+    closed = true
+    sink.close()
+  }
+
   override fun toString() = "GrpcMessageSink"
-}
\ No newline at end of file
+}
```
oldergod: All copied from Wire

dependencies.gradle
```diff
@@ -87,9 +87,9 @@ ext.dep = [
   "tracingJaeger": "com.uber.jaeger:jaeger-core:0.24.0",
   "tracingZipkin": "io.opentracing.brave:brave-opentracing:0.29.0",
   "vitess": "io.vitess:vitess-jdbc:3.0.0",
-  "wireRuntime": "com.squareup.wire:wire-runtime:3.0.0-rc01",
-  "wireGrpcClient": "com.squareup.wire:wire-grpc-client:3.0.0-rc01",
-  "wirePlugin": "com.squareup.wire:wire-gradle-plugin:3.0.0-rc01",
+  "wireRuntime": "com.squareup.wire:wire-runtime-multiplatform:3.0.0-rc02",
```
swankjesse: why does this change? I don’t like this because it means we could have both!

misk-grpc-tests/build.gradle
```diff
@@ -47,7 +49,7 @@ sourceSets {
   main.java.srcDirs += 'build/generated/source/proto/main/java'
 
   // TODO(jwilson): we do this to make IntelliJ happy but the Wire Gradle plugin should do that.
-  main.java.srcDirs += 'build/generated/src/main/java'
+  main.java.srcDirs += 'build/generated/source/wire'
```
swankjesse: nice

misk-grpc-tests/src/main/proto/routeguide/RouteGuideProto.proto
```diff
@@ -11,7 +11,7 @@
 // WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 // See the License for the specific language governing permissions and
 // limitations under the License.
-syntax = "proto3";
+syntax = "proto2";
```
swankjesse: nice

misk/src/main/kotlin/misk/grpc/GrpcMessageSink.kt
```diff
@@ -2,26 +2,30 @@ package misk.grpc
...
  * @param messageAdapter a proto adapter for each message.
+ * @param callForCancel the HTTP call that can be canceled to signal abnormal termination.
  * @param grpcEncoding the content coding for the stream body.
  */
 internal class GrpcMessageSink<T : Any> constructor(
   private val sink: BufferedSink,
   private val messageAdapter: ProtoAdapter<T>,
-  private val grpcEncoding: String = "identity"
-) : MessageSink<T>, Closeable by sink {
+  private val callForCancel: Call?,
```
swankjesse: For follow-up, I think we’re going to want to cancel the Jetty request.

dependencies.gradle
oldergod: I was just playing around, I think I'll revert this whole file (except for the wire's version upgrade) and will keep everything the same otherwise

misk-grpc-tests/src/main/proto/routeguide/RouteGuideProto.proto
oldergod: Well, Wire didn't leave me much choice...

PR #1220, opened 2019-10-09
[CLOSES #933] Add enum support to web-actions tab form

### LINKED ISSUE #933: Add support for Enum types in Web Actions Tab
Existing response returns the the class (ie. "com.squareup.protos.cash.raven.api.SmsDeliveryType") as the `type` for a field. But that enum class definition isn't included so it can't be resolved. Enum class definitions will need to be parsed and included in the WebActionMetadataAction response in a way that Web Actions tab can use them to present a `<select>` field.

WebActionMetadataAction Desired Response
```
{
 ...
 "types": {
        "com.squareup.protos.cash.raven.api.SendSmsNotificationRequest": {
          "fields": [
            {
              "name": "token",
              "type": "String",
              "repeated": false
            },
            ...
            {
              "name": "sms_delivery_type",
              "type": "com.squareup.protos.cash.raven.api.SmsDeliveryType",
              "repeated": false
            },
           ...
         ]
       },
       "com.squareup.protos.cash.raven.api.SmsDeliveryType": [
           "SMS", "VOICE"
         ]
   }
...
}
```

adrw:
Screen Shot 2019-10-09 at 13 11 11

### DIFF
--- a/misk/src/main/kotlin/misk/web/RequestTypes.kt
+++ b/misk/src/main/kotlin/misk/web/RequestTypes.kt
@@ -93,15 +93,34 @@ class RequestTypes {
         // TODO: Support maps
         fields.add(Field(fieldName, fieldClass.qualifiedName!!, repeated))
       }
+      Enum::class -> {
+        handleEnumField(fieldClass, fields, fieldName, repeated)
+      }
       else -> {
         if (fieldClass.superclasses.contains(WireEnum::class)) {
-          // TODO: Communicate back enum values
-          fields.add(Field(fieldName, fieldClass.qualifiedName!!, repeated))
+          handleEnumField(fieldClass, fields, fieldName, repeated)
         } else {
           fields.add(Field(fieldName, fieldClass.qualifiedName!!, repeated))
           stack.push(fieldClass)
         }
       }
     }
+  }
+
+  /**
+   * Adds a field with a type that has the class name and enum values embedded
+   * Example: "Enum<app.cash.backfila.BackfillType,ISOLATED,PARALLEL>"
+   */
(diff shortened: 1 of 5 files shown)


misk/src/main/kotlin/misk/web/RequestTypes.kt
```diff
@@ -93,15 +93,34 @@ class RequestTypes {
         // TODO: Support maps
         fields.add(Field(fieldName, fieldClass.qualifiedName!!, repeated))
       }
+      Enum::class -> {
```
wesleyk: let's update RequestTypesTest

adrw: Will do!

adrw: Added!

misk/src/test/kotlin/misk/web/RequestTypesTest.kt
```diff
@@ -37,5 +37,8 @@ internal class RequestTypesTest {
     // Check repeated types
     assertThat(shipmentType.fields).contains(Field("notes", "String", true))
     assertThat(warehouseType.fields).contains(Field("alternates", Warehouse::class.qualifiedName!!, true))
+
+    // Check enum types
+    assertThat(shipmentType.fields).contains(Field("status", "Enum<${Shipment.State::class.qualifiedName!!},${Shipment.State.values().joinToString(",")}>", false))
```
wesleyk: /nit how about we actually assert the full values rather than programatically calculating it? I prefer that as being more assertive than having to reason about what the expression resolves to

PR #1221, opened 2019-10-09
Add TransformedType

yan:
TransformedType allows arbitrary transformations to entity columns.


misk-hibernate/src/main/kotlin/misk/hibernate/SessionFactoryService.kt
```diff
@@ -160,7 +165,32 @@ internal class SessionFactoryService(
     } else if (BoxedStringType.isBoxedString(field.type.kotlin)) {
       value.typeName = BoxedStringType::class.java.name
       value.setTypeParameter("boxedStringField", field)
+    } else {
+      for (annotation : Annotation in field.annotations) {
```
zcross: This reads like a search, can it be done without iteration (e.g., using a map type?)

misk-hibernate/src/main/kotlin/misk/hibernate/TransformedColumnType.kt
```diff
@@ -0,0 +1,122 @@
...
+import java.sql.PreparedStatement
+import java.sql.ResultSet
+import java.sql.Types
+import java.util.Properties
+import kotlin.reflect.KClass
+import kotlin.reflect.full.primaryConstructor
+
+internal class TransformedColumnType : UserType, ParameterizedType, TypeConfigurationAware {
+  companion object {
+    val TRANSFORMER_CLASS = "transformerClass"
```
zcross: Maybe I haven't digested this enough yet, but do you have a convention in mind for versioning transformers? I.e., should transformer classes never be deleted or changed, and all changes yield a new class? Or should transformers be version aware to enable backwards compatible changes to transformer business logic?

misk-hibernate/src/test/kotlin/misk/hibernate/TransformedColumnTest.kt
```diff
@@ -0,0 +1,451 @@
...
+class DelegatingTransformer(private val ctx: TransformerContext) : Transformer(ctx) {
+  @Inject
+  lateinit var transformer: SwappableTransformer
+
+  override fun assemble(owner: Any?, value: Serializable): Any = transformer.assemble(ctx, owner, value)
+  override fun disassemble(value: Any): Serializable = transformer.disassemble(ctx, value)
+}
+
+@MiskTest(startService = true)
+class TransformedColumnTest {
```
zcross: I'm still ramping up on misk, and on top of that I've mainly used jOOQ and very little hibernate.
Between misk and hibernate, are there good facilities for observing queries made in test so that you can assert things like "exactly 1 query was made against this table," etc? That sort of assertion may be very valuable for a generic/automagic feature like this. Apologies if that's already tested for here and I just haven't recognized it yet :)

misk-hibernate/src/main/kotlin/misk/hibernate/SessionFactoryService.kt
yan: I'm not sure if it can since I need to check if any of those annotations have an annotation on them; i can't just query by type since they can be any type. I'll look into other mechanisms for getting that list however

zcross: Still getting used to the GitHub UI (I'm used to Bitbucket), so maybe I missed the broader context. Is this happening once (e.g., at app startup), or at the rate of DB connections, or some other repeated frequency? If it's repeated, can the results of this lookup be memoized by field?

yan: This gets invoked when Hibernate is initialized.

misk-hibernate/src/main/kotlin/misk/hibernate/TransformedColumnType.kt
yan: Versioning would happen on top of transformers; this makes it able to create something like a `JsonColumn` or `ProtobufColumn` external to misk-hibernate.

misk-hibernate/src/test/kotlin/misk/hibernate/TransformedColumnTest.kt
zcross: That seems sufficient from a correctness standpoint, performance related assertions may be overkill at this point

misk-hibernate/src/main/kotlin/misk/hibernate/TransformedColumnType.kt
```diff
@@ -0,0 +1,122 @@
...
+
+  override fun returnedClass(): Class<*> = retClass.java
+
+  override fun assemble(cached: Serializable, owner: Any): Any = transformer.assemble(owner, cached)
+
+  override fun disassemble(value: Any): Serializable = transformer.disassemble(value)
+
+  override fun nullSafeSet(st: PreparedStatement, value: Any?, index: Int, session: SharedSessionContractImplementor?) {
+    if (value == null) {
+      // TODO: Replace this with actual class
```
zcross: Blocking TODO?

yan: That's actually a stale comment (the actual class is used). Just pushed to remove.

PR #1223, opened 2019-10-10
http metrics do not pre-aggregate

mightyguava:
We pre-aggregated status codes to `all`, `4xx`, `5xx` etc. This caused
confusion for users when they tried to do aggregations on the metric
like `sum(rate(http_request_latency_ms_count[1m]))` because it ended up
tripling the rate.

This PR removes pre-aggregation, prefering to let the monitoring system
to do aggregation.


tirsen:
yay!

PR #1224, opened 2019-10-11
Add count/latency metric for PendingTasks

ryanhall07:
Also added a RepeatedTaskQueueFactory to make it easier for clients to
construct a RepeatedTaskQueue. In the future we don't have to update
every call site when we add a new dependency to RepeatedTaskQueue.

### DIFF
--- a/misk-aws/src/main/kotlin/misk/jobqueue/sqs/AwsSqsJobQueueModule.kt
+++ b/misk-aws/src/main/kotlin/misk/jobqueue/sqs/AwsSqsJobQueueModule.kt
@@ -17,6 +17,7 @@ import misk.jobqueue.JobQueue
 import misk.jobqueue.QueueName
 import misk.jobqueue.TransactionalJobQueue
 import misk.tasks.RepeatedTaskQueue
+import misk.tasks.RepeatedTaskQueueFactory
 import java.time.Clock
 import java.util.concurrent.Executors
 import javax.inject.Inject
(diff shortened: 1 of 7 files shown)


misk/src/main/kotlin/misk/tasks/RepeatedTaskQueue.kt
```diff
@@ -31,22 +37,30 @@ class RepeatedTaskQueue @VisibleForTesting internal constructor(
...
     clock: Clock,
     taskExecutor: ExecutorService,
     config: RepeatedTaskQueueConfig = RepeatedTaskQueueConfig()
   ) :
-      this(name, clock, taskExecutor, null, DelayQueue<DelayedTask>(), config)
+      this(name, clock, taskExecutor, null, DelayQueue<DelayedTask>(), null, config)
 
   private val running = AtomicBoolean(false)
+  internal val taskLatency = metrics?.histogram(
+      "task_latency",
```
mightyguava: This isn’t latency, it’s duration. task_duration

Also I wonder if there should be a prefix: task_queue_task_duration, since Prometheus metrics aren’t namespaced

ryanhall07: I wonder if we should have prefixed all misk metrics with `misk_`. oh well

misk-aws/src/main/kotlin/misk/jobqueue/sqs/AwsSqsJobQueueModule.kt
```diff
@@ -68,13 +69,10 @@ class AwsSqsJobQueueModule(
...
   @Provides @ForSqsConsumer @Singleton
-  fun consumerRepeatedTaskQueue(clock: Clock, config: AwsSqsJobQueueConfig): RepeatedTaskQueue {
-    return RepeatedTaskQueue(
+  fun consumerRepeatedTaskQueue(queueFactory: RepeatedTaskQueueFactory,
+    config: AwsSqsJobQueueConfig): RepeatedTaskQueue {
+    return queueFactory.new(
         "sqs-consumer-poller",
-        clock,
-        Executors.newCachedThreadPool(ThreadFactoryBuilder()
-            .setNameFormat("sqs-consumer-%d")
```
ryanhall07: the name of the thread is going to change to `sqs-consumer-poller-%d`. shouldn't be a big deal

misk-zookeeper/src/main/kotlin/misk/clustering/zookeeper/ZkLeaseModule.kt
```diff
@@ -22,7 +20,6 @@ import javax.inject.Singleton
 class ZkLeaseModule : KAbstractModule() {
   override fun configure() {
     install(ZkLeaseCommonModule())
-    install(ExecutorServiceModule.withFixedThreadPool(ForZkLease::class, "zk-lease-poller", 1))
```
ryanhall07: the `RepeatedQueueFactory` uses a `singleThreadedExecutor`, which is the same as fixed size 1

PR #1225, opened 2019-10-11
Allow a configurable num_of_parallel tasks

ryanhall07:


ryanhall07:
> This may reveal race conditions? Might be worth an announcement.

this is just preserving existing behavior for clients that are already using multiple threads. before I introduced the `Factory`, clients were directly passing the `Executor` they wanted to use. most were passing a single thread (which is the default config), but some were relying on multiple threads. didn't realize that until now.

PR #1228, opened 2019-10-11
DashboardMetadataAction to support multiple tab groupings

adrw:
* Deprecate `AdminDashboardTabAction` in favor of more generic `DashboardMetadataAction` that returns a list of authenticated dashboard tabs per a path parameter for the dashboard
* For example, this now supports from one endpoint, returning tabs for the `_admin/` dashboard and an `app/` dashboard


adrw:
Going to split into multiple PRs and reissue

PR #1231, opened 2019-10-14
Support multiple dashboards with new DashboardMetadataAction

adrw:
* Action returns a list of authenticated dashboard tabs per a path parameter for the dashboard
* This allows support from one endpoint, of tabs for the `_admin/` dashboard and an `app/` dashboard


misk/src/main/kotlin/misk/web/DashboardTab.kt
```diff
@@ -8,47 +8,89 @@ import kotlin.reflect.KClass
...
+  slug = slug,
+  url_path_prefix = url_path_prefix,
+  name = name,
+  category = category,
+  dashboardAnnotation = DA::class,
+  accessAnnotation = AA::class
+)
+
+/** Binds a DashboardTab for Dashboard Group annotation [DA] */
+inline fun <reified DA : Annotation> DashboardTabProvider(
```
wesleyk: what is the access annotation used in this case? I find it rather confusing having two flavors

misk/src/main/kotlin/misk/web/DashboardTab.kt
```diff
@@ -8,47 +8,89 @@ import kotlin.reflect.KClass
 class DashboardTab(
```
wesleyk: any reason these aren't data classes?

misk/src/main/kotlin/misk/web/DashboardTab.kt
adrw: Binding capabilities/service authentication is optional on a DashboardTab but Kotlin doesn't allow for an optional reified type parameter in the function signature, hence the unfortunate two flavors.

misk/src/main/kotlin/misk/web/DashboardTab.kt
adrw: It doesn't seem to work with the way it extends WebTab class which also extends ValidWebEntry

wesleyk: gotcha

misk/src/main/kotlin/misk/web/actions/DashboardMetadataAction.kt
```diff
@@ -23,24 +24,27 @@ import javax.inject.Singleton
...
 
 @Singleton
-class AdminDashboardTabAction @Inject constructor() : WebAction {
-  @Inject @AdminDashboardTab private lateinit var adminDashboardTabs: List<DashboardTab>
+class DashboardMetadataAction @Inject constructor() : WebAction {
+  @Inject private lateinit var allDashboardTabs: List<DashboardTab>
   @Inject lateinit var callerProvider: @JvmSuppressWildcards ActionScoped<MiskCaller?>
 
-  @Get("/api/admindashboardtabs")
+  @Get("/api/dashboard/metadata/{dashboard}")
```
rharter: Would it make more sense for this to be `/api/dashboard/{dashboard}/metadata`, or even `/api/dashboard/{dashboard}/tabs`?

adrw: The path continues the pattern of `ServiceMetadataAction` which has the path `/api/service/metadata`, hope that makes sense!

rharter: That makes sense re: `/metadata` vs. `/tabs`, but I think the rearrangement still stands.

The `/api/service/metadata` is returning metadata for all services, not a specific one. Now that the `DashboardMetadataAction` is returning results for a **specific** dashboard, as opposed to all admin dashboard tabs, IMHO it would make sense for the URI to follow the `parent/child` model.  In this case, the metadata you're requesting is a child of the identified dashboard.

The current structure, `/api/dashboard/metadata/{dashboard}`, is effectively using the dashboard path param as a filter param.  The alternative, `/api/dashboard/{dashboard}/metadata`, communicates that you're accessing child properties of the dashboard identified by the path param.

adrw: Thanks for expanding your thoughts on this, I think they make sense. I've updated the path accordingly to now be `/api/dashboard/{dashboard}/metadata`.

misk/src/test/kotlin/misk/web/actions/DashboardTabTest.kt
```diff
@@ -1,67 +1,67 @@
 package misk.web.actions
 
-import misk.web.DashboardTab
+import misk.web.DashboardTabProvider
 import org.junit.jupiter.api.Test
 import kotlin.test.assertFailsWith
 
 internal class DashboardTabTest {
```
wesleyk: gonna rework this a bit, looks like it's failing now

misk/src/main/kotlin/misk/web/DashboardTab.kt
```diff
@@ -1,46 +1,78 @@
...
- * Sets the tab's authentication based on the injected AdminDashboardAccess access annotation entry
+ * Sets the tab's dashboardId by annotation and authentication by injected access annotation entry
  */
 class DashboardTabProvider(
   val slug: String,
   val url_path_prefix: String,
   val name: String,
   val category: String = "Admin",
-  val accessAnnotation: KClass<out Annotation>
+  val dashboardSlug: String,
```
wesleyk: hmm can we be consistent on snake case and camelcase

misk/src/main/kotlin/misk/web/DashboardTab.kt
```diff
@@ -1,46 +1,78 @@
...
 import javax.inject.Provider
 import kotlin.reflect.KClass
 
+/**
+ * A [WebTab] with additional fields to bind to a specific DashboardId, with a menu name and category
+ */
 class DashboardTab(
   slug: String,
   url_path_prefix: String,
+  val dashboardSlug: String,
```
wesleyk: the distinction of slug to dashboardSlug is confusing to me as a reader. Is slug the slug of the dashboard tab, and dashboardSlug the slug of its corresponding dashboard?

misk/src/test/kotlin/misk/web/actions/DashboardTabTest.kt
adrw: Thanks for fixing this!

PR #1238, opened 2019-10-14
Add home_url, navbar_items, and navbar_status support

adrw:
* Support multibindings and return in DashboardMetadataAction `navbar_items` to add dynamic navbar elements, and `navbar_status` to set status text below the navbar

### DIFF
--- a/misk/src/main/kotlin/misk/web/actions/DashboardMetadataAction.kt
+++ b/misk/src/main/kotlin/misk/web/actions/DashboardMetadataAction.kt
@@ -25,24 +25,94 @@ import javax.inject.Singleton
 
 @Singleton
 class DashboardMetadataAction @Inject constructor() : WebAction {
-  @Inject private lateinit var allDashboardTabs: List<DashboardTab>
+  @Inject private lateinit var allTabs: List<DashboardTab>
+  @Inject private lateinit var allNavbarItems: List<DashboardNavbarItem>
+  @Inject private lateinit var allNavbarStatus: List<DashboardNavbarStatus>
+  @Inject private lateinit var allHomeUrls: List<DashboardHomeUrl>
   @Inject lateinit var callerProvider: @JvmSuppressWildcards ActionScoped<MiskCaller?>
 
-  @Get("/api/dashboard/metadata/{dashboard}")
+  @Get("/api/dashboard/metadata/{dashboardId}")
   @RequestContentType(MediaTypes.APPLICATION_JSON)
   @ResponseContentType(MediaTypes.APPLICATION_JSON)
   @Unauthenticated
   fun getAll(
-    @PathParam dashboard: String
+    @PathParam dashboardId: String
   ): Response {
     val caller = callerProvider.get() ?: return Response()
-    val dashboardTabs = allDashboardTabs.filter { it.dashboard == dashboard }
-    val authorizedDashboardTabs =
-      dashboardTabs.filter { caller.isAllowed(it.capabilities, it.services) }
-    return Response(tabs = authorizedDashboardTabs)
+
+    val authorizedDashboardTabs = allTabs
+      .filter { it.dashboardId == dashboardId }
+      .filter { caller.isAllowed(it.capabilities, it.services) }
+
+    val homeUrl = allHomeUrls
+      .find { it.dashboardId == dashboardId }?.url ?: ""
+
+    val navbarItems = allNavbarItems
+      .filter { it.dashboardId == dashboardId }
+      .sortedBy { it.order }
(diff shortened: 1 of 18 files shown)


misk/src/main/kotlin/misk/web/actions/DashboardMetadataAction.kt
```diff
@@ -36,13 +38,55 @@ class DashboardMetadataAction @Inject constructor() : WebAction {
...
+
+    val dashboardTabs = allTabs.filter { it.dashboard == dashboard }
     val authorizedDashboardTabs =
       dashboardTabs.filter { caller.isAllowed(it.capabilities, it.services) }
-    return Response(tabs = authorizedDashboardTabs)
+
+    val navbarItems = allNavbarItems
+      .filter { it.dashboard == dashboard }
+      .sortedBy { it.order }
+      .map {it.item}
```
zcross: nit: `.map { it.item }` (whitespace)

misk/src/main/kotlin/misk/web/actions/DashboardMetadataAction.kt
```diff
@@ -36,13 +38,55 @@ class DashboardMetadataAction @Inject constructor() : WebAction {
...
+
+    val navbarStatus = allNavbarStatus.find { it.dashboard == dashboard }?.status ?: ""
```
zcross: nit: imo, more readable to add a few newlines (e.g., before `.find` and before `.status`

misk/src/main/kotlin/misk/web/actions/DashboardMetadataAction.kt
```diff
@@ -36,13 +38,55 @@ class DashboardMetadataAction @Inject constructor() : WebAction {
...
+
+    return Response(navbar_items = navbarItems, navbar_status = navbarStatus, tabs = authorizedDashboardTabs)
   }
 
-  data class Response(val tabs: List<DashboardTab> = listOf())
+  data class DashboardNavbarItem(
+    val dashboard: String,
```
zcross: `dashboardLabel` or `dashboardTitle`?

misk/src/main/kotlin/misk/web/actions/DashboardMetadataAction.kt
```diff
@@ -36,13 +38,55 @@ class DashboardMetadataAction @Inject constructor() : WebAction {
...
+    val navbar_status: String = "",
+    val tabs: List<DashboardTab> = listOf()
+  )
+
+  companion object {
+    inline fun <reified DA : Annotation> DashboardNavbarItem(
+      item: String,
+      order: Int
+    ) = DashboardNavbarItem(
+      dashboard = DA::class.simpleName!!,
```
zcross: This reads like we're creating a label from a class name, can we just make an optional field in which the dashboard can provide a human readable label? It could fallback on `simpleName` if not implemented?

misk/src/test/kotlin/misk/web/actions/DashboardMetadataActionTest.kt
```diff
@@ -63,6 +63,28 @@ class DashboardMetadataActionTest {
...
 
+  @Test fun loadDashboardNavbarItems() {
+    val dashboard = DashboardMetadataActionTestDashboard::class.simpleName!!
+    val response = executeRequest(
+      path = path + dashboard,
+      user = "sandy",
+      capabilities = "test_admin_access"
+    )
+    assertEquals(1, response.navbar_items.size)
+    assertNotNull(response.navbar_items.find { it.contains("Test Navbar Link") })
```
zcross: Would be nice to use a non-empty href and assert that the anchor tag is formed as expected with the correct href

misk/src/main/kotlin/misk/web/actions/DashboardMetadataAction.kt
adrw: Updated to `dashboardId`!

misk/src/main/kotlin/misk/web/actions/DashboardMetadataAction.kt
adrw: Updated and added documentation so that pattern is more apparent

misk/src/test/kotlin/misk/web/actions/DashboardMetadataActionTest.kt
adrw: I'll update it to a full `assertEquals` snapshot since the main test is that the bound string is present, not necessarily that it is a `<a>` link.

Issue #1119, opened 2019-07-25
Hibernate count queries


mark-karantayer:
I've been looking through misk and haven't found any mapping for hibernate count queries. Did I overlook something?

Is there a design reason for that? If not, I'd be happy to throw up a PR.

jjestrel:
it's currently missing, `ReflectionQueryFactory` is the place to start if you're interested in adding it

PR #1241, opened 2019-10-16
[Misk #1119] Support for count queries in Query abstraction

wesleyk:


misk-hibernate/src/main/kotlin/misk/hibernate/ReflectionQueryFactory.kt
```diff
@@ -104,19 +102,36 @@ internal class ReflectionQuery<T : DbEntity<T>>(
...
       }
     }
     checkRowCount(returnList, rows.size)
     return rows
   }
 
+  override fun count(session: Session): Long {
+    check(!predicatesOnly) { "cannot select on this query" }
+
+    check(orderFactories.size == 0) { "orderBy shouldn't be used for a count" }
```
cwmyers: 👍

PR #1243, opened 2019-10-17
Tighter timeouts

tirsen:
We drop down the timeouts significantly so that can we quicker react to
issues around blackholed vtgates in Kubernetes and decrease downtime.


misk-hibernate/src/main/kotlin/misk/jdbc/PingDatabaseService.kt
```diff
@@ -44,6 +44,7 @@ class PingDatabaseService @Inject constructor(
   }
 
   private fun connectToDataSource(dataSource: DriverDataSource) {
+    dataSource.setLoginTimeout((config.connection_timeout.toMillis() / 1000).toInt())
```
swankjesse: If we move Misk to Java 11 (we should if we haven't already) this can be toSeconds().

tirsen: I know I tried that first. It failed in the javadocs.

misk-hibernate/src/main/kotlin/misk/jdbc/DataSourceConfig.kt
```diff
@@ -41,8 +41,9 @@ data class DataSourceConfig(
   val username: String? = null,
   val password: String? = null,
   val fixed_pool_size: Int = 10,
-  val connection_timeout: Duration = Duration.ofSeconds(30),
-  val connection_max_lifetime: Duration = Duration.ofMinutes(30),
+  val connection_timeout: Duration = Duration.ofSeconds(10),
+  val validation_timeout: Duration = Duration.ofSeconds(3),
+  val connection_max_lifetime: Duration = Duration.ofMinutes(1),
```
ryanhall07: we should probably set the `idleTimeout` as well, since we are changing the `maxLifetime`

PR #1245, opened 2019-10-22
Add mget and mset methods

tlulu:


tlulu:
https://redis.io/commands/mget#examples
https://redis.io/commands/mset#examples

PR #1246, opened 2019-10-22
Fix flaky SqsJobQueueServiceTest

hamdanjaveed:
The test was flaky because it wasn't guaranteed that the JobConsumer had had a chance to run the job
in between the polls. The fix here is to add a CountDownLatch that'll wait until the job has had a
chance to run.

To test flaky-ness, I ran this test 300 times without failure.

image


misk-aws/src/test/kotlin/misk/jobqueue/sqs/SqsJobQueueServiceTest.kt
```diff
@@ -46,33 +46,36 @@ internal class SqsJobQueueServiceTest {
     serviceManager.awaitStopped(20, TimeUnit.SECONDS)
   }
 
-  @Test fun jobsNotHandledUntilAllServicesAreRunning() {
+  @Test fun `jobs are not handled until all services are running`() {
     val log = LinkedBlockingDeque<String>()
+    val jobRan = CountDownLatch(1)
 
-    sleep(100)
```
adrw: Great to remove all these racey sleeps!

PR #1249, opened 2019-10-23
Upgrade Mockito to 3.1.0

swankjesse:
https://github.com/mockito/mockito/blob/release/3.x/doc/release-notes/official.md


swankjesse:
Changed from Mockito 3.1.2 to Mockito 3.1.0 because they don’t publish all releases to Maven Central.
https://search.maven.org/search?q=g:org.mockito%20AND%20a:mockito-core
https://github.com/mockito/mockito#current-version-is-3x

PR #1252, opened 2019-10-23
Add an idempotenceKey to the JobQueue interface

ryanhall07:
Most likely the underlying queueing system (e.g SQS) supports at least
once delivery of a message. Consumers are going to need an idempotence
key in order to filter duplicate messages.

The eventual goal is to require the idempotence key, but for backwards
compatibility a random token is generated for messages without one. Once
we upgrade all clients to the new methods (and deploy the changes) we
can remove the old methods.

For the SQS impl, the idempotence key is added as a message attribute.

### DIFF
--- a/misk-aws/src/main/kotlin/misk/jobqueue/sqs/SqsJob.kt
+++ b/misk-aws/src/main/kotlin/misk/jobqueue/sqs/SqsJob.kt
@@ -9,13 +9,19 @@ internal class SqsJob(
   override val queueName: QueueName,
   private val queues: QueueResolver,
   private val metrics: SqsMetrics,
-  private val message: Message
+  private val message: Message,
+    // TODO: remove after this change has been deployed everywhere. This handles the case of existing
+    // queued jobs during the deploy.
+  randomIdempotenceKey: String
 ) : Job {
   override val body: String = message.body
   override val id: String = message.messageId
-  override val attributes: Map<String, String> = message.messageAttributes.map { (key, value) ->
-    key to value.stringValue
-  }.toMap()
+  override val idempotenceKey =
+      message.messageAttributes[IDEMPOTENCY_KEY_ATTR]?.stringValue ?: randomIdempotenceKey
+  override val attributes: Map<String, String> = message.messageAttributes
+      .map { (key, value) -> key to value.stringValue }
+      .filter { (key, _) -> key != IDEMPOTENCY_KEY_ATTR }
+      .toMap()
 
   private val queue: ResolvedQueue = queues[queueName]
 
@@ -40,5 +46,6 @@ internal class SqsJob(
 
   companion object {
     const val ORIGINAL_TRACE_ID_ATTR = "x-original-trace-id"
+    const val IDEMPOTENCY_KEY_ATTR = "idempotencey_key"
   }
 }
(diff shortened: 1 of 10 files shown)


misk-aws/src/main/kotlin/misk/jobqueue/sqs/SqsJob.kt
```diff
@@ -40,5 +46,6 @@ internal class SqsJob(
 
   companion object {
     const val ORIGINAL_TRACE_ID_ATTR = "x-original-trace-id"
+    const val IDEMPOTENCY_KEY_ATTR = "idempotencey_key"
```
wesleyk: idempotency_key

misk-aws/src/main/kotlin/misk/jobqueue/sqs/SqsJobQueue.kt
```diff
@@ -15,10 +16,20 @@ import javax.inject.Singleton
...
-  private val tracer: Tracer
+  private val tracer: Tracer,
+  private val tokenGenerator: TokenGenerator
 ) : JobQueue {
+
+  override fun enqueue(
+    queueName: QueueName,
+    body: String,
+    deliveryDelay: Duration?,
+    attributes: Map<String, String>
```
wesleyk: worth validating that idempotency key isn't provided?

misk-aws/src/main/kotlin/misk/jobqueue/sqs/SqsJob.kt
ryanhall07: lol i probably triple checked the spelling

misk-aws/src/test/kotlin/misk/jobqueue/sqs/SqsJobQueueTest.kt
```diff
@@ -90,10 +98,10 @@ internal class SqsJobQueueTest {
     )
 
     // Confirm metrics
-    assertThat(sqsMetrics.jobsEnqueued.labels(queueName.value).get()).isEqualTo(10.0)
+    assertThat(sqsMetrics.jobsEnqueued.labels(queueName.value).get()).isEqualTo(20.0)
     assertThat(sqsMetrics.jobEnqueueFailures.labels(queueName.value).get()).isEqualTo(0.0)
-    assertThat(sqsMetrics.jobsReceived.labels(queueName.value).get()).isEqualTo(10.0)
-    assertThat(sqsMetrics.jobsAcknowledged.labels(queueName.value).get()).isEqualTo(10.0)
+    assertThat(sqsMetrics.jobsReceived.labels(queueName.value).get()).isEqualTo(20.0)
+    assertThat(sqsMetrics.jobsAcknowledged.labels(queueName.value).get()).isEqualTo(20.0)
```
wesleyk: ah this makes sense. SQS would double count the metrics and the app itself will be responsible for filtering

misk-aws/src/main/kotlin/misk/jobqueue/sqs/SqsJobQueue.kt
ryanhall07: good point. should probably also namespace the key like `_misk_idempotency_key`

PR #1254, opened 2019-10-28
tracing: do not extract span from headers if already done

mightyguava:
Datadog has reflection + byte-code generation magic to do this automatically, I think https://github.com/DataDog/dd-trace-java/blob/master/dd-java-agent/instrumentation/jetty-8/src/main/java/datadog/trace/instrumentation/jetty8/JettyHandlerInstrumentation.java


mightyguava:
Not sure if this'll actually work, their impl is for jetty 8 and we are on jetty 9

mightyguava:
> Not sure if this'll actually work, their impl is for jetty 8 and we are on jetty 9

Confirmed with datadog that there's jetty 9 support

PR #1257, opened 2019-10-29
dynamic flag wrapper for feature flags

mightyguava:
Convenience wrapper for feature flags where we just want to apply config
values and stable percentage rollouts are not needed.


misk-launchdarkly-core/src/main/kotlin/misk/feature/launchdarkly/LaunchDarklyFeatureFlags.kt
```diff
@@ -111,6 +111,9 @@ class LaunchDarklyFeatureFlags @Inject constructor(
     if (attributes.number != null) {
       attributes.number!!.forEach { (k, v) -> builder.privateCustom(k, v) }
     }
+    if (attributes.anonymous) {
+      builder.anonymous(true)
```
mightyguava: this makes it so that launchdarkly doesn't create a "User" on their side

misk-feature/src/main/kotlin/misk/feature/DynamicFlags.kt
```diff
@@ -0,0 +1,31 @@
+package misk.feature
+
+/**
+ * Interface for evaluating dynamic flags. Dynamic flags are similar to feature flags, but they
+ * don't support different variations for different keys.
+ */
+interface DynamicFlags {
```
AlexSzlavik: I don't love the naming. We already have a concept of DynamicFlags. Could we not introduce a different name for this to avoid confusion?

misk-feature/src/main/kotlin/misk/feature/FeatureFlags.kt
```diff
@@ -70,5 +70,6 @@ data class Attributes(
   // NB: LaunchDarkly uses typed Gson attributes. We could leak that through, but that could make
   // code unwieldly. Numerical attributes are likely to be rarely used, so we make it a separate,
   // optional field rather than trying to account for multiple possible attribute types.
-  val number: Map<String, Number>? = null
+  val number: Map<String, Number>? = null,
+  val anonymous: Boolean = false
```
AlexSzlavik: I think this field would benefit from a little documentation. It is not clear that anonymous means that there are no attributes to be associated.

misk-launchdarkly-core/src/main/kotlin/misk/feature/launchdarkly/LaunchDarklyFeatureFlags.kt
AlexSzlavik: I think this is worth noting in a comment

misk-feature/src/main/kotlin/misk/feature/DynamicFlags.kt
mightyguava: This is meant to be a convenience class to make feature flags compatible with dynamic flags. What do you suggest?

ryanhall07: `DynamicConfig`?

mightyguava: DynamicConfig it is then

PR #1262, opened 2019-10-31
Revert "Add an idempotenceKey to the JobQueue interface (#1252)"

ryanhall07:
This reverts commit 44e18ba2a63d8b469b3c7da265909c46060d5b50.


ryanhall07:
nobody should be relying on these new interfaces, so it should be safe to revert

PR #1272, opened 2019-11-05
Handle graceful vttablet terminations

tirsen:


misk-hibernate/src/main/kotlin/misk/hibernate/RealTransacter.kt
```diff
@@ -321,11 +321,33 @@ internal class RealTransacter private constructor(
...
-      // See com/zaxxer/hikari/pool/ProxyConnection.java:493
+      isConnectionClosed(th) || isVitessTransactionNotFound(th)
+
+  /**
+   * This is thrown as a raw SQLException from Hikari even though it is most certainly a
+   * recoverable exception.
+   * See com/zaxxer/hikari/pool/ProxyConnection.java:493
+   */
+  private fun isConnectionClosed(th: SQLException) =
       th.message.equals("Connection is closed")
```
gak: Could `th.message` be null here, as checked below?

tirsen: Kotlin isn't complaining (otherwise it would be a compile time error) so I think `equals` is defined on `String?` (i.e. nullable String).

PR #1273, opened 2019-11-05
bump datadog to 0.35.0

mightyguava:


mightyguava:
This only bumps the opentracing bindings. The agent was already bumped 0.35.0 a while ago

PR #1278, opened 2019-11-08
add validation to feature flag key

mightyguava:
Only allow things that look like keys to be passed in to `FeatureFlags`. I'm sure this validation doesn't really belong in misk, but the effort to make this configurable is a bit high and not worthwhile.

I'm also not sure we _should_ have this level of validation. Theoretically users should be testing their code and making sure they are passing in the right key. But, evidence shows that this hasn't been happening. The risk of a bug here is high, since we could potentially launch features to the wrong customers, or worse accidentally send PII to the feature flag vendor when we don't mean to.

So I think this validation is a good bad way of preventing mistakes.

### DIFF
--- a/misk-feature-testing/src/main/kotlin/misk/feature/testing/FakeFeatureFlags.kt
+++ b/misk-feature-testing/src/main/kotlin/misk/feature/testing/FakeFeatureFlags.kt
@@ -4,6 +4,7 @@ import com.google.common.util.concurrent.AbstractIdleService
 import misk.feature.Attributes
 import misk.feature.DynamicConfig
 import misk.feature.Feature
+import misk.feature.FeatureFlagValidation
 import misk.feature.FeatureFlags
 import misk.feature.FeatureService
 import java.util.concurrent.ConcurrentHashMap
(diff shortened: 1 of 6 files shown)


misk-feature-testing/src/main/kotlin/misk/feature/testing/FakeFeatureFlags.kt
```diff
@@ -56,12 +57,14 @@ class FakeFeatureFlags @Inject constructor() : AbstractIdleService(),
...
 
   private fun get(feature: Feature, key: String): Any? {
+    FeatureFlagValidation.checkValidKey(feature, key)
     return overrides.getOrElse(MapKey(feature, key)) {
       overrides[MapKey(feature)]
     }
   }
 
   private fun <V> getOrDefault(feature: Feature, key: String, defaultValue: V): V {
+    FeatureFlagValidation.checkValidKey(feature, key)
```
ryanhall07: nice

mightyguava:
> would it really be that much work to add a `FeatureFlagVaildator` interface that defaults to an `AllowAll` impl. You would then move your impl to `skim`

Spoke offline. At the moment Franklin can't pull in Misk guice modules for the most part. We'd need to publish a new misk jar and it's not really worth the effort.

PR #1283, opened 2019-11-13
Extract misk-events-core from misk-events

zcross:
Using misk-events directly from a project that doesn't use all of misk introduces an unnecessary dependency on misk-hibernate due to usage of hibernate's `Gid` and `Session` in `SpooledProducer`. This PR extracts an API-only model that does not depend on hibernate.

BTW: should this really be introduced at the top level like `root/misk-events-api` instead of `root/misk-events/api` or `root/misk-events/misk-events-api`?


zcross:
At @mightyguava 's suggestion, I'm going to rename this new subproject to `misk-events-core` for consistency (e.g., with `misk-launchdarkly-core`)

wesleyk:
misk-events-api maybe?

zcross:
Good follow on from this would be splitting the producer logic into its own project and moving the consumer into its own (or into core) because consuming does not require `misk-hibernate`

PR #1287, opened 2019-11-19
Explicitly override default attibutes in Launchdarkly

acostama:
Seeing this in deployed java app using launchdarkly. 
```
java.lang.AbstractMethodError: Receiver class misk.feature.launchdarkly.LaunchDarklyFeatureFlags 
does not define or inherit an implementation of the resolved method abstract 
getEnum(Lmisk/feature/Feature;Ljava/lang/String;Ljava/lang/Class;)Ljava/lang/Enum; of interface 
misk.feature.FeatureFlags.

```
Not sure how to confirm if this fixes it, but I assume explicitly overriding it will help. One thing to note, this was not showing up in tests.


mightyguava:
Hmm this doesn't quite make sense. I wonder if you have the wrong version of misk in your production jar. This PR fixed a wrong return type for the `getEnum()` call and could have potentially caused this problem if you had an older version https://github.com/cashapp/misk/pull/1276

PR #1290, opened 2019-11-21
Add chacterEncoding=utf8 on MySQL connection string

tso:


tirsen:
Can you reproduce the issue in a test?

misk-hibernate/src/test/kotlin/misk/hibernate/TransacterTest.kt
```diff
@@ -729,6 +729,13 @@ abstract class TransacterTest {
         "retried Movies transaction succeeded \\(attempt 3, same connection\\)")
   }
 
+  @Test
+  fun emojisCanBeInserted() {
```
tirsen: nit: use backticks \`emojis can be inserted\`

misk-hibernate/src/main/kotlin/misk/jdbc/DataSourceConfig.kt
```diff
@@ -120,6 +120,8 @@ data class DataSourceConfig(
     return when (type) {
       DataSourceType.MYSQL, DataSourceType.VITESS_MYSQL -> {
         var queryParams = "?useLegacyDatetimeCode=false"
+        queryParams += "&useUnicode=true"
+        queryParams += "&characterEncoding=utf8"
```
tirsen: it shouldn't be utf8mb4?

PR #1291, opened 2019-11-26
Widen acceptable characters in feature flag keys

swankjesse:
I've added everything in token68 for the same reasons it's used in HTTP
authentication challenges:

   The token68 syntax allows the 66 unreserved URI characters
   ([RFC3986]), plus a few others, so that it can hold a base64,
   base64url (URL and filename safe alphabet), base32, or base16 (hex)
   encoding, with or without padding, but excluding whitespace
   ([RFC4648]).

https://tools.ietf.org/html/rfc7235#section-2.1

### DIFF
--- a/misk-feature-testing/src/test/kotlin/misk/feature/testing/FakeFeatureFlagsTest.kt
+++ b/misk-feature-testing/src/test/kotlin/misk/feature/testing/FakeFeatureFlagsTest.kt
@@ -5,7 +5,6 @@ import misk.feature.getEnum
 import org.assertj.core.api.Assertions.assertThat
 import org.junit.jupiter.api.Test
 import org.junit.jupiter.api.assertThrows
-import java.lang.IllegalArgumentException
 
 internal class FakeFeatureFlagsTest {
   val FEATURE = Feature("foo")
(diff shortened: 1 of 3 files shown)


misk-feature-testing/src/test/kotlin/misk/feature/testing/FakeFeatureFlagsTest.kt
```diff
@@ -58,19 +57,31 @@ internal class FakeFeatureFlagsTest {
 
   @Test
   fun invalidKeys() {
+    subject.override(FEATURE, 3)
```
swankjesse: Previously this test wasn’t exercising what it wanted to exercise. Because the feature wasn’t provided, even if the token was valid these tests would have passed.

```
java.lang.IllegalArgumentException: String flag Feature(name=which-dinosaur) must be overridden with override() before use
	at misk.feature.testing.FakeFeatureFlags.getString(FakeFeatureFlags.kt:41)
	at misk.feature.FeatureFlags$DefaultImpls.getString(FeatureFlags.kt:67)
	at misk.feature.testing.FakeFeatureFlags.getString(FakeFeatureFlags.kt:18)
	at misk.feature.testing.FakeFeatureFlagsTest$invalidKeys$e$1.invoke(FakeFeatureFlagsTest.kt:62)
	at misk.feature.testing.FakeFeatureFlagsTest$invalidKeys$e$1.invoke(FakeFeatureFlagsTest.kt:10)
	at misk.feature.testing.FakeFeatureFlagsTest$inlined$sam$i$org_junit_jupiter_api_function_Executable$0.execute(Assertions.kt)
	at org.junit.jupiter.api.AssertThrows.assertThrows(AssertThrows.java:54)
	at org.junit.jupiter.api.AssertThrows.assertThrows(AssertThrows.java:36)
	at org.junit.jupiter.api.Assertions.assertThrows(Assertions.java:1179)
```

PR #1293, opened 2019-11-27
Allow for dollar signs in feature flag keys

hamdanjaveed:


mightyguava:
ugh, should i maybe just get rid of validation at this point...? Why is `$` needed? It's not used in any common encodings and is more on the side of something that could be passed in as a mistake imo.

PR #1295, opened 2019-11-27
Add an optional GzipHandler to JettyService configurable via WebConfig

tso:


misk/src/main/kotlin/misk/web/jetty/JettyService.kt
```diff
@@ -164,6 +167,13 @@ class JettyService @Inject internal constructor(
     server.stopTimeout = 25_000
     ServerConnectionStatistics.addToAllConnectors(server)
 
+    if (webConfig.gzip) {
+      gzipHandler.server = server
+      gzipHandler.minGzipSize = webConfig.minGzipSize
+      gzipHandler.addIncludedMethods("POST")
```
swankjesse: what does this do?

misk/src/test/kotlin/misk/web/GzipTest.kt
```diff
@@ -0,0 +1,90 @@
...
+  private fun call(request: Request.Builder): Response {
+    val httpClient = OkHttpClient()
+    val response = httpClient.newCall(request.build()).execute()
+    assertThat(response.code).isEqualTo(200)
+    return response
+  }
+
+  private fun Response.assertGzipEncoding(gzipped: Boolean) {
+    use {
+      assertThat(networkResponse?.headers?.toMap()?.get("Content-Encoding") == "gzip")
```
swankjesse: you can shrink this to `networkResponse?.headers?.get("Content-Encoding") == "gzip"`

swankjesse: (otherwise it needs to create a case-insensitive map!)

swankjesse:
Awesome. For follow-up at some point is accepting Gzipped requests.

misk/src/main/kotlin/misk/web/jetty/JettyService.kt
tso: By default the gzip handler only runs on GET requests. This tells it to also run on POSTs

PR #1299, opened 2019-11-29
WebActions: add a synthetic json binding for proto web actions

acostama:
For web actions that receive and/or respond with proto, add a second binding over json.

### DIFF
--- a/misk/src/main/kotlin/misk/web/actions/WebActionFactory.kt
+++ b/misk/src/main/kotlin/misk/web/actions/WebActionFactory.kt
@@ -3,6 +3,7 @@ package misk.web.actions
 import com.google.inject.Injector
 import com.google.inject.Provider
 import com.squareup.wire.WireRpc
+import misk.Action
 import misk.ApplicationInterceptor
 import misk.MiskDefault
 import misk.asAction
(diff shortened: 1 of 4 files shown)


misk/src/test/kotlin/misk/web/JsonForProtoEndpointsTest.kt
```diff
@@ -0,0 +1,102 @@
...
+  }
+
+  class TestModule : KAbstractModule() {
+    override fun configure() {
+      install(WebTestingModule())
+      install(WebActionModule.create<PostHello>())
+    }
+  }
+
+  class PostHello @Inject constructor() : WebAction {
```
swankjesse: Rename?

misk/src/main/kotlin/misk/web/actions/WebActionFactory.kt
```diff
@@ -76,35 +79,68 @@ internal class WebActionFactory @Inject constructor(
...
+   * Returns a copy of [action] that pretends JSON was the request and response content types.
+   * Returns null if neither of the input action's content types were protobuf.
+   */
+  private fun transformActionIntoJson(action: Action): Action? {
+    var jsonVariant = action
+
+    if (action.acceptedMediaRanges == protobufMediaRanges) {
+      jsonVariant = jsonVariant.copy(acceptedMediaRanges = jsonMediaRanges)
+    }
+    if (action.responseContentType == protobufMediaType) {
```
wesleyk: what if only one of these is hit? Do we still want the json variant? I think yes, because the goal is to have a purely json request/response flavor

wesleyk:
cc @zcross who was interested in this

PR #1301, opened 2019-12-04
Add sane jetty server defaults

jjestrel:


misk/src/main/kotlin/misk/web/WebConfig.kt
```diff
@@ -30,7 +30,10 @@ data class WebConfig(
   val queue_size: Int? = null,
 
   /** Maximum number of threads in Jetty's thread pool. */
-  val jetty_max_thread_pool_size: Int? = null,
+  val jetty_max_thread_pool_size: Int = 200,
```
jjestrel: This was the default value, and I shouldn't have made it nullable in the first place

misk/src/main/kotlin/misk/web/MiskWebModule.kt
```diff
@@ -150,7 +151,12 @@ class MiskWebModule(private val config: WebConfig) : KAbstractModule() {
 
   @Provides @Singleton
   fun provideJettyThreadPool(): QueuedThreadPool {
-    return config.jetty_max_thread_pool_size?.let { QueuedThreadPool(it) } ?: QueuedThreadPool()
+    val threadPool = QueuedThreadPool(config.jetty_max_thread_pool_size,
+        8,
```
jjestrel: These are the defaults today that jetty uses when we don't specify things

tlulu: can we use named params here

jjestrel: Added vals

misk/src/main/kotlin/misk/web/MiskWebModule.kt
```diff
@@ -150,7 +151,17 @@ class MiskWebModule(private val config: WebConfig) : KAbstractModule() {
...
+    val maxThreads = config.jetty_max_thread_pool_size
+    val minThreads = Math.min(8, maxThreads)
+    val idleTimeout = 60_000
+
+    val threadPool = QueuedThreadPool(
+        maxThreads,
+        minThreads,
+        idleTimeout,
+        ArrayBlockingQueue(config.jetty_max_thread_pool_queue_size))
+    threadPool.name = "jetty-thread"
```
ryanhall07: lol this can't be provided int he constructor?

jjestrel: No, but you can pass in a ThreadGroup with a name but that seems different than the actual name value that gets set in the `QueuedThreadPool`

misk/src/main/kotlin/misk/web/WebConfig.kt
```diff
@@ -30,16 +30,18 @@ data class WebConfig(
...
 
   val action_exception_log_level: ActionExceptionLogLevelConfig = ActionExceptionLogLevelConfig(),
 
   /** The maximum number of streams per HTTP/2 connection. */
   val jetty_max_concurrent_streams: Int? = null,
 
   /** A value in [0.0..100.0]. Include 'Connection: close' in this percentage of responses. */
-  // TODO(jayestrella): Add a sane default value to this.
-  val close_connection_percent: Double = 0.0,
+  val close_connection_percent: Double = 0.01,
```
swankjesse: will be really happy when we can remove this

misk/src/main/kotlin/misk/web/MiskWebModule.kt
```diff
@@ -150,7 +151,17 @@ class MiskWebModule(private val config: WebConfig) : KAbstractModule() {
...
-    return config.jetty_max_thread_pool_size?.let { QueuedThreadPool(it) } ?: QueuedThreadPool()
+    val maxThreads = config.jetty_max_thread_pool_size
+    val minThreads = Math.min(8, maxThreads)
+    val idleTimeout = 60_000
+
+    val threadPool = QueuedThreadPool(
+        maxThreads,
+        minThreads,
+        idleTimeout,
+        ArrayBlockingQueue(config.jetty_max_thread_pool_queue_size))
```
swankjesse: OMG there’s two different classes:
```
org.eclipse.jetty.util.BlockingArrayQueue
java.util.concurrent.ArrayBlockingQueue
```
I think you’re choosing the right one, despite the fact that it’s different from Jetty’s default.

PR #1304, opened 2019-12-05
Add a network interceptor that does concurrency limiting

swankjesse:
This uses the Netflix concurrency-limits library that we're already using
elsewhere at Square. It works great.
https://github.com/Netflix/concurrency-limits/

This is currently opt-in. In a follow-up I'd like to make it opt-out.

### DIFF
--- a/misk/src/main/kotlin/misk/web/Http.kt
+++ b/misk/src/main/kotlin/misk/web/Http.kt
@@ -32,3 +32,28 @@ annotation class RequestContentType(vararg val value: String)
 
 @Target(AnnotationTarget.FUNCTION)
 annotation class ResponseContentType(val value: String)
+
+/**
+ * When the service is overloaded Misk will intervene and reject calls by returning "HTTP 503
+ * Service Unavailable". We call this load shedding and it works similarly to flow control in TCP.
+ *
+ * We must not shed calls to status endpoints like health checks as doing so may create cascading
+ * failures: overloaded instances that do not respond to health checks will be killed, and this
+ * further overloads the remaining peers.
+ *
+ * Only put this on endpoints that must never be rejected. Such endpoints must not do RPCs, database
+ * queries or other I/O because unexpected latency there can take down the entire service.
+ */
+@Retention(AnnotationRetention.RUNTIME)
+@Target(AnnotationTarget.FUNCTION)
+annotation class AvailableWhenDegraded
+
+/**
+ * Opt-in to concurrency limits.
+ *
+ * TODO(jwilson): make this the default once we're comfortable with the behavior and remove this
+ *     annotation.
+ */
+@Retention(AnnotationRetention.RUNTIME)
+@Target(AnnotationTarget.FUNCTION)
+annotation class ConcurrencyLimitsOptIn
--- a/misk/src/main/kotlin/misk/web/interceptors/ConcurrencyLimitsInterceptor.kt
+++ b/misk/src/main/kotlin/misk/web/interceptors/ConcurrencyLimitsInterceptor.kt
@@ -1,0 +1,80 @@
+package misk.web.interceptors
+
+import com.netflix.concurrency.limits.Limiter
+import com.netflix.concurrency.limits.limiter.SimpleLimiter
+import misk.Action
+import misk.exceptions.StatusCode
+import misk.logging.getLogger
+import misk.web.AvailableWhenDegraded
+import misk.web.ConcurrencyLimitsOptIn
+import misk.web.NetworkChain
+import misk.web.NetworkInterceptor
+import java.time.Clock
+import javax.inject.Inject
+import javax.inject.Singleton
+import kotlin.reflect.full.findAnnotation
+
+/**
+ * Detects degraded behavior and sheds requests accordingly. Internally this uses adaptive limiting
+ * as implemented by Netflix's [concurrency-limits][concurrency_limits] library. See
+ * [AvailableWhenDegraded] for further documentation.
+ *
+ * [concurrency_limits]: https://github.com/Netflix/concurrency-limits/
+ */
+internal class ConcurrencyLimitsInterceptor internal constructor(
+  private val actionName: String,
+  private val limiter: Limiter<String>
+) : NetworkInterceptor {
+
+  override fun intercept(chain: NetworkChain) {
+    val listener: Limiter.Listener? = limiter.acquire(actionName).orElse(null)
+
+    if (listener == null) {
+      logger.error { "concurrency limits interceptor shedding $actionName" }
+      chain.httpCall.statusCode = StatusCode.SERVICE_UNAVAILABLE.code
+      chain.httpCall.takeResponseBody()?.use { sink ->
+        sink.writeUtf8("service unavailable")
+      }
+      return
+    }
+
+    try {
+      chain.proceed(chain.httpCall)
+    } catch (unexpected: Throwable) {
+      listener.onIgnore()
+      throw unexpected
+    }
+
+    try {
+      when (chain.httpCall.statusCode) {
+        in 400 until 500 -> listener.onIgnore()
+        in 500 until 600 -> listener.onDropped() // Count 5XX errors as drops.
+        else -> listener.onSuccess()
+      }
+    } catch (e: IllegalArgumentException) {
+      // Service container does a ignores an RTT == 0 exception here.
+      // TODO(jwilson): report this upstream and get it fixed.
+      logger.debug { "ignoring concurrency-limits exception: $e" }
+    }
+  }
+
+  @Singleton
+  class Factory @Inject constructor(val clock: Clock) : NetworkInterceptor.Factory {
+    override fun create(action: Action): NetworkInterceptor? {
+      if (action.function.findAnnotation<AvailableWhenDegraded>() != null) return null
+
+      // TODO(jwilson): make this the default behavior and remove this annotation.
+      if (action.function.findAnnotation<ConcurrencyLimitsOptIn>() == null) return null
+
+      val limiter = SimpleLimiter.Builder()
+          .clock { clock.millis() }
+          .build<String>()
+
+      return ConcurrencyLimitsInterceptor(action.name, limiter)
+    }
+  }
+
+  private companion object {
+    val logger = getLogger<ConcurrencyLimitsInterceptor>()
+  }
+}
--- a/misk/src/test/kotlin/misk/time/wait_notify.kt
+++ b/misk/src/test/kotlin/misk/time/wait_notify.kt
@@ -1,0 +1,35 @@
+package misk.time
+
+/**
+ * Wait until the duration has elapsed. Unlike [java.lang.Object.wait] this interprets 0 as
+ * "don't wait" instead of "wait forever".
+ */
+fun Any.waitNanosIgnoreNotifies(nanos: Long) {
+  check(nanos >= 0)
+  var now = System.nanoTime()
+  val waitUntil = now + nanos
+  while (now < waitUntil) {
+    waitNanos(waitUntil - now)
+    now = System.nanoTime()
+  }
+}
+
+/**
+ * Wait until either a duration is elapsed or this is notified. Unlike [java.lang.Object.wait] this
+ * interprets 0 as "don't wait" instead of "wait forever".
+ */
+@Throws(InterruptedException::class)
+@Suppress("PLATFORM_CLASS_MAPPED_TO_KOTLIN")
+fun Any.waitNanos(nanos: Long) {
+  val ms = nanos / 1_000_000L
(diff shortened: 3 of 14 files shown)


tirsen:
Yay to "opt-in followed by opt-out". I think that's exactly what we want to do!

misk/src/main/kotlin/misk/web/interceptors/ConcurrencyLimitsInterceptor.kt
```diff
@@ -0,0 +1,79 @@
...
+  private val actionName: String,
+  private val limiter: Limiter<String>
+) : NetworkInterceptor {
+
+  override fun intercept(chain: NetworkChain) {
+    val listener: Limiter.Listener? = limiter.acquire(actionName).orElse(null)
+
+    if (listener == null) {
+      logger.error { "concurrency limits interceptor shedding $actionName" }
+      chain.httpCall.statusCode = StatusCode.SERVICE_UNAVAILABLE.code
```
tirsen: We want to use 429
https://httpstatuses.com/429

misk/src/main/kotlin/misk/web/Http.kt
```diff
@@ -32,3 +32,28 @@ annotation class RequestContentType(vararg val value: String)
...
+
+/**
+ * Opt-in to concurrency limits.
+ *
+ * TODO(jwilson): make this the default once we're comfortable with the behavior and remove this
+ *     annotation.
+ */
+@Retention(AnnotationRetention.RUNTIME)
+@Target(AnnotationTarget.FUNCTION)
+annotation class ConcurrencyLimitsOptIn
```
tirsen: Shouldn't this be a config option in the yaml so that you can opt in without code changes?

adrw: Once this is an opt out default, would this annotation be deleted and implicitly be present on all web actions?

misk/src/main/kotlin/misk/web/interceptors/ConcurrencyLimitsInterceptor.kt
swankjesse: Really hard to say if it’s client’s fault or server’s fault that the server is overloaded. Franklin does 503 here. Feel strongly on this?

misk/src/main/kotlin/misk/web/Http.kt
swankjesse: Exactly

swankjesse: Actually I think what I’d prefer over YAML config is runtime flags like LaunchDarkly. For follow-up though.

misk/src/main/kotlin/misk/web/interceptors/ConcurrencyLimitsInterceptor.kt
swankjesse: (I’m going to merge with 503 to get things unstuck, and I’m totally open to changing it to 429.)

misk/src/main/kotlin/misk/web/Http.kt
adrw: I agree, being able to turn this on with a flag is much better than having to edit YAML and re-deploy.

misk/src/main/kotlin/misk/web/interceptors/ConcurrencyLimitsInterceptor.kt
eeSeeGee: We went with 503 because we were nervous about leaking load problems to the outside world. I think ideally we use 429 and Envoy scrubs on the way back out to the internet

misk/src/main/kotlin/misk/web/interceptors/ConcurrencyLimitsInterceptor.kt
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    try {
+      when (chain.httpCall.statusCode) {
+        in 400 until 500 -> listener.onIgnore()
+        in 500 until 600 -> listener.onDropped() // Count 5XX errors as drops.
+        else -> listener.onSuccess()
+      }
+    } catch (e: IllegalArgumentException) {
+      // Service container does a ignores an RTT == 0 exception here. Is this still necessary?
```
eeSeeGee: Last I checked, yes :(

misk/src/main/kotlin/misk/web/interceptors/ConcurrencyLimitsInterceptor.kt
ryanhall07: lol "leaking load problems". They'll never know!

misk/src/main/kotlin/misk/web/interceptors/ConcurrencyLimitsInterceptor.kt
```diff
@@ -0,0 +1,79 @@
...
+internal class ConcurrencyLimitsInterceptor internal constructor(
+  private val actionName: String,
+  private val limiter: Limiter<String>
+) : NetworkInterceptor {
+
+  override fun intercept(chain: NetworkChain) {
+    val listener: Limiter.Listener? = limiter.acquire(actionName).orElse(null)
+
+    if (listener == null) {
+      logger.error { "concurrency limits interceptor shedding $actionName" }
```
ryanhall07: we probably don't want log everytime we shed. ideally we are incrementing a metric instead

misk/src/main/kotlin/misk/web/interceptors/ConcurrencyLimitsInterceptor.kt
ryanhall07: update comment? seems strange in misk

misk/src/main/kotlin/misk/web/interceptors/ConcurrencyLimitsInterceptor.kt
```diff
@@ -0,0 +1,79 @@
...
+
+    try {
+      when (chain.httpCall.statusCode) {
+        in 400 until 500 -> listener.onIgnore()
+        in 500 until 600 -> listener.onDropped() // Count 5XX errors as drops.
+        else -> listener.onSuccess()
+      }
+    } catch (e: IllegalArgumentException) {
+      // Service container does a ignores an RTT == 0 exception here. Is this still necessary?
+      logger.info { "ignoring concurrency-limits exception: $e" }
```
ryanhall07: warn?

misk/src/main/kotlin/misk/web/interceptors/ConcurrencyLimitsInterceptor.kt
```diff
@@ -0,0 +1,79 @@
...
+    }
+  }
+
+  @Singleton
+  class Factory @Inject constructor(val clock: Clock) : NetworkInterceptor.Factory {
+    private val limiter = SimpleLimiter.Builder()
```
eeSeeGee: We tried running a single limiter in vttabs, we ran into a lot of problems when a few long rpcs would result in a whole bunch of dropped little ones, it seems pretty intolerant to rangey response times.

eeSeeGee:
> nice! we should add some metrics (and potentially remove or throttle the logging)

The nice thing about per-action metrics is in practice they show which endpoint is misbehaving really cleanly

misk/src/main/kotlin/misk/web/interceptors/ConcurrencyLimitsInterceptor.kt
zhxnlai: Input signal to the limiter should be homogeneous. I would use one limiter per action or use a [AbstractPartitionedLimiter](https://github.com/Netflix/concurrency-limits/blob/master/concurrency-limits-core/src/main/java/com/netflix/concurrency/limits/limiter/AbstractPartitionedLimiter.java)

swankjesse: Fixed to use a separate limiter for each action.

misk/src/main/kotlin/misk/web/interceptors/ConcurrencyLimitsInterceptor.kt
swankjesse: I actually changed it the opposite way to `debug` because it was very chatty.

swankjesse: (And I’m gonna report a bug upstream.)

misk/src/main/kotlin/misk/web/interceptors/ConcurrencyLimitsInterceptor.kt
swankjesse: Fixed.

misk/src/main/kotlin/misk/web/interceptors/ConcurrencyLimitsInterceptor.kt
swankjesse: Will do, for follow-up though.

misk/src/main/kotlin/misk/web/interceptors/ConcurrencyLimitsInterceptor.kt
swankjesse: Fixed to do many limiters. The partitioned one blocks so I don’t think it’s a good fit.

misk/src/main/kotlin/misk/web/interceptors/ConcurrencyLimitsInterceptor.kt
swankjesse: I’m going to leave this as-is for now.

swankjesse:
@tirsen @ryanhall07 @eeSeeGee @zhxnlai 
I did a big update to this PR to add a real stress test that uses Jetty and everything.

It runs a service that can process 25 requests/second and sends it 50 requests a second. Here’s the latest test-run output after a 5-second run:
```
successCount=116, shedCount=127, failureCount=7
```

I think this is totally awesome!

eeSeeGee:
> @tirsen @ryanhall07 @eeSeeGee @zhxnlai
> I did a big update to this PR to add a real stress test that uses Jetty and everything.
> 
> It runs a service that can process 25 requests/second and sends it 50 requests a second. Here’s the latest test-run output after a 5-second run:
> 
> ```
> successCount=116, shedCount=127, failureCount=7
> ```
> 
> I think this is totally awesome!

Heck yeah! It's super fun to watch this thing go in a test, and is a lot easier to understand what it's up to

misk/src/test/kotlin/misk/web/DegradedHealthStressTest.kt
```diff
@@ -0,0 +1,139 @@
...
+    shedCount.set(0)
+    failureCount.set(0)
+
+    // Observe 5 seconds of data.
+    Thread.sleep(5_000)
+    recurringTask.cancel(false)
+
+    // Expect 125 successful calls, most of the rest shed, and the remainder failed.
+    println("successCount=$successCount, shedCount=$shedCount, failureCount=$failureCount")
+    assertThat(successCount.get().toDouble()).isCloseTo(125.0, Offset.offset(50.0))
```
zhxnlai: 💯

misk/src/main/kotlin/misk/web/interceptors/ConcurrencyLimitsInterceptor.kt
tirsen: The nice thing with using 429 is that it signals to the clients that they should back off. For example there is a method to call on Limiter.Listener when the client sends 429 which makes the limiter decrease rate significantly:
https://github.com/Netflix/concurrency-limits/blob/master/concurrency-limits-core/src/main/java/com/netflix/concurrency/limits/Limiter.java#L45

I think unfortunately that means we do have to announce this to the world because ideally we would also want our mobile app clients to implement back off as well in these situations. If Envoy scrubs it that's no good.

tirsen: But I'm fine with merging this as is if we can keep the discussion going. What's so bad about announcing over capacity to the world?

ryanhall07: In theory I agree with @tirsen, 429 is more appropriate. In reality, probably doesn't really matter. Clients should really back off on all errors, including 5xxs.

misk/src/test/kotlin/misk/time/wait_notify.kt
```diff
@@ -0,0 +1,35 @@
+package misk.time
+
+/**
+ * Wait until the duration has elapsed. Unlike [java.lang.Object.wait] this interprets 0 as
+ * "don't wait" instead of "wait forever".
```
ryanhall07: lol finally!

PR #1305, opened 2019-12-05
Allow an unbounded RepeatedTaskQueue

ryanhall07:
An unbounded queue can be nice if an App doesn't know the number of
tasks statically. For example an Event consumer might dynamically
subscribe to topics.

### DIFF
--- a/misk/src/main/kotlin/misk/tasks/RepeatedTaskQueueConfig.kt
+++ b/misk/src/main/kotlin/misk/tasks/RepeatedTaskQueueConfig.kt
@@ -20,7 +20,11 @@ data class RepeatedTaskQueueConfig(
   val default_max_delay_sec: Long = 60,
 
   /**
-   * The number of parallel tasks to run.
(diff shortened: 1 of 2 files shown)


wesleyk:
"For example"

misk/src/main/kotlin/misk/tasks/RepeatedTaskQueueConfig.kt
```diff
@@ -20,7 +20,11 @@ data class RepeatedTaskQueueConfig(
   val default_max_delay_sec: Long = 60,
 
   /**
-   * The number of parallel tasks to run.
+   * The fixed number of parallel tasks to run.
+   *
+   * If -1 then an unbounded number of parallel tasks are allowed. An unbounded number of tasks can
+   * be useful for an App that needs to dynamically compute the number of tasks at runtime. However,
+   * the App is then responsible for ensuring an upper bound for the number of tasks submitted.
```
swankjesse: nice comment!

PR #1307, opened 2019-12-06
Add method to close lease

zhxnlai:


misk/src/main/kotlin/misk/clustering/lease/Lease.kt
```diff
@@ -24,6 +24,12 @@ interface Lease {
    */
   fun addListener(listener: StateChangeListener)
 
+  /**
+   * Release the lease if it is owned by this process instance. This may involve remote calls,
+   * so it is marked as a function rather than a property to make the potential expense clearer.
+   */
+  fun close()
```
swankjesse: Implement closeable so it can be used with `use` ?

mmihic:
Not sure what the intent is here but this doesn’t do anything. Who owns the
lease is determined by the hashring derived from the kubernetes watch set;
the zk node data simply prevents race conditions in delivery of watch
events from causing two pods to both think they own a lease. Deleting the
zk ownership data won’t allow another pod to acquire the lease. To do that
you’d need a more complex impl with eg a markdown list that causes the
releasing pod to be skipped from consideration in the ownership hashring

ryanhall07:
Hi @mmihic ! `ZkLease` already had a `close` method, so this was just promoting it to the interface and adding a fake for tests. At least that was my read.

hi

mmihic:
Yeah the method being internal is intentional - it’s not something you want
to call outside of the pod shutting down, so it doesn’t make sense to be an
application visible method.

FWIW the effective result of an application calling close() outside of
shutdown is to make it so that the lease is unable to be acquired by any
pod in the cluster until the cluster topology changes in a way that adjusts
the hash ring to point the lease to another pod. Those are kind of weird
semantics and I’m guessing aren’t the intent? But if they are I’d recommend
renaming the public method to be a bit more explicit about them - close()
as a public method sounds like you are releasing the lease so others can
acquire it, which this does not do.

(close() as a name made more sense when it was just an internal method used
at shutdown, since it was just closing the local pod’s connection to the
lease in preparation for termination, but even then should probably have
been something like prepareForShutdown)

anyway, gonna go unsubscribe from misk email notifs now :) ✌🏼

zhxnlai:
@mmihic Thanks for the insight! I'd be using this method to rebalance event consumers as more pods get added to the cluster. This semantic is exactly what I was looking for. I will rename the method name to reveal the intent.

mightyguava:
Event consumers are automatically rebalanced when more pods are added to the cluster. That is the current implementation without this change.

The only effect of calling `close()` from a pod is to make the lease un-acquirable by any pods until a rebalance happens. That doesn't sound like what you want

zhxnlai:
@mightyguava According to the source code, a cloud evently consumer holds on to leases forever and even after the cluster layout changes. Let's continue the discussion in slack

PR #1308, opened 2019-12-06
Support multiple prioritized queues

river-jade:


ryanhall07:
can you elaborate a bit about this change? I don't understand why you would need this over just installing multiple queues, with the same `JobHandler`.

river-jade:
We need to have prioritised queues. Ie we need to consume events from queue
1 before queue 2. This is so that we can process real time order events
before scheduled order events, before other lower priority events.

ryanhall07:
> We need to have prioritised queues. Ie we need to consume events from queue 1 before queue 2. This is so that we can process real time order events before scheduled order events, before other lower priority events.

Why can't we just process them all at the same time? Are we resource limited by something? This seems like a resource limiting problem and we should solve it with a permit acquisition solution.

misk-aws/src/main/kotlin/misk/jobqueue/sqs/SqsJobConsumer.kt
```diff
@@ -38,8 +38,12 @@ internal class SqsJobConsumer @Inject internal constructor(
 ) : JobConsumer {
   private val subscriptions = ConcurrentHashMap<QueueName, QueueReceiver>()
 
-  override fun subscribe(queueName: QueueName, handler: JobHandler) {
-    val receiver = QueueReceiver(queueName, handler)
+  override fun subscribe(queueNames: List<QueueName>, handler: JobHandler) {
```
ryanhall07: this doesn't need to take a list

misk-aws/src/main/kotlin/misk/jobqueue/sqs/SqsJobConsumer.kt
```diff
@@ -63,12 +67,22 @@ internal class SqsJobConsumer @Inject internal constructor(
...
+    queueNames: List<QueueName>,
     private val handler: JobHandler
   ) {
-    private val queue = queues[queueName]
+    private val prioritizedQueues = queueNames.map { queues[it] }
 
     fun runOnce(): Status {
+      prioritizedQueues.forEach { resolvedQueue ->
+        val status = runOnce(resolvedQueue)
+        if (status != Status.NO_WORK) {
```
ryanhall07: how should we handle failures from a queue? still skip processing lower prioritized queues?

misk-aws/src/main/kotlin/misk/jobqueue/sqs/AwsSqsJobHandlerModule.kt
```diff
@@ -13,45 +14,48 @@ import kotlin.reflect.KClass
  * Install this module to register a handler for an SQS queue.
```
ryanhall07: this doc needs to be updated to describe how Prioritized queues work

misk-aws/src/main/kotlin/misk/jobqueue/sqs/AwsSqsJobHandlerModule.kt
```diff
@@ -13,45 +14,48 @@ import kotlin.reflect.KClass
...
  */
 class AwsSqsJobHandlerModule<T : JobHandler> private constructor(
-  private val queueName: QueueName,
+  private val queueNames: List<QueueName>,
```
ryanhall07: this is not backwards compatible with existing clients that install a single queue. Plus we should have a flavor of the constructor that allows a single queue, since that's the common case

misk-jobqueue/src/main/kotlin/misk/jobqueue/JobConsumer.kt
```diff
@@ -7,11 +7,11 @@ interface JobConsumer {
    * begin receiving messages from the underlying job queue and dispatch them to the provided
    * handler. A service may only have one subscription outstanding per queue
    */
-  fun subscribe(queueName: QueueName, handler: JobHandler)
+  fun subscribe(queueNames: List<QueueName>, handler: JobHandler)
```
ryanhall07: again not backwards compatible

misk-jobqueue/src/main/kotlin/misk/jobqueue/QueueKey.kt
```diff
@@ -0,0 +1,9 @@
+package misk.jobqueue
+
+/**
+ * Represents an ordered list of queue names, for building a priority queue using underlying queues.
+ * Allows us to workaround limitations with Guice's handling of generics.
+ */
+data class QueueKey(
```
ryanhall07: nit: should probably call this `QueuesKey`

misk-jobqueue/src/main/kotlin/misk/jobqueue/QueueKey.kt
```diff
@@ -0,0 +1,9 @@
+package misk.jobqueue
+
+/**
+ * Represents an ordered list of queue names, for building a priority queue using underlying queues.
+ * Allows us to workaround limitations with Guice's handling of generics.
```
ryanhall07: what does generics have to do with it?

misk-aws/src/main/kotlin/misk/jobqueue/sqs/SqsJobConsumer.kt
ryanhall07: similar question, what about exceptions?

misk-aws/src/main/kotlin/misk/jobqueue/sqs/AwsSqsJobHandlerModule.kt
ryanhall07: also need to update `JobConsumer` docs

misk-aws/src/main/kotlin/misk/jobqueue/sqs/SqsJobConsumer.kt
river-jade: what are you proposing instead? we'd need to use a builder pattern or similar if we didn't take a list. or am i misunderstanding you?

misk-aws/src/main/kotlin/misk/jobqueue/sqs/SqsJobConsumer.kt
river-jade: on failures / exceptions we retry the same queue again. i believe this is the correct behaviour. the bad items should be pushed to the DLQ and we get the next batch.

misk-aws/src/main/kotlin/misk/jobqueue/sqs/AwsSqsJobHandlerModule.kt
river-jade: The constructor is private. There's a overload of the factory method for the single-queue case.

misk-jobqueue/src/main/kotlin/misk/jobqueue/JobConsumer.kt
river-jade: the `subscribe` extension method handles this case.

misk-jobqueue/src/main/kotlin/misk/jobqueue/QueueKey.kt
river-jade: Guice doesn't support binding to a generic key (ie List<QueueName>) in Kotlin without some pretty ugly annotations. Comment is probably more confusing than helpful though, so I'll just remove it.

misk-aws/src/main/kotlin/misk/jobqueue/sqs/SqsJobConsumer.kt
ryanhall07: but you potentially starve the lower priority queues. is that ok?

misk-aws/src/main/kotlin/misk/jobqueue/sqs/SqsJobConsumer.kt
ryanhall07: iterate outside. doesn't really matter

misk-jobqueue/src/main/kotlin/misk/jobqueue/QueueKey.kt
ryanhall07: hmm ok. we do it all over the place in misk. do you mean just adding the `@JvmWildcard`?

misk-aws/src/main/kotlin/misk/jobqueue/sqs/SqsJobConsumer.kt
river-jade: yes, i think that's correct, since you shouldn't get stuck processing the same message -- it should get pushed to the DLQ. there's no way to know if the top priority queue is full of error-causing messages, or if it's just one, so i think you need to keep processing them.

misk-jobqueue/src/main/kotlin/misk/jobqueue/QueueKey.kt
river-jade: yes. it was jesse's suggestion to pull out a class. i don't feel that strongly about it, though the @JvmWildcard is pretty ugly imo.

river-jade:
sorry, forgot to push. fixing now.

misk-jobqueue/src/main/kotlin/misk/jobqueue/JobConsumer.kt
```diff
@@ -3,15 +3,21 @@ package misk.jobqueue
...
 interface JobConsumer {
   /**
-   * Registers a handler to receive messages. Once registered, the consumer will immediately
-   * begin receiving messages from the underlying job queue and dispatch them to the provided
-   * handler. A service may only have one subscription outstanding per queue
+   * Associates a message handler to a prioritized list of queues. Once registered, the consumer
+   * will immediately begin receiving messages from the underlying job queues and dispatch them to
+   * the provided handler.
+   *
+   * If more than one queue name is provided, these will be treated as a list of prioritized queues
```
ryanhall07: we should be explicit about the chance of starvation for the lower priority queues. for example in the case of the higher priority queues being over saturated or if the higher priority queues are failing to process.

misk-aws/src/main/kotlin/misk/jobqueue/sqs/SqsJobConsumer.kt
ryanhall07: > there's no way to know if the top priority queue is full of error-causing messages, or if it's just one, so i think you need to keep processing them.

that's not true. we could do something, like inspect the current error rate.

river-jade: We can know how many errors we've processed, but not how many are still to come. Might be a reasonable heuristic there, but I'd worry that it's overcomplicating things.

misk-jobqueue/src/main/kotlin/misk/jobqueue/JobConsumer.kt
river-jade: Yeah, good call. I'll update the docs. Thanks!

mightyguava: It seems like that as long as there's a constant trickle of messages going into the higher priority queues, the lower priorities queues will never processed?

misk-aws/src/main/kotlin/misk/jobqueue/sqs/SqsJobConsumer.kt
river-jade: we already do that. it has a different semantic though. calling subscribe multiple times with different queues (and the same handler) subscribes the same handler to different queues, with no prioritisation.

misk-jobqueue/src/main/kotlin/misk/jobqueue/JobConsumer.kt
river-jade: this should only happen if we're processing messages at the maximum rate (so it's correct to starve the lower-priority queue in this case), though there is an issue where we currently wait for all workers to finish before pulling the next batch. i have another PR to address that.

mightyguava: Each Receive request can contain _up to_ 10 messages. If you get 1 message in each Receive request (a trickle), you'll be processing at 1/10 rate and starve the low-priority queues.

PR #1316, opened 2019-12-10
Add the QueueName label to jobqueue metrics

hamdanjaveed:
This matches the equivalent SQS metric which lets us filter for queues using 1 metric.

Opted to add a new label instead of replacing the existing one as to not break dashboards
and detectors.

### DIFF
--- a/misk-aws/src/main/kotlin/misk/jobqueue/sqs/SqsMetrics.kt
+++ b/misk-aws/src/main/kotlin/misk/jobqueue/sqs/SqsMetrics.kt
@@ -4,65 +4,73 @@ import misk.metrics.Metrics
 import javax.inject.Inject
 import javax.inject.Singleton
 
+/**
+ * SQS Jobqueue metrics.
+ *
+ * NB: we use the capitalized "QueueName" label to stay consistent with SQS' similarly named label. This lets us filter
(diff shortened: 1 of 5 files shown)


misk-aws/src/main/kotlin/misk/jobqueue/sqs/SqsMetrics.kt
```diff
@@ -9,60 +9,62 @@ internal class SqsMetrics @Inject internal constructor(metrics: Metrics) {
   val jobsEnqueued = metrics.counter(
       "jobs_enqueued_total",
       "total # of jobs sent to a queueName",
-      listOf("queueName")
+      // Duplicate labels so we don't break existing dashboards/detectors
+      // TODO(hamdan): remove queueName once no one's using it
+      listOf("queueName", "QueueName")
```
ryanhall07: don't you have to update the callers to provide both?

hamdanjaveed: Updated callers to provide both labels

ryanhall07: leave a comment we are using `QueueName` to be consistent with AWS SQS

PR #1320, opened 2019-12-13
Rename ResourceMapper to Partitioner

zhxnlai:
Metrics and tests have shown that `ClusterHashRing` does not uniformly distribute resources to cluster members when the resource to cluster size ratio is low.

Inspired by the Kafka client, I would like to add alternative ways to assign resources to cluster members that are more suitable for event consumers.
- [AbstractPartitionAssignor](https://github.com/apache/kafka/blob/trunk/clients/src/main/java/org/apache/kafka/clients/consumer/internals/AbstractPartitionAssignor.java)
- [RoundRobinAssignor](https://github.com/apache/kafka/blob/trunk/clients/src/main/java/org/apache/kafka/clients/consumer/RoundRobinAssignor.java)
- [RangeAssignor](https://github.com/apache/kafka/blob/trunk/clients/src/main/java/org/apache/kafka/clients/consumer/RangeAssignor.java#L68)

I was going to add a `RoundRobinResourceMapper` or  a`RangeResourceMapper`.

This made me wonder if what `ResourceMapper` essentially does is partitioning resources to cluster members, would `Partitioner` or `PartitionStrategy` be a better name?

### DIFF
--- a/misk-zookeeper/src/main/kotlin/misk/clustering/zookeeper/ZkLease.kt
+++ b/misk-zookeeper/src/main/kotlin/misk/clustering/zookeeper/ZkLease.kt
@@ -13,7 +13,7 @@ import kotlin.concurrent.withLock
 /**
  * Zookeeper based load balanced lease.
  *
- * <p>Multiple servers use a [misk.clustering.ClusterResourceMapper] to determine which process should
+ * <p>Multiple servers use a [misk.clustering.Partitioner] to determine which process should
  * own a lease. As long as each process has the same view of the cluster, this mapping will be
  * consistent and processes will not actively compete for the same lease.
  *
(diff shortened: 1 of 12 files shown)


ryanhall07:
can you provide some context in the commit message? why is this a better name?

mightyguava:
My recollection of the investigation so far is that we are unsure of the current hash ring algorithm is correct. There might be bugs. There have been bugs. Can we prove that it's a flaw of the algorithm and not a bug before moving on to alternate algorithms?

ryanhall07:
why don't we first rollout new impl of `ClusterResourceMapper` and prove out your theory. Also IMO `ClusterResourceMapper` is a pretty generic and easy to understand name. You would just provide a new impl `ClusterRangeAssigner`

PR #1321, opened 2019-12-13
Prepare for Moshi 1.9.x

swankjesse:
It requires Kotlin adapters for Kotlin types.


swankjesse:
It’ll add a dependency on Kotlin-stdlib but we already have that.

PR #1324, opened 2019-12-15
Remove jetty dependency for clustering

jjestrel:
use unbounded LinkedBlockingQueue instead of BlockingArrayQueue for DefaultCluster

### DIFF
--- a/misk/src/main/kotlin/misk/clustering/DefaultCluster.kt
+++ b/misk/src/main/kotlin/misk/clustering/DefaultCluster.kt
@@ -2,7 +2,7 @@ package misk.clustering
 
 import com.google.common.util.concurrent.AbstractExecutionThreadService
 import misk.logging.getLogger
-import org.eclipse.jetty.util.BlockingArrayQueue
+import java.util.concurrent.LinkedBlockingQueue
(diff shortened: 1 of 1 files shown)


misk/src/main/kotlin/misk/clustering/DefaultCluster.kt
```diff
@@ -26,7 +26,7 @@ internal class DefaultCluster(
       resourceMapper = newResourceMapper(setOf())
   ))
   private val running = AtomicBoolean(false)
-  private val actions = BlockingArrayQueue<(MutableSet<ClusterWatch>) -> Unit>()
+  private val actions = LinkedBlockingQueue<(MutableSet<ClusterWatch>) -> Unit>()
```
jjestrel: `BlockingArrayQueue` is effectively unbounded and grows up to `Integer.MAX_VALUE`. `LinkedBlockingQueue` defaults to allow `Integer.MAX_VALUE`, behaviour should be the same. 

This is what `Executors.newSingleThreadExecutor` uses by default

ryanhall07:
You should explain why you're doing this in the commit message. It appears like you're going to be doing several refactorings, so maybe create a github issue that you can link to in your commit message

PR #1325, opened 2019-12-16
Move most of misk.clustering misk.random, misk.backoff, and misk.logging in to misk-core for #1330

jjestrel:
Part breaking things out of `misk` module, for #1330 

I'm not sure clustering will stay in misk-core but I think this is iterating in the right direction


misk-core/src/main/kotlin/misk/clustering/DefaultCluster.kt
```diff
@@ -14,19 +14,20 @@ import javax.inject.Singleton
  * that need to be triggered as the cluster changes.
  */
 @Singleton
-internal class DefaultCluster(
```
jjestrel: This is 1 of 2 non file moving changes here

misk-core/src/main/kotlin/misk/clustering/DefaultCluster.kt
```diff
@@ -99,7 +100,7 @@ internal class DefaultCluster(
    * Triggers a callback once all of the actions on the queue have been processed. Useful
    * for writing deterministic tests
    */
-  internal fun syncPoint(callback: () -> Unit) {
+  fun syncPoint(callback: () -> Unit) {
```
jjestrel: This is not internal now either (change 2 of 2)

ryanhall07:
update the commit message with why you are doing this

jjestrel:
> update the commit message with why you are doing this

@ryanhall07 Updated commit message to reference #1330 and added to title

PR #1331, opened 2019-12-16
Remove FeatureFlagValidation

ryanhall07:
This does not fail gracefully and failing to lookup a feature flag
should not impact a production system.


ryanhall07:
for now we are just updating our code to add a try-catch and fallback, so I'll hold off merging this until tomorrow when we can have a broader discussion on how this library should fail

misk-feature-testing/src/main/kotlin/misk/feature/testing/FakeFeatureFlags.kt
```diff
@@ -57,14 +56,15 @@ class FakeFeatureFlags @Inject constructor() : AbstractIdleService(),
   override fun <T : Enum<T>> getEnum(feature: Feature, clazz: Class<T>): T = getEnum(feature, KEY, clazz, Attributes())
 
   private fun get(feature: Feature, key: String): Any? {
-    FeatureFlagValidation.checkValidKey(feature, key)
+    require(key.isNotEmpty()) { "Key to flag $feature must not be empty" }
```
ryanhall07: FYI @mightyguava kept the not-empty validation

mightyguava:
@ryanhall07 is this still happening?

PR #1332, opened 2019-12-17
Retry Cockroach transactions

tirsen:


misk-hibernate/src/main/kotlin/misk/hibernate/RealTransacter.kt
```diff
@@ -323,7 +323,7 @@ internal class RealTransacter private constructor(
   }
 
   private fun isMessageRetryable(th: SQLException) =
-      isConnectionClosed(th) || isVitessTransactionNotFound(th)
+      isConnectionClosed(th) || isVitessTransactionNotFound(th) || isCockroachRestartTransaction(th)
```
mightyguava: I'm curious why this hasn't been needed for TiDB

tirsen: I bet we gotta do something similar but maybe it doesn't fail as often?

mightyguava: TiDB does some internal retries for this kind of conflict. I'm curious if cockroach expects the user to retry.

tirsen: Oh right it was those retries that Jepsen commented on. I thought they were disabled by default?

mightyguava: Uh, true, maybe they don't do that anymore

PR #1333, opened 2019-12-17
Add config parameter for SQS pull batch size

river-jade:


misk-aws/src/main/kotlin/misk/jobqueue/sqs/SqsJobConsumer.kt
```diff
@@ -88,7 +88,7 @@ internal class SqsJobConsumer @Inject internal constructor(
           .withAttributeNames("All")
           .withMessageAttributeNames("All")
           .withQueueUrl(queue.url)
-          .withMaxNumberOfMessages(10)
+          .withMaxNumberOfMessages(config.message_batch_size)
```
ryanhall07: not sure I see a reason to make this configurable, since the max is 10. are you trying to pull less than 10 messages at time? this could dramatically affect the SQS bill (10x more!) if lowered to 1.

even if we keep it, we should validate this config value is between 1 and 10 at startup.

river-jade: The reason to make it configurable is to work around the issue that my other PR (on the thread pool consumption) attempts to address. If we only pull one message at a time, we can have a predictable level of concurrency. It's definitely not the ideal solution, but as per my comments on the other PR, I think solving this in a more efficient way is non-trivial.

Re: cost, requests are charged at 0.40c / million requests, so I think we're a fair way off this being a significant issue. Good call to put in a check for the bounds on the parameter. I'm happy to put that in. Where do we normally do checks like that in Misk?

ryanhall07: actually we're not that far away. with our internal switch to SQS it was a diff of 100k/year.

river-jade: i mean for investing specifically. we are doing less than 1M messages per day. also, fwiw, we're using DW's queue, so iiuc, they pay for it.

PR #1334, opened 2019-12-18
Improves default configuration of `AwsSqsJobQueueConfig`.

frojasg:
* By default we're using one thread which is problematic as it slows down
the consumption of messages when you have more than one queue.

### DIFF
--- a/misk-aws/src/main/kotlin/misk/jobqueue/sqs/AwsSqsJobQueueModule.kt
+++ b/misk-aws/src/main/kotlin/misk/jobqueue/sqs/AwsSqsJobQueueModule.kt
@@ -16,6 +16,7 @@ import misk.jobqueue.JobQueue
 import misk.jobqueue.QueueName
 import misk.jobqueue.TransactionalJobQueue
 import misk.tasks.RepeatedTaskQueue
+import misk.tasks.RepeatedTaskQueueConfig
 import misk.tasks.RepeatedTaskQueueFactory
 import javax.inject.Inject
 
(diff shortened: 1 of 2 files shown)


frojasg:
@mightyguava This is what -1 is doing based on https://github.com/cashapp/misk/pull/1305/files#diff-2e81bbece5fd45711cf43307687a626b.

mightyguava:
Ooohhhh, ok. I think -1 for unbounded is not a Java-y API, but works for me.

hamdanjaveed:
I was thinking we should bound (and control via a feature flag) the number of consumption threads as that controls how fast our services will poll SQS and process jobs. In this case, will a massive influx of jobs have the potential to spawn tons of consumption threads and possibly degrade our service?

frojasg:
@hamdanjaveed yeah, that's the next PR i will be working on.

misk-aws/src/main/kotlin/misk/jobqueue/sqs/AwsSqsJobQueueModule.kt
```diff
@@ -72,9 +73,12 @@ class AwsSqsJobQueueModule(
   ): RepeatedTaskQueue {
     return queueFactory.new(
         "sqs-consumer-poller",
-        config.task_queue)
+        repeatedTaskQueueConfig(config))
   }
 
+  private fun repeatedTaskQueueConfig(config: AwsSqsJobQueueConfig) =
+      config.task_queue ?: RepeatedTaskQueueConfig(num_parallel_tasks = -1)
```
ryanhall07: i don't see value in allowing somebody to configure a bounded task queue for this. can we just remove the config entirely from the aws job queue config?

frojasg: neither do I. Yes, I can remove it. I just want to make sure I don't break existing clients.

ryanhall07: i would just grep for existing clients. at the very least if you don't remove, mark it clearly deprecated with instructions to remove it so you don't have a single threaded consuming app!

ryanhall07:
> I was thinking we should bound (and control via a feature flag) the number of consumption threads as that controls how fast our services will poll SQS and process jobs. In this case, will a massive influx of jobs have the potential to spawn tons of consumption threads and possibly degrade our service?

This is realistically bounded by the number of `subscribe` calls by the application to register a consuming task. It just allows dynamically controlling the upper bound at runtime, instead of statically defining the upper bound at compile time.

> Ooohhhh, ok. I think -1 for unbounded is not a Java-y API, but works for me.

Yea that was pretty stupid by me. I guess we should just have a `unbounded: Boolean`

Issue #1336, opened 2019-12-19
Add Range ResourceMapper


zhxnlai:
**tl;dr:** consistent hashing is the only way to load balance leases in misk. Although it is a well established way to partition databases, counter intuitively, it is not the best way to load balance event consumers, where the number of total leases is small and each consumer is stateless.

Our team operates a service that consumes an event topic. This topic has 32 partitions and emits 400+ events per second. At some time last week, the service was auto scaled 20 pods. 

Most pods were doing work. But according to the metrics, the load on these pods was not evenly distributed: 4 pods were idle, not consuming any events, while 2 pods were busy, consuming a lot of events.

After doing some math, I found this:
- Assuming the consistent hash ring assigns partitions to a specific pod (let’s call it pod A) 1 time out of 20 (given 20 pods)
- Given 32 partitions, we have the following binomial distribution:

<img width="612" alt="Screen Shot 2019-12-12 at 9 58 22 PM" src="https://user-images.githubusercontent.com/4694896/71133199-ae44ec00-21ae-11ea-8e89-01b8d3d3878d.png">

- the probability that pod A receives 0 partition is 0.19371,
- the probability that pod A receives 1 partitions is 0.32625
- the probability that pod A receives 2 partitions is 0.26615
- the probability that pod A receives 3 partitions is 0.14008
- the probability that pod A receives 4 partitions is 0.05345

This distribution, consistent with the production metrics, is not ideal: 4 pods are idle (consuming 0 partition), 4 pods take twice the load (consuming 3 - 4 partitions). 

As we increase the number of partitions, the distribution becomes more uniform:

**# of partitions = 10000**
- <img width="519" alt="Screen Shot 2019-12-12 at 10 22 32 PM" src="https://user-images.githubusercontent.com/4694896/71133200-ae44ec00-21ae-11ea-893e-52b6aa6454e1.png">

**# of partitions = 100000**
- <img width="522" alt="Screen Shot 2019-12-12 at 10 23 19 PM" src="https://user-images.githubusercontent.com/4694896/71133201-ae44ec00-21ae-11ea-8b58-ade23ba06548.png">

It is common for databases to have more than 1^6 rows. But it is not practical for a topic to have this many partitions.

There are other ways to assign partitions to worker pods. For example,
[...]

PR #1337, opened 2019-12-19
Move ClusterResourceMapper out of cluster snapshot

zhxnlai:
Our cluster abstraction exposes `ClusterResourceMapper` as an instance variable of `Cluster.Snapshot`. This makes sense because the mapper uses only the set of cluster members to generate a mapping from resource ids to cluster members .

However, the Range/Round Robin ResourceMapper described in #1336 requires both the set of cluster members and the set of all resource ids.

This PR separates `ClusterResourceMapper` from `Cluster.Snapshot` and adds a `ClusterResourceMapper.Provider` that provides it based on the current cluster snapshot.

This makes it easy to add a range `ClusterResourceMapper.Provider` that reads both the current cluster members and all leases.

### DIFF
--- a/misk-zookeeper/src/main/kotlin/misk/clustering/zookeeper/ZkLease.kt
+++ b/misk-zookeeper/src/main/kotlin/misk/clustering/zookeeper/ZkLease.kt
@@ -1,5 +1,6 @@
 package misk.clustering.zookeeper
 
+import misk.clustering.ClusterResourceMapper
 import misk.clustering.NoMembersAvailableException
 import misk.clustering.lease.Lease
 import misk.clustering.weights.ClusterWeightProvider
@@ -39,6 +40,7 @@ internal class ZkLease(
   private val manager: ZkLeaseManager,
   private val leaseResourceName: String,
   private val clusterWeight: ClusterWeightProvider,
+  private val clusterResourceMapper: ClusterResourceMapper.Provider,
   override val name: String
 ) : Lease {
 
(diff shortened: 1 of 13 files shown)


misk-zookeeper/src/main/kotlin/misk/clustering/zookeeper/ZkLease.kt
```diff
@@ -37,8 +35,7 @@ import kotlin.concurrent.withLock
 internal class ZkLease(
   ownerName: String,
   private val manager: ZkLeaseManager,
-  private val leaseResourceName: String,
-  private val clusterWeight: ClusterWeightProvider,
+  val leaseResourceName: String,
```
mightyguava: private

misk/src/main/kotlin/misk/clustering/ClusterResourceMapper.kt
```diff
@@ -9,7 +9,7 @@ package misk.clustering
 interface ClusterResourceMapper {
   /**
    * @throws NoMembersAvailableException if there are no members in the cluster
-   * @return The [Cluster.Member] that should own the given resource id
+   * @return A map of resource IDs to [Cluster.Member].
    */
-  operator fun get(resourceId: String): Cluster.Member
+  fun getMappings(resourceIds: Set<String>, members: Set<Cluster.Member>): Map<String, Cluster.Member>
```
mightyguava: > Because Range ResourceMapper makes assignments in batches, this PR also makes ClusterResourceMapper return a set of mappings instead of a single entry.

I don't understand, could you explain this?

Issue #1330, opened 2019-12-16
Break up misk module


jjestrel:
See also #1044 
The misk module currently contains several groups of things:
- Relatively dependency free things: clustering, config, environment, metrics(?), token gen, etc
- Clients: misk.client, including TypedHttpClient, etc
- Server: misk.web.jetty, including some other helpers
- Actions: most of the existing misk module code; anything that touches WebAction. clients also depend on this in tests today.

We should separate the misk module into something like these four pieces:
misk-core
misk-server
misk-client
misk-actions

Additionally we should merge misk-inject into misk-core

ryanhall07:
> Additionally we should merge misk-inject into misk-core

what is the benefit of doing this? just less modules and there isn't a high probability you just want our guice code?

> misk-actions

why can't this be in `misk-server`? tests probably need the `misk-server` stuff if they want to run a real app

this sounds reasonable to me. It would be nice to give guidance about the modules, especially something like `misk-core`. is that just everything that is shared between `misk-client` and `misk-server`?

jjestrel:
> > Additionally we should merge misk-inject into misk-core
> 
> what is the benefit of doing this? just less modules and there isn't a high probability you just want our guice code?

misk-inject is two files in a module that have a dependency on guice and kotlin. All our other modules will have the same dependencies on guice and kotlin. It seems worth rolling it in to something common (misk-core) if we promise to keep the dependencies to guice, kotlin, and logging

> 
> > misk-actions
> 
> why can't this be in `misk-server`? tests probably need the `misk-server` stuff if they want to run a real app

I think there may be benefits if we structure the app as actions being built on top of something misk-server provides

> 
> this sounds reasonable to me. It would be nice to give guidance about the modules, especially something like `misk-core`. is that just everything that is shared between `misk-client` and `misk-server`?

I think that's exactly the distinction. The two entrypoints of misk are either including misk-client and building a client, or misk-server and building a server. misk-core is everything shared by both

zhxnlai:
> Additionally we should merge misk-inject into misk-core

I actually wrote a library whose public api contains a subclass of `KAbstractModule`. This library has an [api dependency](https://docs.gradle.org/current/userguide/java_library_plugin.html#sec:java_library_separation) on misk-inject. I think we should keep `misk-inject` a standalone module for this purpose.

swankjesse:
[...]

PR #1341, opened 2019-12-20
Add empty misk-core module to make a filled in misk-core a non-breaking change

jjestrel:
-> Make everything include an empty misk-core
-> Fill in misk-core with https://github.com/cashapp/misk/pull/1325
-> bump misk-core everywhere

For #1330

### DIFF
--- a/misk-core/src/build.gradle
+++ b/misk-core/src/build.gradle
@@ -1,0 +1,27 @@
+buildscript {
+  dependencies {
+    classpath dep.kotlinNoArgPlugin
+  }
+}
+
+dependencies {
+  compile dep.guice
+  compile dep.guiceMultibindings
+  compile dep.kotlinStdLib
+  compile dep.kotlinReflection
+  compile dep.loggingApi
+  compile dep.logbackClassic
+  compile dep.slf4jApi
+
+  compile project(':misk-inject')
+  compile project(':misk-service')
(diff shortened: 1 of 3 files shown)


misk-core/src/build.gradle
```diff
@@ -0,0 +1,27 @@
+buildscript {
+  dependencies {
+    classpath dep.kotlinNoArgPlugin
+  }
+}
+
+dependencies {
+  compile dep.guice
```
adrw: meta: we should change these to `api`

adrw:
```
    When you switch to the new Android Gradle plugin 3.X.X, you should replace all your compile with the implementation keyword (1*). Then try to compile and test your app. If everything it's ok leave the code as is, if you have problems you probably have something wrong with your dependencies or you used something that now is private and not more accessible. Suggestion by Android Gradle plugin engineer Jerome Dochez (1)*)

    If you are a library mantainer you should use api for every dependency which is needed for the public API of your library, while use implementation for test dependencies or dependencies which must not be used by the final users.
```

- https://stackoverflow.com/questions/44413952/gradle-implementation-vs-api-configuration

jjestrel: All of misk is doing this, and I think we should get to that later

PR #1348, opened 2020-01-03
Add transaction token to sessions

drashkov:
Adds an identifier to a hibernate session which stays the same across retries. Can be used to determine whether something happened in the same transaction.

### DIFF
--- a/misk-hibernate/src/main/kotlin/misk/hibernate/RealTransacter.kt
+++ b/misk-hibernate/src/main/kotlin/misk/hibernate/RealTransacter.kt
@@ -11,6 +11,7 @@ import misk.jdbc.DataSourceType
 import misk.jdbc.map
 import misk.jdbc.uniqueString
 import misk.logging.getLogger
+import misk.tokens.TokenGenerator
 import misk.tracing.traceWithSpan
 import org.hibernate.FlushMode
 import org.hibernate.SessionFactory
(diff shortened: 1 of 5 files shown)


mightyguava:
Can you add a description? What are you planning to use this for? What can other people use it for?

misk-hibernate/src/main/kotlin/misk/hibernate/RealTransacter.kt
```diff
@@ -46,7 +47,8 @@ internal class RealTransacter private constructor(
   private val threadLatestSession: ThreadLocal<RealSession>,
   private val options: TransacterOptions,
   private val queryTracingListener: QueryTracingListener,
-  private val tracer: Tracer?
+  private val tracer: Tracer?,
+  private val tokenGenerator: TokenGenerator
```
alanpaulin: nit: Should nullable parameters go below required ones? that is, should we move TokenGenerator above Tracer?

misk-hibernate/src/main/kotlin/misk/hibernate/Session.kt
```diff
@@ -5,6 +5,9 @@ import kotlin.reflect.KClass
 
 interface Session {
   val hibernateSession: org.hibernate.Session
+  // Token to identify a session. Tokens stay the same across transaction retries.
```
alanpaulin: nice

tirsen: Yes but this should be a ktdoc, not a single line comment :-)

alanpaulin: 👍 True!

PR #1350, opened 2020-01-07
Add JSON FeatureFlag

ryanhall07:
The LDFeatureFlag impl uses the guice registered Moshi instance to
deserialize the JSON into a POJO.

Most likely app developers will screw up the JSON values in LD,
resulting in Moshi deserialization Exceptions in their app.
However, it should be a rather quick change to update the value in LD.
I believe the usefulness of JSON flags for DynamicConfigs outweighs this chance
of human error.

I want to use this to dynamically configure jobqueue and event
consumers. Something like a JobqueueConsumerConfig:

{
  "enabled": Boolean
  "batch_size": Int
  "schedule": TaskSchedule
}

TaskSchedule:

{
  "more_work_ms": Int
  "no_more_work_backoff: Backoff
  "error_backoff": Backoff
}

This would be extremely difficult to manage with individual flattened
flags like "jobqueue_consumer_schedule_more_work_ms".

### DIFF
--- a/misk-feature-testing/src/main/kotlin/misk/feature/testing/FakeFeatureFlags.kt
+++ b/misk-feature-testing/src/main/kotlin/misk/feature/testing/FakeFeatureFlags.kt
@@ -1,6 +1,7 @@
 package misk.feature.testing
 
 import com.google.common.util.concurrent.AbstractIdleService
+import com.squareup.moshi.Moshi
 import misk.feature.Attributes
 import misk.feature.DynamicConfig
 import misk.feature.Feature
@@ -9,13 +10,14 @@ import misk.feature.FeatureFlags
 import misk.feature.FeatureService
 import java.util.concurrent.ConcurrentHashMap
 import javax.inject.Inject
+import javax.inject.Provider
 import javax.inject.Singleton
 
 /**
  * In-memory test implementation of [FeatureFlags] that allows flags to be overridden.
  */
 @Singleton
-class FakeFeatureFlags @Inject constructor() : AbstractIdleService(),
+class FakeFeatureFlags @Inject constructor(private val moshi : Provider<Moshi>) : AbstractIdleService(),
     FeatureFlags,
     FeatureService,
     DynamicConfig {
@@ -51,10 +53,23 @@ class FakeFeatureFlags @Inject constructor() : AbstractIdleService(),
     return getOrDefault(feature, key, clazz.enumConstants[0]) as T
   }
 
+  override fun <T> getJson(
+    feature: Feature,
+    key: String,
+    clazz: Class<T>,
+    attributes: Attributes
+  ): T {
+    val json = get(feature, key) as? String ?: throw IllegalArgumentException(
+        "JSON flag $feature must be overridden with override() before use")
+    return moshi.get().adapter(clazz).fromJson(json)
+        ?: throw IllegalArgumentException("null value deserialized from $feature")
+  }
+
   override fun getBoolean(feature: Feature) = getBoolean(feature, KEY)
   override fun getInt(feature: Feature) = getInt(feature, KEY)
   override fun getString(feature: Feature) = getString(feature, KEY)
-  override fun <T : Enum<T>> getEnum(feature: Feature, clazz: Class<T>): T = getEnum(feature, KEY, clazz, Attributes())
+  override fun <T : Enum<T>> getEnum(feature: Feature, clazz: Class<T>): T = getEnum(feature, KEY, clazz)
+  override fun <T> getJson(feature: Feature, clazz: Class<T>): T = getJson(feature, KEY, clazz)
 
   private fun get(feature: Feature, key: String): Any? {
     FeatureFlagValidation.checkValidKey(feature, key)
@@ -87,6 +102,14 @@ class FakeFeatureFlags @Inject constructor() : AbstractIdleService(),
     overrides[MapKey(feature)] = value
   }
 
+  fun <T> override(feature: Feature, value: T, clazz: Class<T>) {
+    overrides[MapKey(feature)] = moshi.get().adapter(clazz).toJson(value)
+  }
+
+  inline fun <reified T> overrideJson(feature: Feature, value: T) {
+    override(feature, value, T::class.java)
+  }
+
   fun overrideKey(feature: Feature, key: String, value: Boolean) {
     overrides[MapKey(feature, key)] = value
   }
@@ -101,6 +124,14 @@ class FakeFeatureFlags @Inject constructor() : AbstractIdleService(),
 
   fun overrideKey(feature: Feature, key: String, value: Enum<*>) {
     overrides[MapKey(feature, key)] = value
+  }
+
+  fun <T> overrideKey(feature: Feature, key: String, value: T, clazz : Class<T>) {
+    overrides[MapKey(feature, key)] = moshi.get().adapter(clazz).toJson(value)
+  }
+
+  inline fun <reified T> overrideKeyJson(feature: Feature, key: String, value: T) {
+    overrideKey(feature, key, value, T::class.java)
   }
 
   fun reset() {
--- a/misk-feature-testing/src/test/kotlin/misk/feature/testing/FakeFeatureFlagsModuleTest.kt
+++ b/misk-feature-testing/src/test/kotlin/misk/feature/testing/FakeFeatureFlagsModuleTest.kt
@@ -1,8 +1,11 @@
 package misk.feature.testing
 
 import com.google.inject.Guice
+import com.squareup.moshi.Moshi
 import misk.feature.Feature
 import misk.feature.FeatureFlags
+import misk.inject.KAbstractModule
+import misk.moshi.MoshiAdapterModule
 import org.junit.jupiter.api.Test
 import kotlin.test.assertEquals
 
@@ -11,6 +14,12 @@ class FakeFeatureFlagsModuleTest {
   fun testModule() {
     val injector = Guice.createInjector(FakeFeatureFlagsModule().withOverrides {
       override(Feature("foo"), 24)
+    }, object : KAbstractModule() {
+      override fun configure() {
+        // Misk services automatically get this binding, but no need to depend on all of misk to
+        // test this.
+        bind<Moshi>().toInstance(Moshi.Builder().build())
+      }
     })
 
     val flags = injector.getInstance(FeatureFlags::class.java)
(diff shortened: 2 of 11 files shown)


misk-feature-testing/src/main/kotlin/misk/feature/testing/FakeFeatureFlags.kt
```diff
@@ -9,13 +10,14 @@ import misk.feature.FeatureFlags
...
 import javax.inject.Inject
+import javax.inject.Provider
 import javax.inject.Singleton
 
 /**
  * In-memory test implementation of [FeatureFlags] that allows flags to be overridden.
  */
 @Singleton
-class FakeFeatureFlags @Inject constructor() : AbstractIdleService(),
+class FakeFeatureFlags @Inject constructor(private val moshi : Provider<Moshi>) : AbstractIdleService(),
```
ryanhall07: unfortunately needs to be a Provider, since the guice module manually constructs the `FakeFeatureFlags` instance

misk-feature-testing/src/test/kotlin/misk/feature/testing/FakeFeatureFlagsModuleTest.kt
```diff
@@ -11,9 +14,15 @@ class FakeFeatureFlagsModuleTest {
   fun testModule() {
     val injector = Guice.createInjector(FakeFeatureFlagsModule().withOverrides {
       override(Feature("foo"), 24)
+    }, object : KAbstractModule() {
+      override fun configure() {
+        // Misk services automatically get this binding, but no need to depend on all of misk to
```
zcross: "no need to depend on all of misk" is a recurring theme in my experience

misk-launchdarkly-core/src/test/kotlin/misk/feature/launchdarkly/LaunchDarklyFeatureFlagsTest.kt
```diff
@@ -100,6 +103,28 @@ internal class LaunchDarklyFeatureFlagsTest {
...
+  fun getJson() {
+    val json = """{
+      "value" : "dino"
+    }""".trimIndent()
+    // LD uses Gson internally, so we need to do the same for mocking.
+    val gson = Gson()
+    val jsonElement = gson.toJsonTree(gson.fromJson(json, JsonFeature::class.java))
+    @Suppress("DEPRECATION")
+    Mockito
+        .`when`(client.jsonValueVariationDetail(anyString(), any(LDUser::class.java),
```
zcross: Nice, haven't run into much mockito usage yet in my misk project work

misk-launchdarkly-core/src/main/kotlin/misk/feature/launchdarkly/LaunchDarklyFeatureFlags.kt
```diff
@@ -75,6 +78,22 @@ class LaunchDarklyFeatureFlags @Inject constructor(
...
+    feature: Feature,
+    key: String,
+    clazz: Class<T>,
+    attributes: Attributes
+  ): T {
+    checkInitialized()
+    val result = ldClient.jsonValueVariationDetail(
+        feature.name,
+        buildUser(feature, key, attributes),
+        LDValue.ofNull())
```
zcross: Named arg for this null arg?

zcross: BTW, this PR isn't the place for this debate, but the lack of in-code default values for flags is still hard to get used to. As long as nobody resolved flags during app startup, and only certain code paths will fail due to unregistered flags, I guess it could be worse!

misk-launchdarkly-core/src/main/kotlin/misk/feature/launchdarkly/LaunchDarklyFeatureFlags.kt
```diff
@@ -75,6 +78,22 @@ class LaunchDarklyFeatureFlags @Inject constructor(
...
+    checkDefaultNotUsed(feature, result)
```
zcross: What's `checkDefaultNotUsed` doing?

misk-launchdarkly-core/src/test/kotlin/misk/feature/launchdarkly/LaunchDarklyFeatureFlagsTest.kt
ryanhall07: not a fan on mockito, but I'm being consistent

misk-feature-testing/src/main/kotlin/misk/feature/testing/FakeFeatureFlags.kt
```diff
@@ -87,6 +102,14 @@ class FakeFeatureFlags @Inject constructor() : AbstractIdleService(),
     overrides[MapKey(feature)] = value
   }
 
+  fun <T> override(feature: Feature, value: T, clazz: Class<T>) {
+    overrides[MapKey(feature)] = moshi.get().adapter(clazz).toJson(value)
```
swankjesse: there’s a dangerous feature that might be useful here...
```
 moshi.get().adapter(clazz).failOnUnknown().toJson(value)
```
The upside is it’ll catch typos where you say `shardCount` but wanted `shard_count`. The downside is if somebody does a deploy that takes away a field it’ll crash here rather than silently failing.

misk-launchdarkly-core/src/main/kotlin/misk/feature/launchdarkly/LaunchDarklyFeatureFlags.kt
ryanhall07: basically working around the LD convention of returning a default value when the flag doesn't exist. so instead of returning the default value, we blow up instead :(

misk-feature-testing/src/main/kotlin/misk/feature/testing/FakeFeatureFlags.kt
swankjesse: ... actually maybe we can have our cake and eat it too.

```
  val decoded = try { 
      moshi.get().adapter(clazz).failOnUnknown().toJson(value)
  } catch (e: JsonDataException) {
      logger.error(e)
      moshi.get().adapter(clazz).toJson(value)
  }
```


```

File: misk-feature-testing/src/test/kotlin/misk/feature/testing/FakeFeatureFlagsModuleTest.kt (hunk 2 above)
Comment (swankjesse) [reply misk-feature-testing/src/test/kotlin/misk/feature/testing/FakeFeatureFlagsModuleTest.kt]: We will modularize Misk!

File: misk-feature-testing/src/test/kotlin/misk/feature/testing/FakeFeatureFlagsModuleTest.kt [hunk 7]
```diff
@@ -11,9 +14,15 @@ class FakeFeatureFlagsModuleTest {
...
+        // test this.
+        bind<Moshi>().toInstance(Moshi.Builder().build())
```
swankjesse: .... might want the Kotlin adapter too?

swankjesse:
Nice!

zcross:
Serialization exceptions are a good example of why in-code default values could be useful.
How controversial would it be to expand the `Feature` API to _also_ include a safe get-or-default accessor that would bark loudly (e.g., with `log.warn`) and use the specified default. Maybe making it something you have to explicitly use would make it less controversial than making default value fallbacks a standard for all feature flag usage?

E.g., 
```
const val SOME_SAFE_DEFAULT = makeValidGsonThatWontExplode()
...
fun Feature.getOrDefault(..., defaultVal) {
  try {
    this.get(...)
  } catch (e: Exception) {
    log.warn("Flag not registered or value was invalid", e)
    SOME_SAFE_DEFAULT
  }
} 
...
...
val definitelyValidParams = zachsJsonFeature.getOrDefault(flagInput, SOME_SAFE_DEFAULT)
doStuff(definitelyValidParams)
```

misk-launchdarkly-core/src/test/kotlin/misk/feature/launchdarkly/LaunchDarklyFeatureFlagsTest.kt
mightyguava: your other option would have been to implement 20+ methods just to mock out a single method on a fake :)

ryanhall07: pretty easy to do with Intellij, just defaults to `TODO` in kt or `UnsupportedOperationException` in Java

mightyguava: But the clutter 😢

swankjesse:
@zcross I still feel like that default value is a little bomb in the codebase. Lots of feature flags are created with a initial value, and then get turned on in production to a target value. But the default in the code will always be the initial value.

misk-feature-testing/src/main/kotlin/misk/feature/testing/FakeFeatureFlags.kt
ryanhall07: nice. I like that. I initial avoided `failOnUnknown` because it makes rollouts of changes difficult.

misk-feature/src/main/kotlin/misk/feature/FeatureFlags.kt
```diff
@@ -50,6 +49,20 @@ interface FeatureFlags {
     attributes: Attributes = Attributes()
   ): T
 
+  /**
+   * Calculates the value of a JSON feature flag for the given key and attributes.
+   *
+   * @param clazz the type to convert the JSON string into. It is expected the a Moshi type adapter
```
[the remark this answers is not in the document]
ryanhall07: there are default via reflection, but it's kind of messy with kotlin. with moshi 1.9 you need to explicitly provide a kotlin reflection adapter. I'll do that.

ryanhall07:
> @zcross I still feel like that default value is a little bomb in the codebase. Lots of feature flags are created with a initial value, and then get turned on in production to a target value. But the default in the code will always be the initial value.

@zcross @swankjesse there is never going to be a golden answer to this problem, it is very use case specific. sometimes a default is totally sane, sometimes it could cause catastrophic consequences. additionally, failing a request that _could_ use a sane default could also lead to catastrophic consequences. 

The question is should we provide overloaded methods that accept an optional default. I think the concern there is it makes providing a default too easy, without forcing a developer to think through what happens if my code unexpectedly falls back to the default in the future. Is it really safe? 

After initially being an advocate for defaults, and implementing my own default fallback logic in a library, I'm starting to realize defaults are hard to reason about. Ultimately it's much easier to think about the failure scenario of "failing to talk to LD and failing the request" vs "failing to talk to LD and running an old code path". 

I think for now we force developers to implement their own try-catch-return default logic and reconsider adding it later.

swankjesse:
Yeah, I think you got it exactly right. Running old code is a dangerous failure mode, whereas 503ing is a well-known failure mode.

PR #1352, opened 2020-01-08
Try to use failOnUnknown when parsing JSON FeatureFlags

ryanhall07:
This will notify the application owner they have an incorrect json field
in LaunchDarkly, without failing the request.

In addition, addressed a few additional followups from https://github.com/cashapp/misk/pull/1350

### DIFF
--- a/misk-feature-testing/src/test/kotlin/misk/feature/testing/FakeFeatureFlagsModuleTest.kt
+++ b/misk-feature-testing/src/test/kotlin/misk/feature/testing/FakeFeatureFlagsModuleTest.kt
@@ -2,6 +2,7 @@ package misk.feature.testing
 
 import com.google.inject.Guice
 import com.squareup.moshi.Moshi
+import com.squareup.moshi.kotlin.reflect.KotlinJsonAdapterFactory
 import misk.feature.Feature
 import misk.feature.FeatureFlags
 import misk.inject.KAbstractModule
@@ -18,7 +19,9 @@ class FakeFeatureFlagsModuleTest {
       override fun configure() {
         // Misk services automatically get this binding, but no need to depend on all of misk to
         // test this.
-        bind<Moshi>().toInstance(Moshi.Builder().build())
+        bind<Moshi>().toInstance(Moshi.Builder()
+            .add(KotlinJsonAdapterFactory())
+            .build())
       }
     })
 
--- a/misk-feature/src/main/kotlin/misk/feature/MoshiExt.kt
+++ b/misk-feature/src/main/kotlin/misk/feature/MoshiExt.kt
@@ -1,0 +1,37 @@
+package misk.feature
+
+import com.squareup.moshi.JsonAdapter
+import com.squareup.moshi.JsonDataException
+import misk.logging.getLogger
+
+private val logger = getLogger<FeatureFlags>()
+
+/**
+ * Attempts to use [JsonAdapter.failOnUnknown] and logs any issues before falling back to ignoring
+ * the unknown fields.
+ */
+fun <T> JsonAdapter<T>.toSafeJson(value: T): String {
+  return try {
+    failOnUnknown().toJson(value)
+  } catch (e: JsonDataException) {
+    logger.error(e) {
+      "failed to serialize JSON due to unknown fields. ignoring those fields and trying again"
+    }
+    return toJson(value)
+  }
+}
+
+/**
+ * Attempts to use [JsonAdapter.failOnUnknown] and logs any issues before falling back to ignoring
+ * the unknown fields.
+ */
+fun <T> JsonAdapter<T>.fromSafeJson(json: String): T? {
+  return try {
+    failOnUnknown().fromJson(json)
+  } catch (e: JsonDataException) {
+    logger.error(e) {
+      "failed to parse JSON due to unknown fields. ignoring those fields and trying again"
+    }
+    return fromJson(json)
+  }
+}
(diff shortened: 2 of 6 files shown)


misk-feature-testing/src/test/kotlin/misk/feature/testing/FakeFeatureFlagsModuleTest.kt
```diff
@@ -18,7 +19,9 @@ class FakeFeatureFlagsModuleTest {
       override fun configure() {
         // Misk services automatically get this binding, but no need to depend on all of misk to
         // test this.
-        bind<Moshi>().toInstance(Moshi.Builder().build())
+        bind<Moshi>().toInstance(Moshi.Builder()
+            .add(KotlinJsonAdapterFactory())
```
swankjesse: thanks!

misk-feature/src/main/kotlin/misk/feature/MoshiExt.kt
```diff
@@ -0,0 +1,37 @@
...
+import com.squareup.moshi.JsonDataException
+import misk.logging.getLogger
+
+private val logger = getLogger<FeatureFlags>()
+
+/**
+ * Attempts to use [JsonAdapter.failOnUnknown] and logs any issues before falling back to ignoring
+ * the unknown fields.
+ */
+fun <T> JsonAdapter<T>.toSafeJson(value: T): String {
```
swankjesse: internal?

misk-feature/src/main/kotlin/misk/feature/MoshiExt.kt
```diff
@@ -0,0 +1,37 @@
...
+    }
+    return toJson(value)
+  }
+}
+
+/**
+ * Attempts to use [JsonAdapter.failOnUnknown] and logs any issues before falling back to ignoring
+ * the unknown fields.
+ */
+fun <T> JsonAdapter<T>.fromSafeJson(json: String): T? {
```
swankjesse: same

misk-feature/src/main/kotlin/misk/feature/MoshiExt.kt
ryanhall07: actually no, the fakes and launchdarkly impls are in different modules :(

misk-feature/src/main/kotlin/misk/feature/MoshiExt.kt
```diff
@@ -0,0 +1,37 @@
...
+
+private val logger = getLogger<FeatureFlags>()
+
+/**
+ * Attempts to use [JsonAdapter.failOnUnknown] and logs any issues before falling back to ignoring
+ * the unknown fields.
+ */
+fun <T> JsonAdapter<T>.toSafeJson(value: T): String {
+  return try {
+    failOnUnknown().toJson(value)
```
zcross: Converting a `T` to JSON, with the type `T` itself defining the schema, how can there be extraneous field that necessitate `failOnUnknown()`?

misk-feature/src/main/kotlin/misk/feature/MoshiExt.kt
```diff
@@ -0,0 +1,37 @@
...
+  }
+}
+
+/**
+ * Attempts to use [JsonAdapter.failOnUnknown] and logs any issues before falling back to ignoring
+ * the unknown fields.
+ */
+fun <T> JsonAdapter<T>.fromSafeJson(json: String): T? {
+  return try {
+    failOnUnknown().fromJson(json)
```
zcross: Makes sense to me that we will let operators know when they have specified an extraneous (or more likely, mistyped) field, without failing the entire parsing operation.

misk-feature/src/main/kotlin/misk/feature/MoshiExt.kt
```diff
@@ -0,0 +1,37 @@
...
+  } catch (e: JsonDataException) {
+    logger.error(e) {
+      "failed to parse JSON due to unknown fields. ignoring those fields and trying again"
+    }
+    return fromJson(json)
```
zcross: This will still fail if they neglect to specify a field and the type is `Type!`, right? Guess our hands are tied in that case (goes back to the debate of potentially confusing/unsafe default values versus exception throwing)

misk-feature/src/main/kotlin/misk/feature/MoshiExt.kt
ryanhall07: > LGTM but test cases over the failure modes would be helpful (e.g, in answering the questions I had around when a non-optional field on a `T` is not specified in the input JSON string)

@zcross tests to show what happens https://github.com/cashapp/misk/pull/1353

ryanhall07: you manually create the JSON in launch darkly, so you can easily add a field that doesn't exist, especially if you screw something up like `shardKey` vs `shard_key`

misk-feature/src/main/kotlin/misk/feature/MoshiExt.kt
ryanhall07: no it actually doesn't, it just sets the value to `null` and most likely an NPE will happen when you to access it in KT

ryanhall07: so moshi is actually breaking the "null type safety" of kotlin :(

ryanhall07: actually that's not true if you use the `KotlinJsonAdapterFactory`. See https://github.com/cashapp/misk/pull/1354/files

PR #1354, opened 2020-01-08
Configure Kotlin support in Moshi

swankjesse:
Without this Moshi will incorrectly decode null into non-nullable
fields which causes all kinds of problems.

### DIFF
--- a/misk-feature-testing/src/test/kotlin/misk/feature/testing/FakeFeatureFlagsTest.kt
+++ b/misk-feature-testing/src/test/kotlin/misk/feature/testing/FakeFeatureFlagsTest.kt
@@ -1,6 +1,7 @@
 package misk.feature.testing
 
 import com.google.inject.util.Providers
+import com.squareup.moshi.JsonDataException
 import com.squareup.moshi.Moshi
 import misk.feature.Feature
 import misk.feature.getEnum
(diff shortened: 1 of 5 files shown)


swankjesse:
Moshi started out as a Java library and this isn't strictly backwards-compatible due to differences in nullability between Java and Kotlin.

In Moshi 1.8 this is opt-in. In Moshi 1.9 we refuse to handle Kotlin types without being opted-in, so you can't really forget it.

misk-feature-testing/src/test/kotlin/misk/feature/testing/FakeFeatureFlagsTest.kt
```diff
@@ -91,12 +95,14 @@ internal class FakeFeatureFlagsTest {
   }
 
   @Test
-  fun `sets null for missing fields`() {
-    subject.overrideJsonString(FEATURE, "{}")
-    val feature = subject.getJson<JsonFeature>(FEATURE, TOKEN)
-    assertThat(feature.value).isNull()
-    assertThat(feature.optional).isNull()
-    assertThrows<NullPointerException> { feature.value.length }
+  fun `fails for missing required fields`() {
```
ryanhall07: @swankjesse it works!

PR #1355, opened 2020-01-09
Monitor external SQS Queue attributes and record them as metrics

frojasg:
We consume the following attribtues and record them as metrics:
- ApproximateNumberOfMessages
- ApproximateNumberOfMessagesNotVisible

Reference: https://docs.aws.amazon.com/AWSSimpleQueueService/latest/APIReference/API_GetQueueAttributes.html

### DIFF
--- a/misk-aws/src/main/kotlin/misk/jobqueue/sqs/AwsSqsQueueAttributeImporter.kt
+++ b/misk-aws/src/main/kotlin/misk/jobqueue/sqs/AwsSqsQueueAttributeImporter.kt
@@ -22,8 +22,8 @@ internal class AwsSqsQueueAttributeImporter @Inject constructor(
    * Spawn a new thread that will consume the [queueName] SQS attributes and record them as metrics
    *
    * The metric recorded are:
-   * - jobs_sqs_aproximate_number_of_messages
-   * - jobs_sqs_aproximate_number_of_messages_not_visible
+   * - ApproximateNumberOfMessagesVisible
+   * - ApproximateNumberOfMessagesNotVisible
    */
   fun import(queueName: QueueName) {
     val frequency = Duration.ofMillis(config.queue_attribute_importer_frequency_ms)
@@ -31,7 +31,10 @@ internal class AwsSqsQueueAttributeImporter @Inject constructor(
     taskQueue.scheduleWithBackoff(frequency) {
       val queue = queues[queueName]
       val lease = leaseManager.requestLease("sqs-queue-attributes-${queue.queueName}")
-      if (lease.checkHeld()) {
+      if (!lease.checkHeld()) {
+        metrics.sqsApproxNumberOfMessages.clear()
+        metrics.sqsApproxNumberOfMessagesNotVisible.clear()
+      } else {
         log.info {
           "recording metrics for queue ${queueName.value}"
         }
@@ -46,11 +49,11 @@ internal class AwsSqsQueueAttributeImporter @Inject constructor(
         }
 
         attributes[QueueAttributeName.ApproximateNumberOfMessages.toString()]?.let {
-          metrics.sqsApproxNumberOfMessages.labels(queue.queueName, queue.queueName)
+          metrics.sqsApproxNumberOfMessages.labels(metricNamespace, metricStat, queue.queueName, queue.queueName)
               .set(it.toDouble())
         }
         attributes[QueueAttributeName.ApproximateNumberOfMessagesNotVisible.toString()]?.let {
-          metrics.sqsApproxNumberOfMessagesNotVisible.labels(queue.queueName, queue.queueName)
+          metrics.sqsApproxNumberOfMessagesNotVisible.labels(metricNamespace, metricStat, queue.queueName, queue.queueName)
               .set(it.toDouble())
         }
       }
@@ -60,5 +63,7 @@ internal class AwsSqsQueueAttributeImporter @Inject constructor(
 
   companion object {
     private val log = getLogger<AwsSqsQueueAttributeImporter>()
+    internal const val metricNamespace = "AWS/SQS"
+    internal const val metricStat = "sum"
   }
 }
--- a/misk-aws/src/main/kotlin/misk/jobqueue/sqs/AwsSqsQueueAttributeImporter.kt
+++ b/misk-aws/src/main/kotlin/misk/jobqueue/sqs/AwsSqsQueueAttributeImporter.kt
@@ -1,0 +1,64 @@
+package misk.jobqueue.sqs
+
+import com.amazonaws.services.sqs.model.GetQueueAttributesRequest
+import com.amazonaws.services.sqs.model.QueueAttributeName
+import misk.clustering.lease.LeaseManager
+import misk.jobqueue.QueueName
+import misk.logging.getLogger
+import misk.tasks.RepeatedTaskQueue
+import misk.tasks.Status
+import java.time.Duration
+import javax.inject.Inject
+
+internal class AwsSqsQueueAttributeImporter @Inject constructor(
+  private val config: AwsSqsJobQueueConfig,
+  private val leaseManager: LeaseManager,
+  private val metrics: SqsMetrics,
+  private val queues: QueueResolver,
+  @ForSqsHandling private val taskQueue: RepeatedTaskQueue
+){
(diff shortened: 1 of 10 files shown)


misk-aws/src/test/kotlin/misk/jobqueue/sqs/SqsJobQueueTest.kt
```diff
@@ -29,7 +29,7 @@ internal class SqsJobQueueTest {
 
   @Inject private lateinit var sqs: AmazonSQS
   @Inject private lateinit var queue: JobQueue
-  @Inject private lateinit var consumer: JobConsumer
```
frojasg: I don't think we should expose `JobConsumer` interface.

misk-aws/src/main/kotlin/misk/jobqueue/sqs/SqsJobConsumer.kt
```diff
@@ -64,6 +66,50 @@ internal class SqsJobConsumer @Inject internal constructor(
...
+   * The metric recorded are:
+   * - jobs_sqs_aproximate_number_of_messages
+   * - jobs_sqs_aproximate_number_of_messages_not_visible
+   */
+  fun subscribeToMetrics(queueName: QueueName) {
+    log.info {
+      "recording metrics for queue ${queueName.value}"
+    }
+
+    taskQueue.scheduleWithBackoff(Duration.ZERO) {
```
frojasg: should we make this value configurable?

ryanhall07: at the very least don't make it zero

misk-aws/src/main/kotlin/misk/jobqueue/sqs/SqsJobConsumer.kt
```diff
@@ -64,6 +66,50 @@ internal class SqsJobConsumer @Inject internal constructor(
...
   }
 
+  /**
+   * Spawn a new thread that will consume the [queueName] SQS attributes and record them as metrics
+   *
+   * The metric recorded are:
+   * - jobs_sqs_aproximate_number_of_messages
+   * - jobs_sqs_aproximate_number_of_messages_not_visible
+   */
+  fun subscribeToMetrics(queueName: QueueName) {
```
ryanhall07: let's move this to another class.

misk-aws/src/main/kotlin/misk/jobqueue/sqs/SqsJobConsumer.kt
```diff
@@ -64,6 +66,50 @@ internal class SqsJobConsumer @Inject internal constructor(
...
+    log.info {
```
ryanhall07: i think you should log _after_ the lease has been checked

misk-aws/src/main/kotlin/misk/jobqueue/sqs/SqsJobConsumer.kt
```diff
@@ -64,6 +66,50 @@ internal class SqsJobConsumer @Inject internal constructor(
...
+            val request = GetQueueAttributesRequest()
+                .withQueueUrl(queue.url)
+                .withAttributeNames(
+                    QueueAttributeName.ApproximateNumberOfMessages,
+                    QueueAttributeName.ApproximateNumberOfMessagesNotVisible)
+            val response = client.getQueueAttributes(request)
+            response.attributes
+          }
+        } catch (e: ClientExecutionTimeoutException) {
+          log.info("timed out long polling for queue attributes: ${queue.queueName}")
```
ryanhall07: you shouldn't be long polling for queue attributes, this should have a reasonable timeout, like 5s and there isn't a reason to catch this exception because it's not normal.

we catch the Exception when receiving messages because we use long polling and sometimes the long poll exceeds the timeout, but we don't care.

misk-aws/src/main/kotlin/misk/jobqueue/sqs/SqsMetrics.kt
```diff
@@ -69,8 +69,22 @@ internal class SqsMetrics @Inject internal constructor(metrics: Metrics) {
...
   val sqsDeleteTime = metrics.histogram(
-    "jobs_sqs_delete_latency",
-    "the round trip time to delete messages from SQS",
-    listOf("queueName", "QueueName")
+      "jobs_sqs_delete_latency",
+      "the round trip time to delete messages from SQS",
+      listOf("queueName", "QueueName")
+  )
+
+  val sqsApproxNumberOfMessages = metrics.gauge(
```
ryanhall07: I think ideally these should match the metrics that we consume directly from Cloudwatch into signalfx, so the same dashboards can be used for external queues.

frojasg: updated

misk-aws/src/main/kotlin/misk/jobqueue/sqs/SqsJobConsumer.kt
```diff
@@ -1,7 +1,9 @@
 package misk.jobqueue.sqs
 
 import com.amazonaws.http.timers.client.ClientExecutionTimeoutException
+import com.amazonaws.services.sqs.model.GetQueueAttributesRequest
```
ryanhall07: remove?

misk-aws/src/main/kotlin/misk/jobqueue/sqs/AwsSqsQueueAttributeImporter.kt
```diff
@@ -0,0 +1,64 @@
...
+   * - jobs_sqs_aproximate_number_of_messages_not_visible
+   */
+  fun import(queueName: QueueName) {
+    val frequency = Duration.ofMillis(config.queue_attribute_importer_frequency_ms)
+
+    taskQueue.scheduleWithBackoff(frequency) {
+      val queue = queues[queueName]
+      val lease = leaseManager.requestLease("sqs-queue-attributes-${queue.queueName}")
+      if (lease.checkHeld()) {
+        log.info {
```
ryanhall07: you probably want to clear the metrics if you don't own the lease, since they are gauges

misk-aws/src/main/kotlin/misk/jobqueue/sqs/AwsSqsQueueAttributeImporter.kt
```diff
@@ -0,0 +1,64 @@
...
+  private val metrics: SqsMetrics,
+  private val queues: QueueResolver,
+  @ForSqsHandling private val taskQueue: RepeatedTaskQueue
+){
+
+  /**
+   * Spawn a new thread that will consume the [queueName] SQS attributes and record them as metrics
+   *
+   * The metric recorded are:
+   * - jobs_sqs_aproximate_number_of_messages
```
ryanhall07: update the metrics, or maybe just a code pointer instead

misk-aws/src/main/kotlin/misk/jobqueue/sqs/SqsMetrics.kt
```diff
@@ -69,8 +69,22 @@ internal class SqsMetrics @Inject internal constructor(metrics: Metrics) {
...
+      "ApproximateNumberOfMessagesVisible",
+      "the approximate number of messages available for retrieval from SQS",
+      listOf("queueName", "QueueName")
```
ryanhall07: I think we'll also need to include the dimensions that cloudwatch is setting. `namespace` and `stat`

misk-aws/src/test/kotlin/misk/jobqueue/sqs/AwsSqsQueueAttributeImporterTest.kt
```diff
@@ -0,0 +1,49 @@
...
+  }
+
+  @Test fun importQueueAttributes() {
+    importer.import(queueName)
+    queue.enqueue(queueName, "ok")
+    queue.enqueue(queueName, "ok")
+    queue.enqueue(queueName, "ok")
+    queue.enqueue(queueName, "ok")
+
+    Thread.sleep(100)
```
ryanhall07: this is going to be flaky. You should await the condition to be true

PR #1356, opened 2020-01-09
Lazily provide the JSON overrides in FakeFeatureFlags

ryanhall07:
This handles the case where the override is set in the
FakeFeatureFlagsModule. When called from the module, the Moshi instance
cannot be accessed yet, so we need to lazily serialize into JSON
when fetching.

### DIFF
--- a/misk-feature-testing/src/main/kotlin/misk/feature/testing/FakeFeatureFlags.kt
+++ b/misk-feature-testing/src/main/kotlin/misk/feature/testing/FakeFeatureFlags.kt
@@ -61,8 +61,12 @@ class FakeFeatureFlags @Inject constructor(private val moshi : Provider<Moshi>) 
     clazz: Class<T>,
     attributes: Attributes
   ): T {
-    val json = get(feature, key) as? String ?: throw IllegalArgumentException(
-        "JSON flag $feature must be overridden with override() before use")
+    val jsonFn = get(feature, key) as? Function0<*> ?: throw IllegalArgumentException(
+        "JSON flag $feature must be overridden with override() before use: ${get(feature, key)}")
+    // The JSON is lazily provided to handle the case where the override is provided by the
+    // FakeFeatureFlagModule and the Moshi instance cannot be accessed inside the module.
+    val json = jsonFn.invoke() as? String ?: throw IllegalArgumentException(
+        "JSON function did not provide a string")
     return moshi.get().adapter(clazz).fromSafeJson(json)
         ?: throw IllegalArgumentException("null value deserialized from $feature")
   }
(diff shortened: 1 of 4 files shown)


misk-feature-testing/src/test/kotlin/misk/feature/testing/FakeFeatureFlagsModuleTest.kt
```diff
@@ -15,17 +17,11 @@ class FakeFeatureFlagsModuleTest {
...
-    }, object : KAbstractModule() {
-      override fun configure() {
-        // Misk services automatically get this binding, but no need to depend on all of misk to
-        // test this.
-        bind<Moshi>().toInstance(Moshi.Builder()
-            .add(KotlinJsonAdapterFactory())
-            .build())
-      }
-    })
+      overrideJson(Feature("jsonFeature"), JsonFeature("testValue"))
```
ryanhall07: this exposes the previous bug. calling override from within the module.

misk-feature-testing/src/main/kotlin/misk/feature/testing/FakeFeatureFlags.kt
```diff
@@ -61,8 +61,12 @@ class FakeFeatureFlags @Inject constructor(private val moshi : Provider<Moshi>)
     clazz: Class<T>,
     attributes: Attributes
   ): T {
-    val json = get(feature, key) as? String ?: throw IllegalArgumentException(
-        "JSON flag $feature must be overridden with override() before use")
+    val jsonFn = get(feature, key) as? Function0<*> ?: throw IllegalArgumentException(
+        "JSON flag $feature must be overridden with override() before use: ${get(feature, key)}")
+    // The JSON is lazily provided to handle the case the override is provided by the
+    // FakeFeatureFlagModule and the Moshi instance cannot be accessed inside the module.
```
swankjesse: supernit: to handle the case _where_  the override

misk-feature-testing/src/test/kotlin/misk/feature/testing/MoshiTestingModule.kt
```diff
@@ -0,0 +1,18 @@
...
+import com.squareup.moshi.Moshi
+import com.squareup.moshi.kotlin.reflect.KotlinJsonAdapterFactory
+import misk.inject.KAbstractModule
+
+/**
+ * Binds a [Moshi] instance for testing.
+ *
+ * Misk services automatically get this binding, but no need to depend on all of misk to test this.
+ */
+internal class MoshiTestingModule : KAbstractModule() {
```
swankjesse: awesome

PR #1357, opened 2020-01-10
Constrain dependency versions with BOMs

swankjesse:
Imagine the following dependency graph:

    libFoo
      io.grpc:grpc-core:1.0
      io.grpc.grpc-context:1.0

    libBar
      io.grpc.grpc-core:1.5

An application that depends on libFoo and libBar will get this:

    io.grpc:grpc-core:1.5
    io.grpc.grpc-context:1.0

It's quite possible that running these two versions side-by-side will cause problems.
In particular, grpc-context might use internal APIs of grpc-core, and these might not
be strongly versioned.

The BOM attempts to solve this problem by advising versions.

This PR starts adopting BOMs of upstream projects where they exist.

Next steps:
 - publish BOMs for Misk (and other Square projects)
 - remove explicit versions for things also in BOMs


dependencies.gradle
```diff
@@ -13,10 +13,11 @@ ext.dep = [
   "gcpKms": "com.google.apis:google-api-services-cloudkms:v1-rev30-1.23.0",
   "gcpLogback": "com.google.cloud:google-cloud-logging-logback:0.39.0-alpha",
   "gradleBuildScan": "com.gradle:build-scan-plugin:2.3",
-  "grpcGenJava": "io.grpc:protoc-gen-grpc-java:1.14.0",
-  "grpcNetty": "io.grpc:grpc-netty:1.14.0",
-  "grpcProtobuf": "io.grpc:grpc-protobuf:1.14.0",
-  "grpcStub": "io.grpc:grpc-stub:1.14.0",
+  "grpcGenJava": "io.grpc:protoc-gen-grpc-java:1.21.0",
```
ryanhall07: and you're also updating the version?

build.gradle
```diff
@@ -77,6 +77,15 @@ subprojects {
   dependencies {
     testImplementation dep.junitApi
     testRuntimeOnly dep.junitEngine
+
+    // Platform/BOM dependencies constrain versions only.
+    api enforcedPlatform(dep.grpcBom)
```
ryanhall07: TIL what a BOM is. thanks!

dependencies.gradle
swankjesse: this is the version we end up using anyway, via one of our transitive deps

tirsen:
What does "BOM" mean?

swankjesse:
Bill of materials. I think it's a manufacturing metaphor.

PR #1359, opened 2020-01-14
Implement repeated tasks in FakeScheduledExecutorService

keeferrourke:
Implemented the  `scheduleWithFixedDelay` and `scheduleWithFixedRate` functions for the _FakeScheduledExecutorService_.

Since this is a single-threaded executor,  and the service always has to catch up to the clock, fixed delay is no different from fixed rate to my understanding.

### DIFF
--- a/misk-testing/src/main/kotlin/misk/concurrent/FakeScheduledExecutorService.kt
+++ b/misk-testing/src/main/kotlin/misk/concurrent/FakeScheduledExecutorService.kt
@@ -1,13 +1,12 @@
 package misk.concurrent
 
-import com.google.common.collect.MultimapBuilder
 import com.google.common.util.concurrent.MoreExecutors
 import java.time.Clock
 import java.util.concurrent.Callable
 import java.util.concurrent.Delayed
 import java.util.concurrent.ExecutorService
 import java.util.concurrent.FutureTask
-import java.util.concurrent.RunnableFuture
+import java.util.concurrent.PriorityBlockingQueue
 import java.util.concurrent.ScheduledExecutorService
 import java.util.concurrent.ScheduledFuture
 import java.util.concurrent.TimeUnit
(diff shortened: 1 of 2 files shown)


misk-testing/src/main/kotlin/misk/concurrent/FakeScheduledExecutorService.kt
```diff
@@ -46,39 +72,80 @@ class FakeScheduledExecutorService @Inject constructor(private val clock: Clock)
...
+    val isRepeated: Boolean get() = fixedDelay > 0
+
+    override fun run() {
+      if (isCancelled) throw CancellationException("FutureTask was cancelled")
+      super.run()
+    }
+
+    public override fun runAndReset(): Boolean {
+      return super.runAndReset()
+    }
```
mightyguava: This seems superfluous?

misk-testing/src/main/kotlin/misk/concurrent/FakeScheduledExecutorService.kt
```diff
@@ -17,21 +17,47 @@ import javax.inject.Singleton
...
 
   /** Check the current time on the clock and run any scheduled tasks that are due. */
   fun tick() {
-    futures.entries().stream()
-        .filter { entry -> entry.key <= clock.millis() }
-        .forEach { entry -> entry.value.run() }
+    var future = futures.poll()
+    while (future != null) {
+      if (this.isShutdown) {
+        future.cancel(false)
```
mightyguava: Why did you need to add cancellation? If tick isn't called, no futures will run.

misk-testing/src/main/kotlin/misk/concurrent/FakeScheduledExecutorService.kt
```diff
@@ -46,39 +72,80 @@ class FakeScheduledExecutorService @Inject constructor(private val clock: Clock)
...
+      if (isCancelled) throw CancellationException("FutureTask was cancelled")
+      super.run()
+    }
+
+    public override fun runAndReset(): Boolean {
+      return super.runAndReset()
+    }
+
+    internal fun reschedule() {
+      if (!isRepeated) return
```
mightyguava: redundant check. Might be better as an assert.

misk-testing/src/main/kotlin/misk/concurrent/FakeScheduledExecutorService.kt
keeferrourke: I would agree, but unfortunately it seems that `runAndReset()` needs to be overridden to make it a public method, so it may be used by the enclosing class.

misk-testing/src/main/kotlin/misk/concurrent/FakeScheduledExecutorService.kt
keeferrourke: Good point! Removed the cancellation.

misk-testing/src/main/kotlin/misk/concurrent/FakeScheduledExecutorService.kt
keeferrourke: Changed this to a `check` call.

PR #1364, opened 2020-01-20
Emit queuing latency metrics from the jetty threadpool queue

lcabancla:
Create a `TimedBlockingQueue` that will measure queuing latencies in the jetty threadpool queue when the config `enable_thread_pool_queue_metrics` is set to true.

### DIFF
--- a/misk-core/src/main/kotlin/misk/queuing/TimedBlockingQueue.kt
+++ b/misk-core/src/main/kotlin/misk/queuing/TimedBlockingQueue.kt
@@ -57,7 +57,7 @@ class TimedBlockingQueue<T>(
   }
 
   override fun removeAll(elements: Collection<T>): Boolean {
-    return removeItems(elements) > 0
+    return removeItems(elements)
   }
 
   override fun add(element: T): Boolean {
@@ -93,7 +93,7 @@ class TimedBlockingQueue<T>(
   }
 
   override fun remove(element: T): Boolean {
-    return removeItems(listOf(element)) > 0
+    return removeItems(listOf(element))
   }
 
   override fun containsAll(elements: Collection<T>): Boolean {
@@ -102,8 +102,9 @@ class TimedBlockingQueue<T>(
 
   override fun retainAll(elements: Collection<T>): Boolean {
     val toRemove = queue.filter {wrappedItem -> !elements.contains(wrappedItem.value)}
-        .map(this::unwrap)
-    return removeItems(toRemove) > 0
+    val result = queue.retainAll(wrapCollection(elements))
+    invokeDelayHandlerOnAll(toRemove)
+    return result
   }
 
   override fun remainingCapacity(): Int {
@@ -113,6 +114,7 @@ class TimedBlockingQueue<T>(
   override fun drainTo(c: MutableCollection<in T>): Int {
     val collection = mutableListOf<TimedQueueItem<T>>()
     val result = queue.drainTo(collection)
+    c.addAll(collection.map(this::unwrap))
     invokeDelayHandlerOnAll(collection)
     return result
   }
@@ -120,6 +122,7 @@ class TimedBlockingQueue<T>(
   override fun drainTo(c: MutableCollection<in T>, maxElements: Int): Int {
     val collection = mutableListOf<TimedQueueItem<T>>()
     val result = queue.drainTo(collection, maxElements)
+    c.addAll(collection.map(this::unwrap))
     invokeDelayHandlerOnAll(collection)
     return result
   }
@@ -158,11 +161,11 @@ class TimedBlockingQueue<T>(
     collection.forEach {item -> invokeDelayHandler(item)}
   }
 
-  private fun removeItems(collection: Collection<T>): Int {
+  private fun removeItems(collection: Collection<T>): Boolean {
     val itemsToRemove = find(collection)
-    queue.removeAll(wrapCollection(collection))
+    val result = queue.removeAll(wrapCollection(collection))
     invokeDelayHandlerOnAll(itemsToRemove)
-    return itemsToRemove.size
+    return result
   }
 
 }
--- a/misk-core/src/main/kotlin/misk/queuing/TimedBlockingQueue.kt
+++ b/misk-core/src/main/kotlin/misk/queuing/TimedBlockingQueue.kt
@@ -1,5 +1,6 @@
 package misk.queuing
 
+import java.time.Duration
 import java.util.concurrent.ArrayBlockingQueue
 import java.util.concurrent.BlockingQueue
 import java.util.concurrent.TimeUnit
@@ -7,18 +8,15 @@ import java.util.concurrent.TimeUnit
 /**
  * This BlockingQueue implementation allows measuring how long dequeued items have been
  * in the queue. It requires a delayHandler which consumes the latency every time an item is
- * removed from the queue by {@link BlockingQueue#take()},
- * {@link BlockingQueue#poll(long, TimeUnit)},
- * {@link BlockingQueue#poll()} or
- * {@link BlockingQueue#remove()}.
- * The operations
- * {@link BlockingQueue#drainTo(Collection)} and
- * {@link BlockingQueue#drainTo(Collection, int)} are not supported
+ * removed from the queue.
+ * Note that the the actual removal of an item from the queue and the delayHandler invocation
+ * are not atomic, hence it may not be appropriate for applications which require
+ * a higher level of accuracy.
  * @param <T> The type of the elements held in this queue
  */
 class TimedBlockingQueue<T>(
   maxQueueSize: Int,
-  private val delayHandler: (Long) -> Unit
+  private val delayHandler: (Duration) -> Unit
 ) : BlockingQueue<T> {
   private val queue: BlockingQueue<TimedQueueItem<T>> =
       ArrayBlockingQueue<TimedQueueItem<T>>(maxQueueSize)
@@ -50,6 +48,7 @@ class TimedBlockingQueue<T>(
   }
 
   override fun clear() {
+    invokeDelayHandlerOnAll(queue)
     queue.clear()
   }
 
@@ -58,7 +57,8 @@ class TimedBlockingQueue<T>(
   }
 
   override fun removeAll(elements: Collection<T>): Boolean {
-    return queue.removeAll(wrapCollection(elements))
+    val numRemoved = removeItems(elements)
+    return numRemoved > 0
   }
 
   override fun add(element: T): Boolean {
@@ -77,8 +77,12 @@ class TimedBlockingQueue<T>(
     return queue.map(this::unwrap).toMutableList().iterator()
   }
 
-  override fun peek(): T {
-    return unwrap(queue.peek())
+  override fun peek(): T? {
+    val item = queue.peek()
+    if (item != null) {
+      return unwrap(item)
+    }
+    return null
   }
 
   override fun put(e: T) {
@@ -90,7 +94,7 @@ class TimedBlockingQueue<T>(
   }
 
   override fun remove(element: T): Boolean {
-    return queue.remove(wrap(element))
+    return removeItems(listOf(element)) > 0
   }
 
   override fun containsAll(elements: Collection<T>): Boolean {
@@ -98,7 +102,9 @@ class TimedBlockingQueue<T>(
   }
 
   override fun retainAll(elements: Collection<T>): Boolean {
-    return queue.retainAll(wrapCollection(elements))
+    val toRemove = queue.filter {wrappedItem -> !elements.contains(wrappedItem.value)}
+        .map(this::unwrap)
+    return removeItems(toRemove) > 0
   }
 
   override fun remainingCapacity(): Int {
@@ -106,11 +112,17 @@ class TimedBlockingQueue<T>(
   }
 
   override fun drainTo(c: MutableCollection<in T>): Int {
-    throw UnsupportedOperationException()
+    val collection = mutableListOf<TimedQueueItem<T>>()
+    val result = queue.drainTo(collection)
+    invokeDelayHandlerOnAll(collection)
+    return result
   }
 
   override fun drainTo(c: MutableCollection<in T>, maxElements: Int): Int {
-    throw UnsupportedOperationException()
+    val collection = mutableListOf<TimedQueueItem<T>>()
+    val result = queue.drainTo(collection, maxElements)
+    invokeDelayHandlerOnAll(collection)
+    return result
   }
 
   override val size: Int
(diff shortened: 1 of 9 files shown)


misk-core/src/main/kotlin/misk/queuing/TimedBlockingQueue.kt
```diff
@@ -0,0 +1,136 @@
...
+ * {@link BlockingQueue#poll()} or
+ * {@link BlockingQueue#remove()}.
+ * The operations
+ * {@link BlockingQueue#drainTo(Collection)} and
+ * {@link BlockingQueue#drainTo(Collection, int)} are not supported
+ * @param <T> The type of the elements held in this queue
+ */
+class TimedBlockingQueue<T>(
+  maxQueueSize: Int,
+  private val delayHandler: (Long) -> Unit
```
ryanhall07: prefer `(Duration) -> Unit`

misk-core/src/main/kotlin/misk/queuing/TimedBlockingQueue.kt
```diff
@@ -0,0 +1,136 @@
...
+  override fun retainAll(elements: Collection<T>): Boolean {
+    return queue.retainAll(wrapCollection(elements))
+  }
+
+  override fun remainingCapacity(): Int {
+    return queue.remainingCapacity()
+  }
+
+  override fun drainTo(c: MutableCollection<in T>): Int {
+    throw UnsupportedOperationException()
```
ryanhall07: why not support these? isn't the impl just removing and adding to the provided collection?

misk/src/main/kotlin/misk/web/jetty/ThreadPoolQueueMetrics.kt
```diff
@@ -0,0 +1,20 @@
...
+
+import misk.metrics.Metrics
+import javax.inject.Inject
+import javax.inject.Singleton
+
+@Singleton
+class ThreadPoolQueueMetrics @Inject internal constructor(
+  metrics: Metrics
+) {
+  private val queueLatencyHistogram = metrics.histogram(
```
ryanhall07: don't we have other jetty metrics? Ideally all `jetty_thread_pool_` metrics would be co-located.

misk-core/src/main/kotlin/misk/queuing/TimedBlockingQueue.kt
lcabancla: The problem is with the type parameter `<in T>`. Since we are wrapping `T` to become `TimedQueueItem<T>`, we cannot directly use `queue.drainTo()`. Something like the following will have to be done as a workaround:

```
    val collection = mutableListOf<TimedQueueItem<T>>()
    val result = queue.drainTo(collection)
    for (item in collection) {
      c.add(unwrap(item))
    }
    return result
```

My initial thought is that the `drainTo()` operation is rarely used enough to justify not supporting it. I could be wrong though and if needed I can add the workaround.

misk-core/src/main/kotlin/misk/queuing/TimedBlockingQueue.kt
lcabancla: Sounds good.

misk-core/src/main/kotlin/misk/queuing/TimedBlockingQueue.kt
```diff
@@ -0,0 +1,136 @@
...
+  override fun clear() {
+    queue.clear()
+  }
+
+  override fun element(): T {
+    return unwrap(queue.element())
+  }
+
+  override fun removeAll(elements: Collection<T>): Boolean {
+    return queue.removeAll(wrapCollection(elements))
```
ryanhall07: shouldn't this invoke the `delayHandler`?

misk-core/src/main/kotlin/misk/queuing/TimedBlockingQueue.kt
```diff
@@ -0,0 +1,136 @@
...
+  override fun put(e: T) {
+    queue.put(wrap(e))
+  }
+
+  override fun isEmpty(): Boolean {
+    return queue.isEmpty()
+  }
+
+  override fun remove(element: T): Boolean {
+    return queue.remove(wrap(element))
```
ryanhall07: again shouldn't this invoke the handler?

misk-core/src/main/kotlin/misk/queuing/TimedBlockingQueue.kt
ryanhall07: I think it's worth it since it seems pretty trivial to support. You probably don't want to just `unwrap`, but rather also invoke the delay handler too right?

misk/src/main/kotlin/misk/web/jetty/ThreadPoolQueueMetrics.kt
lcabancla: We have [ThreadPoolMetrics](https://github.com/cashapp/misk/blob/master/misk/src/main/kotlin/misk/web/jetty/ThreadPoolMetrics.kt#L9) but this depends on `QueuedThreadPool` which is created inside `MiskWebModule`. Hence, adding a dependency to `ThreadPoolMetrics` in `MiskWebModule` is not possible as it will be circular.

misk-core/src/main/kotlin/misk/queuing/TimedBlockingQueue.kt
lcabancla: The problem is that there is no way to get the delay since these remove operations do not return the queued value (this one returns a boolean).

misk-core/src/main/kotlin/misk/queuing/TimedBlockingQueue.kt
lcabancla: same comment as above.

lcabancla: It's probably worth adding a comment on the javadoc that the delayHandler won't be invoked on these.

misk-core/src/main/kotlin/misk/queuing/TimedBlockingQueue.kt
ryanhall07: ah right, well that's lame. That would be really surprising to me as a caller, that some methods that remove elements don't update the metrics. I do see the javadoc makes it clear only a few methods actually update metrics, which is probably fine. An alternative is to just not support these other methods?

lcabancla: I am working on a workaround to find all matching items before calling `queue.removeAll()` so that we can invoke the delayHandler on each of them.

lcabancla: All operations that remove should now invoke the `delayHandler` in the latest commit

misk-core/src/main/kotlin/misk/queuing/TimedBlockingQueue.kt
lcabancla: All remove operations should now invoke the handler.

misk-core/src/main/kotlin/misk/queuing/TimedBlockingQueue.kt
```diff
@@ -0,0 +1,169 @@
...
+  override fun put(e: T) {
+    queue.put(wrap(e))
+  }
+
+  override fun isEmpty(): Boolean {
+    return queue.isEmpty()
+  }
+
+  override fun remove(element: T): Boolean {
+    return removeItems(listOf(element)) > 0
```
ryanhall07: nit: be consistent when you inline calls to `removeItems`

misk-core/src/test/kotlin/misk/queuing/TimedBlockingQueueTest.kt
```diff
@@ -0,0 +1,136 @@
...
+    assertEquals(arrayQueue.isEmpty(), timedQueue.isEmpty())
+    assertEquals(arrayQueue.remainingCapacity(), timedQueue.remainingCapacity())
+    assertEquals(arrayQueue.peek(), timedQueue.peek())
+    shouldThrowException({ timedQueue.element() }, "Exception expected")
+    shouldThrowException({ timedQueue.remove() }, "Exception expected")
+    assertEquals(arrayQueue.poll(), timedQueue.poll())
+    assertEquals(arrayQueue.poll(1, TimeUnit.SECONDS), timedQueue.poll(1, TimeUnit.SECONDS))
+  }
+
+  private fun shouldThrowException(lambda: () -> Unit, failureMsg: String) {
```
ryanhall07: use kt `assertFailsWith`? Specifying the exception type would be nice to prevent swallowing unintended errors

misk-core/src/main/kotlin/misk/queuing/TimedBlockingQueue.kt
lcabancla: Are you referring to this:
```
  override fun removeAll(elements: Collection<T>): Boolean {
    val numRemoved = removeItems(elements)
    return numRemoved > 0
  }
```

ryanhall07: yes. sorry I just made the comment as I was scrolling through the code and felt it wasn't consistent.

misk-core/src/main/kotlin/misk/queuing/TimedBlockingQueue.kt
```diff
@@ -0,0 +1,171 @@
...
+  override fun contains(element: T): Boolean {
+    return queue.contains(wrap(element))
+  }
+
+  override fun addAll(elements: Collection<T>): Boolean {
+    return queue.addAll(wrapCollection(elements))
+  }
+
+  override fun clear() {
+    invokeDelayHandlerOnAll(queue)
```
swankjesse: This isn’t atomic, so two concurrent calls could trigger multiple calls into the delay hander. Instead repeatedly call poll ?

misk-core/src/main/kotlin/misk/queuing/TimedBlockingQueue.kt
```diff
@@ -0,0 +1,171 @@
...
+    c.addAll(collection.map(this::unwrap))
+    invokeDelayHandlerOnAll(collection)
+    return result
+  }
+
+  override val size: Int
+    get() = queue.size
+
+  private fun <E> wrap(obj: E): TimedQueueItem<E> {
+    return TimedQueueItem(obj, System.currentTimeMillis())
```
swankjesse: Please use nanoTime(). This one can go backwards.

misk-core/src/main/kotlin/misk/queuing/TimedQueueItem.kt
```diff
@@ -0,0 +1,24 @@
+package misk.queuing
+
+import java.util.Objects
+
+/**
+ * A wrapper class around any queued type which provides a startTime field for computing
+ * the time it spent in the queue
+ */
+data class TimedQueueItem<T>(
```
swankjesse: internal?

misk-core/src/main/kotlin/misk/queuing/TimedBlockingQueue.kt
```diff
@@ -0,0 +1,171 @@
...
+/**
+ * This BlockingQueue implementation allows measuring how long dequeued items have been
+ * in the queue. It requires a delayHandler which consumes the latency every time an item is
+ * removed from the queue.
+ * Note that the the actual removal of an item from the queue and the delayHandler invocation
+ * are not atomic, hence it may not be appropriate for applications which require
+ * a higher level of accuracy.
+ * @param <T> The type of the elements held in this queue
+ */
+class TimedBlockingQueue<T>(
```
swankjesse: internal

misk/src/main/kotlin/misk/web/WebConfig.kt
```diff
@@ -35,6 +35,9 @@ data class WebConfig(
   /** Maximum number of items in the queue for Jetty's thread pool. */
   val jetty_max_thread_pool_queue_size: Int = 300,
 
+  /** Flag to enable thread pool queue metrics */
+  val enable_thread_pool_queue_metrics: Boolean = false,
```
swankjesse: add a todo to make true the default?

lcabancla: Will add

misk-core/src/main/kotlin/misk/queuing/TimedBlockingQueue.kt
lcabancla: Will change.

misk-core/src/main/kotlin/misk/queuing/TimedQueueItem.kt
lcabancla: Will make this internal. Thanks.

misk-core/src/main/kotlin/misk/queuing/TimedBlockingQueue.kt
lcabancla: This is being used by one of the modules in `misk.web`. Adding internal here causes a compilation error unless we move this class to the same package as the module. I feel that putting this class in a separate package and making it public allows reuse.

misk-core/src/main/kotlin/misk/queuing/TimedBlockingQueue.kt
lcabancla: The same problem applies to `remove(element: T)`, `removeAll()` and `retainAll()`. Will think about this and see if we can solve it in a more generic way.

lcabancla: Here are my thoughts:

We can use `poll()` for `clear()` as suggested. 

For `removeAll(collection: Collection<T>)`, `retainAll(collection: Collection<T>)`, we can follow a similar approach by calling `remove(element)` for each item in the collection. `remove(element)` in turn will invoke the delayHandler and the underlying `remove(element)`. This will work but since the delayHandler and the `remove(element)` calls are not protected by a lock, they can become out of sync. 

However, since the primary use case is for request queues, I think it is fine to leave as is. For example, the jetty QueuedThreadPool does not use the above operations. I have added comments about the limitations so that users are aware.

misk-core/src/main/kotlin/misk/queuing/TimedBlockingQueue.kt
swankjesse: Wanna move it to the misk.web module and keep it in this package? note that internal is module-scoped and is unrelated to the package name.

misk-core/src/main/kotlin/misk/queuing/TimedBlockingQueue.kt
swankjesse: I like the idea of keeping this internal-only and not spending too much energy implementing general-purpose features.

lcabancla: Agreed. Will move this to misk-web and make internal only.

misk-core/src/main/kotlin/misk/queuing/TimedBlockingQueue.kt
lcabancla: Done.

PR #1366, opened 2020-01-22
Add miskweb step to ensure tabs are built in public releases

adrw:


adrw:
@shellderp Most recent public release probably doesn't have web code but the next one will. Thanks for your patience!

shellderp:
sweet

PR #1369, opened 2020-01-26
Add support for HTTP DELETE method

biasedbit:
Adds support for HTTP DELETE method via `@Delete` annotation.

### DIFF
--- a/misk/src/main/kotlin/misk/web/DispatchMechanism.kt
+++ b/misk/src/main/kotlin/misk/web/DispatchMechanism.kt
@@ -8,9 +8,9 @@ import org.eclipse.jetty.http.HttpMethod
  */
 enum class DispatchMechanism {
   GET,
-  GRPC,
   POST,
   DELETE,
+  GRPC,
   WEBSOCKET;
 
   /**
@@ -23,12 +23,11 @@ enum class DispatchMechanism {
     get() {
       return when (this) {
         GET -> HttpMethod.GET
+        POST -> HttpMethod.POST
+        DELETE -> HttpMethod.DELETE
 
         // gRPC layers over POST. https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md
         GRPC -> HttpMethod.POST
-
-        POST -> HttpMethod.POST
-        DELETE -> HttpMethod.DELETE
 
         // WebSocket upgrades from GET.
         // https://developer.mozilla.org/en-US/docs/Web/HTTP/Protocol_upgrade_mechanism
--- a/misk/src/main/kotlin/misk/web/DispatchMechanism.kt
+++ b/misk/src/main/kotlin/misk/web/DispatchMechanism.kt
@@ -10,6 +10,7 @@ enum class DispatchMechanism {
   GET,
   GRPC,
   POST,
+  DELETE,
   WEBSOCKET;
 
   /**
@@ -27,6 +28,7 @@ enum class DispatchMechanism {
         GRPC -> HttpMethod.POST
 
         POST -> HttpMethod.POST
+        DELETE -> HttpMethod.DELETE
 
         // WebSocket upgrades from GET.
         // https://developer.mozilla.org/en-US/docs/Web/HTTP/Protocol_upgrade_mechanism
--- a/misk/src/main/kotlin/misk/web/actions/WebActionFactory.kt
+++ b/misk/src/main/kotlin/misk/web/actions/WebActionFactory.kt
@@ -10,6 +10,7 @@ import misk.asAction
 import misk.scope.ActionScope
 import misk.web.BoundAction
 import misk.web.ConnectWebSocket
+import misk.web.Delete
 import misk.web.DispatchMechanism
 import misk.web.Get
 import misk.web.NetworkInterceptor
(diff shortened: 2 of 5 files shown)


misk/src/test/kotlin/misk/web/actions/SupportedHttpMethodsTest.kt
```diff
@@ -0,0 +1,103 @@
...
+import misk.web.mediatype.MediaTypes
+import okhttp3.OkHttpClient
+import okhttp3.Request
+import okhttp3.RequestBody.Companion.toRequestBody
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.jupiter.api.Test
+import java.util.UUID
+import javax.inject.Inject
+
+// Tests all supported HTTP methods.
```
biasedbit: Couldn't seem to find a similar test to add the DELETE test to. No strong reason to test GET and POST (since there's a bunch of other code already testing it), just erring on the side of having a test for all HTTP methods.

Happy to drop GET/POST tests if y'all feel it's redundant.

misk/src/main/kotlin/misk/web/jetty/WebActionsServlet.kt
```diff
@@ -82,7 +86,7 @@ internal class WebActionsServlet @Inject constructor(
       val candidateActions = boundActions.mapNotNull {
         it.match(httpCall.dispatchMechanism, requestContentType, requestAccepts, httpCall.url)
       }
-      val bestAction = candidateActions.sorted().firstOrNull()
+      val bestAction = candidateActions.min()
```
biasedbit: I don't know what's our stance on minor, somewhat-unrelated, cosmetic changes (no contribution guidelines) so I threw this in (separate commit). Happy to drop/yank into separate PR.

misk/src/main/kotlin/misk/web/actions/WebActionFactory.kt
```diff
@@ -40,11 +41,12 @@ internal class WebActionFactory @Inject constructor(
     webActionClass: KClass<A>,
     pathPrefix: String = "/"
   ): List<BoundAction<A>> {
-    // Find the function with Get, Post, or ConnectWebSocket annotation. Only one such function is
-    // allowed.
+    // Find the function with Get, Post, Delete or ConnectWebSocket annotation.
+    // Only one such function is allowed.
     val actionFunctions = webActionClass.functions.mapNotNull {
       if (it.findAnnotationWithOverrides<Get>() != null ||
```
biasedbit: Wouldn't be a terrible idea to turn this into a loop over a list but a) I'll likely be adding support for other methods (PUT/PATCH) soon, and b) might as well make that change as its own PR.

misk/src/main/kotlin/misk/web/DispatchMechanism.kt
```diff
@@ -8,8 +8,9 @@ import org.eclipse.jetty.http.HttpMethod
  */
 enum class DispatchMechanism {
```
biasedbit: Grouping HTTP methods together in this enum was done as a separate commit. Trivial to drop if we feel strongly about alphabetical order.

misk/src/main/kotlin/misk/web/jetty/WebActionsServlet.kt
swankjesse: Tidying up along the way is nice.

misk/src/test/kotlin/misk/web/actions/SupportedHttpMethodsTest.kt
swankjesse: I think the most useful thing to test here is whether a body is required or expected. Could test that GET declarations can't have a body, and we should decide whether we permit/require bodies for delete.

misk/src/test/kotlin/misk/web/actions/SupportedHttpMethodsTest.kt
```diff
@@ -0,0 +1,103 @@
...
+        .build()
+
+    val response = httpClient.newCall(request).execute()
+    assertThat(response.isSuccessful).isTrue()
+    assertThat(response.body?.string()).isEqualTo(content)
+  }
+
+  @Test
+  fun delete() {
+    val content = UUID.randomUUID().toString()
```
swankjesse: Please, nothing random in tests. It brings about flakiness.

misk/src/test/kotlin/misk/web/actions/SupportedHttpMethodsTest.kt
```diff
@@ -0,0 +1,103 @@
...
+      install(WebActionModule.create<PostAction>())
+      install(WebActionModule.create<DeleteAction>())
+    }
+  }
+
+  internal class GetAction @Inject constructor() : WebAction {
+    @Get(PATH)
+    fun get(): String = CONTENT
+    companion object {
+      const val PATH = "/resources/id"
```
swankjesse: My preference is to avoid the constants and just use strings here. The benefits of the indirection don't pay for themselves.

In particular we really want to be able to copy & paste test code without refactoring other tests. Duplication across tests is okay; it means each test is more self-contained.

misk/src/test/kotlin/misk/web/actions/SupportedHttpMethodsTest.kt
biasedbit: Texas-size 10-4; will change.

misk/src/test/kotlin/misk/web/actions/SupportedHttpMethodsTest.kt
biasedbit: Hard agreement and will do!

misk/src/test/kotlin/misk/web/actions/SupportedHttpMethodsTest.kt
biasedbit: Can do. The spec doesn't explicitly forbid message body for DELETE requests (okhttp also seems to accept one in the `RequestBuilder`, though it defaults to empty), but the [section on Message Body](https://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html#sec4.3) has something on ignoring them when there's no defined semantics:

> A server SHOULD read and forward a message-body on any request; if the request method does not include defined semantics for an entity-body, then the message-body SHOULD be ignored when handling the request.

I recall coming across libraries that don't even allow specifying a body so I don't mind being opinionated about disallowing body for DELETEs. How do you reckon it's the best way to disallow content? Validations on `WebActionServlet#doDelete`, prior to delegating to `#handleCall`?

swankjesse: I think probably we should just forbid DELETE actions from declaring a request body. That way we blame the programmer for making a mistake, and don’t worry too much about what crazy internet users send us.

Maybe we do that in WebActionBinding.newWebActionBinder(). That can be a follow-up PR!

PR #1371, opened 2020-01-29
Moved QueryLimitsConfig outside internal ReflectionQuery

stanislaw89:
`QueryLimitsConfig` seems intended to be configurable but it's not because it's defined inside `internal ReflectionQuery`, which makes it `internal` as well. Thus `QueryLimitsConfig` is not accessible outside of the `misk-hibernate` module.


chris-ryan-square:

PR #1373, opened 2020-01-30
Add option to install retry queue

tlulu:


misk-aws/src/main/kotlin/misk/jobqueue/sqs/AwsSqsJobHandlerModule.kt
```diff
@@ -11,38 +11,47 @@ import kotlin.reflect.KClass
...
 /**
  * Install this module to register a handler for an SQS queue,
- * along with its corresponding retry queue
+ * and if specified, registers its corresponding retry queue.
  */
 class AwsSqsJobHandlerModule<T : JobHandler> private constructor(
   private val queueName: QueueName,
-  private val handler: KClass<T>
+  private val handler: KClass<T>,
+  private val installRetryQueue: Boolean
```
minostro: @tlulu what about having the `handler` and the `dlqHandler`, that way we are always sure that we will have a handler for the dead letter queue. Also I would add a constructor where the `handler` will be used as the `dlqHandler`

misk-aws/src/main/kotlin/misk/jobqueue/sqs/AwsSqsJobHandlerModule.kt
```diff
@@ -11,38 +11,47 @@ import kotlin.reflect.KClass
...
+      newMapBinder<QueueName, JobHandler>().addBinding(queueName.retryQueue).to(handler.java)
+    }
+
     install(ServiceModule<AwsSqsJobHandlerSubscriptionService>())
   }
 
   companion object {
-    inline fun <reified T : JobHandler> create(queueName: QueueName):
-        AwsSqsJobHandlerModule<T> = create(queueName, T::class)
+    inline fun <reified T : JobHandler> create(
```
ryanhall07: this is not going to be backwards binary compatible with existing callers. just add an overloaded method, or tag with `@JvmOverloads`

tlulu: Got it. Done

misk-aws/src/main/kotlin/misk/jobqueue/sqs/AwsSqsJobHandlerModule.kt
tlulu: I think we want to install it automatically without every service having to specify the `dlqHandler` right?

minostro: I thought that the intention was to force everybody to have a handler on their dead letter queues. If that's the intention, I think my proposal is one way to accomplish that. @ryanhall07 can you help us clarifying this?

ryanhall07: there is no point in having an automatic processor for your dlq. the entire point of the dlq is to move it _out_ of the queue for a human to manually inspect. once a human has decided to re-process an entry in the dlq, it is deleted from the dlq and moved to a retrq. the retryq has an automatic processor.

the reason for a separate retryq vs main queue is so we can have a different redrive policy on the retryq (deadletter after only a few attempts).

the idea is the retryq is completely transparent to the app owner, which is why we automatically install a handler for you.

misk-aws/src/main/kotlin/misk/jobqueue/sqs/AwsSqsJobHandlerModule.kt
ryanhall07: same for the other methods too

misk-aws/src/main/kotlin/misk/jobqueue/sqs/AwsSqsJobHandlerModule.kt
minostro: > the idea is the retryq is completely transparent to the app owner, which is why we automatically install a handler for you.

Yes, and that's what I am proposing. To keep that semantics as is, but let the client provide their handler for their dlq if they want.

minostro: My use of dlq was misleading, I was actually referring to the retry queue.

minostro: synced up with @tlulu on this. The boolean flag makes sense because we only want to install a handler for "regular" queues and not for dead-letter one.

misk-aws/src/main/kotlin/misk/jobqueue/sqs/AwsSqsJobHandlerModule.kt
tlulu: Oops sorry. Added @JvmOverloads

PR #1377, opened 2020-02-03
bump datadog version to 0.42.0

isafan:
Updating datadog version from 0.35.0 to 0.42.0. 

https://github.com/DataDog/dd-trace-java/releases/tag/v0.42.0


misk-hibernate-testing/build.gradle
```diff
@@ -6,12 +6,6 @@ dependencies {
   testImplementation dep.openTracingDatadog
 }
 
-test {
-  // To help that with parallel execution by reusing test setup,
-  // misk-hibernate-testing is made dependent on misk-hibernate
-  dependsOn ':misk-hibernate:test'
-}
```
mightyguava: add context on why you are doing this?

isafan: The test dependency on misk-hibernate causes gradle to assume that SpanInjectorTest is inside misk-hibernate when it's actually in misk-hibernate-testing. As a result SpanInjector Test fails to run.

misk-hibernate-testing/src/test/kotlin/misk/jdbc/SpanInjectorTest.kt
```diff
@@ -33,16 +33,17 @@ class SpanInjectorTest {
 
   @Test
   fun testDatadog() {
-    DDTracer(NoopWriter()).use { tracer ->
```
isafan: For context, this constructor is now deprecated so switched to builder

misk-hibernate-testing/build.gradle
mightyguava: Looks like this makes it so that if you attempt to run just tests in `misk-hibernate-testing` it'll trigger tests in `misk-hibernate` to run first. And then this messes up because Intellij added a filtering expression to gradle for a test in `misk-hibernate-testing` for a class that isn't in `misk-hibernate`.

I think this optimization may not even be needed anymore. Seems safe to remove.

PR #1378, opened 2020-02-05
Test that we enforce uniqueness constraints on TiDB and Cockroach

tirsen:


misk-hibernate/src/test/kotlin/misk/hibernate/TransacterTest.kt
```diff
@@ -364,8 +364,12 @@ abstract class TransacterTest {
   }
 
   @Test
-  @Disabled("Uniqueness constraints aren't reliably enforced on Vitess")
   fun constraintViolationCausesTransactionToRollback() {
+    // Uniqueness constraints aren't reliably enforced on Vitess
+    if (transacter.config().type.isVitess) {
+      return
```
swankjesse, line 370: If you do assumeTrue you'll get output that shows the test was skipped!

tirsen: ohhh nice, I've been looking for that!

PR #1382, opened 2020-02-07
Adding an KInstallOnceModule class.

mpawliszyn:
For situations where you want the flexiblity of ensuring a dependent module exists.

### DIFF
--- a/misk-inject/src/test/kotlin/misk/inject/InstallOnceTest.kt
+++ b/misk-inject/src/test/kotlin/misk/inject/InstallOnceTest.kt
@@ -1,0 +1,40 @@
+package misk.inject
+
+import misk.testing.MiskTest
+import misk.testing.MiskTestModule
+import org.assertj.core.api.Assertions
+import org.junit.jupiter.api.Test
+import javax.inject.Inject
+
+@MiskTest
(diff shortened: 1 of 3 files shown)


misk-inject/src/test/kotlin/misk/inject/InstallOnceTest.kt
```diff
@@ -0,0 +1,40 @@
...
+@MiskTest
+class InstallOnceTest {
+  @MiskTestModule
+  val module = TestInstallModule()
+
+  @Inject @TestAnnotation private lateinit var installedOnce: String
+
+  @Test
+  fun testInstallOnce() {
+    Assertions.assertThat(installedOnce).startsWith("only one of me with identity")
```
nt3rp: Doesn't this test need to also print the identity result? (i.e. `${System.identityHashCode(this)}`)

mpawliszyn: startsWith?

I just have to put something that is different with each instance. Because otherwise String already overloads equals and hashcode to be the same with the same contents.

adrw:
Cool if we close this for now @mpawliszyn ?

PR #1385, opened 2020-02-07
Force use of Python 3 for mkdocs CI

adrw:
mkdocs is fatally complaining about being run with Python 2. We can fix this by forcing CI to run mkdocs against Python 3.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -115,20 +115,10 @@ jobs:
     parallelism: 1
     steps:
       - checkout # check out source code to working directory
-      - run: |
-          sudo mkdir -p /usr/local/bin
-          sudo chown -R circleci:circleci /usr/local/bin
-      - run: |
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -115,20 +115,10 @@ jobs:
...
-      - run: |
-          sudo mkdir -p /usr/local/bin
-          sudo chown -R circleci:circleci /usr/local/bin
-      - run: |
-          sudo mkdir -p /usr/local/lib/python3.6/site-packages
-          sudo chown -R circleci:circleci /usr/local/lib/python3.6/site-packages
-      - restore_cache:
-          key: pipdeps-{{ .Branch }}
-      - run: sudo pip install mkdocs mkdocs-material
-      - save_cache:
```
adrw: Going to abandon caching for now since it was taking too long to get it working and it hasn't been speeding up builds in any meaningful way.

PR #1386, opened 2020-02-12
remove spammy no members available warning

mightyguava:
This is an expected state in all 1 pod clusters and on first service
launch.

### DIFF
--- a/misk-zookeeper/src/main/kotlin/misk/clustering/zookeeper/ZkLease.kt
+++ b/misk-zookeeper/src/main/kotlin/misk/clustering/zookeeper/ZkLease.kt
@@ -1,5 +1,6 @@
 package misk.clustering.zookeeper
 
+import misk.clustering.Cluster
 import misk.clustering.NoMembersAvailableException
 import misk.clustering.lease.Lease
 import misk.clustering.weights.ClusterWeightProvider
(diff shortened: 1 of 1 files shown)


misk-zookeeper/src/main/kotlin/misk/clustering/zookeeper/ZkLease.kt
```diff
@@ -201,7 +201,7 @@ internal class ZkLease(
     val desiredLeaseOwner = try {
       clusterSnapshot.resourceMapper[leaseResourceName]
     } catch (e: NoMembersAvailableException) {
-      log.warn { e.message }
+      // no healthy members in the cluster yet
```
ryanhall07: can we instead add `tryResourceMapper` method that returns a nullable owner?

mightyguava: sure, I added it to this class, didn't want to change the interface

tirsen: just so I understand this correctly: what does returning null here mean? does that mean the current process is the owner? or it means there is no owner so it tries to grab the lock?

mightyguava: It means there is no healthy pod to act as the owner, so no one will try to grab the lock

PR #1387, opened 2020-02-12
Add AppScopedDynamicConfigResolver and related module

zcross:
It's an unsurprising usage pattern to rely on feature flags to decouple config changes from deploys for _app scoped_ functionality. We don't technically have a feature gap around this, but this PR makes an addition to make the process of implementing app config overrides a little less repetitive for service owners.

For example, it's convenient to:
- Dynamically override configuration related to job queue handlers (e.g,. disable them, change batch size, etc)
- Dynamically disable or change configurations related to event consumers/publishers
- Make other "config only" changes _to an app_ without deploying.

I began to write the equivalent of this for the specific case of consumer config, but thought it may save other people some time to go ahead and implement the generic form of it. I'm anticipating some backlash on the "default value" usage for invalid feature flag value settings, but I'm happy to discuss it!


misk-feature/src/main/kotlin/misk/feature/RealAppScopedDynamicConfigResolver.kt
```diff
@@ -0,0 +1,31 @@
+package misk.feature
+
+import misk.logging.getLogger
+
+internal class RealAppScopedDynamicConfigResolver<T : ValidatableConfig<T>> constructor(
```
ryanhall07: i'm having a hard time parsing what all the words mean in this class name. what is meant by app scoped? aren't all configs apps scoped. seems like `DynamicConfigResolver` is a pretty name IMO.

misk-feature/src/main/kotlin/misk/feature/RealAppScopedDynamicConfigResolver.kt
```diff
@@ -0,0 +1,31 @@
...
+  private val appName: String,
+  private val configName: String,
+  private val dynamicConfig: DynamicConfig,
```
ryanhall07: i could see this being useful for `FeatureFlags` as well. For example you have a single feature flag for all your topic configs and the flag is keyed on topic. that's what we did for jobqueues.

misk-feature/src/main/kotlin/misk/feature/ValidatableConfig.kt
```diff
@@ -0,0 +1,10 @@
+package misk.feature
+
+import misk.config.Config
+
+/**
+ * Interface used by [ConsumerConfigResolver] to support validation of retrieved configs.
+ */
+interface ValidatableConfig<T> : Config {
```
ryanhall07: nit: `ValidatedConfig`

misk-feature/src/main/kotlin/misk/feature/ValidatableConfig.kt
```diff
@@ -0,0 +1,10 @@
...
+  fun validate()
```
ryanhall07: what are semantics of this method? does it throw? what does it throw?

misk-feature/src/main/kotlin/misk/feature/AppScopedDynamicConfigResolver.kt
```diff
@@ -0,0 +1,13 @@
+package misk.feature
+
+/**
+ * Abstraction for dynamically retrieving dynamic config overrides, scoped to the calling app.
+ */
+interface AppScopedDynamicConfigResolver<T : ValidatableConfig<T>> {
+  /**
+   * Resolves the named configuration scoped to the calling application.
+   * If the underlying feature is not found, or its JSON representation cannot be interpreted,
```
ryanhall07: this seems very specific to the `Real` impl. why do you even have an interface? would we ever need a fake resolver. seems like `FakeFeatureFlags` is enough.

misk-feature/src/main/kotlin/misk/feature/RealAppScopedDynamicConfigResolver.kt
```diff
@@ -0,0 +1,31 @@
...
+  private val failureModeDefault: T
+) : AppScopedDynamicConfigResolver<T> {
+  private val logger = getLogger<AppScopedDynamicConfigResolver<T>>()
+
+  override fun resolveConfig(): T {
+    require(!configName.startsWith(appName)) {
+      "Redundant '$appName' prefix for ConsumerConfigResolver managed app config '$configName'"
```
ryanhall07: update

misk-feature/src/main/kotlin/misk/feature/RealAppScopedDynamicConfigResolver.kt
```diff
@@ -0,0 +1,31 @@
...
+    }
+
+    val featureName = listOf(appName, configName).joinToString("-")
```
ryanhall07: it's not clear that the feature must be named like this

ryanhall07: why enforce this anyways?

misk-feature/src/main/kotlin/misk/feature/RealAppScopedDynamicConfigResolver.kt
```diff
@@ -0,0 +1,31 @@
...
+    return try {
+      val feature = Feature(featureName)
+      val config = dynamicConfig.getJson(feature, failureModeDefault::class.java)
+      config.validate()
+      config
+    } catch (e: Exception) {
+      logger.warn(e) {
+        "Failed to retrieve or parse JSON for config override '$featureName', using $failureModeDefault"
```
ryanhall07: you could probably catch a more typed exception to make it clear it was a JSON parsing error.

misk-feature-testing/src/test/kotlin/misk/feature/testing/TestModule.kt
```diff
@@ -0,0 +1,14 @@
+package misk.feature.testing
+
+import misk.config.AppNameModule
+import misk.inject.KAbstractModule
+import misk.logging.LogCollectorModule
+
+class TestModule : KAbstractModule() {
```
ryanhall07: internal ?

misk-feature/src/main/kotlin/misk/feature/AppScopedDynamicConfigModule.kt
```diff
@@ -0,0 +1,50 @@
...
+import javax.inject.Provider
+import kotlin.reflect.KClass
+
+/**
+ * A module that binds a single instance of [AppScopedDynamicConfigResolver] for the given type,
+ * registering the name suffix (following appName-) to be used to retrieve the backing feature.
+ * Registration also involves the specification of a failure mode default (failure modes including
+ * invalid feature value specification that makes the output fail parsing).
+ *
+ * Use [AppScopedDynamicConfigModule.create] to register a config name with a config type.
```
ryanhall07: this seems quite complicated to use. you should probably add some code snippets

misk-feature/src/main/kotlin/misk/feature/RealAppScopedDynamicConfigResolver.kt
zcross: Yeah, this is poorly based on an implementation detail (the underlying feature's name being prefixed with the app name) that isn't relevant to the abstraction. Will rename to your much better recommendation

misk-feature/src/main/kotlin/misk/feature/RealAppScopedDynamicConfigResolver.kt
zcross: Hm, this makes me think that `FeatureFlags` would be preferable to `DynamicConfig` as the basis of this... Forcing access without key inputs might force app owners to maintain a single flag mega-config object (e.g., nested config) _or_ to add numerous feature flags (instead of single variations using a single key). 

I could see it being nice to have a few separate features for configuring different app components, but it would be way better to have N event consumer's `ConsumerConfig`s provided by a single flag given the topic as an input than to compose all of them into one global "all event consumers" config feature.

misk-feature/src/main/kotlin/misk/feature/ValidatableConfig.kt
zcross: Is it too loose to just suggest IllegalState/IllegalArgument exceptions?

misk-feature/src/main/kotlin/misk/feature/ValidatableConfig.kt
zcross: Much better, good nit

misk-feature/src/main/kotlin/misk/feature/AppScopedDynamicConfigResolver.kt
zcross: Yeah, no need considering this is nothing more than a shim

misk-feature/src/main/kotlin/misk/feature/RealAppScopedDynamicConfigResolver.kt
zcross: Busted!!

mightyguava:
I have mixed feelings about this.

1. Adding JSON flag support at all to feature flags was a reluctant decision. By using JSON flags, we lose type safety and documentation and other nice things that come from each flag controlling only a single behavior. Here you are validating the data only at the consumer side. It's actually more important to have validation where you enter values for the flag, but that support is missing. If we are going to be extending this type of pattern, we should probably look into disabling JSON flags in LaunchDarkly UI, and build tooling to create flags via commandline, having types documented and validated using protobuf or something.
2. Default values in code is bad for dynamic config is bad. Complex JSON default values in code is really bad. The defaults will inevitably become stale, and become an un-exercised code path. And in the unlikely case that we ever use a default, we can run into bugs and not know it. Why do you need defaults? I think hear pretty often: we need a value when LaunchDarkly is down. That's actually a situation that doesn't exist. If LaunchDarkly is down when you app is starting, Misk is currently configured to crash. If LaunchDarkly is down while your app is running, it already has the latest known copy. So why do you need a default?

I see the benefits to having JSON flags. In the restricted use case where pubsub team wants to provide a dynamically configurable config for job queues, this makes sense. I think the benefits could potentially outweigh the cons, especially if the config is kept simple, and well-documented.

I don't see the benefit of having default values. If the worry is that people will pull in the library but forget to create a flag in LaunchDarkly, causing their apps to crash on startup, that is an exceptional case. I'd be better if the code that calls into this handles the exception and then provides a default.

misk-feature/src/main/kotlin/misk/feature/RealAppScopedDynamicConfigResolver.kt
zcross: This is probably overly prescriptive, but I wanted to account for the not-yet-enforced convention of scoping feature flags to a service by prefixing them with the app name.

I want to reduce the chances of collisions between features registered for misk service deployments in an environment sharing one namespace for all their features (as we do with LaunchDarkly)

misk-feature/src/main/kotlin/misk/feature/RealAppScopedDynamicConfigResolver.kt
zcross: Good call, this can be a lot more specific!

zcross:
@mightyguava 

> I don't see the benefit of having default values. If the worry is that people will pull in the library but forget to create a flag in LaunchDarkly, causing their apps to crash on startup, that is an exceptional case. I'd be better if the code that calls into this handles the exception and then provides a default.

Not rhetorically, what do you think about the different error case of an operator (e.g., oncall engineer) tweaking a feature flag but doing so in a way that causes the JSON flag to be difficult to parse?

> By using JSON flags, we lose type safety and documentation and other nice things that come from each flag controlling only a single behavior. 

You were probably getting at the above in this point. In the use case I actually imagined this for, it would be high leverage to merely have a "library-ified" *boolean* flag instead of a JSON flag. Specifically, I want the ability to disable event consumption for a topic without making a deploy.
However, the presence of other useful knobs like batch sizes, backoff policy, etc (in my example) are good clues that having JSON feature flags may be worth having. The alternative would be having a set of keyed feature flags of the primitive types that are logically read piecemeal by services, and maybe this would happen organically if we didn't provide structured (JSON) flags.

misk-feature/src/main/kotlin/misk/feature/RealAppScopedDynamicConfigResolver.kt
zcross: What if I just require this by validation instead of constructing it myself here? 
I.e., I'll just start with `require(configName.startsWith(appName))` instead of `require(!configName.startsWith(appName))`

Then, implementors will be on the same page and the code that registers these config names will both have to start with the app name *and* it will appear in code as it is registered in the feature flag service (e.g., LD)

mightyguava:
> Not rhetorically, what do you think about the different error case of an operator (e.g., oncall engineer) tweaking a feature flag but doing so in a way that causes the JSON flag to be difficult to parse?

In that case, the app should definitely start throwing exceptions. We don't want a fallback to a default that may or may not be good. We would mostly solve this by having input validation as I was talking about above. JSON is too dangerous to use as a general configuration mechanism without proper tooling.

> The alternative would be having a set of keyed feature flags of the primitive types that are logically read piecemeal by services, and maybe this would happen organically if we didn't provide structured (JSON) flags.

You do have a good use case, yes, but so far we have only the 1. Your consumer config is the same use case as the previously added publisher config. When we run into more cases where libraries want to have dynamic, complex configs, that'd be a call to add tooling for it to make it safe.

I recommend removing default value support from this PR, and add some warning signs in the code documentation on how JSON can be dangerous. Since your use case is within the same events library, you could go even a step further and just have this code in that library, where you can feel free to keep the default value support.

zcross:
> > Not rhetorically, what do you think about the different error case of an operator (e.g., oncall engineer) tweaking a feature flag but doing so in a way that causes the JSON flag to be difficult to parse?
> 
> In that case, the app should definitely start throwing exceptions. We don't want a fallback to a default that may or may not be good. We would mostly solve this by having input validation as I was talking about above. JSON is too dangerous to use as a general configuration mechanism without proper tooling.
> 
> > The alternative would be having a set of keyed feature flags of the primitive types that are logically read piecemeal by services, and maybe this would happen organically if we didn't provide structured (JSON) flags.
> 
> You do have a good use case, yes, but so far we have only the 1. Your consumer config is the same use case as the previously added publisher config. When we run into more cases where libraries want to have dynamic, complex configs, that'd be a call to add tooling for it to make it safe.
> 
> Since your use case is within the same events library, you could go even a step further and just have this code in that library, where you can feel free to keep the default value support.

Sadly, this is where I started (before thinking I should make this a more general feature in Misk). Ha!

misk-feature/src/main/kotlin/misk/feature/RealAppScopedDynamicConfigResolver.kt
ryanhall07: just expose an api that allows both key and non-key lookups. under the hood you'll need to use `FeatureFlags` probably.

misk-feature/src/main/kotlin/misk/feature/ValidatableConfig.kt
ryanhall07: i think throwing IllegalArgumentException is fine.

misk-feature/src/main/kotlin/misk/feature/RealAppScopedDynamicConfigResolver.kt
mightyguava: For this use case, you should probably inject `LDClientInterface` directly. You don't need the `FeatureFlags` abstraction

ryanhall07:
@mightyguava is pretty spot on and I agree with his points. '

JSON flags are pretty scary and we should probably limit them for now, until we can build proper support for validating the input to LD. But yes JSON flags are super valuable for pubsub stuff (jobqueue and event consumers), so maybe we just limit them for now. there is actually a lot of overlap between jobqueue and event stuff, so we should probably start a separate repo for that stuff.

re: default values. yea using a cached value from LD is a far better solution :P

zcross:
Ok, I'll move this closer to its intended use case and PR something specific to events and/or job queues

PR #1388, opened 2020-02-13
TiDB or Cockroach don't need a separate connection for replica reads

tirsen:


misk-hibernate/src/main/kotlin/misk/hibernate/RealTransacter.kt
```diff
@@ -253,6 +253,9 @@ internal class RealTransacter private constructor(
     if (config.type.isVitess) {
       return sessionFactory
     }
+    if (config.type == DataSourceType.COCKROACHDB || config.type == DataSourceType.TIDB) {
```
kalfonso: nit: can it be collapsed into one single conditional for vitess, crdb and tidb?

mightyguava: For TiDB, we should probably directly use follower reads and don't even call this function? Something like:

```
  override fun <T> replicaRead(block: (session: Session) -> T): T {
    check(!inTransaction) { "Can't do replica reads inside a transaction" }
    try {
      session.useConnection { connection ->
        connection.createStatement().use { s ->
          s.execute("set @@tidb_replica_read = 'follower'")
        }
      }
      block(session)
    } finally {
      session.useConnection { connection ->
        connection.createStatement().use { s ->
          s.execute("set @@tidb_replica_read = ''")
        }
      }
    }
  }
```

tirsen: OH!! I was looking for something like that but couldn't find anything googling "tidb replica reads"

PR #1389, opened 2020-02-14
Fixing kotlin moshi use for slack integrations.

mpawliszyn:


misk-slack/src/main/kotlin/misk/slack/SlackModule.kt
```diff
@@ -33,4 +36,9 @@ class SlackModule(private val config: SlackConfig) : KAbstractModule() {
         .build()
     return retrofit.create(SlackWebhookApi::class.java)
   }
+  @Provides @com.google.inject.Singleton @Named("emailBridge") fun provideMoshi(): Moshi? {
```
swankjesse: Moshi? –> Moshi

Why fully-qualify `@Singleton` ?

swankjesse: emailBridge ?!

mpawliszyn: 🤦‍♂ ...
Better?
Am I missing anything that you know of?

PR #1395, opened 2020-02-19
Close the Jaeger tracer when service shuts down

swankjesse:
This was leaking threads, two per test.

Also upgrade all of our tracing-related dependencies including:
 - Open Tracing
 - Prometheus
 - Datadog Tracing
 - Zipkin


misk-jaeger/src/main/kotlin/misk/tracing/backends/jaeger/JaegerTracingService.kt
```diff
@@ -18,5 +19,6 @@ internal class JaegerTracingService @Inject internal constructor(
   }
 
   override fun shutDown() {
+    tracer.close()
```
mpawliszyn: 👍

misk/src/test/kotlin/misk/web/interceptors/TracingInterceptorTest.kt
```diff
@@ -92,11 +92,15 @@ class TracingInterceptorTest {
...
-              chain.proceed(chain.httpCall)
-            }
+        val span = tracer.buildSpan("parent-span-exists").start()
+        val scope = tracer.scopeManager().activate(span)
+        try {
+          tracer.activeSpan().setBaggageItem("hello", "world")
+          chain.proceed(chain.httpCall)
+        } finally {
+          scope.close()
+          span.finish()
```
swankjesse: FYI @mpawliszyn this didn’t remain as nice as it was originally, thanks to the need to finish()

misk/src/test/kotlin/misk/web/interceptors/TracingInterceptorTest.kt
```diff
@@ -92,11 +92,15 @@ class TracingInterceptorTest {
     val chain = RealNetworkChain(TracingTestAction::call.asAction(DispatchMechanism.GET),
```
mpawliszyn: Do we want a test for shutdown to make sure it works?

PR #1398, opened 2020-02-20
Adding DynamoDb to Misk

abakes:
- real and fake modules
- base entities for common logic (convention of `Dy` prefix for entities for Dynamo instead of `Db` for SQL
- a local version of dynamoDb to run in Docker (as external dependency) for testing


misk-aws/src/main/kotlin/misk/dynamodb/DyEntity.kt
```diff
@@ -0,0 +1,14 @@
+package misk.dynamodb
+
+import com.amazonaws.services.dynamodbv2.datamodeling.DynamoDBTable
+import com.amazonaws.services.dynamodbv2.datamodeling.DynamoDBVersionAttribute
+
+/*
+ * This version field is used for optimistic locking.
+ */
+@DynamoDBTable(tableName = "")
```
mightyguava: do abstract entities need table names?

misk-aws/src/main/kotlin/misk/dynamodb/DyTimestampedEntity.kt
```diff
@@ -0,0 +1,19 @@
+package misk.dynamodb
+
+import com.amazonaws.services.dynamodbv2.datamodeling.DynamoDBAttribute
+import com.amazonaws.services.dynamodbv2.datamodeling.DynamoDBAutoGenerateStrategy
+import com.amazonaws.services.dynamodbv2.datamodeling.DynamoDBAutoGeneratedTimestamp
+import com.amazonaws.services.dynamodbv2.datamodeling.DynamoDBTable
+import java.util.Date
+
+@DynamoDBTable(tableName = "")
+abstract class DyTimestampedEntity : DyEntity() {
```
mightyguava: Do we have a use case for a non-timestamped entity?

misk-aws/src/main/kotlin/misk/dynamodb/RealDynamoDbModule.kt
```diff
@@ -0,0 +1,46 @@
...
+        .build()
+  }
+
+  @Provides @Singleton
+  fun providesDynamoDBmapper(dynamoDb: AmazonDynamoDB): DynamoDBMapper {
+    return dynamoDbMapper(dynamoDb)
+  }
+
+  companion object {
+    internal fun dynamoDbMapper(dynamoDb: AmazonDynamoDB): DynamoDBMapper {
```
mightyguava: inline this function?

misk-aws/src/main/kotlin/misk/dynamodb/RealDynamoDbModule.kt
```diff
@@ -0,0 +1,46 @@
...
+      return DynamoDBMapper(
+          dynamoDb,
+          DynamoDBMapperConfig.Builder()
+              .withSaveBehavior(DynamoDBMapperConfig.SaveBehavior.CLOBBER)
```
mightyguava: Is this an option that should be exposed via a config?

misk-aws/src/main/kotlin/misk/dynamodb/RealDynamoDbModule.kt
```diff
@@ -0,0 +1,46 @@
...
+import com.amazonaws.services.dynamodbv2.AmazonDynamoDB
+import com.amazonaws.services.dynamodbv2.AmazonDynamoDBClientBuilder
+import com.amazonaws.services.dynamodbv2.datamodeling.DynamoDBMapper
+import com.amazonaws.services.dynamodbv2.datamodeling.DynamoDBMapperConfig
+import com.google.inject.Provides
+import misk.cloud.aws.AwsRegion
+import misk.inject.KAbstractModule
+import javax.inject.Singleton
+
+internal class RealDynamoDbModule : KAbstractModule() {
```
mightyguava: why is this internal?

misk-aws/src/main/kotlin/misk/dynamodb/FakeDynamoDbModule.kt
```diff
@@ -0,0 +1,73 @@
...
+import com.amazonaws.services.dynamodbv2.document.DynamoDB
+import com.amazonaws.services.dynamodbv2.model.ProvisionedThroughput
+import com.amazonaws.services.dynamodbv2.model.ResourceInUseException
+import com.google.inject.Provides
+import misk.cloud.aws.FakeAwsEnvironmentModule
+import misk.inject.KAbstractModule
+import javax.inject.Singleton
+import kotlin.reflect.KClass
+
+internal class FakeDynamoDbModule(
```
mightyguava: why is this internal?

misk-aws/src/test/kotlin/misk/dynamodb/DockerDynamoDb.kt
```diff
@@ -0,0 +1,81 @@
...
+import com.github.dockerjava.api.model.Ports
+import misk.containers.Composer
+import misk.containers.Container
+import misk.logging.getLogger
+import misk.testing.ExternalDependency
+
+/**
+ * A test DynamoDb Service. Tests can connect to the service at 127.0.0.1:[clientPort]
+ */
+internal object DockerDynamoDb : ExternalDependency {
```
mightyguava: public?

misk-aws/src/main/kotlin/misk/dynamodb/FakeDynamoDbModule.kt
```diff
@@ -0,0 +1,73 @@
+package misk.dynamodb
```
mightyguava: We usually put fakes and testing-related code in a separate module so production code doesn't need to pull them in. See `misk-jobqueue-testing`, `misk-hibernate-testing`, etc

mpawliszyn: THIS!!!!!!!!!!!!!!!!!!!!!

misk-aws/src/test/kotlin/misk/dynamodb/DockerDynamoDb.kt
```diff
@@ -0,0 +1,81 @@
...
+
+  private val log = getLogger<DockerDynamoDb>()
+  private const val clientPort = 8000
```
mpawliszyn: Is that a good default?

mpawliszyn: I'm confused. Is this the internal docker port or the port we use to connect? I'm not sure it is a great default.

mightyguava: Ooohhh good point. It'd be nice if we could have Java auto-assign the port, but that might make it difficult to make this shared across tests. Either way, 8000 is probably a port you'll run into conflicts on :). I think this is currently both the external and internal port

misk-aws/src/main/kotlin/misk/dynamodb/DyEntity.kt
abakes: It doesn't need a actual name, but does need the table attribute for the member attributes to work

misk-aws/src/main/kotlin/misk/dynamodb/DyTimestampedEntity.kt
abakes: I do not. But I know there are use cases of non-timestamped entities in SQL, so I took the same approach here

misk-aws/src/main/kotlin/misk/dynamodb/RealDynamoDbModule.kt
abakes: Oops. It was internal when I was developing for my needs. Will fix

misk-aws/src/main/kotlin/misk/dynamodb/FakeDynamoDbModule.kt
abakes: Oops. It was internal when I was developing for my needs. Will fix

misk-aws/src/test/kotlin/misk/dynamodb/DockerDynamoDb.kt
abakes: 👍

misk-aws/src/main/kotlin/misk/dynamodb/RealDynamoDbModule.kt
abakes: What's the benefit of that? (newer to kotlin and the syntax is a tad confusing)

misk-aws/src/main/kotlin/misk/dynamodb/RealDynamoDbModule.kt
abakes: After some slack discussion, yes

misk-aws/src/test/kotlin/misk/dynamodb/DockerDynamoDb.kt
abakes: It is both internal and external. I'm actually having trouble remapping it :(

misk-aws/src/main/kotlin/misk/dynamodb/RealDynamoDbModule.kt
mightyguava: By inline I mean just removing the function and put the logic in `providesDynamoDBmapper`. Why was it a separate function to start with?

abakes: Currently it is also used by the Fake module (its in the same package currently )

misk-aws/src/main/kotlin/misk/dynamodb/RealDynamoDbModule.kt
abakes: I was thinking about whether this should be available from a config file, or just as a parameter for the Module. But then I wondered if the module should even be providing this at all..

If we ever wanted the possibility of having 2 tables with different save behaviors, this wouldn't work. Is it better to leave this configuration to the app @mightyguava ?

mightyguava: Yeah, it sounds like this doesn't belong in the module

misk-aws/src/main/kotlin/misk/dynamodb/FakeDynamoDbModule.kt
```diff
@@ -0,0 +1,73 @@
...
+internal class FakeDynamoDbModule(
+  private val entities: List<Pair<String, KClass<*>>>
+) : KAbstractModule() {
+  override fun configure() {
+    install(FakeAwsEnvironmentModule())
+  }
+
+  @Provides @Singleton
+  fun providesAmazonDynamoDB(): AmazonDynamoDB {
+    val dynamoDb = AmazonDynamoDBClientBuilder
```
zhxnlai: Would you consider returning a `DynamoDBEmbedded.create().amazonDynamoDB()` instead of the docker one?

zhxnlai: @mightyguava do you have a preference?

misk-aws/src/main/kotlin/misk/dynamodb/FakeDynamoDbModule.kt
```diff
@@ -0,0 +1,73 @@
...
+
+  companion object {
+    fun createTableForEntity(
+      amazonDynamoDB: AmazonDynamoDB,
+      entity: KClass<*>,
+      tableName: String
+    ) {
+      val tableRequest = DynamoDBMapper(amazonDynamoDB)
+          .generateCreateTableRequest(entity.java)
+          .withTableName(tableName)
```
zhxnlai: I believe DynamoDBMapper#describeTable gives you the table name

misk-aws/src/main/kotlin/misk/dynamodb/FakeDynamoDbModule.kt
```diff
@@ -0,0 +1,73 @@
...
+    return dynamoDb
+  }
+
+  @Provides @Singleton
+  fun providesDynamoDBmapper(dynamoDb: AmazonDynamoDB): DynamoDBMapper {
+    return RealDynamoDbModule.dynamoDbMapper(dynamoDb)
+  }
+
+  companion object {
+    fun createTableForEntity(
```
zhxnlai: This could be private?

misk-aws/src/main/kotlin/misk/dynamodb/FakeDynamoDbModule.kt
mightyguava: I don’t have a preference. But for sake of expediency I don’t think it should block this PR. Since you got embedded dynamo working why don’t we merge this pr and you can submit a follow up?

misk-aws/src/main/kotlin/misk/dynamodb/DyEntity.kt
```diff
@@ -0,0 +1,14 @@
...
+abstract class DyEntity {
```
zhxnlai: I would not use inheritance to model dynamo db tables because:
- it is difficult to gather a set of good defaults at this point
  - not every table wants a version field: declaring a version field prevents you from using condition expressions in write transactions. [reference](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/DynamoDBMapper.OptimisticLocking.html)
- it is not straightforward to override what the parent class customizes

misk-aws/src/main/kotlin/misk/dynamodb/DyEntity.kt
```diff
@@ -0,0 +1,14 @@
+package misk.dynamodb
+
+import com.amazonaws.services.dynamodbv2.datamodeling.DynamoDBTable
+import com.amazonaws.services.dynamodbv2.datamodeling.DynamoDBVersionAttribute
+
+/*
```
tirsen: I think you need another `*` here to make it a ktdoc?

misk-aws/src/main/kotlin/misk/dynamodb/DyEntity.kt
tirsen: Agreed. Prefer interfaces over abstract classes.

misk-aws/src/main/kotlin/misk/dynamodb/FakeDynamoDbModule.kt
tirsen: I prefer embedded over docker (faster, easier to debug, easier to support, etc) but I also agree that it shouldn't be blocking the PR.

misk-aws/src/main/kotlin/misk/dynamodb/FakeDynamoDbModule.kt
abakes: I cant seem to see this in intelliSense. And how do you descrbe a table which doesn't exist?

misk-aws/src/main/kotlin/misk/dynamodb/FakeDynamoDbModule.kt
abakes: deleted code

misk-aws/src/test/kotlin/misk/dynamodb/DockerDynamoDb.kt
tirsen: I think you can NOT have port 8000 :-) It's taken :-)

misk-aws/src/main/kotlin/misk/dynamodb/DyEntity.kt
abakes: I've deleted this entity, but think I should keep the `DyTimeStampedEntity`.

 At first there was more code associated with making them work. Now its more simple (2 annotations on each member). Unfortunately if its an interface, I cannot have the attributes on the members; basically defeating the purpose IMO (know how to use these attributes is the tricky part).

If we're against abstract classes, then I will also kill the `DyTimeStampedEntity`

misk-aws/src/test/kotlin/misk/dynamodb/DockerDynamoDb.kt
abakes: It worked for me 

This file will be removed

misk-aws/src/main/kotlin/misk/dynamodb/FakeDynamoDbModule.kt
zhxnlai: I was referring to [this method](https://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/com/amazonaws/services/dynamodbv2/datamodeling/DynamoDBTableMapper.html#describeTable--)

However I did not use it on my branch. I had the following working (with out specifying table names)
```kotlin
@Singleton
internal class CreateTablesService @Inject constructor(
  private val dynamoDb: AmazonDynamoDB,
  private val dynamoDbMapper: DynamoDBMapper,
  private val tables: Set<DynamoDbTable>
) : AbstractIdleService() {

  override fun startUp() {
    for (tableName in dynamoDb.listTables().tableNames) {
      dynamoDb.deleteTable(tableName)
    }

    val provisionedThroughput = ProvisionedThroughput(1000L, 1000L)
    for (table in tables) {
      try {
        val createTableRequest = dynamoDbMapper.generateCreateTableRequest(table.tableClass.java)
            .withProvisionedThroughput(provisionedThroughput)
        dynamoDb.createTable(createTableRequest)
      } catch (e: ResourceInUseException) {
        logger.error(e) { "Cannot create table." }
      }
    }
  }

  override fun shutDown() {
    dynamoDb.shutdown()
  }
}
/**
 * [DynamoDbTable] is a wrapper class that allows for unqualified binding of dynamo table classes
 * without collision.
 */
internal data class DynamoDbTable(val tableClass: KClass<*>)
```

misk-aws/src/main/kotlin/misk/dynamodb/DyEntity.kt
zhxnlai: Yeah I think we avoid using abstract classes when possible. 

BTW do we want to prefix the table class like we do with hibernate classes?

misk-aws-dynamodb-testing/src/main/kotlin/misk/aws/dynamodb/testing/DockerDynamoDbModule.kt
```diff
@@ -0,0 +1,68 @@
...
+import com.amazonaws.services.dynamodbv2.document.DynamoDB
+import com.amazonaws.services.dynamodbv2.model.ProvisionedThroughput
+import com.amazonaws.services.dynamodbv2.model.ResourceInUseException
+import com.google.inject.Provides
+import misk.cloud.aws.FakeAwsEnvironmentModule
+import misk.inject.KAbstractModule
+import javax.inject.Singleton
+import kotlin.reflect.KClass
+
+class DockerDynamoDbModule(
```
abakes: I renamed the FakeModule. I just wanted something to be in the testing module. We can delete this when the Embedded gets added

misk-aws/src/main/kotlin/misk/dynamodb/DyVersionedEntity.kt
```diff
@@ -0,0 +1,16 @@
+package misk.dynamodb
+
+/**
+ * This version field is used for optimistic locking.
+ */
+interface DyEntity {
```
abakes: These are now interfaces, but I left the needed annotations in comments to help people figure out how to use them.

abakes:
I think this should be good now

misk-aws-dynamodb-testing/src/main/kotlin/misk/aws/dynamodb/testing/DockerDynamoDbModule.kt
```diff
@@ -0,0 +1,68 @@
...
+  private val entities: List<Pair<String, KClass<*>>>
+) : KAbstractModule() {
+  override fun configure() {
+    install(FakeAwsEnvironmentModule())
```
mightyguava: This should get installed by the test, not here

misk-aws/src/main/kotlin/misk/dynamodb/DyTimestampedEntity.kt
```diff
@@ -0,0 +1,21 @@
+package misk.dynamodb
+
+import java.util.Date
+
+/**
+ * Interface for a timestamped entity.
+ */
+interface DyTimestampedEntity {
+
+  /**
```
mightyguava: docs

misk-aws/src/main/kotlin/misk/dynamodb/RealDynamoDbModule.kt
```diff
@@ -0,0 +1,30 @@
...
+import com.amazonaws.services.dynamodbv2.AmazonDynamoDB
+import com.amazonaws.services.dynamodbv2.AmazonDynamoDBClientBuilder
+import com.amazonaws.services.dynamodbv2.datamodeling.DynamoDBMapper
+import com.amazonaws.services.dynamodbv2.datamodeling.DynamoDBMapperConfig
+import com.google.inject.Provides
+import misk.cloud.aws.AwsRegion
+import misk.inject.KAbstractModule
+import javax.inject.Singleton
+
+class RealDynamoDbModule : KAbstractModule() {
```
mightyguava: docs

misk-aws-dynamodb-testing/src/main/kotlin/misk/aws/dynamodb/testing/DockerDynamoDbModule.kt
abakes: why is that? It seems odd to me to include a module for something, and it not work because it has another dependancy.

mightyguava: That is one exactly what you want with dependency injection. Tests should be isolated and provide its own dependencies. As applied to Guice, modules should only ever install other modules that are part of its own private API.

More practically, if something else installs `FakeAwsEnvironmentModule()`, Guice will throw an exception.

misk-aws-dynamodb-testing/build.gradle
```diff
@@ -0,0 +1,21 @@
+dependencies {
+  compile dep.awsApi
```
zhxnlai: `api dep.awsApi`?

misk-aws-dynamodb-testing/build.gradle
```diff
@@ -0,0 +1,21 @@
...
+  compile dep.awsSqs
```
zhxnlai: This is not needed right?

misk-aws-dynamodb-testing/build.gradle
```diff
@@ -0,0 +1,21 @@
...
+  compile dep.tracingJaeger
+  compile project(':misk')
```
zhxnlai: Could you replace `compile` with `implementation` and `testCompile` with `testImplementation`?

misk-aws-dynamodb-testing/build.gradle
abakes: I don't know what that does... This was a mostly copy paste from the `misk-aws` `build.gradle`

misk-aws-dynamodb-testing/build.gradle
abakes: Correct, copy/paste error

misk-aws-dynamodb-testing/build.gradle
abakes: Is that what we should use going forward? Again, this is what `misk-aws` does

zhxnlai: Yes: https://stackoverflow.com/questions/44493378/whats-the-difference-between-implementation-and-compile-in-gradle

misk-aws-dynamodb-testing/build.gradle
zhxnlai: > The plugin exposes two configurations that can be used to declare dependencies: api and implementation. The api configuration should be used to declare dependencies which are exported by the library API, whereas the implementation configuration should be used to declare dependencies which are internal to the component.

https://docs.gradle.org/current/userguide/java_library_plugin.html#sec:java_library_separation

misk-aws-dynamodb-testing/build.gradle
abakes: 👍

misk-aws-dynamodb-testing/build.gradle
zhxnlai: because this library exposes `AmazonDynamoDB`, it should have API dependency on dep.awsApi. Since this is copy paste let's leave it as is and change it later?

abakes: 👍

PR #1413, opened 2020-02-25
Handle explicit nulls in WireMessageAdapter

swankjesse:


misk/src/main/kotlin/misk/moshi/wire/WireMessageAdapter.kt
```diff
@@ -38,9 +38,12 @@ internal class WireMessageAdapter(
     reader.beginObject()
     while (reader.hasNext()) {
       val fieldName = reader.nextName()
-      fieldBindings[fieldName]?.let { binding ->
+      val binding = fieldBindings[fieldName]
+      if (binding != null) {
         binding.adapter.fromJson(reader)?.let { binding.set(builder, it) }
```
clare-lsy: I'm not very familiar with `let`.. If `binding.adapter.fromJson(reader)` returns `null` Do we want to do `reader.skipValue()` as well?

swankjesse: I think the else clause will catch it.

clare-lsy: oh.. I missed matched the brackets :) Looks good to me!

PR #1414, opened 2020-02-25
Replace DockerDynamoDbModule with EmbeddedDynamoDbModule

zhxnlai:
A follow up of https://github.com/cashapp/misk/pull/1398

Per our discussion:
> prefer embedded over docker (faster, easier to debug, easier to support, etc)


ryanhall07:
can we do this with docker instead?

https://hub.docker.com/r/amazon/dynamodb-local/

abakes:
A docker implementation is already in the repo

zhxnlai:
> can we do this with docker instead?
> 
> https://hub.docker.com/r/amazon/dynamodb-local/

The intention is to use embedded version of dynamodb-local instead of its docker image.

ryanhall07:
> The intention of this PR is to use embedded version of dynamodb-local instead of its docker image.

can you please explain why in the commit message? it seems strange to rely on something that is so system dependent and probably error prone for a future developer.

mightyguava:
+1 I thought embedded only required the jvm. If it needs sqlite we are probably better off just using the docker version.

abakes:
If we decide to not use this then, we should cleanup the DockerDynamo... 
1) clean up table creation to not require tablename
2) make the port mapping configureable (hardcoded to 8000 currently)

zhxnlai:
Here is what I found:
- `com.almworks.sqlite4java:sqlite4java` is a wrapper around SQLite
- `com.almworks.sqlite4java` provides 6 different native libraries containing SQLite for different platform. e.g. `*:libsqlite4java-linux-amd64`, `*:libsqlite4java-osx`
- `com.almworks.sqlite4java:sqlite4java` uses the `sqlite4java.library.path` system property to find the correct native libraries at runtime.
- DynamoDBLocal depends on `com.almworks.sqlite4java:sqlite4java`
- DynamoDBLocal also [depends on](https://s3-us-west-2.amazonaws.com/dynamodb-local/release/com/amazonaws/DynamoDBLocal/1.12.0/DynamoDBLocal-1.12.0.pom) all native libraries of `com.almworks.sqlite4java`
- At compile time, gradle downloads all native libraries and makes them available in `java.class.path`
- However, `sqlite4java.library.path` needs to be manually set. Hence the need for [`AwsDynamoDbLocalTestUtils`](https://github.com/redskap/aws-dynamodb-java-example-local-testing).

This is really fragile. I agree that the docker image is more preferable 👍

PR #1417, opened 2020-02-26
Add a proof of concept driver for sqldelight on top of hibernate

AlecStrong:
This is only meant to be a proof of concept, but should largely be usable, save for two things:

- `MiskDriver` is currently backed by a hibernate `SessionFactory`. My understanding is a proper implementation would have a pure-jdbc backing but didn't want to scope that here. Hibernate-fu ends up leaking all over the place in the test module because of this.

- All sql funnels through a single session created by the driver, which is probably not what we want - I need a bit more clarity on what ideal consumer APIs would look like, for now its deliberately straightforward and consumer doesn't need a reference to their own session.

- mysql dialect in sqldelight is still snapshot release, will need to publish if/before anything gets merged

Anyways the test passes so something works! The first three files in the diff are the consumer-facing experience with sqld


swankjesse:
Don’t take my lack of response as disinterest! I’m eager to look into this and to adopt it!

chris-ryan-square:

chris-ryan-square:

PR #1418, opened 2020-02-26
Add DockerDynamoDb

zhxnlai:
A follow up of #1398 and #1414

- Adds the DockerDynamoDb external dependency, hard coded to use port 58000
- Adds tests for DockerDynamoDbModule

### DIFF
--- a/misk-aws-dynamodb-testing/src/main/kotlin/misk/aws/dynamodb/testing/CreateTablesService.kt
+++ b/misk-aws-dynamodb-testing/src/main/kotlin/misk/aws/dynamodb/testing/CreateTablesService.kt
@@ -1,0 +1,50 @@
+package misk.aws.dynamodb.testing
+
+import com.amazonaws.services.dynamodbv2.AmazonDynamoDB
+import com.amazonaws.services.dynamodbv2.datamodeling.DynamoDBMapper
+import com.amazonaws.services.dynamodbv2.document.DynamoDB
+import com.amazonaws.services.dynamodbv2.model.ProvisionedThroughput
+import com.amazonaws.services.dynamodbv2.model.ResourceInUseException
+import com.google.common.util.concurrent.AbstractIdleService
+import misk.logging.getLogger
+import javax.inject.Inject
+import javax.inject.Singleton
+import kotlin.reflect.KClass
+
+@Singleton
+internal class CreateTablesService @Inject constructor(
+  private val dynamoDbClient: AmazonDynamoDB,
+  private val tables: Set<DynamoDbTable>
+) : AbstractIdleService() {
+
+  override fun startUp() {
+    // Cleans up the tables before each run.
+    for (tableName in dynamoDbClient.listTables().tableNames) {
+      dynamoDbClient.deleteTable(tableName)
+    }
+
+    for (table in tables) {
+      dynamoDbClient.createTable(table.tableClass)
+    }
+  }
+
+  override fun shutDown() {
+    dynamoDbClient.shutdown()
+  }
+
+  private fun AmazonDynamoDB.createTable(
+    table: KClass<*>
+  ) {
+    val tableRequest = DynamoDBMapper(this)
+        .generateCreateTableRequest(table.java)
+        // Provisioned throughput needs to be specified when creating the table. However,
+        // DynamoDB Local ignores your provisioned throughput settings. The values that you specify
+        // when you call CreateTable and UpdateTable have no effect. In addition, DynamoDB Local
+        // does not throttle read or write activity.
+        .withProvisionedThroughput(ProvisionedThroughput(1L, 1L))
+    DynamoDB(this).createTable(tableRequest).waitForActive()
+  }
+}
+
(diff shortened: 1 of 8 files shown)


misk-aws-dynamodb-testing/src/test/resources/moviestestmodule-common.yaml
```diff
@@ -0,0 +1,3 @@
+docker_dynamodb_config:
+  host: "localhost"
+  port: 8000
```
mightyguava: I don't think this needs to be configurable. Rather, it should automatically discover an unused port and use it. E.g. something like this https://gist.github.com/vorburger/3429822

ryanhall07: the main downside to a random port is debugging using the aws cli. although if you don't use a random port, you could easily create a flaky tests if multiple CI branches are running at the same host.

I think we should have a convention that a random port is used in CI and a static known (although uncommon to avoid local conflicts) for local development.

zhxnlai: If we assign a random unused port to this DynamoDB local container, do we risk occupying a port required by another test, such as 27000?

Launching a container takes a few seconds. It'd be great if we can reuse the same DynamoDB local container for multiple tests.

mightyguava: > If we assign a random unused port to this DynamoDB local container, do we risk occupying a port required by another test, such as 27000?

Possible, not likely. Feel free to choose a static port too, as long as it doesn't collide.

> Launching a container takes a few seconds. It'd be great if we can reuse the same DynamoDB local container for multiple tests.

Now that I look at this PR in detail, I think something got lost. `StartDatabaseService` is unnecessary. The Hibernate StartDatabaseService is a pattern I'd like to get rid of but just haven't gotten around to. The right way to instantiate a fake DynamoDB is to add this to your test. This already works

```
@MiskExternalDependency private val dockerDynamoDb = DockerDynamoDb
```

But this still has the issue that you can no longer share the same container across test runs because the port name is not hard-coded. There are ways to work around this but I think it's a bit complicated and we are better off just using static ports for now.

misk-aws-dynamodb-testing/src/main/kotlin/misk/aws/dynamodb/testing/CreateTablesService.kt
```diff
@@ -0,0 +1,54 @@
...
+
+  private fun AmazonDynamoDB.createTable(
+    table: KClass<*>
+  ) {
+    val tableRequest = DynamoDBMapper(this)
+        .generateCreateTableRequest(table.java)
+        // DynamoDB Local ignores your provisioned throughput settings. The values that you specify
+        // when you call CreateTable and UpdateTable have no effect. In addition, DynamoDB Local
+        // does not throttle read or write activity.
+        .withProvisionedThroughput(ProvisionedThroughput(1L, 1L))
```
mightyguava: why do you need `withProvisionedThroughput` at all?

misk-aws-dynamodb-testing/src/main/kotlin/misk/aws/dynamodb/testing/CreateTablesService.kt
```diff
@@ -0,0 +1,54 @@
...
+
+    try {
+      DynamoDB(this).createTable(tableRequest).waitForActive()
+    } catch (e: ResourceInUseException) {
+      logger.error(e) { "Cannot create table." }
```
mightyguava: why does this swallow the error instead of re-throwing?

misk-aws-dynamodb-testing/src/test/kotlin/misk/aws/dynamodb/testing/MoviesTestModule.kt
```diff
@@ -0,0 +1,19 @@
...
+import misk.MiskTestingServiceModule
+import misk.environment.Environment
+import misk.environment.EnvironmentModule
+import misk.inject.KAbstractModule
+import misk.logging.LogCollectorModule
+
+class MoviesTestModule: KAbstractModule() {
+
+  override fun configure() {
+    install(LogCollectorModule())
```
mightyguava: I don't think you need this module?

misk-aws-dynamodb-testing/src/main/kotlin/misk/aws/dynamodb/testing/CreateTablesService.kt
zhxnlai: I believe the create table request operation it

misk-aws-dynamodb-testing/src/test/kotlin/misk/aws/dynamodb/testing/MoviesTestModule.kt
zhxnlai: removed

PR #1420, opened 2020-02-27
Introduce Deployment type to eventually replace Environment enum

alecholmes:
Not meant to be fully complete, but shows the Misk changes for starting to remove the `Environment` enum in favor of a `Deployment` type.

The environment (in plain string form) is still used for things like parsing config, at least for now.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -52,13 +52,17 @@ subprojects { subProject ->
   compileKotlin {
     kotlinOptions {
       jvmTarget = "1.8"
-      allWarningsAsErrors = true
+
+      // TODO(alec): Enable again once Environment enum is deleted
+       allWarningsAsErrors = false
     }
   }
   compileTestKotlin {
     kotlinOptions {
       jvmTarget = "1.8"
-      allWarningsAsErrors = true
+
+      // TODO(alec): Enable again once Environment enum is deleted
+       allWarningsAsErrors = false
     }
   }
   sourceSets {
--- a/misk/src/main/kotlin/misk/config/MiskConfig.kt
+++ b/misk/src/main/kotlin/misk/config/MiskConfig.kt
@@ -24,12 +24,13 @@ import org.apache.commons.lang3.StringUtils
 import java.io.File
 import java.io.FilenameFilter
 import java.net.URL
+import java.util.Locale
 
 object MiskConfig {
   @JvmStatic
   inline fun <reified T : Config> load(
     appName: String,
-    environment: Environment,
+    environment: String,
     overrideFiles: List<File> = listOf(),
     resourceLoader: ResourceLoader = ResourceLoader.SYSTEM
   ): T {
@@ -37,10 +38,21 @@ object MiskConfig {
   }
 
   @JvmStatic
+  @Deprecated("the Environment enum is deprecated")
+  inline fun <reified T : Config> load(
+    appName: String,
+    environment: Environment,
+    overrideFiles: List<File> = listOf(),
+    resourceLoader: ResourceLoader = ResourceLoader.SYSTEM
+  ): T {
+    return load(T::class.java, appName, environment.name, overrideFiles, resourceLoader)
+  }
+
+  @JvmStatic
   fun <T : Config> load(
     configClass: Class<out Config>,
     appName: String,
-    environment: Environment,
+    environment: String,
     overrideFiles: List<File> = listOf(),
     resourceLoader: ResourceLoader = ResourceLoader.SYSTEM
   ): T {
@@ -57,7 +69,7 @@ object MiskConfig {
 
     val jsonNode = flattenYamlMap(configYamls)
 
-    val configFile = "$appName-${environment.name.toLowerCase()}.yaml"
+    val configFile = "$appName-${environment.toLowerCase(Locale.US)}.yaml"
     try {
       @Suppress("UNCHECKED_CAST")
       return mapper.readValue(jsonNode.toString(), configClass) as T
@@ -134,7 +146,7 @@ object MiskConfig {
    */
   fun loadConfigYamlMap(
     appName: String,
-    environment: Environment,
+    environment: String,
     overrideFiles: List<File>
   ): Map<String, String?> {
     // Load from jar files first, starting with the common config and then env specific config
@@ -157,8 +169,8 @@ object MiskConfig {
   }
 
   /** @return the list of config file names in the order they should be read */
-  private fun embeddedConfigFileNames(appName: String, environment: Environment) =
-      listOf("common", environment.name.toLowerCase()).map { "$appName-$it.yaml" }
+  private fun embeddedConfigFileNames(appName: String, environment: String) =
+      listOf("common", environment.toLowerCase(Locale.US)).map { "$appName-$it.yaml" }
 
   class SecretJacksonModule(val resourceLoader: ResourceLoader, val mapper: ObjectMapper) :
       SimpleModule() {
--- a/misk/src/main/kotlin/misk/environment/Deployment.kt
+++ b/misk/src/main/kotlin/misk/environment/Deployment.kt
@@ -1,0 +1,49 @@
+package misk.environment
+
+/**
+ * The raw environment name. The set of possibilities is unbounded so this should rarely be used.
+ * Instead, use the Deployment type for behavior based on characteristics of an environment.
+ */
+data class Env(val name: String)
+
+/** Deployment describes the context in which the application is running */
+data class Deployment(
+  /**
+   * The name of this deployment. This is used for debugging and should not be parsed.
+   *
+   * All pods in the same deployment will have this same name.
+   */
+  val name: String,
+
+  /**
+   * Whether the service is running in a production environment, having an SLA or handling customer data.
+   */
+  val isProduction: Boolean = false,
+
+  /**
+   * Whether the service is running in a test environment, either locally or in a CI.
+   */
+  val isTest: Boolean = false,
+
+  /**
+   * Whether the service is running on a local developer machine, including as a Docker image.
+   */
+  val isLocalDevelopment: Boolean = false
+) {
+  init {
+    if (isProduction) check(!isTest && !isLocalDevelopment)
+    if (isTest) check(!isLocalDevelopment)
+  }
+
+  /**
+   * Returns true if running in a managed cluster, such as a staging or production cluster. Mutually exclusive with isFake.
+   */
+  val isReal: Boolean
+    get() = !isFake
+
+  /**
+   * Returns true if running outside of a cluster (CI or local development). Mutually exclusive with isReal.
+   */
+  val isFake: Boolean
+    get() = isTest || isLocalDevelopment
+}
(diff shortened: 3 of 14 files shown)


misk/src/main/kotlin/misk/environment/Deployment.kt
```diff
@@ -0,0 +1,39 @@
+package misk.environment
+
+/** Deployment describes the context in which the application is running */
+data class Deployment(
```
alecholmes: Also thinking about

```
// Raw environment name
data class Env(val name: String)
data class Deployment(
  val name: String,

  val environment: Env, // Include raw env in Deployment

  // ...
}
```

misk/src/main/kotlin/misk/config/MiskConfig.kt
```diff
@@ -57,7 +68,7 @@ object MiskConfig {
 
     val jsonNode = flattenYamlMap(configYamls)
 
-    val configFile = "$appName-${environment.name.toLowerCase()}.yaml"
+    val configFile = "$appName-${environment.toLowerCase()}.yaml"
```
swankjesse: not a new problem, but `toLowerCase()` doesn’t work here or anywhere; you need `toLowerCase(Locale.US)`

https://en.wikipedia.org/wiki/Dotted_and_dotless_I

misk/src/main/kotlin/misk/config/MiskConfig.kt
```diff
@@ -157,8 +168,8 @@ object MiskConfig {
   }
 
   /** @return the list of config file names in the order they should be read */
-  private fun embeddedConfigFileNames(appName: String, environment: Environment) =
-      listOf("common", environment.name.toLowerCase()).map { "$appName-$it.yaml" }
+  private fun embeddedConfigFileNames(appName: String, environment: String) =
+      listOf("common", environment.toLowerCase()).map { "$appName-$it.yaml" }
```
swankjesse: same

misk/src/main/kotlin/misk/environment/Deployment.kt
swankjesse: What would it be used for? And can we create something more formal for that?

misk/src/main/kotlin/misk/environment/Deployment.kt
```diff
@@ -0,0 +1,39 @@
...
+   * Whether the service is running on a local developer machine, including as a Docker image.
+   */
+  val isLocalDevelopment: Boolean = false
+) {
+  init {
+    if (isProduction) check(!isTest && !isLocalDevelopment)
+    if (isTest) check(!isLocalDevelopment)
+  }
+
+  /** Returns true if running in a managed cluster. Mutually exclusive with isFake. */
```
swankjesse: Maybe mention staging environments?

Maybe mention that 'real' means that the code that’s running in process, and the infrastructure it’s sitting on, is also real? Real means the same components as production, whether or not there’s the same data or SLA.

misk/src/main/kotlin/misk/environment/Environment.kt
```diff
@@ -4,39 +4,55 @@ import com.google.common.base.Preconditions
...
+  fun isFake(): Boolean = this == TESTING || this == DEVELOPMENT
+
+  fun isReal(): Boolean = !isFake()
+
   companion object {
     internal val logger = getLogger<Environment>()
     private const val ENV_ENVIRONMENT = "ENVIRONMENT"
-    private lateinit var env: Environment
+
+    private lateinit var rawEnv: String
```
swankjesse: Rename to just `string` ?

misk/src/main/kotlin/misk/environment/Environment.kt
```diff
@@ -4,39 +4,55 @@ import com.google.common.base.Preconditions
...
+      if (::rawEnv.isInitialized) {
+        Preconditions.checkState(rawEnv == TESTING.name, "Environment already set to $rawEnv")
       }
-      env = TESTING
+      rawEnv = TESTING.name
     }
 
     @JvmStatic
     fun fromEnvironmentVariable(): Environment {
+      return valueOf(when (val rawEnv = rawEnvironment()) {
```
swankjesse: this line has a lot of density to it. Maybe separate string mapping from valueOf ?

misk/src/main/kotlin/misk/web/dashboard/WebTabResourceModule.kt
```diff
@@ -26,24 +27,44 @@ import misk.web.resources.StaticResourceEntry
  * @property resourcePath JVM path for non-Development environment static resources (includes `classpath:/` prefix)
  */
 class WebTabResourceModule(
-  val environment: Environment,
+  private val isDevelopment: Boolean,
```
swankjesse: Pass in the Deployment instead?

misk/src/main/kotlin/misk/web/dashboard/WebTabResourceModule.kt
```diff
@@ -26,24 +27,44 @@ import misk.web.resources.StaticResourceEntry
...
   val slug: String,
   val web_proxy_url: String,
   val url_path_prefix: String = "/_tab/$slug/",
   val resourcePath: String = "classpath:/web/_tab/$slug/"
 ) : KAbstractModule() {
+
+  constructor(
```
adrw: Thanks for these changes and maintaining the existing constructor with deprecation

misk/src/main/kotlin/misk/environment/Deployment.kt
alecholmes: Uses are atypical. Some concrete examples in our internal code:
- map of env to top-level domains. The is where the low-level env like "platform-staging" vs "staging" is relevant.
- in platform-staging environment, fake out some dependencies that aren't available as they are in staging or prod.

WDYT about keeing the `Env` data class above but not including it in the `Deployment`. It can be passed around separately or bound using Guice as needed.

alecholmes:
Misk fails builds on compile warnings, including deprecation. I'm planning on disabling this during the rollout.

misk/src/main/kotlin/misk/environment/Environment.kt
alecholmes: I don't follow. Rename the type or variable name?

misk/src/main/kotlin/misk/web/dashboard/WebTabResourceModule.kt
alecholmes: Keeping for backwards compatibility until `Environment` is deleted. After that, will replace this and have a single ctor (the one currently on line 37).

build.gradle
```diff
@@ -52,13 +52,17 @@ subprojects { subProject ->
   compileKotlin {
     kotlinOptions {
       jvmTarget = "1.8"
-      allWarningsAsErrors = true
+
+      // TODO(alec): Enable again once Environment enum is deleted
+      // allWarningsAsErrors = true
```
alecholmes: Changed since initial PR

misk/src/main/kotlin/misk/environment/DeploymentModule.kt
```diff
@@ -0,0 +1,14 @@
+package misk.environment
+
+import misk.inject.KAbstractModule
+
+/** Binds [Deployment] and [Env] to make them available to services and actions */
+class DeploymentModule(
```
alecholmes: After landing this PR, my plan is to:
1. Update apps to instead this in addition to `EnvironmentModule`, so that all are available during transition period.
2. Update misk and skim components to rely on `Deployment` and raw env instead of `Environment`
3. Remove uses of `EnvironmentModule` and `Environment` in misk, skim, and app code.
4. Delete `EnvironmentModule` and `Environment`.

adrw: Plan sounds thorough and backwards compatible during the migration period 👍

misk/src/main/kotlin/misk/environment/Deployment.kt
swankjesse: This feels very skimmy to me

swankjesse: Or perhaps, things that should be in config files.

misk/src/main/kotlin/misk/environment/Environment.kt
swankjesse: I was thinking the variable name. I don’t believe rawEnv should be a thing long term.

alecholmes: It's not a good name, but fortunately this entire class (and its `rawEnv`) will be going away soon.

misk/src/main/kotlin/misk/config/MiskConfig.kt
```diff
@@ -134,7 +146,7 @@ object MiskConfig {
    */
   fun loadConfigYamlMap(
     appName: String,
-    environment: Environment,
+    environment: String,
```
ryanhall07: why doesn't this use the `Env` data type?

misk/src/main/kotlin/misk/environment/Deployment.kt
ryanhall07: yea I think everything `Env` specific should be in a config. To me the only drawback to a config is yaml kind of sucks, especially for something that is truly global and static (e.g top-level domains). It would be nice if components could register static configs in code per `Env` and they would get magically merged with the yaml configs per `Env`. App code would just see a single unified `Config`.

Until we build something like that (or never cause it's too magical), we should put all `Env` differences in a yaml.

ryanhall07: thinking about it more, we already kind of have yaml + code defaults being merged. we should definitely consider expanding that to allow `Env` defaults in code.

misk/src/main/kotlin/misk/environment/Deployment.kt
```diff
@@ -0,0 +1,49 @@
...
+
+  /**
+   * Whether the service is running in a test environment, either locally or in a CI.
+   */
+  val isTest: Boolean = false,
+
+  /**
+   * Whether the service is running on a local developer machine, including as a Docker image.
+   */
+  val isLocalDevelopment: Boolean = false
```
ryanhall07: what's the reason for `isLocalDev` vs `isTest`

misk/src/main/kotlin/misk/config/MiskConfig.kt
alecholmes: It should. I'll fix.

misk/src/main/kotlin/misk/environment/Deployment.kt
alecholmes: The wiring for running locally seems to differ in enough cases.

adrw: For one example, Admin Dashboard should be run in Development but not Test environments.

ryanhall07: why? don't you want to test the admin dashboard works in tests?

adrw: Fair enough :+1: My initial thoughts were prompted by the current state of not having many tests for admin dashboard

PR #1422, opened 2020-02-27
jdbc: support sslMode

mightyguava:
`useSSL`, `requireSSL`, and `verifyServerCertificate` are all deprecated
and superceded by `sslMode` as of `mysql:mysql-connector-java:8.0.13`.

This PR removes the deprecated options and uses `sslMode`, and also
enables the use of the `VERIFY_IDENTITY` option to verify the server's
hostname against its certificate.

Apparently validating the hostname has never been supported in the
jdbc connector before.

https://dev.mysql.com/doc/connector-j/8.0/en/connector-j-reference-configuration-properties.html

### DIFF
--- a/misk-hibernate/src/main/kotlin/misk/jdbc/DataSourceConfig.kt
+++ b/misk-hibernate/src/main/kotlin/misk/jdbc/DataSourceConfig.kt
@@ -70,6 +70,7 @@ data class DataSourceConfig(
   // going forward
   val trust_certificate_key_store_path: String? = null,
   val client_certificate_key_store_path: String? = null,
+  val verify_server_identity: Boolean = false,
   val show_sql: String? = "false",
   // Consider using this if you want Hibernate to automagically batch inserts/updates when it can.
   val jdbc_statement_batch_size: Int? = null
@@ -181,7 +182,6 @@ data class DataSourceConfig(
           }
           queryParams += "&trustCertificateKeyStoreUrl=$trustStoreUrl"
           queryParams += "&trustCertificateKeyStorePassword=${config.trust_certificate_key_store_password}"
-          queryParams += "&verifyServerCertificate=true"
           useSSL = true
         }
         if (!certStoreUrl.isNullOrBlank()) {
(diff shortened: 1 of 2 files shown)


mightyguava:
@tirsen this should all just work with vitess right? Since everything is client-side?

misk-hibernate/src/test/kotlin/misk/jdbc/DataSourceConfigTest.kt
```diff
@@ -88,15 +88,33 @@ class DataSourceConfigTest {
...
+        "sslMode=VERIFY_IDENTITY",
+        config.buildJdbcUrl(Environment.TESTING))
   }
 
   @Test
   fun buildMysqlJDBCUrlWithNoTls() {
     val config = DataSourceConfig(DataSourceType.MYSQL)
     assertEquals("jdbc:tracing:mysql://127.0.0.1:3306/?useLegacyDatetimeCode=false&" +
-        "createDatabaseIfNotExist=true&connectTimeout=10000&socketTimeout=60000",
+        "createDatabaseIfNotExist=true&connectTimeout=10000&socketTimeout=60000&sslMode=PREFERRED",
```
mightyguava: preferred means use TLS if available, and is the default. Could probably drop this but no harm in being explicit?

misk-hibernate/src/main/kotlin/misk/jdbc/DataSourceConfig.kt
```diff
@@ -194,10 +194,14 @@ data class DataSourceConfig(
...
 
-        if (useSSL) {
-          queryParams += "&useSSL=true"
-          queryParams += "&requireSSL=true"
+        val sslMode = if (useSSL && verify_server_identity) {
+          "VERIFY_IDENTITY"
+        } else if (useSSL) {
+          "VERIFY_CA"
+        } else {
+          "PREFERRED"
```
kalfonso: Don't we require all our servers to have SSL enabled? PREFERRED will default to unencrypted connection if server hasn't enabled it

mightyguava: We don’t enable SSL in local dev, so it’s a matter of which set of config files we add overrides to. Setting to PREFERRED as default preserves current behavior.

PR #1423, opened 2020-02-28
DeploymentModule.forTesting() to replace uses of EnvironmentModule

alecholmes:
Getting rid of `Environment` means also phasing out `EnvironmentModule`. It's common for tests to bind up `EnvironmentModule(Environment.TESTING)` since many Misk components depend on it.

As part of the move to `Deployment`, this introduces `DeploymentModule.forTesting()` to replace unit test cases.

### DIFF
--- a/misk/src/main/kotlin/misk/environment/DeploymentModule.kt
+++ b/misk/src/main/kotlin/misk/environment/DeploymentModule.kt
@@ -1,5 +1,7 @@
 package misk.environment
 
+import com.google.inject.util.Modules
+import com.google.inject.Module
 import misk.inject.KAbstractModule
 
 /** Binds [Deployment] and [Env] to make them available to services and actions */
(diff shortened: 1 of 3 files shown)


misk/src/main/kotlin/misk/environment/DeploymentModule.kt
```diff
@@ -11,4 +13,20 @@ class DeploymentModule(
...
+     * Return a Module that binds a [Deployment] and [Env] for a test environment.
+     *
+     * Until [Environment] is deleted, this also installs an [EnvironmentModule]
+     */
+    fun forTesting(): Module {
+      return Modules.combine(
+          DeploymentModule(
+              deployment = Deployment(name = "testing", isTest = true),
+              env = Env("TESTING")),
+          EnvironmentModule(Environment.TESTING)
```
alecholmes: This will go away at the end of the migration, once Misk components no longer depend on `Environment`

ryanhall07: i'd leave this comment as a TODO in the code

misk/src/main/kotlin/misk/environment/DeploymentModule.kt
```diff
@@ -11,4 +13,20 @@ class DeploymentModule(
...
+  companion object {
+    /**
+     * Return a Module that binds a [Deployment] and [Env] for a test environment.
+     *
+     * Until [Environment] is deleted, this also installs an [EnvironmentModule]
+     */
+    fun forTesting(): Module {
+      return Modules.combine(
+          DeploymentModule(
+              deployment = Deployment(name = "testing", isTest = true),
```
ryanhall07: might need another attribute, like `isReallyTest = true` :trollface:

misk/src/main/kotlin/misk/environment/DeploymentModule.kt
alecholmes: The final step is deleting the `Environment` enum altogether. I figured the compiler would remind me of that TODO :)

PR #1424, opened 2020-02-28
Use Env in config parsing instead of string

alecholmes:


misk/src/main/kotlin/misk/web/metadata/ConfigMetadataAction.kt
```diff
@@ -55,7 +56,7 @@ class ConfigMetadataAction @Inject constructor(
       environment: Environment,
       config: Config
     ): Map<String, String?> {
-      val rawYamlFiles = MiskConfig.loadConfigYamlMap(appName, environment.name, listOf())
+      val rawYamlFiles = MiskConfig.loadConfigYamlMap(appName, Env(environment.name), listOf())
```
alecholmes: This type of awkwardness will go away since all the functions taking `Environment` will be replaced soon with `Env`-only functions.

samples/exemplarchat/src/main/kotlin/com/squareup/chat/ChatService.kt
```diff
@@ -12,15 +15,18 @@ import misk.web.MiskWebModule
...
-  val config = MiskConfig.load<ChatConfig>("chat", environment.name)
+  val env = Env(environment.name)
+  val deployment = Deployment(name = "exemplarchat", isLocalDevelopment = true)
+
+  val config = MiskConfig.load<ChatConfig>("chat", env)
 
   MiskApplication(
       MiskRealServiceModule(),
       MiskWebModule(config.web),
       RealEventRouterModule(environment),
```
alecholmes: Next phase is updating all Misk modules to use `Deployment` or `Env`. Once this is updated, won't need both `env` and `environment` variables above.

PR #1427, opened 2020-03-03
Update WideOpenDevelopmentInterceptor to use Deployment

adrw:


adrw:
Turns out my service's TestingModule was causing this since I had inadvertently deleted EnvironmentModule which this interceptor requires. User error!

PR #1429, opened 2020-03-05
Don't constraint max connection size for SQS receiving clients

wesleyk:


misk-aws/src/main/kotlin/misk/jobqueue/sqs/QueueResolver.kt
```diff
@@ -17,25 +17,35 @@ internal class QueueResolver @Inject internal constructor(
...
   private val crossRegionSQS: Map<AwsRegion, AmazonSQS>,
+  @ForSqsReceiving private val defaultForReceivingSQS: AmazonSQS,
+  @ForSqsReceiving private val crossRegionForReceivingSQS: Map<AwsRegion, AmazonSQS>,
   private val externalQueues: Map<QueueName, AwsSqsQueueConfig>
 ) {
   private val mapping = ConcurrentHashMap<QueueName, ResolvedQueue>()
 
   operator fun get(q: QueueName): ResolvedQueue {
-    return mapping.computeIfAbsent(q) { resolve(it) }
+    return mapping.computeIfAbsent(q) { resolve(it, false) }
```
zcross: Do we want separate `mapping` instances for the two variations (for receiving or for sending)

zcross: (I'm asking because the "for sending" input is only used in the resolution within `computeIfAbsent`, and not as an additional input to the map key during lookups by queue name only)

wesleyk: oh yeah this is totally busted, good point

zcross: cool, glad I didn't totally miss something obvious

wesleyk: updated, how about that?

misk-aws/src/main/kotlin/misk/jobqueue/sqs/AwsSqsJobQueueModule.kt
```diff
@@ -90,12 +101,35 @@ class AwsSqsJobQueueModule(
...
+      val clientConfiguration = ClientConfiguration()
+          .withSocketTimeout(25000)
+          .withConnectionTimeout(1000)
+
+      val builder = AmazonSQSClientBuilder.standard()
+          .withCredentials(credentials)
+          .withRegion(region.name)
+      if (forSqsReceiving) {
+        // Do not artificially constrain the # of connections to SQS. Instead we rely on higher
+        // level resource limiting knobs (e.g # of parallel receivers and # of jobs/sec).
```
ryanhall07: nit: rm # of jobs/sec since you aren't using max connections for sending

PR #1431, opened 2020-03-06
Add idempotence key and jobqueue metadata

biasedbit:
Add idempotence key and jobqueue internal metadata attributes

This change brings internal and misk jobqueue libraries closer in
feature-parity by offering the option to specify an application assigned
`idempotence_key` (as opposed to the system assigned `id`, only available after
a job is enqueued), which can be used by `JobHandler` implementations to detect
duplicate jobs.

## Jobqueue metadata

Internal attributes required for jobqueue-related functionality are stored in a
single message attribute (JSON-encoded `Map<String, String>`), leaving a
maximum of 8 arbitrary attributes for applications (the other reserved
attribute is for trace id; while not always present, it makes sense to assume
it'll exist and just consistently limit app keys to 8).

This change does not introduce dependency on metadata or any of its attributes,
meaning new jobs will begin having these properties as services are deployed.

The two metadata attributes added with this change are:

* Idempotence key: our internal jobqueue libs have always provided idempotence
key support, which some services heavily rely on to avoid duplicate work.

* Origin queue: when a job is transferred from main or retry queue to
dead-letter queue, it does not contain any metadata about its origins.
As we prepare to roll out a(n optional) more functional solution for
dead-lettered job management, which relies on a single, global dead-letter
queue, it's necessary to record the main queue where a job came from (so that
it's known where it should be sent to, in case of reprocessing).

Support for global DLQ (vs per-queue DLQ) will be introduced in a follow-up PR.

### DIFF
--- a/misk-aws/src/main/kotlin/misk/jobqueue/sqs/SqsJob.kt
+++ b/misk-aws/src/main/kotlin/misk/jobqueue/sqs/SqsJob.kt
@@ -2,23 +2,47 @@ package misk.jobqueue.sqs
 
 import com.amazonaws.services.sqs.model.Message
 import com.amazonaws.services.sqs.model.SendMessageRequest
+import com.squareup.moshi.Moshi
 import misk.jobqueue.Job
 import misk.jobqueue.QueueName
+import misk.moshi.adapter
 import misk.time.timed
 
 internal class SqsJob(
   override val queueName: QueueName,
   private val queues: QueueResolver,
   private val metrics: SqsMetrics,
+  private val moshi: Moshi,
   private val message: Message
 ) : Job {
   override val body: String = message.body
   override val id: String = message.messageId
-  override val attributes: Map<String, String> = message.messageAttributes.map { (key, value) ->
-    key to value.stringValue
-  }.toMap()
+  override val idempotenceKey: String by lazy {
+    jobqueueMetadata[JOBQUEUE_METADATA_IDEMPOTENCE_KEY]
+        ?: error("$JOBQUEUE_METADATA_IDEMPOTENCE_KEY not set in $JOBQUEUE_METADATA_ATTR")
+  }
+  override val attributes: Map<String, String> by lazy {
+    message.messageAttributes
+        .filter { (key, _) -> key != JOBQUEUE_METADATA_ATTR }
+        .map { (key, value) -> key to value.stringValue }.toMap()
+  }
 
   private val queue: ResolvedQueue = queues.getForReceiving(queueName)
+  private val jobqueueMetadata: Map<String, String> by lazy {
+    val metadata = message.messageAttributes[JOBQUEUE_METADATA_ATTR]
+    // NB: old/in-flight jobs enqueued prior to the introduction of _jobqueue-metadata won't have this property. In
+    // order to not change the contract later on (optional to non-optional), synthesize a reasonable replacement.
+    if (metadata == null) {
+      // TODO(bruno): drop fallback; error out if metadata is not set.
+      mapOf(
+          JOBQUEUE_METADATA_ORIGIN_QUEUE to queueName.parentQueue.value,
+          // If the job had no metadata, there was no app-specified idempotence key;
+          // any sufficiently random value would work, so just use system-assigned id.
+          JOBQUEUE_METADATA_IDEMPOTENCE_KEY to id)
+    } else {
+      moshi.adapter<Map<String, String>>().fromJson(metadata.stringValue)!!
+    }
+  }
 
   override fun acknowledge() {
     deleteMessage(queue, message)
@@ -33,6 +57,7 @@ internal class SqsJob(
       client.sendMessage(SendMessageRequest()
           .withQueueUrl(dlq.url)
           .withMessageBody(body)
+          // Preserves original jobqueue metadata and trace id.
           .withMessageAttributes(message.messageAttributes))
     }
     deleteMessage(queue, message)
@@ -47,6 +72,16 @@ internal class SqsJob(
   }
 
   companion object {
+    /** Message attribute that captures original trace id for a job, when available. */
     const val ORIGINAL_TRACE_ID_ATTR = "x-original-trace-id"
+    /** Message attribute used to store metadata specific to jobqueue functionality. JSON-encoded. */
+    const val JOBQUEUE_METADATA_ATTR = "_jobqueue-metadata"
+    /**
+     * The name of the queue the job was originally submitted to. Used when operating with a global dead-letter queue,
+     * so that jobs can be returned to their original (or retry) queue when reprocessing.
+     */
+    const val JOBQUEUE_METADATA_ORIGIN_QUEUE = "origin_queue"
+    /** Client-assigned identifier, useful to detect duplicate messages. */
+    const val JOBQUEUE_METADATA_IDEMPOTENCE_KEY = "idempotence_key"
   }
 }
--- a/misk-aws/src/main/kotlin/misk/jobqueue/sqs/SqsJob.kt
+++ b/misk-aws/src/main/kotlin/misk/jobqueue/sqs/SqsJob.kt
@@ -33,7 +33,7 @@ internal class SqsJob(
     // NB: old/in-flight jobs enqueued prior to the introduction of _jobqueue-metadata won't have this property. In
     // order to not change the contract later on (optional to non-optional), synthesize a reasonable replacement.
     if (metadata == null) {
-      // TODO(bruno): drop fallback; error out if metadata is not set.
+      // TODO(bruno): drop fallback after rollout; error when metadata is not set.
       mapOf(
           JOBQUEUE_METADATA_ORIGIN_QUEUE to queueName.parentQueue.value,
           // If the job had no metadata, there was no app-specified idempotence key;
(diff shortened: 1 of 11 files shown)


misk-aws/src/main/kotlin/misk/jobqueue/sqs/SqsJobQueue.kt
```diff
@@ -17,14 +19,21 @@ import javax.inject.Singleton
...
     queueName: QueueName,
     body: String,
+    idempotenceKey: String,
     deliveryDelay: Duration?,
     attributes: Map<String, String>
   ) {
+    // Ensure there are at most 8 attributes; AWS SQS enforces a limit of 10 custom attributes
+    // per message, 2 of which are reserved for this library (trace id and jobqueue metadata).
+    // https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-message-attributes.html
+    check(attributes.size <= 8) { "a maximum of 8 attributes are supported (got ${attributes.size})" }
```
ryanhall07: the trace id should just go in the metadata

misk-jobqueue/src/main/kotlin/misk/jobqueue/JobQueue.kt
```diff
@@ -12,17 +13,23 @@ interface JobQueue {
...
    * @param deliveryDelay If specified, the job will only become visible to the consumer after
    * the provided duration. Used for jobs that should delay processing for a period of time.
-   * @param attributes Arbitrary contextual attributes associated with the job
+   * @param attributes Arbitrary contextual attributes associated with the job. Implementations may limit the number of
+   * attributes per message.
    */
   fun enqueue(
     queueName: QueueName,
     body: String,
+    idempotenceKey: String = UUID.randomUUID().toString(),
```
ryanhall07: this is going to break existing clients because it's not binary compatible. you need to create another overloaded method or can possibly use `@JvmOverloads`

misk-aws/src/main/kotlin/misk/jobqueue/sqs/SqsJobQueue.kt
biasedbit: Initially did that but didn't want to break backwards compatibility or make it harder to inspect jobs; then again, this PR is the perfect time for this change.

Will change.

ryanhall07: well I think you want both for the rollout. keep it a separate attribute _and_ move it into the metadata

misk-jobqueue/src/main/kotlin/misk/jobqueue/JobQueue.kt
biasedbit: I wasn't familiar with `@JvmOverloads`  before but sounds like the only consequence is Java clients would have to explicitly set the idempotence key; from [the docs](https://kotlinlang.org/docs/reference/java-to-kotlin-interop.html#overloads-generation):

> a Kotlin function with default parameter values, it will be visible in Java only as a full signature, with all parameters present.

Since we can't use that annotation in interfaces, I'm okay with forcing Java apps to specify this argument (they already have to for the remaining args). If consistency is a concern, we can drop the default value and just force everyone to pass an explicit value (using docs to recommend a random UUID when no better candidate exists). Thoughts?

misk-aws/src/main/kotlin/misk/jobqueue/sqs/SqsJobQueue.kt
biasedbit: Updated.

misk-jobqueue/src/main/kotlin/misk/jobqueue/JobQueue.kt
ryanhall07: the problem is existing kotlin clients will not have a JVM bytecode method of the signature `enqueue(QueueName, String, Duration?, Map)` anymore and fail to compile when the misk library is updated. if you add `@JvmOverloads`, it will generate the old JVM bytecode to be backwards compatbile.

misk-aws/src/main/kotlin/misk/jobqueue/sqs/SqsJob.kt
```diff
@@ -2,23 +2,47 @@ package misk.jobqueue.sqs
...
+    // order to not change the contract later on (optional to non-optional), synthesize a reasonable replacement.
+    if (metadata == null) {
+      // TODO(bruno): drop fallback after rollout; error when metadata is not set.
+      mapOf(
+          JOBQUEUE_METADATA_ORIGIN_QUEUE to queueName.parentQueue.value,
+          // If the job had no metadata, there was no app-specified idempotence key;
+          // any sufficiently random value would work, so just use system-assigned id.
+          JOBQUEUE_METADATA_IDEMPOTENCE_KEY to id)
+    } else {
+      moshi.adapter<Map<String, String>>().fromJson(metadata.stringValue)!!
```
hamdanjaveed: I'd make the adapter a variable so we don't re-create it everytime

misk-aws/src/main/kotlin/misk/jobqueue/sqs/SqsJobQueue.kt
```diff
@@ -41,15 +51,24 @@ internal class SqsJobQueue @Inject internal constructor(
...
-                MessageAttributeValue()
-                  .withDataType("String")
-                  .withStringValue(it.context().traceId.toString()))
+                MessageAttributeValue().withDataType("String").withStringValue(traceId))
             }
+
+            // Add the internal metadata dictionary, encoded as JSON.
+            addMessageAttributesEntry(SqsJob.JOBQUEUE_METADATA_ATTR, MessageAttributeValue()
+                .withDataType("String")
+                .withStringValue(moshi.adapter<Map<String, String>>().toJson(metadata)))
```
hamdanjaveed: Same here, would make the adapter a variable

misk-aws/src/test/kotlin/misk/jobqueue/sqs/SqsJobTest.kt
```diff
@@ -0,0 +1,73 @@
...
+      addMessageAttributesEntry("foo", MessageAttributeValue().withDataType("String").withStringValue("bar"))
+    }
+
+    val job = SqsJob(QueueName("test"), queueResolver, sqsMetrics, moshi, message)
+
+    assertThat(job.idempotenceKey).isEqualTo(job.id)
+    assertThat(job.attributes)
+        .containsEntry("foo", "bar")
+        .doesNotContainKey(SqsJob.JOBQUEUE_METADATA_ATTR)
+  }
```
hamdanjaveed: Assert that the fallbacks for the parent queue and the idempotence key work?

misk-aws/src/test/kotlin/misk/jobqueue/sqs/SqsJobTest.kt
```diff
@@ -0,0 +1,73 @@
...
+                |  "${SqsJob.JOBQUEUE_METADATA_ORIGINAL_TRACE_ID}": "oti-0"
+                |}""".trimMargin()))
+    }
+    val job = SqsJob(QueueName("test"), queueResolver, sqsMetrics, moshi, message)
+
+    assertThat(job.idempotenceKey).isEqualTo("ik-0")
+    assertThat(job.attributes)
+        .containsEntry("foo", "bar")
+        .doesNotContainKey(SqsJob.JOBQUEUE_METADATA_ATTR)
+  }
```
hamdanjaveed: Assert that the origin queue and idempotence key values are present in the metadata?

misk-aws/src/main/kotlin/misk/jobqueue/sqs/SqsJob.kt
biasedbit: Moshi [caches adapters](https://github.com/square/moshi/blob/master/moshi/src/main/java/com/squareup/moshi/Moshi.java#L66) (just to err on the safe side, double checked with `System.identityHashCode()` just to be sure, and multiple calls with the same arguments result in same object identity).

misk-aws/src/test/kotlin/misk/jobqueue/sqs/SqsJobTest.kt
biasedbit: Can do!

biasedbit: Actually, we're testing that indirectly via the `idempotenceKey` property (which reads its value from the private `jobqueueMetadata` dictionary). `origin_queue` isn't exposed (unnecessary at this point).

misk-aws/src/test/kotlin/misk/jobqueue/sqs/SqsJobTest.kt
biasedbit: This test is for that scenario (notice the absence of idempotence key in the attributes, which causes idempotence key = job id).

misk-jobqueue/src/main/kotlin/misk/jobqueue/JobQueue.kt
biasedbit: Can't do `@JvmOverloads` in an interface so I added another signature.

PR #1437, opened 2020-03-13
Implement connection rebalancing for HTTP/2

swankjesse:
Also upgrade Jetty to the version we use internally


swankjesse:
Tragically, CI failed in exactly the way it shouldn’t! I suspect there’s a race that doesn’t reveal itself on my Mac.

```
Expected size:<2> but was:<1> in:
<[Connection{127.0.0.1:43083, proxy=DIRECT hostAddress=/127.0.0.1:43083 cipherSuite=TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256 protocol=h2}]>
	at misk.web.RebalancingEnabledTest.checkResponse(AbstractRebalancingTest.kt:116)
```

swankjesse:
... I found a simple way to mitigate against the race. By making 3 calls it’s a guarantee the first close will be observed before the 3rd call is attempted.

PR #1442, opened 2020-03-17
Introduce DynamicTypedClientFactory

wesleyk:
Enables services to dynamically build up typed clients
with the same interceptors / listeners that you would
get with TypedHttpClientModule.

Useful for platform services such as backfila that may
not know all the services they talk to upfront.

### DIFF
--- a/misk/src/main/kotlin/misk/client/TypedHttpClientModule.kt
+++ b/misk/src/main/kotlin/misk/client/TypedHttpClientModule.kt
@@ -61,7 +61,7 @@ class TypedHttpClientModule<T : Any>(
     private val name: String,
     private val httpClientProvider: Provider<OkHttpClient>,
     retrofitBuilderProvider: Provider<Retrofit.Builder>?
-  ) : TypedClientFactory<T>(kclass, name, retrofitBuilderProvider), Provider<T> {
+  ) : TypedClientFactoryProvider<T>(kclass, name, retrofitBuilderProvider), Provider<T> {
 
     @Inject private lateinit var httpClientsConfig: HttpClientsConfig
     @Inject private lateinit var httpClientConfigUrlProvider: HttpClientConfigUrlProvider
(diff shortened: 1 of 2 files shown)


wesleyk:
Notably I'm deferring any desired client caching to the application layer, I don't think it's appropriate for Misk / this specific layer to provide that functionality

zcross:
> Notably I'm deferring any desired client caching to the application layer, I don't think it's appropriate for Misk / this specific layer to provide that functionality

I think this is the right call. In terms of the existing usages of the now-renamed module, presumably they are already dealing with this? I didn't see the diff touching any bindings or removing anything that might currently be relied on for caching of these clients?

misk/src/main/kotlin/misk/client/TypedHttpClientModule.kt
```diff
@@ -136,15 +136,47 @@ class TypedPeerHttpClientModule<T : Any>(
...
-private abstract class TypedClientFactory<T : Any>(
-  private val kclass: KClass<T>,
-  private val name: String,
-  private val retrofitBuilderProvider: Provider<Retrofit.Builder>?
-) {
+/**
+ * Build up a typed client dynamically in runtime. This is useful for platform-type services
+ * that cannot statically define all of the services they talk to.
+ *
+ * Services may find it desirable to cache the resulting clients to avoid incurring the construction
```
swankjesse: may find it desirable to --> should

wesleyk:
> Diff is nice and minimal, couldn't think of any line level feedback. Only question I had was whether this would change anything about the existing usages of the formerly static module. My read is that it does not change anything, i.e., that it is strictly additive (the addition being the dynamic usage support).

Exactly. What would change is in services like backfila, that appear to have spun up their own dynamic setup without the benefits of shared interceptors / observability. I'll ping the backfila folks to let them know this is available.

misk/src/main/kotlin/misk/client/TypedHttpClientModule.kt
```diff
@@ -163,7 +155,49 @@ private abstract class TypedClientFactory<T : Any>(
...
-  fun typedClient(client: OkHttpClient, baseUrl: String): T {
+  @Inject private lateinit var httpClientConfigUrlProvider: HttpClientConfigUrlProvider
+
+  @Inject private lateinit var httpClientFactory: HttpClientFactory
+
+  /**
+   * Build up a typed client dynamically in runtime. This is useful for platform-type services
+   * that cannot statically define all of the services they talk to.
+   *
+   * Services should cache the resulting clients to avoid incurring the construction on every call.
```
ryanhall07: misk should probably provide this, just not at this level.

wesleyk: yeah, maybe. I'll build it for plasma and see

PR #1443, opened 2020-03-18
Add @Inject annotation to http config classes

wesleyk:


wesleyk:
Needed to be able to @Provide some default configs as part of building dynamic clients

misk-core/src/main/kotlin/misk/security/ssl/CertStoreConfig.kt
```diff
@@ -1,6 +1,8 @@
 package misk.security.ssl
 
-data class CertStoreConfig(
+import javax.inject.Inject
+
+data class CertStoreConfig @Inject constructor(
   val resource: String,
```
zcross: How are these strings w/out annotations selected for injection?

misk-core/src/main/kotlin/misk/security/ssl/TrustStoreConfig.kt
```diff
@@ -1,8 +1,9 @@
 package misk.security.ssl
 
 import misk.security.ssl.SslLoader.Companion.FORMAT_JCEKS
+import javax.inject.Inject
 
-data class TrustStoreConfig(
+data class TrustStoreConfig @Inject constructor(
```
zcross: Same Q: are these all provided by factories or providers that invoke the constructor manually?

misk/src/main/kotlin/misk/client/HttpClientsConfig.kt
```diff
@@ -50,7 +51,7 @@ data class HttpClientEndpointConfig(
   val ssl: HttpClientSSLConfig? = null
 )
 
-data class HttpClientEnvoyConfig(
+data class HttpClientEnvoyConfig @Inject constructor(
   val app: String,
```
zcross: Does this need `@AppName`?

misk-core/src/main/kotlin/misk/security/ssl/TrustStoreConfig.kt
wesleyk: These aren't injected out of the box. Explicit instances are provided by the app from HttpClientsConfig. However, our strict guice mode means that you need to have this annotation, even if you provide it.

zcross: Ah, that clarifies things. Thanks!

misk/src/main/kotlin/misk/client/HttpClientsConfig.kt
```diff
@@ -29,14 +30,30 @@ data class HttpClientsConfig(
         ssl = endpointConfig.ssl ?: ssl
     )
   }
+
+  /**
+   * Returns a default config with no url/envoy config specified.
+   * Used to build up endpoint config for clients dynamically.
+   */
+  fun getDefault(): HttpClientEndpointConfig {
```
zcross: Ah, this is nice.

misk-core/src/main/kotlin/misk/security/ssl/CertStoreConfig.kt
```diff
@@ -1,6 +1,8 @@
 package misk.security.ssl
 
-data class CertStoreConfig(
+import javax.inject.Inject
+
+data class CertStoreConfig @Inject constructor(
```
swankjesse: I don’t think this is going to inject the strings you want it to inject

misk-core/src/main/kotlin/misk/security/ssl/TrustStoreConfig.kt
swankjesse: Oh wow, that’s a terrible bug. Lets fix

PR #1446, opened 2020-03-23
added ability to customize table creation

abakes:


abakes:
> This LGTM. I wonder if table creation should just be the responsibility of the DockerDynamoDb external dependency.

I'm not sure I see the value in that. This also keeps things consistent with Hibernate modules.

(The issue for me here is that DynamoDBMapper does not default secondary indexes to ProjectionType.ALL, which mean I cannot properly test).

mightyguava:
> I'm not sure I see the value in that.

I like leaner abstractions :)

misk-aws-dynamodb-testing/src/main/kotlin/misk/aws/dynamodb/testing/DockerDynamoDbModule.kt
```diff
@@ -30,7 +35,13 @@ class DockerDynamoDbModule(
...
+ * [DynamoDbTable] is a wrapper class that allows for unqualified binding of
+ * dynamo table classes without collision. It also provides the ability to
+ * customize the table creation request for testing (ie. if a secondary index
+ * required ProjectionType.ALL).
  */
-data class DynamoDbTable(val tableClass: KClass<*>)
\ No newline at end of file
+data class DynamoDbTable(
+  val tableClass: KClass<*>,
+  val customizeRequest: (CreateTableRequest) -> CreateTableRequest =
```
zhxnlai: customize table?

abakes: I like Configure!

abakes:
> > I'm not sure I see the value in that.
> 
> I like leaner abstractions :)

Doesn't that mean that each test would need the table definition?

misk-aws-dynamodb-testing/src/main/kotlin/misk/aws/dynamodb/testing/DockerDynamoDbModule.kt
```diff
@@ -11,14 +12,18 @@ import kotlin.reflect.KClass
...
 class DockerDynamoDbModule(
-  private val tables: List<KClass<*>>
+  private val tables: List<DynamoDbTable>
 ) : KAbstractModule() {
 
-  constructor(vararg tables: KClass<*>) : this(tables.toList())
+  constructor(vararg tables: DynamoDbTable) : this(tables.toList())
+  constructor(vararg tables: KClass<*>) :
+    this(
+        tables.toList().map { DynamoDbTable(it) }
```
zhxnlai: Would `tables.toList().map { DynamoDbTable(it) }` work?

abakes: it does. updating

misk-aws-dynamodb-testing/src/main/kotlin/misk/aws/dynamodb/testing/CreateTablesService.kt
```diff
@@ -47,8 +46,16 @@ internal class CreateTablesService @Inject constructor(
       // Provisioned throughput needs to be specified when creating the table.
       globalSecondaryIndex.provisionedThroughput = ProvisionedThroughput(1L, 1L)
     }
+    tableRequest = table.customizeRequest(tableRequest)
+
     DynamoDB(this).createTable(tableRequest).waitForActive()
   }
+
+  companion object {
+    val customizeRequestNoop: (CreateTableRequest) -> CreateTableRequest = {
```
zhxnlai: object val properties with no custom get function that hold deeply immutable data) should use uppercase underscore-separated names:

 https://kotlinlang.org/docs/reference/coding-conventions.html#property-names

mightyguava:
> Doesn't that mean that each test would need the table definition?

Or a shared function that returns the table definition.

PR #1447, opened 2020-03-25
Optional label selector for pods considered in app cluster

alecholmes:
Not all pods in a namespace may be part of the same Misk application. This adds the ability to specify a pod label selector to specify which pods are part of the running application's cluster.

### DIFF
--- a/misk/src/main/kotlin/misk/clustering/kubernetes/KubernetesConfig.kt
+++ b/misk/src/main/kotlin/misk/clustering/kubernetes/KubernetesConfig.kt
@@ -2,10 +2,17 @@ package misk.clustering.kubernetes
 
 import misk.config.Config
 
+/**
+ * @property clustering_pod_label_selector Optional Kubernetes label selector to filter which pods
+ *     in the namespace are considered to be in the same cluster. If omitted, all healthy pods in
+ *     the namespace are included in the cluster.
+ *     Ex: "app = helloserver".
+ */
 data class KubernetesConfig(
   val my_pod_namespace: String = System.getenv("MY_POD_NAMESPACE") ?: "<invalid-namespace>",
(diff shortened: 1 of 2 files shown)


misk/src/main/kotlin/misk/clustering/kubernetes/KubernetesConfig.kt
```diff
@@ -2,10 +2,17 @@ package misk.clustering.kubernetes
 
 import misk.config.Config
 
+/**
+ * @property clustering_pod_label_selector Optional Kubernetes label selector to filter which pods
+ *                                         in the namespace are considered to be in the same cluster.
```
swankjesse: wanna wrap +4 like source code?

misk/src/main/kotlin/misk/clustering/kubernetes/KubernetesConfig.kt
```diff
@@ -2,10 +2,17 @@ package misk.clustering.kubernetes
...
+ *                                         If omitted, all healthy pods in the namespace are included in the cluster.
+ *                                         Ex: "app = helloserver".
+ */
 data class KubernetesConfig(
   val my_pod_namespace: String = System.getenv("MY_POD_NAMESPACE") ?: "<invalid-namespace>",
   val my_pod_name: String = System.getenv("MY_POD_NAME") ?: "<invalid-pod-name>",
   val my_pod_ip: String = System.getenv("MY_POD_IP") ?: "<invalid-pod-ip>",
+  val clustering_pod_label_selector: String? = null,
```
swankjesse: should this default to `"app = $my_pod_namespace"` instead of null?

alecholmes: I put the opinionated default in `skim`, since having a misk default that doesn't just work feels surprising.

PR #1451, opened 2020-03-26
Make it easier to test jobs that aren't acknowledged

swankjesse:


misk-jobqueue-testing/src/main/kotlin/misk/jobqueue/FakeJobQueue.kt
```diff
@@ -46,30 +46,54 @@ class FakeJobQueue @Inject constructor(
...
   override val queueName: QueueName,
   override val id: String,
   override val idempotenceKey: String,
   override val body: String,
-  override val attributes: Map<String, String>,
-  internal var acknowledged: Boolean = false
+  override val attributes: Map<String, String>
 ) : Job {
+  var acknowledged: Boolean = false
+    private set
```
tso: TIL private set

PR #1453, opened 2020-03-27
add deliverydelay field to fake job

BrandonWeng:
This allows us to test that the delivery delay field is set to an expected value


misk-jobqueue-testing/src/main/kotlin/misk/jobqueue/FakeJobQueue.kt
```diff
@@ -37,7 +37,14 @@ class FakeJobQueue @Inject constructor(
     attributes: Map<String, String>
   ) {
     val id = tokenGenerator.generate("fakeJobQueue")
-    val job = FakeJob(queueName, id, idempotenceKey, body, attributes)
+    val job = FakeJob(
+      queueName = queueName,
+      id = id,
+      idempotenceKey = idempotenceKey,
+      body = body,
+      attributes = attributes,
```
swankjesse: please use a consistent order to these parameters

swankjesse: (attributes is before deliveryDelay here, but after it above)

misk-jobqueue-testing/src/main/kotlin/misk/jobqueue/FakeJobQueue.kt
```diff
@@ -82,7 +89,8 @@ data class FakeJob(
   override val id: String,
   override val idempotenceKey: String,
   override val body: String,
-  override val attributes: Map<String, String>
+  override val attributes: Map<String, String>,
+  val deliveryDelay: Duration?
```
swankjesse: similarly

swankjesse:
We’re also overdue to merge FakeJobQueue and FakeTransactionalJobQueue into one class. We could probably just have one class that implements both interfaces!

BrandonWeng:
added code to actually honor the delivery delay and combined FakeJobQueue and FakeTransactionalJobQueue into one class.  

I would like to delete FakeTransactionalJobQueue entirely but that would break dependencies - not really sure if that's desired

BrandonWeng:
@swankjesse could you take another look whenever you have time?

PR #1456, opened 2020-04-01
Propagate traces or create root spans in ExecutoreServiceFactory

tirsen:


misk/src/main/kotlin/misk/concurrent/RealExecutorServiceFactory.kt
```diff
@@ -62,31 +66,51 @@ internal class RealExecutorServiceFactory @Inject constructor(
     awaitAllTerminated.start()
   }
 
+  // The traceWithActiveSpanOnly: false flag below means we will create a new root span if there is
```
ryanhall07, line 69: what is the problem with "too many root spans"? just potentially more for datadog to index? probably fine since we're sampling.

honestly it's good to surface them so we can attach them.

misk/src/test/kotlin/misk/concurrent/RealExecutorServiceFactoryTest.kt
```diff
@@ -32,10 +34,14 @@ internal class RealExecutorServiceFactoryTest {
         "running happy-0 run 1",
         "running happy-0 run 2"
     )
+    assertThat(tracer.finishedSpans().map { it.operationName() }).containsExactly(
```
ryanhall07, line 37: nice

misk/src/main/kotlin/misk/concurrent/RealExecutorServiceFactory.kt
```diff
@@ -62,31 +66,51 @@ internal class RealExecutorServiceFactory @Inject constructor(
     awaitAllTerminated.start()
   }
 
+  // The traceWithActiveSpanOnly=false flag below means we will create a new root span if there is
+  // not an ongoing active span at the point of handing over a task to the ExecutorService. This
+  // means we run the risk of having too many root spans but I prefer to err on the side of too much
+  // visibility. This may cause unnecessary noise though so we may need to change this as we try it
+  // out in production.
```
swankjesse, line 73: I think this is a comment that wants to be a runtime parameter and some documentation?

swankjesse:
Ship it!

misk/src/main/kotlin/misk/concurrent/RealExecutorServiceFactory.kt
tirsen: Yeah I agree

misk/src/main/kotlin/misk/concurrent/RealExecutorServiceFactory.kt
tirsen: Yeah possibly. I'd like to merge this as is and then see how things look and possibly add a parameter if necessary. I hope that this setting is good for all our services but I may be wrong.

tirsen:
I'm gonna merge with the broken docs thing. I don't think that's on me :-)

PR #1457, opened 2020-04-01
mysql table scan checker 

BrandonWeng:
The listener will catch table scans in mysql databases. 

Refactored ScaleSafetyCheck into an abstract class for the Vitess and MYSQL implementations to inherit from


misk-hibernate-testing/src/main/kotlin/misk/jdbc/AbstractScaleSafetyChecks.kt
```diff
@@ -0,0 +1,154 @@
...
+
+  /**
+   * Digs into the MySQL log to find the last executed DML statement that passed through Vitess.
+   */
+  internal open fun extractQueriesSince(mysqlTime: Timestamp): List<String> {
+    return connect()?.let { c ->
+      c.prepareStatement("""
+                  SELECT argument
+                  FROM mysql.general_log
+                  WHERE command_type in ('Query', 'Execute')
```
BrandonWeng, line 88: For the MYSQL database type, the queries (insert, select) were showing up as `Execute`. It was also the case when I checked the general log in our cloud service.

misk-hibernate-testing/src/main/kotlin/misk/jdbc/AbstractScaleSafetyChecks.kt
```diff
@@ -0,0 +1,154 @@
...
+        s.executeQuery().map { it.getString(1) }
+      }
+    } ?: arrayListOf()
+  }
+
+  /**
+   * Figure out how many total full scatter queries we've executed so far.
+   */
+  internal fun extractScatterQueryCount(): Int {
+    val request = Request.Builder()
```
swankjesse, line 109: should this be in a subtype?

misk-hibernate-testing/src/main/kotlin/misk/jdbc/AbstractScaleSafetyChecks.kt
```diff
@@ -0,0 +1,154 @@
...
+    }
+    return variables.QueriesProcessed["SelectScatter"] ?: 0
+  }
+
+  /**
+   * We don't use this anymore but we may want to introspect the query plans again in the future
+   * so I'm leaving it around for now.
+   */
+  private fun extractScatterQueryCountFromPlan(): Int {
+    val request = Request.Builder()
```
swankjesse, line 124: should this be in a subtype?

misk-hibernate-testing/src/main/kotlin/misk/jdbc/AbstractScaleSafetyChecks.kt
```diff
@@ -0,0 +1,154 @@
...
+  val isScatter: Boolean
+    get() = Instructions.isScatter
+}
+
+internal data class Variables(
+  /** Queries processed by plan type */
+  val QueriesProcessed: Map<String, Int>
+)
+
+abstract class AbstractScaleSafetyChecks @Inject constructor(
```
tirsen, line 39: Rather than reuse by inheritance can we reuse by composition? Inheritance is a blunt tool for reuse and one that I prefer to avoid...

tirsen: Actually I think these could even be static methods where you pass in the connection as a parameter

misk-hibernate-testing/src/main/kotlin/misk/jdbc/AbstractScaleSafetyChecks.kt
```diff
@@ -0,0 +1,154 @@
...
+    val adapter = moshi.adapter<Variables>()
+    val variables = okHttpClient.newCall(request).execute().use {
+      adapter.fromJson(it.body!!.source())!!
+    }
+    return variables.QueriesProcessed["SelectScatter"] ?: 0
+  }
+
+  /**
+   * We don't use this anymore but we may want to introspect the query plans again in the future
+   * so I'm leaving it around for now.
```
tirsen, line 121: Actually I don't think we need this anymore

misk-hibernate-testing/src/main/kotlin/misk/jdbc/AbstractScaleSafetyChecks.kt
```diff
@@ -0,0 +1,154 @@
...
+      .trimStart()
+      .toLowerCase()
+      .takeWhile { !it.isWhitespace() }
+    return DML.contains(first)
+  }
+
+  /**
+   * Digs into the MySQL log to find the last executed DML statement that passed through Vitess.
+   */
+  internal fun extractLastDmlQuery(): String? {
```
tirsen, line 59: This should be in the Vitess scale checker right? I don't think we would use this in a mysql checker...

misk-hibernate-testing/src/main/kotlin/misk/jdbc/MySqlScaleSafetyChecks.kt
```diff
@@ -0,0 +1,117 @@
...
+          .singleOrNull()
+      })
+    }
+
+    override fun afterQuery(query: String) {
+      if (!transacter.isCheckEnabled(Check.TABLE_SCAN)) return
+      val mysqlTime = mysqlTimeBeforeQuery.get() ?: return
+      val queries = extractQueriesSince(mysqlTime)
+
+      for (rawQuery in queries) {
```
tirsen, line 80: I think the rest of this method could be reused between the Vitess and the MySQL checkers. Possibly just as static methods.

misk-hibernate-testing/src/main/kotlin/misk/jdbc/AbstractScaleSafetyChecks.kt
BrandonWeng: i'm not too sure what you mean. any chance you could show me an example?

misk-hibernate-testing/src/main/kotlin/misk/jdbc/ScaleSafetyChecker.kt
```diff
@@ -0,0 +1,111 @@
...
+                  AND event_time > ?
+                  ORDER BY event_time DESC
+                """.trimIndent()).use { s ->
+        s.setTimestamp(1, mysqlTime)
+        s.executeQuery().map { it.getString(1) }
+      }
+    }
+  }
+
+  internal fun parseQueryPlans(moshi: Moshi, data: BufferedSource): Sequence<QueryPlan> {
```
BrandonWeng, line 95: this is just being used by a unit test - is this still needed?

BrandonWeng:
Refactored the common logic for the table scan detector into `ScaleSafetyChecker`

misk-hibernate-testing/src/main/kotlin/misk/jdbc/ScaleSafetyChecker.kt
```diff
@@ -0,0 +1,121 @@
...
+import mu.KotlinLogging
+import okio.BufferedSource
+import okio.ByteString.Companion.encodeUtf8
+import java.io.EOFException
+import java.sql.Connection
+import java.sql.SQLException
+import java.sql.SQLSyntaxErrorException
+import java.sql.Timestamp
+
+object ScaleSafetyChecker {
```
tirsen, line 15: nit: `Checker` makes it sound like it is something but this is really just a library so maybe just call it `ScaleSafetyChecks`

misk-hibernate-testing/src/main/kotlin/misk/jdbc/ScaleSafetyChecker.kt
tirsen: is it just a test to check if it works? then no, you can delete that

misk-hibernate-testing/src/test/kotlin/misk/jdbc/ParseQueryPlansTest.kt
```diff
@@ -58,7 +58,7 @@ class ParseQueryPlansTest {
     val moshi = Moshi.Builder()
         .add(KotlinJsonAdapterFactory())
         .build()
-    val plans = VitessScaleSafetyChecks.parseQueryPlans(
+    val plans = ScaleSafetyChecker.parseQueryPlans(
```
tirsen, line 61: nah delete this test

misk-hibernate/src/test/kotlin/misk/hibernate/MySqlScaleSafetyTest.kt
```diff
@@ -0,0 +1,59 @@
...
+class MySqlScaleSafetyTest {
+  @MiskTestModule
+  val module = MoviesTestModule(DataSourceType.MYSQL)
+
+  @Inject @Movies lateinit var transacter: Transacter
+
+  @Test
+  fun tableScansDetected() {
+    transacter.transaction { session ->
+      session.withoutChecks(Check.FULL_SCATTER, Check.COWRITE) {
```
tirsen, line 26: hmm these `withoutChecks` shouldn't need to be here? FULL_SCATTER and COWRITE can only happen on vitess

misk-hibernate-testing/src/main/kotlin/misk/jdbc/Explanation.kt
```diff
@@ -0,0 +1,196 @@
+package misk.jdbc
```
tirsen, line 1: I assume this is just a plain code move so will not review this in detail

misk-hibernate-testing/src/main/kotlin/misk/jdbc/MySqlScaleSafetyChecks.kt
```diff
@@ -0,0 +1,68 @@
...
+        Properties(),
+        config.username,
+        config.password
+      ).connection
+    } catch (e: SQLException) {
+      throw IllegalStateException("Could not connect to test MySQL server!", e)
+    }
+  }
+
+  inner class TableScanDetector : ExtendedQueryExecutionListener() {
```
tirsen, line 49: Can this whole class be reused between Vitess and MySQL?

tirsen: Oh I suppose the one in Vitess works per keyspace/shard. Nevermind.

tirsen:
YAY! This is great! It will avoid so many issues in production!

tirsen:
you can ignore the circleci docs build failure. Someone else is working on that and it's not related to this PR.

misk-hibernate/src/test/kotlin/misk/hibernate/MySqlScaleSafetyTest.kt
BrandonWeng: you're right... deleted

PR #1459, opened 2020-04-03
Rename vitess package as it hosts docker test infrastructure for othe…

kalfonso:
…r data stores


misk-hibernate-testing/src/main/kotlin/misk/hibernate/HibernateTestingModule.kt
```diff
@@ -14,7 +14,7 @@ import misk.jdbc.MySqlScaleSafetyChecks
 import misk.jdbc.TruncateTablesService
 import misk.jdbc.VitessScaleSafetyChecks
 import misk.time.ForceUtcTimeZoneService
-import misk.vitess.StartDatabaseService
+import misk.databases.StartDatabaseService
```
tirsen, line 17: I'm not a big fan of plural in package name. Can we call it just `database`?

kalfonso: Done

PR #1461, opened 2020-04-08
Allow apps to configure a global dead-letter queue

biasedbit:
Rather than forcing every queue to have its own DLQ, this change introduces the
possibility to configure a global DLQ, where all explicitly dead-lettered jobs
get sent to.

All jobs have a `_jobqueue-metadata` attribute that contains the original main
queue name. Jobs that end up in the global DLQ can be traced back to their main
queue by inspecting this property.

Unless this new property is overridden, apps will preserve existing behavior.

### DIFF
--- a/misk-aws/src/main/kotlin/misk/jobqueue/sqs/QueueResolver.kt
+++ b/misk-aws/src/main/kotlin/misk/jobqueue/sqs/QueueResolver.kt
@@ -19,10 +19,12 @@ internal class QueueResolver @Inject internal constructor(
   private val crossRegionSQS: Map<AwsRegion, AmazonSQS>,
   @ForSqsReceiving private val defaultForReceivingSQS: AmazonSQS,
   @ForSqsReceiving private val crossRegionForReceivingSQS: Map<AwsRegion, AmazonSQS>,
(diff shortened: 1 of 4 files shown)


biasedbit:
Build failure is unrelated to changes (master is also red).

misk-aws/src/main/kotlin/misk/jobqueue/sqs/QueueResolver.kt
```diff
@@ -95,7 +91,6 @@ val QueueName.retryQueue
 
 val QueueName.parentQueue
   get() = when {
-    isDeadLetterQueue -> QueueName(value.removeSuffix(deadLetterQueueSuffix))
```
hamdanjaveed, line 98: Would we still want to keep this in if the queue is using its own DLQ?

biasedbit: Good catch! I thought I reverted that with the (force) push of 9f9ccf9 but looks like I didn't  properly amend that commit.

I believe this particular `when` condition can no longer be met after this change but I didn't mean to remove it; will restore before merging.

PR #1466, opened 2020-04-10
Remove deployable shadowJars from samples

swankjesse:
These conflict with our internal publishing mechanisms.


swankjesse:
This stuff is obsolete. We may want to remove the examples completely.

PR #1468, opened 2020-04-14
Added information about a HTTP request's source socket.

janzantinge:
The SourceSocket data class provides some identifying information about the socket on which a request arrives. Authentication interceptors will sometimes make assumptions based on the source socket (for example: accept a header value as an identity if the request arrives on a Unix domain socket).

### DIFF
--- a/misk/src/main/kotlin/misk/web/HttpCall.kt
+++ b/misk/src/main/kotlin/misk/web/HttpCall.kt
@@ -12,12 +12,21 @@ import okio.BufferedSink
 import okio.BufferedSource
 
 /**
+ * Information about the socket on which a HTTP call arrived.
+ */
+sealed class SocketAddress {
+  class Network(val ipAddress: String, val port: Int) : SocketAddress()
+  class Unix(val path: String) : SocketAddress()
+}
+
+/**
  * A live HTTP call from a client for use by a chain of network interceptors.
  */
 interface HttpCall {
 
   /** Immutable information about the incoming HTTP request. */
   val url: HttpUrl
+  val linkLayerAddress: SocketAddress?
   val dispatchMechanism: DispatchMechanism
   val requestHeaders: Headers
 
--- a/misk/src/main/kotlin/misk/web/ServletHttpCall.kt
+++ b/misk/src/main/kotlin/misk/web/ServletHttpCall.kt
@@ -12,6 +12,7 @@ import javax.servlet.http.HttpServletRequest
 
 internal data class ServletHttpCall(
   override val url: HttpUrl,
+  override val linkLayerAddress: SocketAddress? = null,
   override val dispatchMechanism: DispatchMechanism,
   override val requestHeaders: Headers,
   var requestBody: BufferedSource? = null,
@@ -98,6 +99,7 @@ internal data class ServletHttpCall(
       request: HttpServletRequest,
       dispatchMechanism: DispatchMechanism,
       upstreamResponse: UpstreamResponse,
+      linkLayerAddress: SocketAddress? = null,
       webSocket: WebSocket? = null,
       requestBody: BufferedSource? = null,
       responseBody: BufferedSink? = null
@@ -108,6 +110,7 @@ internal data class ServletHttpCall(
 
       return ServletHttpCall(
           url = request.httpUrl(),
+          linkLayerAddress = linkLayerAddress,
           dispatchMechanism = dispatchMechanism,
           requestHeaders = request.headers(),
           upstreamResponse = upstreamResponse,
--- a/misk/src/main/kotlin/misk/web/jetty/JettyWebSocket.kt
+++ b/misk/src/main/kotlin/misk/web/jetty/JettyWebSocket.kt
@@ -3,6 +3,7 @@ package misk.web.jetty
 import misk.web.BoundAction
 import misk.web.DispatchMechanism
 import misk.web.ServletHttpCall
+import misk.web.SocketAddress
 import misk.web.actions.WebAction
 import misk.web.actions.WebSocket
 import misk.web.actions.WebSocketListener
@@ -10,6 +11,8 @@ import okhttp3.Headers
 import okio.ByteString
 import okio.ByteString.Companion.toByteString
 import okio.utf8Size
+import org.eclipse.jetty.server.ServerConnector
+import org.eclipse.jetty.unixsocket.UnixSocketConnector
 import org.eclipse.jetty.websocket.api.Session
 import org.eclipse.jetty.websocket.api.WebSocketAdapter
 import org.eclipse.jetty.websocket.api.WriteCallback
--- a/misk/src/main/kotlin/misk/web/jetty/WebActionsServlet.kt
+++ b/misk/src/main/kotlin/misk/web/jetty/WebActionsServlet.kt
@@ -5,6 +5,7 @@ import misk.logging.getLogger
 import misk.web.BoundAction
 import misk.web.DispatchMechanism
 import misk.web.ServletHttpCall
+import misk.web.SocketAddress
 import misk.web.actions.WebAction
 import misk.web.actions.WebActionEntry
 import misk.web.actions.WebActionFactory
@@ -20,6 +21,8 @@ import okio.source
 import org.eclipse.jetty.http.HttpMethod
 import org.eclipse.jetty.http2.HTTP2Connection
 import org.eclipse.jetty.server.Response
+import org.eclipse.jetty.server.ServerConnector
+import org.eclipse.jetty.unixsocket.UnixSocketConnector
 import org.eclipse.jetty.websocket.servlet.ServletUpgradeResponse
 import org.eclipse.jetty.websocket.servlet.WebSocketServlet
 import org.eclipse.jetty.websocket.servlet.WebSocketServletFactory
@@ -75,6 +78,13 @@ internal class WebActionsServlet @Inject constructor(
     try {
       val httpCall = ServletHttpCall.create(
           request = request,
+          linkLayerAddress = with ((request as? org.eclipse.jetty.server.Request)?.httpChannel?.connector) {
+            when (this) {
+              is UnixSocketConnector -> SocketAddress.Unix(this.unixSocket)
+              is ServerConnector -> SocketAddress.Network(this.host, this.localPort)
+              else -> throw IllegalStateException("Unknown socket connector.")
+            }
+          },
           dispatchMechanism = request.dispatchMechanism(),
           upstreamResponse = JettyServletUpstreamResponse(response as Response),
           requestBody = request.inputStream.source().buffer(),
--- a/misk/src/test/kotlin/misk/web/jetty/WebActionsServletTest.kt
+++ b/misk/src/test/kotlin/misk/web/jetty/WebActionsServletTest.kt
@@ -1,0 +1,138 @@
+package misk.web.jetty
+
+import com.netflix.concurrency.limits.limit.SettableLimit
+import com.netflix.concurrency.limits.limiter.SimpleLimiter
+import misk.Action
+import misk.ApplicationInterceptor
+import misk.asAction
+import misk.client.UnixDomainSocketFactory
+import misk.inject.KAbstractModule
+import misk.testing.MiskTest
+import misk.testing.MiskTestModule
+import misk.time.FakeClock
+import misk.time.FakeClockModule
+import misk.web.Get
+import misk.web.NetworkChain
+import misk.web.NetworkInterceptor
+import misk.web.ResponseContentType
+import misk.web.SocketAddress
+import misk.web.WebActionModule
+import misk.web.WebTestingModule
+import misk.web.WebUnixDomainSocketConfig
+import misk.web.actions.LivenessCheckAction
+import misk.web.actions.ReadinessCheckAction
+import misk.web.actions.StatusAction
+import misk.web.actions.WebAction
+import misk.web.interceptors.UserInterceptorTest
+import misk.web.mediatype.MediaTypes
+import okhttp3.Headers
+import okhttp3.HttpUrl.Companion.toHttpUrl
+import okhttp3.OkHttpClient
+import okhttp3.Protocol
+import okhttp3.Request
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.jupiter.api.Test
+import java.io.File
+import java.time.Duration
+import java.util.UUID
+import javax.inject.Inject
+
+@MiskTest(startService = true)
+class WebActionsServletTest {
+  @MiskTestModule
+  val module = TestModule()
+
+  @Inject
+  internal lateinit var jettyService: JettyService
+
+  private var socketName: String = "@udstest" + UUID.randomUUID().toString()
+
+  @Test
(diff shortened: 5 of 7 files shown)


misk/src/main/kotlin/misk/web/HttpCall.kt
```diff
@@ -11,13 +11,21 @@ import okhttp3.RequestBody
...
+}
+
 /**
  * A live HTTP call from a client for use by a chain of network interceptors.
  */
 interface HttpCall {
 
   /** Immutable information about the incoming HTTP request. */
   val url: HttpUrl
+  val sourceSocketType: SourceSocketType
```
alecholmes, line 28: Since you'll be able to extract it, you may want to include the listener name using something like:
```
data class Listener(socketType: SourceSocketType, name String)

interface HttpCall {
  // ...
  val listener: Listener
  // ...
}
```

This would let the interceptor trust only a specific listener, not any unix socke tone.

alecholmes:
Looks good so far.

misk/src/main/kotlin/misk/web/ServletHttpCall.kt
```diff
@@ -12,6 +12,7 @@ import javax.servlet.http.HttpServletRequest
 
 internal data class ServletHttpCall(
   override val url: HttpUrl,
+  override val sourceSocket: SourceSocket = SourceSocket(SourceSocket.Type.NETWORK),
```
alecholmes, line 15: Why a default value?

misk/src/test/kotlin/misk/web/jetty/WebActionsServletTest.kt
```diff
@@ -0,0 +1,131 @@
...
+    @ResponseContentType(MediaTypes.TEXT_PLAIN_UTF8)
+    fun call(): TestActionResponse {
+      return TestActionResponse("foo")
+    }
+  }
+
+  internal data class TestActionResponse(val text: String)
+
+  private fun get(path: String, socketType: SourceSocket.Type): okhttp3.Response =
+    if (socketType == SourceSocket.Type.NETWORK)
```
alecholmes, line 85: Misk supports an alternative syntax that will cause build failures if not all enum cases are handled:
```
val request = Request.Builder()
        .url(jettyService.httpServerUrl.newBuilder().encodedPath(path).build())
        .get()

return when (socketType) {
  SourceSocket.Type.NETWORK -> call(request)
  SourceSocket.Type.UNIX -> udsCall(request)
}
```

misk/src/test/kotlin/misk/web/jetty/WebActionsServletTest.kt
```diff
@@ -0,0 +1,131 @@
...
+        .url(jettyService.httpServerUrl.newBuilder().encodedPath(path).build())
+        .get())
+    else
+      udsCall(Request.Builder()
+        .url(jettyService.httpServerUrl.newBuilder().encodedPath(path).build())
+        .get())
+
+  private fun call(request: Request.Builder): okhttp3.Response {
+    val httpClient = OkHttpClient()
+    val response = httpClient.newCall(request.build()).execute()
```
alecholmes, line 96: `return httpClient.newCall(request.build()).execute()`

misk/src/test/kotlin/misk/web/jetty/WebActionsServletTest.kt
```diff
@@ -0,0 +1,131 @@
...
+
+@MiskTest(startService = true)
+class WebActionsServletTest {
+  @MiskTestModule
+  val module = TestModule()
+
+  @Inject
+  internal lateinit var jettyService: JettyService
+
+  private var socketName: String = "@udstest" + UUID.randomUUID().toString()
```
alecholmes, line 41: Does this socket get deleted when the test is done?

misk/src/test/kotlin/misk/web/jetty/WebActionsServletTest.kt
```diff
@@ -0,0 +1,131 @@
...
+  private fun call(request: Request.Builder): okhttp3.Response {
+    val httpClient = OkHttpClient()
+    val response = httpClient.newCall(request.build()).execute()
+    return response
+  }
+
+  private fun udsCall(request: Request.Builder): okhttp3.Response {
+    val httpClient = OkHttpClient().newBuilder()
+        .socketFactory(UnixDomainSocketFactory(File(socketName)))
+        .protocols(listOf(Protocol.HTTP_1_1))
```
alecholmes, line 103: Is this needed?

misk/src/test/kotlin/misk/web/jetty/WebActionsServletTest.kt
```diff
@@ -0,0 +1,131 @@
...
+  fun udsSocketSuccess() {
+    val response = get("/potato", SourceSocket.Type.UNIX)
+    assertThat(response.code).isEqualTo(200)
+    assertThat(response.header("ActualSocketType")).isEqualTo("UNIX")
+    assertThat(response.header("ActualSocketName")).isEqualTo(socketName)
+  }
+
+  internal class WebActionsServletNetworkInterceptor : NetworkInterceptor {
+    override fun intercept(chain: NetworkChain) {
+      chain.httpCall.statusCode = 200
```
alecholmes, line 61: Is this needed?

misk/src/main/kotlin/misk/web/HttpCall.kt
```diff
@@ -11,13 +11,31 @@ import okhttp3.RequestBody
 import okio.BufferedSink
 import okio.BufferedSource
 
+/**
+ * Information about the socket on which a HTTP call arrived.
+ */
+data class SourceSocket (
+  val type: Type = Type.NETWORK,
+  val name: String? = null
```
alecholmes, line 19: I think this will have a name even for network sockets, so can make it a non-optional `String` without a default value.

misk/src/main/kotlin/misk/web/HttpCall.kt
```diff
@@ -11,13 +11,31 @@ import okhttp3.RequestBody
 import okio.BufferedSink
 import okio.BufferedSource
 
+/**
+ * Information about the socket on which a HTTP call arrived.
+ */
+data class SourceSocket (
+  val type: Type = Type.NETWORK,
```
alecholmes, line 18: Preferable to have the caller indicate the type of socket than assume a default. Calling code should have to think about the type of socket.

misk/src/main/kotlin/misk/web/jetty/WebActionsServlet.kt
```diff
@@ -73,8 +75,13 @@ internal class WebActionsServlet @Inject constructor(
 
   private fun handleCall(request: HttpServletRequest, response: HttpServletResponse) {
     try {
+      val connector = (request as? org.eclipse.jetty.server.Request)?.httpChannel?.connector
       val httpCall = ServletHttpCall.create(
           request = request,
+          sourceSocket =
+            if (connector is UnixSocketConnector)
```
alecholmes, line 82: Square style is:
```
sourceSocket = if (connector is UnixSocketConnector) {
    SourceSocket(SourceSocket.Type.UNIX, connector.unixSocket)
  } else {
    SourceSocket(SourceSocket.Type.NETWORK)
  }
```

But you can also use `when`:
```
sourceSocket = when(connector) {
  is UnixSocketConnector -> SourceSocket(SourceSocket.Type.UNIX, connector.unixSocket)
  else -> SourceSocket(SourceSocket.Type.NETWORK)
}
```

misk/src/test/kotlin/misk/web/jetty/WebActionsServletTest.kt
```diff
@@ -0,0 +1,131 @@
...
+      chain.httpCall.addResponseHeaders(
+          Headers.Builder()
+            .set("ActualSocketType", chain.httpCall.sourceSocket.type.toString())
+            .set("ActualSocketName", chain.httpCall.sourceSocket.name?:"null")
```
alecholmes, line 65: Are you formatting the code with the Square styles imported into IJ?

janzantinge: probably not

misk/src/test/kotlin/misk/web/jetty/WebActionsServletTest.kt
janzantinge: presumably when the webserver shuts down, I don't have control over it's lifecycle AFAIK

alecholmes: Worth checking, and making sure that unit tests don't leave a bunch of socket files around. There should be unit test lifecycle hooks to help here, if not.

misk/src/test/kotlin/misk/web/jetty/WebActionsServletTest.kt
alecholmes: Make sure to do that before making a non-draft PR. go/cashdevguide has instructions.

janzantinge: The styles are imported, nothing changed 🦊

misk/src/test/kotlin/misk/web/jetty/WebActionsServletTest.kt
```diff
@@ -0,0 +1,123 @@
...
+import misk.Action
+import misk.ApplicationInterceptor
+import misk.asAction
+import misk.client.UnixDomainSocketFactory
+import misk.inject.KAbstractModule
+import misk.testing.MiskTest
+import misk.testing.MiskTestModule
+import misk.time.FakeClock
+import misk.time.FakeClockModule
+import misk.web.*
```
alecholmes, line 14: Don't use `*` importts

janzantinge: typically I would agree with this but I would make exceptions for tests

misk/src/test/kotlin/misk/web/jetty/WebActionsServletTest.kt
janzantinge: there are no leftover sockets. IIRC this makes sense since I would expect Jetty to shutdown a connector which *should* call close on the abstract unix domain socket. Closing an abstract unix domain socket will always remove it.

misk/src/main/kotlin/misk/web/jetty/JettyWebSocket.kt
```diff
@@ -155,6 +158,7 @@ internal class JettyWebSocket(
 
       val httpCall = ServletHttpCall.create(
           request = request.httpServletRequest,
+          sourceSocket = SourceSocket(SourceSocket.Type.NETWORK, "websocket"),
```
janzantinge, line 161: I don't know what name this should have, the request doesn't appear to have any (network) socket info, only request info. This means there isn't a way to get the host:port (other then parsing it out of the ws:// request URL)

misk/src/test/kotlin/misk/web/jetty/WebActionsServletTest.kt
```diff
@@ -0,0 +1,131 @@
...
+  internal lateinit var jettyService: JettyService
+
+  private var socketName: String = "@udstest" + UUID.randomUUID().toString()
+
+  @Test
+  fun networkSocketSuccess() {
+    val response = get("/potato", SourceSocket.Type.NETWORK)
+    assertThat(response.header("ActualSocketType")).isEqualTo("NETWORK")
+    assertThat(response.header("ActualSocketName")).isEqualTo(
+      jettyService.httpServerUrl.toString().removeSurrounding("http://", "/"))
```
alecholmes, line 55: `"${jettyService.httpServerUrl.host}:${jettyService.httpServerUrl.port}"`

alecholmes:
This LGTM. I think it's ready for a wider review.

misk/src/test/kotlin/misk/web/jetty/WebActionsServletTest.kt
```diff
@@ -0,0 +1,131 @@
...
+
+@MiskTest(startService = true)
+class WebActionsServletTest {
+  @MiskTestModule
+  val module = TestModule()
+
+  @Inject
+  internal lateinit var jettyService: JettyService
+
+  private var socketName: String = "@udstest" + UUID.randomUUID().toString()
```
alecholmes, line 48: I think an earlier comment got lost: add a test hook to delete these sockets after tests complete.

janzantinge: I checked, they're deleted, no need for a hook

misk/src/test/kotlin/misk/web/jetty/WebActionsServletTest.kt
janzantinge: that's much better, forgot httpServerUrl wasn't a string

misk/src/main/kotlin/misk/web/HttpCall.kt
```diff
@@ -11,13 +11,31 @@ import okhttp3.RequestBody
 import okio.BufferedSink
 import okio.BufferedSource
 
+/**
+ * Information about the socket on which a HTTP call arrived.
+ */
+data class SourceSocket (
```
swankjesse, line 17: Rename to SocketAddress. This is definitely not a socket, but merely it’s identifier

misk/src/main/kotlin/misk/web/HttpCall.kt
```diff
@@ -11,13 +11,31 @@ import okhttp3.RequestBody
...
+}
+
 /**
  * A live HTTP call from a client for use by a chain of network interceptors.
  */
 interface HttpCall {
 
   /** Immutable information about the incoming HTTP request. */
   val url: HttpUrl
+  val sourceSocket: SourceSocket?
```
swankjesse, line 38: rename the field to `linkLayerAddress`. “Source” suggests origination to me, and I think this is much weaker than that.

Can you also document a disclaimer that this is the raw address known to our service, but it is very unlikely to be the originating caller’s host address? Perhaps say that this is the nearest middlebox address? (I’d be very sad if somebody did risk analysis based on these addresses.)

For follow up it’d be rad to include an originAddress field, perhaps populated by something that understands the corresponding HTTP headers in a service’s deployment topology.

misk/src/main/kotlin/misk/web/HttpCall.kt
```diff
@@ -11,13 +11,31 @@ import okhttp3.RequestBody
 import okio.BufferedSink
 import okio.BufferedSource
 
+/**
+ * Information about the socket on which a HTTP call arrived.
+ */
+data class SourceSocket (
+  val type: Type,
+  val name: String
```
swankjesse, line 19: I think `name` is too weak. This would be worth properly modeling.

```
sealed class SocketAddress {
  class NetworkAddress(val ipAddress: String, val port: Int) : SocketAddress
  class UnixSocketAddress(val path: String) : SocketAddress
}
```

swankjesse: (With sealed classes you also don’t need the enum!)

PR #1470, opened 2020-04-15
Expose all gRPC endpoints as JSON POST endpoints

swankjesse:
We use the same mechanism we were using to expose proto endpoints
as JSON endpoints.

This also changes misk-web to use httpMethod instead of
dispatchMechanism to describe server-side endpoints. These
two concepts are related but different and conflating them
was broken for gRPC.

### DIFF
--- a/misk/src/main/kotlin/misk/web/actions/WebActionFactory.kt
+++ b/misk/src/main/kotlin/misk/web/actions/WebActionFactory.kt
@@ -16,6 +16,7 @@ import misk.web.Get
 import misk.web.NetworkInterceptor
 import misk.web.PathPattern
 import misk.web.Post
+import misk.web.RequestBody
 import misk.web.WebActionBinding
 import misk.web.mediatype.MediaRange
 import misk.web.mediatype.MediaTypes
@@ -23,6 +24,7 @@ import javax.inject.Inject
 import javax.inject.Singleton
 import kotlin.reflect.KClass
 import kotlin.reflect.KFunction
+import kotlin.reflect.KParameter
 import kotlin.reflect.full.functions
 
 @Singleton
@@ -130,8 +132,20 @@ internal class WebActionFactory @Inject constructor(
    * Returns null if neither of the input action's content types were protobuf.
    */
   private fun transformActionIntoJson(action: Action): Action? {
-    var jsonVariant = action
+    // Map gRPC actions to the equivalent JSON actions. This changes the dispatch mechanism from
+    // gRPC to POST which prevents gRPC framing from being used.
+    if (action.dispatchMechanism == DispatchMechanism.GRPC && action.parameters.size == 1) {
+      val mappedParameters = action.parameters.toMutableList()
+      mappedParameters[0] = mappedParameters[0].withAnnotations(listOf(requestBodyAnnotation()))
+      return action.copy(
+          dispatchMechanism = DispatchMechanism.POST,
+          acceptedMediaRanges = jsonMediaRanges,
+          responseContentType = jsonMediaType,
+          parameters = mappedParameters
+      )
+    }
 
+    var jsonVariant = action
     if (action.acceptedMediaRanges == protobufMediaRanges) {
       jsonVariant = jsonVariant.copy(acceptedMediaRanges = jsonMediaRanges)
     }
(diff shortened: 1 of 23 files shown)


misk/web/tabs/web-actions/src/ducks/webActions.ts
```diff
@@ -79,7 +79,7 @@ export interface IWebActionAPI {
   allowedServices: string[]
   allowedCapabilities: string[]
   applicationInterceptors: string[]
-  dispatchMechanism: HTTPMethod
```
swankjesse, line 82: this was problematic because the HTTPMethod enum didn’t really support GRPC.

adrw:
I'm unfamiliar with gRPC, does it still have concepts like method (Get, Post, Put...) or is relying on that `HTTPMethod` enum in Web Actions tab incompatible with how we think about gRPC? We can migrate from that enum (3rd party library) and roll our own in Misk-Web that better supports how we think about gRPC.

misk/src/test/kotlin/misk/web/JsonForProtoEndpointsTest.kt
```diff
@@ -79,14 +129,64 @@ internal class JsonForProtoEndpointsTest {
...
+      )))
+      install(WebActionModule.create<ProtoEchoShipmentToken>())
+      install(WebActionModule.create<GrpcEchoShipmentToken>())
     }
   }
 
-  class EchoShipmentToken @Inject constructor() : WebAction {
+  // NB: The server doesn't get a port until after it starts so we create the client module
+  // _after_ we start the services
+  class ClientModule(val jetty: JettyService) : KAbstractModule() {
```
acostama, line 163: does this class still need to be here? can't see where it used..

misk/web/tabs/web-actions/src/ducks/webActions.ts
```diff
@@ -154,7 +154,7 @@ export const WebActionInternalLabel: { [key: string]: string } = {
   "Allowed Capabilities": "allowedCapabilities",
   "Allowed Services": "allowedServices",
   "Application Interceptor": "applicationInterceptors",
-  "Dispatch Mechanism": "dispatchMechanism",
+  "Dispatch Mechanism": "httpMethod",
```
adrw, line 157: Can we update this to "HTTP Method" as well?

acostama:
> This also changes misk-web to use httpMethod instead of
> dispatchMechanism to describe server-side endpoints.

a little unfamiliar here but, if I have a gRPC endpoint and the same POST endpoint is there potential for ambiguity?

misk/src/test/kotlin/misk/web/JsonForProtoEndpointsTest.kt
swankjesse: ooooh good catch. Fixed!

misk/web/tabs/web-actions/src/ducks/webActions.ts
swankjesse: YES!

swankjesse: Fixed.

swankjesse:
@acostama there is in theory, but unlikely in practice. If you do have a collision Misk will crash. (It’s unlikely in practice because gRPC path declarations are unlikely to be what you’d pick naturally.)

adrw:
Looks like node build failures are snapshot tests. Update them with `miskweb test-update -e` and install the CLI if you don't have it with `npm install -g @misk/cli`

misk/src/main/kotlin/misk/web/actions/WebActionFactory.kt
```diff
@@ -130,8 +132,20 @@ internal class WebActionFactory @Inject constructor(
    * Returns null if neither of the input action's content types were protobuf.
    */
   private fun transformActionIntoJson(action: Action): Action? {
-    var jsonVariant = action
+    // Map gRPC actions to the equivalent JSON actions. This changes the dispatch mechanism from
+    // gRPC to POST which prevents gRPC framing from being used.
+    if (action.dispatchMechanism == DispatchMechanism.GRPC && action.parameters.size == 1) {
```
ryanhall07, line 137: what happens if it's a grpc stream?

ryanhall07, line 137: what does the parameter.size == 1 mean?

misk/src/test/kotlin/misk/web/JsonForProtoEndpointsTest.kt
```diff
@@ -96,7 +159,37 @@ internal class JsonForProtoEndpointsTest {
...
+      ShippingGetDestinationWarehouseBlockingServer, WebAction {
+    @Unauthenticated
+    override fun GetDestinationWarehouse(shipment: Shipment): Warehouse {
+      return Warehouse.Builder()
+          .warehouse_token(shipment.shipment_token)
+          .build()
+    }
+  }
+
+  // TODO(jwilson): get Wire to generate this interface.
```
ryanhall07, line 172: so at the moment you have to manually define these if you want to use http/json?

misk/src/main/kotlin/misk/web/actions/WebActionFactory.kt
swankjesse: parameters.size won’t be 1 and it won’t get any special treatment!

swankjesse: gRPC streams have two parameters (inbound and outbound streams)

misk/src/test/kotlin/misk/web/JsonForProtoEndpointsTest.kt
swankjesse: no definitely not. I just need to change how Misk generates its test .proto files from that Bash script to the Wire plugin. Seemed like too much for this PR.

misk/src/main/kotlin/misk/web/actions/WebActionFactory.kt
ryanhall07: it seems like we could at least support server streaming with long polling

ryanhall07: k. a think a code comment would be helpful here

PR #1471, opened 2020-04-16
Swap global DLQ AWS SQS config for a DLQ provider

biasedbit:
At SQ Cash all our queues have their redrive policy sending dead-lettered jobs
to a global DLQ. To avoid forcing this approach on non-sqcash misk clients,
`QueueResolver` previously relied on an optional config param that would
override the DLQ for all queues with a global DLQ. However, this approach was
leading to lots of repetition when updating all Cash services.

This change swaps that config property with a more generic dead-letter queue
provider, which receives a `QueueName` and returns the matching dead-letter
queue's `QueueName`, allowing applications to use an existing impl (e.g.
"<main_queue>_dlq" or static global DLQ) or build their own logic (e.g. read
from config file, map different queues to different DLQs, mix & match, etc.)

The default implementation preserves status quo — adds a "_dlq" suffix to the
main queue's name — but allows apps to bind a different provider.

h/t to rhall@ for the suggestion.


misk-aws/src/main/kotlin/misk/jobqueue/sqs/AwsSqsJobQueueConfig.kt
```diff
@@ -36,19 +36,5 @@ class AwsSqsJobQueueConfig(
...
-   * from. When undefined, explicitly dead-lettered jobs are sent to a dead-letter queue named after
-   * the parent queue, with a "_dlq" suffix (e.g. "barb-queue" -> "barb-queue_dlq").
-   *
-   * All jobs contain a `_jobqueue-metadata` custom attribute. Any jobs sent to the global
-   * dead-letter queue can be traced to their original main queue by inspecting this property.
-   *
-   * Make sure this property is set or unset according to the AWS redrive policy for queues used
-   * by your app.
-   */
-  val global_dead_letter_queue_name: String? = null
```
biasedbit, line 53: This property didn't live long enough to warrant deprecation; 🔪'ing it.

misk-aws/src/test/kotlin/misk/jobqueue/sqs/QueueResolverTest.kt
```diff
@@ -1,51 +0,0 @@
...
-import misk.mockito.Mockito.mock
-import org.assertj.core.api.Assertions.assertThat
-import org.junit.jupiter.api.Test
-import org.mockito.ArgumentMatchers.any
-import org.mockito.Mockito.`when`
-
-internal class QueueResolverTest {
-
-  @Test
-  fun `getDeadLetter returns global DLQ when one is configured`() {
```
biasedbit, line 18: One the fence about keeping one of these, not so much about testing multiple outputs but rather that it uses the specified DLQ provider. Thoughts?

PR #1473, opened 2020-04-18
Add support for HTTP PATCH method

JGulbronson:
Need to fix up the formatting, and look at adding an enum for HTTP methods in DispatchMechanism.

### DIFF
--- a/misk/src/main/kotlin/misk/web/DispatchMechanism.kt
+++ b/misk/src/main/kotlin/misk/web/DispatchMechanism.kt
@@ -1,7 +1,5 @@
 package misk.web
 
-import org.eclipse.jetty.http.HttpMethod
-
 /**
  * Describes how an action is processed. This is like the HTTP method but has special cases for web
  * sockets and gRPC. These protocols layer on top of HTTP methods and have different semantics.
@@ -9,6 +7,7 @@ import org.eclipse.jetty.http.HttpMethod
 enum class DispatchMechanism {
   GET,
   POST,
+  PATCH,
   DELETE,
   GRPC,
   WEBSOCKET;
@@ -19,19 +18,21 @@ enum class DispatchMechanism {
    * This can be misleading: web sockets don't behave like normal GETs (they are upgraded), and GRPC
    * calls don't behave like normal POSTS (they are duplex).
    */
-  val method: HttpMethod
+  // TODO(jeff) - Make this an enum?
+  val method: String
     get() {
       return when (this) {
-        GET -> HttpMethod.GET
-        POST -> HttpMethod.POST
-        DELETE -> HttpMethod.DELETE
+        GET -> "GET"
+        POST -> "POST"
+        PATCH -> "PATCH"
+        DELETE -> "DELETE"
 
         // gRPC layers over POST. https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md
-        GRPC -> HttpMethod.POST
+        GRPC -> "POST"
 
         // WebSocket upgrades from GET.
         // https://developer.mozilla.org/en-US/docs/Web/HTTP/Protocol_upgrade_mechanism
-        WEBSOCKET -> HttpMethod.GET
+        WEBSOCKET -> "GET"
       }
     }
 }
--- a/misk/src/main/kotlin/misk/web/DispatchMechanism.kt
+++ b/misk/src/main/kotlin/misk/web/DispatchMechanism.kt
@@ -18,7 +18,6 @@ enum class DispatchMechanism {
    * This can be misleading: web sockets don't behave like normal GETs (they are upgraded), and GRPC
    * calls don't behave like normal POSTS (they are duplex).
    */
-  // TODO(jeff) - Make this an enum?
   val method: String
     get() {
       return when (this) {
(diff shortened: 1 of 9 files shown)


JGulbronson:
I mostly just followed what was done in #1369 

However, since Jetty doesn't have built-in support for PATCH the same way it does for other verbs, I had to hack that on a bit.

JGulbronson:
@swankjesse I can't seem to add reviewers for some reason, but would love your thoughts on this.

misk/src/main/kotlin/misk/web/DispatchMechanism.kt
```diff
@@ -19,19 +18,20 @@ enum class DispatchMechanism {
    * This can be misleading: web sockets don't behave like normal GETs (they are upgraded), and GRPC
    * calls don't behave like normal POSTS (they are duplex).
    */
-  val method: HttpMethod
+  val method: String
```
JGulbronson, line 21: We possibly want this as a new enum to avoid comparing strings.

JGulbronson:
It seems I'm having trouble reproducing the failing tests locally

misk/src/main/kotlin/misk/web/DispatchMechanism.kt
shellderp: do we really need that if `DispatchMechanism` is already an enum? also this value could be a constructor param on this enum instead

JGulbronson: That's a good point, method is really just a subset of `DispatchMechanism`, so method could probably return a `DispatchMechanism` instead for cases like `WEBSOCKET` and `GET`

swankjesse: Why not use HttpMethod?

misk/src/main/kotlin/misk/web/jetty/WebActionsServlet.kt
```diff
@@ -59,6 +59,13 @@ internal class WebActionsServlet @Inject constructor(
     }
   }
 
+  override fun service(request: HttpServletRequest?, response: HttpServletResponse?) {
+    if (request?.method == "PATCH" && response != null) {
+      doPatch(request, response)
```
swankjesse, line 64: You need a `return` here

misk/src/main/kotlin/misk/web/jetty/WebActionsServlet.kt
```diff
@@ -59,6 +59,13 @@ internal class WebActionsServlet @Inject constructor(
...
+    }
+    super.service(request, response)
```
swankjesse, line 66: should we rename `handleCall` to `service` and remove the doGet/doPost/doDelete boilerplate?

misk/src/main/kotlin/misk/web/DispatchMechanism.kt
swankjesse: OMG, it’s because HttpMethod doesn’t have PATCH! Craxy

JGulbronson: Yeah, Jetty stuck a little too close to the spec on that one. Should we keep it as a string? Or try to use an enum?

misk/src/main/kotlin/misk/web/jetty/WebActionsServlet.kt
JGulbronson: Yeah maybe, though follow-up PR?

misk/src/main/kotlin/misk/web/jetty/WebActionsServlet.kt
JGulbronson: I'm actually confused how the test passes, it should be calling `doPatch` and then end up throwing later in the function. Will debug locally to make sure something else isn't happening.

JGulbronson: Yeah Jetty throws this in the background but the tests happily passes
```
java.lang.IllegalStateException: COMPLETED
	at org.eclipse.jetty.server.HttpChannelState.sendError(HttpChannelState.java:893)
	at org.eclipse.jetty.server.Response.sendError(Response.java:461)
	at javax.servlet.http.HttpServlet.service(HttpServlet.java:732)
	at org.eclipse.jetty.websocket.servlet.WebSocketServlet.service(WebSocketServlet.java:176)
	at misk.web.jetty.WebActionsServlet.service(WebActionsServlet.kt:66)
	at javax.servlet.http.HttpServlet.service(HttpServlet.java:790)
	at org.eclipse.jetty.servlet.ServletHolder.handle(ServletHolder.java:755)
	at org.eclipse.jetty.servlet.ServletHandler.doHandle(ServletHandler.java:547)
```

Adding the `return` fixes

JGulbronson:
@swankjesse thanks for the review! Added the `return`, it's a little concerning that exception deep down in Jetty didn't fail the test. Happy to add a new enum if we want that for `method()`, and then remove the boilerplate in a followup if that's the route we want to take.

misk/src/main/kotlin/misk/web/DispatchMechanism.kt
swankjesse: String is fine.

PR #1477, opened 2020-04-22
Make FakeTokenGenerator thread safe

Synesso:


misk/src/test/kotlin/misk/tokens/FakeTokenGeneratorTest.kt
```diff
@@ -51,6 +52,14 @@ class FakeTokenGeneratorTest {
     assertThat(token3).isEqualTo("payment000000000000000002")
   }
 
+  @Test
+  fun generationIsThreadSafe() {
```
swankjesse, line 56: This isn’t testing what you think it’s testing. Because you aren’t passing 100 unique prefixes to tokenGenerator.generate() you’re making it extremely unlikely that unsynchronized accesses to the map will contend in a meaningful way.

I think you want `tokenGenerator.generate("token$it")`

misk/src/main/kotlin/misk/tokens/FakeTokenGenerator.kt
```diff
@@ -1,13 +1,14 @@
...
 import misk.tokens.TokenGenerator.Companion.canonicalize
+import java.util.concurrent.ConcurrentHashMap
 import java.util.concurrent.atomic.AtomicLong
 import javax.inject.Inject
 import javax.inject.Singleton
 
 @Singleton
 class FakeTokenGenerator @Inject constructor() : TokenGenerator {
-  internal val nextByLabel = mutableMapOf<String, AtomicLong>()
+  internal val nextByLabel = ConcurrentHashMap<String, AtomicLong>()
```
swankjesse, line 11: I claim you almost never want ConcurrentHashMap for things that aren’t extremely highly concurrent. It’s a big slow datastructure designed for very high concurrency. If you only need safety I recommend this:
```
internal val nextByLabel = Collections.synchronizedMap(mutableMapOf<String, AtomicLong>())
```

misk/src/test/kotlin/misk/tokens/FakeTokenGeneratorTest.kt
Synesso: I red/green tested this and it exposes the bug (revert to `mutableMapOf` to see it). The bug being contention around first access for a single key.

misk/src/main/kotlin/misk/tokens/FakeTokenGenerator.kt
Synesso: I had considered (googled) both. My understanding is that `ConcurrentHashMap` locks writes and `SynchronizedMap` locks all access, giving the former the behaviour I'm looking for. Is that not the case?

misk/src/test/kotlin/misk/tokens/FakeTokenGeneratorTest.kt
swankjesse: Got it. Works for me.

misk/src/main/kotlin/misk/tokens/FakeTokenGenerator.kt
swankjesse: CHM is higher concurrency, synchronized map is lower memory. I think for tests the high concurrency won't pay for itself.

Synesso: 👍

PR #1481, opened 2020-04-24
Improve vitess scale safety checks.

kalfonso:
These PR addresses two issues:
- don't explain explain queries
- match tables by word boundaries. This avoid exceptions analyzing queries for sequence tables where ```select next, val from customer_token_lookup_seq``` will match ```customer_token_lookup``` in the ```service_lookup``` keyspace, for instancr

### DIFF
--- a/misk-hibernate-testing/src/main/kotlin/misk/jdbc/ScaleSafetyChecks.kt
+++ b/misk-hibernate-testing/src/main/kotlin/misk/jdbc/ScaleSafetyChecks.kt
@@ -23,7 +23,7 @@ object ScaleSafetyChecks {
   }
 
   fun checkQueryForTableScan(connection: Connection, query: String) {
-    if (isDml(query)) return
+    if (isDml(query) || query.startsWith("EXPLAIN", true)) return
 
     val explanations = connection.createStatement().use { statement ->
       try {
--- a/misk-hibernate-testing/src/main/kotlin/misk/jdbc/VitessScaleSafetyChecks.kt
+++ b/misk-hibernate-testing/src/main/kotlin/misk/jdbc/VitessScaleSafetyChecks.kt
@@ -13,6 +13,9 @@ import okhttp3.Request
 import java.sql.Connection
 import java.sql.Timestamp
 import java.util.ArrayDeque
+import java.util.Collections
+import java.util.regex.Matcher
+import java.util.regex.Pattern
 import javax.inject.Singleton
 import javax.sql.DataSource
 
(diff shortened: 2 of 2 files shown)


misk-hibernate-testing/src/main/kotlin/misk/jdbc/ScaleSafetyChecks.kt
```diff
@@ -23,7 +23,7 @@ object ScaleSafetyChecks {
   }
 
   fun checkQueryForTableScan(connection: Connection, query: String) {
-    if (isDml(query)) return
+    if (isDml(query) || query.startsWith("EXPLAIN", true)) return
```
swankjesse, line 26: nice

misk-hibernate-testing/src/main/kotlin/misk/jdbc/VitessScaleSafetyChecks.kt
```diff
@@ -103,6 +106,8 @@ class VitessScaleSafetyChecks(
   }
 
   inner class TableScanDetector : ExtendedQueryExecutionListener() {
+    private val tablePattern = Maps.newConcurrentMap<String, Pattern>()
```
swankjesse, line 109: For what it’s worth, I’ve been encouraging lighter-weight datastructures instead of ConcurrentHashMap
https://github.com/cashapp/misk/pull/1477#discussion_r413028978

swankjesse: In this case I suspect you could avoid caching altogether and get good enough performance

misk-hibernate-testing/src/main/kotlin/misk/jdbc/VitessScaleSafetyChecks.kt
```diff
@@ -139,6 +144,16 @@ class VitessScaleSafetyChecks(
...
     }
+
+    private fun containsTable(query: String, table: String): Boolean {
+      val pattern = tablePattern.computeIfAbsent(table) {
+        val pattern = "\\b$it\\b"
+        Pattern.compile(pattern)
+      }
+
+      val m: Matcher = pattern.matcher(query)
+        return m.find()
```
swankjesse, line 155: why is this indented?

misk-hibernate-testing/src/main/kotlin/misk/jdbc/VitessScaleSafetyChecks.kt
tirsen: Not sure about no need for caching :-) This runs on EVERY query so compiling a regexp pattern might get really expensive...

I don't think we need a full blown CHM though, it's not very concurrent but does SHM have the very convenient `computeIfAbsent` method?

kalfonso: makes sense!. Changed to ```Collections.synchronizedMap(...)```

PR #1485, opened 2020-04-28
Upgrade Misk to use wire 3.2.0 and construct gRPC clients with updated APIs

wesleyk:
To accommodate for updates made in Wire from https://github.com/square/wire/issues/1437


misk/src/main/kotlin/misk/client/GrpcClientModule.kt
```diff
@@ -64,7 +67,8 @@ class GrpcClientModule<T : Service>(
           .baseUrl(baseUrl)
           .build()
 
-      return grpcClient.create(kclass)
+      // There should be *exactly one constructor* that takes in a grpcClient
+      return grpcClientClass.constructors.first().call(grpcClient)
```
wesleyk, line 71: could probably add some sanity checks here if folks would find it useful; mainly that there's:
- One constructor
- One parameter on the constructor
- That parameter implements Service

swankjesse: This should be fine.

swankjesse: Looking at this and how it used to work, I think it was a mistake to take this API out of Wire in the first place. Wire can figure out what the naming pattern is and do the right thing, so the application never sees the implementation, only the interface.

FYI @egorand @oldergod

Egorand: Yeah, let's definitely revisit.

swankjesse: https://github.com/square/wire/pull/1527

PR #1488, opened 2020-04-29
Fix LD user setup

biasedbit:
Pass values to `LSUser` fields rather than keys.


misk-launchdarkly-core/src/test/kotlin/misk/feature/launchdarkly/LaunchDarklyFeatureFlagsTest.kt
```diff
@@ -138,6 +138,52 @@ internal class LaunchDarklyFeatureFlagsTest {
...
+        .name("name value")
+        .avatar("avatar value")
+        .firstName("firstName value")
+        .lastName("lastName value")
+        .country("US")
+        .privateCustom("custom1", "custom1 value")
+        .privateCustom("custom2", "custom2 value")
+        .build()
+
+    // isEqualTo() would be more appropriate, since LDUser overrides equals(). However, failures would offer no
```
ryanhall07, line 181: you can override the error message with assertj

PR #1492, opened 2020-05-01
Remove deprecated Environment from DataSourceConfig.kt

JGulbronson:
This would be a breaking change, since injection would fail if `Deployment` is not provided. That said, it would be a fairly easy change for users to accommodate.


chris-ryan-square:

PR #1493, opened 2020-05-01
Add ExceptionMapper when client leaves before finishing sending request

frojasg:
We record a "client closed request" (499) when a request close the
connection before sending the entire payload.


misk/src/test/kotlin/misk/web/ssl/Http2ConnectivityTest.kt
```diff
@@ -129,6 +139,33 @@ class Http2ConnectivityTest {
...
+    val http1Client = client.newBuilder()
+        .protocols(listOf(Protocol.HTTP_1_1))
+        .writeTimeout(1, TimeUnit.MILLISECONDS)
+        .build()
+    val requestBody = object : RequestBody() {
+      override fun contentType() = "text/plain;charset=utf-8".toMediaType()
+
+      override fun writeTo(sink: BufferedSink) {
+        for (i in 0 until 1024 * 1024) {
+          sink.writeUtf8("impossible\n")
```
frojasg, line 154: I'm not sure if this is the best way to trigger this scenario. I'm open to other ideas.

swankjesse: another option: override contentLength() to be 1, and call `sink.close()` without writing any bytes?

frojasg: I tried that but okhttp wasn't happy with that.

PR #1494, opened 2020-05-01
Plaintext HTTP listener support for h2c

alecholmes:
If the config specifies http2 support, adds h2c listeners to non-TLS listeners.

### DIFF
--- a/misk/src/main/kotlin/misk/web/jetty/JettyService.kt
+++ b/misk/src/main/kotlin/misk/web/jetty/JettyService.kt
@@ -11,6 +11,7 @@ import misk.web.WebSslConfig
 import misk.web.mediatype.MediaTypes
 import okhttp3.HttpUrl
 import org.eclipse.jetty.alpn.server.ALPNServerConnectionFactory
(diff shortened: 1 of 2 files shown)


misk/src/main/kotlin/misk/web/jetty/JettyService.kt
```diff
@@ -63,6 +64,10 @@ class JettyService @Inject internal constructor(
     }
     httpConnectionFactories += HttpConnectionFactory(httpConfig)
 
+    if (webConfig.http2) {
```
swankjesse, line 67: Does this break HTTP/1 over plaintext? It would be better to create a new http2c config?

swankjesse: ... and potentially remove the regular HTTP config there.

alecholmes: HTTP/1 still works since this appends the h2c connection factory instead of replacing the HTTP/1 factory. Given that, why have a second Misk config?

PR #1497, opened 2020-05-04
Require h2c for plaintext gRPC clients

alecholmes:
Use case: clients proxying through a sidecar.

### DIFF
--- a/misk/src/main/kotlin/misk/client/GrpcClientModule.kt
+++ b/misk/src/main/kotlin/misk/client/GrpcClientModule.kt
@@ -7,6 +7,7 @@ import com.squareup.wire.GrpcClient
 import com.squareup.wire.Service
 import misk.inject.KAbstractModule
 import okhttp3.OkHttpClient
+import okhttp3.Protocol
 import javax.inject.Provider
 import javax.inject.Singleton
 import kotlin.reflect.KClass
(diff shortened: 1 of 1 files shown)


misk/src/main/kotlin/misk/client/GrpcClientModule.kt
```diff
@@ -58,10 +59,19 @@ class GrpcClientModule<T : Service, G : T>(
     @Inject private lateinit var httpClientConfigUrlProvider: HttpClientConfigUrlProvider
 
     override fun get(): T {
-      val client = httpClientProvider.get()
       val endpointConfig = httpClientsConfigProvider.get()[name]
       val baseUrl = httpClientConfigUrlProvider.getUrl(endpointConfig)
 
+      // Since gRPC uses HTTP/2, force h2c when calling an unencrypted endpoint
+      val protocols = when {
+        baseUrl.startsWith("http://") -> listOf(Protocol.H2_PRIOR_KNOWLEDGE)
```
swankjesse, line 67: awesome

swankjesse:
@alecholmes does anyone want h2c over TCP? 

I like HTTPS for almost everything always. Then cleartext over domain sockets because we know we aren’t talking to something untrusted.

alecholmes:
> @alecholmes does anyone want h2c over TCP?
> 
> I like HTTPS for almost everything always. Then cleartext over domain sockets because we know we aren’t talking to something untrusted.

No, it's useful for testing but that's about it.

PR #1498, opened 2020-05-05
[Hibernate] Add hasNext method to Pager api

acostama:
This allows a pager to be used similar to an iterator, with `hasNext` and `nextPage`

### DIFF
--- a/misk-hibernate/src/main/kotlin/misk/hibernate/pagination/Pager.kt
+++ b/misk-hibernate/src/main/kotlin/misk/hibernate/pagination/Pager.kt
@@ -8,6 +8,9 @@ import misk.hibernate.Transacter
 interface Pager<T> {
   /** Returns null when there are no more pages. */
   fun nextPage(session: Session): Page<T>?
+
+  /** Returns true when there are more pages to load. */
+  fun hasNext(): Boolean
 }
 
 /** Use a null [initialOffset] to start at the beginning. */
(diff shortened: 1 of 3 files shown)


misk-hibernate/src/main/kotlin/misk/hibernate/pagination/Pager.kt
```diff
@@ -24,12 +27,10 @@ fun <T : DbEntity<T>, R> Pager<T>.listAll(
   transform: (T) -> R
 ): List<R> {
   val results = mutableListOf<R>()
-  var offset: Offset?
```
acostama, line 27: this way, it isn't necessary to track the offset

misk-hibernate/src/main/kotlin/misk/hibernate/pagination/RealPager.kt
```diff
@@ -39,4 +41,6 @@ internal class RealPager<T : DbEntity<T>, Q: Query<T>>(
       contents to false
     }
   }
+
+  override fun hasNext() = hasMore
```
zcross, line 45: Can we pick one phrasing and use it uniformly?

misk-hibernate/src/test/kotlin/misk/hibernate/pagination/RealPagerTest.kt
```diff
@@ -84,6 +84,58 @@ class RealPagerTest {
     assertThat(actualCharacterNames).containsExactlyElementsOf(expectedCharacterNames)
   }
 
+  @Test fun `hasNext always true on first page`() {
```
zcross, line 87: Thanks for the comprehensive tests!

misk-hibernate/src/main/kotlin/misk/hibernate/pagination/RealPager.kt
acostama: changed

PR #1499, opened 2020-05-05
Added an equals function to allow two X500Names to be compared.

janzantinge:
The X500Name objects cannot currently be compared in tests in a nice way. Converting `X500Name` to a data class will allow comparisons in tests like:
`assertThat(actualSubject).isEqualTo(X500Name(mapOf("O" to "Square, Inc.", "OU" to "potato"))`

### DIFF
--- a/misk/src/main/kotlin/misk/security/cert/X500Name.kt
+++ b/misk/src/main/kotlin/misk/security/cert/X500Name.kt
@@ -1,16 +1,21 @@
 package misk.security.cert
 
-class X500Name(private val components: Map<String, String>) {
-  val commonName = get("CN")
-  val organization = get("O")
-  val organizationalUnit = get("OU")
-  val state = get("ST")
-  val locality = get("L")
-  val country = get("C")
-
-  operator fun get(componentName: String) = components[componentName.toUpperCase()]
-
-  fun asMap() = components
+data class X500Name(
+  val commonName: String?,
+  val organizationalUnit: String?,
+  val organization: String?,
+  val locality: String?,
+  val state: String?,
+  val country: String?
+) {
+  constructor(components: Map<String, String>) : this(
+      commonName = components.get("CN"),
+      organizationalUnit = components.get("OU"),
+      organization = components.get("O"),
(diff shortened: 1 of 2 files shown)


misk/src/main/kotlin/misk/security/cert/X500Name.kt
```diff
@@ -12,6 +12,26 @@ class X500Name(private val components: Map<String, String>) {
 
   fun asMap() = components
 
+  override fun equals(other: Any?): Boolean {
```
ryanhall07, line 15: you should also be overriding hashCode

ryanhall07: this should probably just be a `data` class right? would get all of this for free

janzantinge: will look in to making this a data class as discussed

misk/src/main/kotlin/misk/security/cert/X500Name.kt
```diff
@@ -1,16 +1,21 @@
...
-  val organization = get("O")
-  val organizationalUnit = get("OU")
-  val state = get("ST")
-  val locality = get("L")
-  val country = get("C")
-
-  operator fun get(componentName: String) = components[componentName.toUpperCase()]
-
-  fun asMap() = components
+data class X500Name(
```
swankjesse, line 3: Nice

misk/src/main/kotlin/misk/security/cert/X500Name.kt
```diff
@@ -1,16 +1,21 @@
...
+  val commonName: String?,
```
swankjesse, line 4: I wonder if these fields would be better as `cn`, `ou`, `o`, `l`, `st`, `c`. The words are almost irrelevant!

swankjesse: or perhaps the uppercase form, YOLO, `CN`, `OU`, `O` etc.!

PR #1500, opened 2020-05-05
Do not randomly close Unix domain socket connections

alecholmes:
Connections over a Unix domain do not need to be randomly closed. The proxy in front of it should handle this if it's needed, although in our case Istio will load balance properly for us.

### DIFF
--- a/misk/src/main/kotlin/misk/web/interceptors/RebalancingInterceptor.kt
+++ b/misk/src/main/kotlin/misk/web/interceptors/RebalancingInterceptor.kt
@@ -5,6 +5,7 @@ import misk.logging.getLogger
 import misk.random.ThreadLocalRandom
 import misk.web.NetworkChain
 import misk.web.NetworkInterceptor
+import misk.web.SocketAddress
 import misk.web.WebConfig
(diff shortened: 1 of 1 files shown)


misk/src/main/kotlin/misk/web/interceptors/RebalancingInterceptor.kt
```diff
@@ -15,12 +16,19 @@ private val logger = getLogger<RebalancingInterceptor>()
  * When we're deploying and redeploying our pods, we want to make sure that clients rebalance onto
  * the new pods. This randomly closes connections so they will be recreated, naturally balancing
  * connections across pods.
+ *
+ * This does not close Unix domain socket connections.
```
wesleyk, line 20: can we describe why here?

wesleyk: it feels like a skim-type of thing to configure TBH, but seems reasonable to start with it here

ryanhall07:
@alecholmes just quietly rearchitecting your entire network when you're not looking

PR #1506, opened 2020-05-14
Add support for hybrid encryption

yoavamit:
Hybrid encryption is another primitive provided by the Tink library.
It combines the speed and efficiency of symmetric encryption with the convenience of public-key cryptography.

This PR adds the configuration, Guice providers and key managers to handle and supply the `HybridEncrypt` and `HybridDecrypt` interfaces for misk users with the same ease of use we have for other key types.


misk-crypto/src/main/kotlin/misk/crypto/CiphertextFormatter.kt
```diff
@@ -0,0 +1,74 @@
...
+ * This class contains utility functions to serialize the deserialize the ciphertext
+ * generated by the [misk.crypto.CryptoModule] extension functions.
+ */
+internal sealed class CiphertextFormatter {
+  companion object {
+    /**
+     * Serialize the encryption context associated with the ciphertext and wrap both in a ByteString
+     * in the following format:
+     *
+     * `ciphertext length | ciphertext [ encryption context length | encryption context ]`
```
ddz, line 19: When we are decrypting data, how do we differentiate ciphertext without this change (i.e. created before this PR was merged) from ciphertext with this change?

Is there a unit test to make sure that we can always properly handle the ciphertext format before this change is merged?

misk-crypto/src/main/kotlin/misk/crypto/CryptoModule.kt
```diff
@@ -79,52 +102,137 @@ class CryptoModule(
...
+ */
+fun Aead.encrypt(
+  plaintext: ByteString,
+  encryptionContext: Map<String, String>? = null
+) : ByteString {
+  val serializedEncryptionContext = encryptionContext?.let { serialize(it) }
+  val plaintextBytes = plaintext.toByteArray()
+  val encryptedByteString = this.encrypt(plaintextBytes, serializedEncryptionContext)
+  plaintextBytes.fill(0)
+  return CiphertextFormatter.serialize(encryptedByteString, encryptionContext)
```
ddz, line 130: It looks like our ciphertext format wraps the Tink ciphertext format. Can we document what our ciphertext wire format looks like in a markdown file (e.g. [here is Tink's as an example](https://github.com/google/tink/blob/master/docs/WIRE-FORMAT.md)) to help for debugging? For example, it has been helpful to look for Tink's prefix byte, key id hint, and length in a hexdump of data to identify that it is Tink-encrypted data. A complicated example could document what our ciphertext+EncryptionContext format looks like for an envelope-encrypted Tink ciphertext.

misk-crypto/src/main/kotlin/misk/crypto/CryptoModule.kt
```diff
@@ -79,52 +102,137 @@ class CryptoModule(
...
+/**
+ * Extension function for convenient decryption of [ByteString]s
+ * using a map of strings as encryption context.
+ * This function also makes sure that no extra copies of the plaintext data are kept in memory.
+ */
+fun Aead.decrypt(
+  ciphertext: ByteString,
+  encryptionContext: Map<String, String>? = null
+) : ByteString {
+  val ciphertextByteString = CiphertextFormatter.deserialize(ciphertext, encryptionContext)
```
ddz, line 158: What should happen if we call `Aead.decrypt` on a ciphertext that was created with a version of misk-crypto before this PR was merged?

misk-crypto/src/main/kotlin/misk/crypto/InvalidEncryptionContextException.kt
```diff
@@ -0,0 +1,8 @@
+package misk.crypto
+
+import java.security.GeneralSecurityException
+
+class InvalidEncryptionContextException : GeneralSecurityException {
```
yan, line 5: Is this from a different pr?

yoavamit: Yeah, I'll remove it from here

misk-crypto/src/main/kotlin/misk/crypto/KeyManager.kt
```diff
@@ -112,6 +115,39 @@ class DigitalSignatureKeyManager @Inject internal constructor(
...
+/**
+ * Holds a map of every [HybridEncrypt] key name to its corresponding primitive listed
+ * in the configuration for this app.
+ *
+ * Users may use this class to obtain a [HybridEncrypt] object dynamically:
+ * ```
+ * val hybridEncrypt: HybridEncrypt = hybridKeyManager["myHybridKey"]
+ * ```
+ * Note: Hybrid encryption is intentionally divided to 2 separate key managers,
+ * [HybridEncryptKeyManager] and [HybridDecryptKeyManager], so that the public portion of the keyset
```
ddz, line 127: This is good separation to surface, it increases safety and mis-use resistance.

PR #1507, opened 2020-05-16
Allow opting out of scale safety checks

biasedbit:
This option severely impacts test suite performance.

A misk app I was updating went from 20s to run the full test suite to 12m,
while causing a test to systematically fail.

Comparison used the exact same code as `HibernateTestModule` but without
`bindScaleSafetyCheck`.

---

In understand the value of having this (hence the change to make this opt-out, rather than opt-in) but the impact to tests shouldn't have to be this high.

### DIFF
--- a/misk-hibernate-testing/src/main/kotlin/misk/hibernate/HibernateTestingModule.kt
+++ b/misk-hibernate-testing/src/main/kotlin/misk/hibernate/HibernateTestingModule.kt
@@ -34,7 +34,9 @@ class HibernateTestingModule(
   private val qualifier: KClass<out Annotation>,
   private val config: DataSourceConfig? = null,
   private val startUpStatements: List<String> = listOf(),
-  private val shutDownStatements: List<String> = listOf()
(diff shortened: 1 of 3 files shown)


misk-hibernate-testing/src/main/kotlin/misk/hibernate/HibernateTestingModule.kt
```diff
@@ -34,7 +34,8 @@ class HibernateTestingModule(
   private val qualifier: KClass<out Annotation>,
   private val config: DataSourceConfig? = null,
   private val startUpStatements: List<String> = listOf(),
-  private val shutDownStatements: List<String> = listOf()
+  private val shutDownStatements: List<String> = listOf(),
+  private val enableScaleSafetyChecks: Boolean = true
```
swankjesse, line 38: Wanna flip the default? These are broken and should be off til they're fixed.

biasedbit: Makes sense; will do!

PR #1508, opened 2020-05-18
Make create tables service public

zhxnlai:
So that other services can depend on it


swankjesse:
I don’t think this is what you want. When you depend on `TransacterService` you depend on this for free.

swankjesse:
Whoops, I was thinking `SchemaMigratorService`. This is fine!

swankjesse:
One more thought – consider renaming this to DynamoDbService. That opens the door to potentially making table creation optional, for production when creating new tables on-demand could be surprising.

That way when my production code depends on DynamoDbService it’s obvious that I’m depending on the datastore, not how it’s initialized.

wesleyk:
@swankjesse given this is a testing module that didn't seem like a concern to me

mightyguava:
I think it's better to set this up as an `ExternalDependency`. Table creation has no business being part of the service's own dependency graph. Though I know I'm in the minority here.

PR #1512, opened 2020-05-27
Add configurable SQS timeout for jobqueues

wesleyk:


misk-aws/src/main/kotlin/misk/jobqueue/sqs/AwsSqsJobQueueConfig.kt
```diff
@@ -36,5 +36,12 @@ class AwsSqsJobQueueConfig(
   /**
    * Frequency used to import Queue Attributes in milliseconds.
    */
-  val queue_attribute_importer_frequency_ms: Long = 1000
+  val queue_attribute_importer_frequency_ms: Long = 1000,
+
+  /**
+   * Timeout to reach SQS with, not including retries.
```
ryanhall07, line 42: it should be clear this is only the socket timeout. it's also quite strange we are only setting the socket timeout and not the connection and especially the request timeout.

wesleyk: request timeout is actually unbounded by default. I can make the connection timeout configurable too

ryanhall07: we should only have an unbounded request for receiving (with long polls), although even that should be bounded at 20ish seconds. unbounded sending is a good way to tie up resources in a misk app, like http threads

wesleyk:
updated to set upper bound defaults of 20 seconds for socket / request timeouts

PR #1515, opened 2020-05-27
Only apply SQS timeout configs to sending

wesleyk:


misk-aws/src/main/kotlin/misk/jobqueue/sqs/AwsSqsJobQueueConfig.kt
```diff
@@ -56,5 +58,5 @@ class AwsSqsJobQueueConfig(
    * We use the default retry strategy with SQS, which retries 3 times.
    * As a result, your app could potentially spend 3 x this timeout talking to SQS.
    */
-  val sqs_sending_request_timeout_ms: Int = 20000
+  val sqs_sending_request_timeout_ms: Int = ClientConfiguration.DEFAULT_REQUEST_TIMEOUT
```
ryanhall07, line 61: I think we should be opinionated about a sending timeout. My thought is 5s seems reasonable

wesleyk: yeah I'm getting tired of iterating through misk though, I'll come back after I've actually seen this working as expected

PR #1516, opened 2020-05-28
Set more opinionated default timeouts for SQS sending

wesleyk:
Confirmed things work as expected in Raven, so we can set more opinionated defaults now.

### DIFF
--- a/misk-aws/src/main/kotlin/misk/jobqueue/sqs/AwsSqsJobQueueConfig.kt
+++ b/misk-aws/src/main/kotlin/misk/jobqueue/sqs/AwsSqsJobQueueConfig.kt
@@ -46,7 +46,7 @@ class AwsSqsJobQueueConfig(
    * We use the default retry strategy with SQS, which retries 3 times.
    * As a result, your app could potentially spend 3 x this timeout talking to SQS.
    */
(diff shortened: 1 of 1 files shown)


misk-aws/src/main/kotlin/misk/jobqueue/sqs/AwsSqsJobQueueConfig.kt
```diff
@@ -58,5 +58,5 @@ class AwsSqsJobQueueConfig(
    * We use the default retry strategy with SQS, which retries 3 times.
    * As a result, your app could potentially spend 3 x this timeout talking to SQS.
    */
-  val sqs_sending_request_timeout_ms: Int = ClientConfiguration.DEFAULT_REQUEST_TIMEOUT
+  val sqs_sending_request_timeout_ms: Int = 5000
```
ryanhall07, line 61: nit: it's not just sending. also deleting too right?

ryanhall07: maybe just remove `_sending_` and make it clear in the doc it doesn't apply to `receive` which uses long polling, a common practice with receiving

PR #1518, opened 2020-05-29
Specify utf8 on mysql data source config

shellderp:
explicitly set to utf8 for the connection even if server is not configured to default to it. Was required to get emojis working


swankjesse:
🦖

swankjesse:
FYI @tso

PR #1519, opened 2020-05-29
Add with*Timeout methods to ClientNetworkChain

ryanhall07:
This will support a future NetworkInterceptor that updates the outgoing
timeouts based on a propagated deadline.

Currently prototyping this in another project and will move over once I
prove it works (including a test).


ryanhall07:
the eventual idea is to land something like this https://github.com/cashapp/misk/pull/1130/files which is going to require this

PR #1523, opened 2020-06-05
Allow using DbEntity directly

drashkov:
Instead of requiring an app to explicitly mark a DbEntity as sharded or unsharded, allow using DbEntity directly to denote that sharding is not relevant (suitable for TiDB).

### DIFF
--- a/misk-hibernate/src/main/kotlin/misk/hibernate/RealTransacter.kt
+++ b/misk-hibernate/src/main/kotlin/misk/hibernate/RealTransacter.kt
@@ -432,10 +432,7 @@ internal class RealTransacter private constructor(
       check(!readOnly) { "Saving isn't permitted in a read only session." }
       return when (entity) {
         is DbChild<*, *> -> (hibernateSession.save(entity) as Gid<*, *>).id
-        is DbRoot<*> -> hibernateSession.save(entity)
-        is DbUnsharded<*> -> hibernateSession.save(entity)
-        else -> throw IllegalArgumentException(
-            "You need to sub-class one of [DbChild, DbRoot, DbUnsharded]")
+        else -> hibernateSession.save(entity)
       } as Id<T>
     }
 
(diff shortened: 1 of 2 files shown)


misk-hibernate/src/main/kotlin/misk/hibernate/RealTransacter.kt
```diff
@@ -432,10 +432,7 @@ internal class RealTransacter private constructor(
       check(!readOnly) { "Saving isn't permitted in a read only session." }
       return when (entity) {
         is DbChild<*, *> -> (hibernateSession.save(entity) as Gid<*, *>).id
-        is DbRoot<*> -> hibernateSession.save(entity)
-        is DbUnsharded<*> -> hibernateSession.save(entity)
-        else -> throw IllegalArgumentException(
-            "You need to sub-class one of [DbChild, DbRoot, DbUnsharded]")
```
alanpaulin, line 438: Is there any special logic in these 3 classes (or specifically in @DbUnsharded) that we lose by not using one of them?

misk-hibernate/src/main/kotlin/misk/hibernate/RealTransacter.kt
```diff
@@ -432,10 +432,7 @@ internal class RealTransacter private constructor(
...
+        else -> hibernateSession.save(entity)
```
alanpaulin, line 435: Does the transacter know if its in a sharded environment? If it does, then you could leave the old logic if there are shards or keyspaces, and allow `DbEntity` otherwise.

misk-hibernate/src/main/kotlin/misk/hibernate/RealTransacter.kt
drashkov: They're mostly just marker interfaces.

misk-hibernate/src/main/kotlin/misk/hibernate/RealTransacter.kt
drashkov: I don't think it'll make a difference here, hibernate handles them the same way.

PR #1532, opened 2020-06-10
Extract HttpClientConfig from HttpClientEndpointConfig

bezmax:
1. Extract HttpClientConfig from HttpClientEndpointConfig. This will make it more reuseable/mergeable. 
2. Kotlin backwards compatibility code for constructor/accessors. Will remove later when everything is safely migrated.
3. Jackson backwards compatibility code to support both old config file format and new one. Will remove after migrating all configs.
4. Added two new fields to HttpClientConfig to replace `envoy` setting with:
    * `unixSocketFile` - this simplifies code a lot, as currently we have to inject EnvoyClientEndpointProvider just to figure out which socket file to use. Putting it directly into config seems more sane and flexible.
    * `protocols` - this is also to enable migration from envoy, as currently we explicitly limit it to http1.1. 

Majority of code changes are just refactor of constructor/accessors.

### DIFF
--- a/misk/src/main/kotlin/misk/client/HttpClientsConfig.kt
+++ b/misk/src/main/kotlin/misk/client/HttpClientsConfig.kt
@@ -1,61 +1,16 @@
 package misk.client
 
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize
+import com.fasterxml.jackson.databind.type.TypeFactory
 import misk.config.Config
 import misk.security.ssl.CertStoreConfig
 import misk.security.ssl.TrustStoreConfig
+import okhttp3.OkHttpClient
+import java.io.File
+import java.net.URL
 import java.time.Duration
 
-data class HttpClientsConfig(
-  private val defaultConnectTimeout: Duration? = null,
-  private val defaultWriteTimeout: Duration? = null,
-  private val defaultReadTimeout: Duration? = null,
-  private val ssl: HttpClientSSLConfig? = null,
-  private val defaultPingInterval: Duration? = null,
-  private val defaultCallTimeout: Duration? = null,
-  private val endpoints: Map<String, HttpClientEndpointConfig> = mapOf()
-) : Config {
-  /** @return The [HttpClientEndpointConfig] for the given client, populated with defaults as needed */
-  operator fun get(clientName: String): HttpClientEndpointConfig {
-    // TODO(mmihic): Cache, proxy, etc
-    val endpointConfig = endpoints[clientName] ?: throw IllegalArgumentException(
-        "no client configuration for endpoint $clientName")
-
-    return HttpClientEndpointConfig(
-        endpointConfig.url,
-        endpointConfig.envoy,
-        connectTimeout = endpointConfig.connectTimeout ?: defaultConnectTimeout,
-        writeTimeout = endpointConfig.writeTimeout ?: defaultWriteTimeout,
-        readTimeout = endpointConfig.readTimeout ?: defaultReadTimeout,
-        pingInterval = endpointConfig.pingInterval ?: defaultPingInterval,
-        callTimeout = endpointConfig.callTimeout ?: defaultCallTimeout,
-        ssl = endpointConfig.ssl ?: ssl
-    )
-  }
-
-  /**
-   * Returns a default config with no url/envoy config specified.
-   * Used to build up endpoint config for clients dynamically.
-   */
-  fun getDefault(): HttpClientEndpointConfig {
-    return HttpClientEndpointConfig(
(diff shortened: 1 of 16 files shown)


bezmax:
@swankjesse Added few more commits addressing your feedback and some new code. Please re-stamp.

1. Added additional backwards compatibility layer to support parsing old config files. The plan is to have this in place, then migrate projects to new config and finally remove this layer. There's also a test for parsing in old format, and another for parsing in new format.
2. Sneaked into this PR some minor downstream changes:
* HttpClientConfig now has new field `unixSocketFile`. Plan is to replace `envoy` with this property. This would simplify code a lot, as currently we have to inject `EnvoyClientEndpointProvider` just to figure out which socket file to use. Putting it directly into config seems more sane and flexible.  
* HttpClientConfig now also has field `protocols`. This is also to enable migration from `envoy`, as currently we explicitly limit it to `http1.1`. Therefore `envoy` will be replaced with these 2 properties.

misk/src/main/kotlin/misk/client/HttpClientsConfig.kt
```diff
@@ -1,34 +1,96 @@
...
 
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize
+import com.fasterxml.jackson.databind.type.TypeFactory
 import misk.config.Config
 import misk.security.ssl.CertStoreConfig
 import misk.security.ssl.TrustStoreConfig
+import java.io.File
 import java.time.Duration
 
+data class BackwardsCompatibleEndpointConfig(
```
swankjesse, line 11: Nice.

misk/src/main/kotlin/misk/client/HttpClientsConfig.kt
```diff
@@ -1,34 +1,96 @@
...
+    //Common fields
+  val url: String? = null,
+  val envoy: HttpClientEnvoyConfig? = null,
+
+    //Legacy fields
```
swankjesse, line 16: These are all deprecated? Maybe worth noting that in a comment

misk/src/main/kotlin/misk/client/HttpClientsConfig.kt
```diff
@@ -1,34 +1,96 @@
...
+  val maxRequestsPerHost: Int = 32,
+  val maxIdleConnections: Int = 100,
+  val keepAliveDuration: Duration = Duration.ofMinutes(5),
+  val ssl: HttpClientSSLConfig? = null,
+
+    //New fields
+  val clientConfig: HttpClientConfig? = null
+)
+
+class BackwardsCompatibleEndpointConfigConverter : com.fasterxml.jackson.databind.util.Converter<BackwardsCompatibleEndpointConfig, HttpClientEndpointConfig> {
```
swankjesse, line 32: Neat

misk/src/main/kotlin/misk/client/HttpClientsConfig.kt
```diff
@@ -67,9 +129,108 @@ data class HttpClientEndpointConfig(
...
+        callTimeout,
+        maxRequests,
+        maxRequestsPerHost,
+        maxIdleConnections,
+        keepAliveDuration,
+        ssl
+    )
 )
 
+data class HttpClientEndpointConfig(
```
swankjesse, line 168: Sweet

misk/src/test/kotlin/misk/config/HttpClientsConfigBackwardsCompatibilityTest.kt
```diff
@@ -0,0 +1,75 @@
...
+import misk.resources.ResourceLoader
+import misk.web.WebConfig
+import okio.ByteString
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.jupiter.api.Test
+import java.io.File
+import java.time.Duration
+import kotlin.test.assertFailsWith
+
+class HttpClientsConfigBackwardsCompatibilityTest {
```
swankjesse, line 19: Wonderful. Thanks for doing this

misk/src/main/kotlin/misk/client/HttpClientsConfig.kt
bezmax: Tough question. This is a backwards compatibility layer, so the layer itself is not deprecated (yet) and will never be - it will just be deleted when we do not care about backwards compatibility anymore. I think I won't mark them `@Deprecated`, as to me `@Deprecated` sounds as "do not use", while nobody will be using any of these classes anyway. I would have made them private/internal, but Jackson does not like that idea.

bezmax:
Force pushed rebase onto master.

PR #1534, opened 2020-06-11
Allow no queueing in Jetty

ryanhall07:
The QueuedThreadPool first submits all requests to the queue and the
thread pool pulls from the queue. With this impl, there is no way to
disable queueing on the server.

If config sets queue size to 0, a different impl is used which direclty
hands the reques to the ThreadPool and rejects the request if max
threads are in use.

### DIFF
--- a/misk/src/main/kotlin/misk/web/WebConfig.kt
+++ b/misk/src/main/kotlin/misk/web/WebConfig.kt
@@ -36,7 +36,14 @@ data class WebConfig(
   /** Maximum number of threads in Jetty's thread pool. */
   val jetty_max_thread_pool_size: Int = 200,
 
-  /** Maximum number of items in the queue for Jetty's thread pool. */
+  /**
+   * Maximum number of items in the queue for Jetty's thread pool.
+   *
+   * If 0, no queueing is used and requests are directly handed off to the thread pool. If a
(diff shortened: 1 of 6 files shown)


misk/src/main/kotlin/misk/web/WebConfig.kt
```diff
@@ -36,7 +36,14 @@ data class WebConfig(
...
   val jetty_max_thread_pool_size: Int = 200,
 
-  /** Maximum number of items in the queue for Jetty's thread pool. */
+  /**
+   * Maximum number of items in the queue for Jetty's thread pool.
+   *
+   * If 0, no queueing is used and requests are directly handed off to the thread pool. If a
+   * thread is not available (i.e max threads in use) the request is rejected. Unfortunately Jetty
+   * rejects requests by closing the socket instead of returning a 429. This can lead to confusing
+   * EOFExceptions for the client.
```
swankjesse, line 45: 😻 nice docs

ryanhall07:
@swankjesse had to introduce a new `MeasuredThreadPool` interface so we can monitor both `QueuedThreadPool` and `ExecutorThreadPool`. Found in 482bb6097e757d22304eba81a69d3ba2b1a37095. Nothing too controversial, so I'm going to merge and can follow up

PR #1539, opened 2020-06-16
Add support for host-based http client settings

bezmax:
Second part of HttpClientEndpointConfig overhaul (first part [here](https://github.com/cashapp/misk/pull/1532)). This is the last piece required to start the migration off `envoy` style configs towards the host-based configs with `unixSocketFile` explicitly specified.
This change is still mostly backwards compatible with original config format. 

1. Made all of the fields in `HttpClientsConfig` nullable. This is to avoid accidentally overwriting default settings by no specifying any override (thus using constructor defaults, which would in fact count as an override).
2. Added `hostConfigs` attribute to `HttpClientsConfig`. This allows overriding settings based on host regular expressions. Uses `LinkedHashMap` as the order of definition defines the order the overrides are evaluated/applied.
3. Removed all the `default*` attributes from `HttpClientsConfig` constructor. The default settings can now be specified as `.*` host pattern. The backwards compatibility layer does exactly this, when it sees `default*` specified - it creates a synthetic "hosts" entry with `.*` in it.
4. `HttpClientFactory` uses the `hostConfigs` to build final set of settings. For `url` based endpoints it checks regular expressions for all matches. For `envoy` based endpoints, only hosts with `.*` in them are applied - this is to provide backwards compatibility for `envoy` endpoints with default configuration settings.

You can look at example of how new http client settings look like [here](https://github.com/cashapp/misk/blob/79d986aa8a1caa04c7a52199be499340bafefc69/misk/src/test/resources/http_clients_config_new-testing.yaml). The order of application priority is:
1. Hardcoded defaults.
2. Host-pattern based configuration (`hosts` section in Yaml), by order of declaration (lowest priority first).
3. Endpoint-based configuration. This would allow to have 2 differently named clients using some variation of settings.

Further planned steps (in future PRs):
1. Migrate all projects to not use `envoy` endpoints.
2. Remove support for `envoy` endpoints from Misk.
3. Remove all backwards-compatibility code.

### DIFF
--- a/misk/src/main/kotlin/misk/client/HttpClientFactory.kt
+++ b/misk/src/main/kotlin/misk/client/HttpClientFactory.kt
@@ -7,6 +7,7 @@ import okhttp3.Dispatcher
 import okhttp3.Interceptor
 import okhttp3.OkHttpClient
 import okhttp3.Protocol
+import java.io.File
 import java.net.Proxy
 import java.time.Duration
 import java.util.concurrent.TimeUnit
@@ -73,11 +74,9 @@ class HttpClientFactory @Inject constructor(
       builder.proxy(Proxy.NO_PROXY)
     }
 
-    config.clientConfig.protocols?.let {
-      builder.protocols(
-          config.clientConfig.protocols.map { Protocol.get(it) }
-      )
-    }
+    config.clientConfig.protocols
+        ?.map { Protocol.get(it) }
+        ?.let { builder.protocols(it) }
 
     config.envoy?.let {
       builder.socketFactory(
--- a/misk/src/main/kotlin/misk/client/HttpClientFactory.kt
+++ b/misk/src/main/kotlin/misk/client/HttpClientFactory.kt
@@ -8,11 +8,25 @@ import okhttp3.Interceptor
 import okhttp3.OkHttpClient
 import okhttp3.Protocol
 import java.net.Proxy
+import java.time.Duration
 import java.util.concurrent.TimeUnit
 import javax.inject.Inject
 import javax.inject.Provider
 import javax.inject.Singleton
 import javax.net.ssl.X509TrustManager
+
+
+private object Defaults {
+  /*
+    Copied from okhttp3.ConnectionPool, as it does not provide "use default" option
+   */
+  val maxIdleConnections = 5
+
+  /*
+    Copied from okhttp3.ConnectionPool, as it does not provide "use default" option
+   */
+  val keepAliveDuration = Duration.ofMinutes(5)
+}
 
 @Singleton
 class HttpClientFactory @Inject constructor(
@@ -52,7 +66,7 @@ class HttpClientFactory @Inject constructor(
     }
 
     config.clientConfig.unixSocketFile?.let {
-      builder.socketFactory(UnixDomainSocketFactory(it))
+      builder.socketFactory(UnixDomainSocketFactory(File(it)))
       // No DNS lookup needed since we're just sending the request over a socket.
       builder.dns(NoOpDns)
       // Proxy config not supported
@@ -62,7 +76,6 @@ class HttpClientFactory @Inject constructor(
     config.clientConfig.protocols?.let {
       builder.protocols(
           config.clientConfig.protocols.map { Protocol.get(it) }
-              ?: listOf(Protocol.HTTP_1_1) //Safe default
       )
     }
 
@@ -78,13 +91,17 @@ class HttpClientFactory @Inject constructor(
     }
 
     val dispatcher = Dispatcher()
-    dispatcher.maxRequests = config.clientConfig.maxRequests
-    dispatcher.maxRequestsPerHost = config.clientConfig.maxRequestsPerHost
+    config.clientConfig.maxRequests?.let { maxRequests ->
+      dispatcher.maxRequests = maxRequests
+    }
+    config.clientConfig.maxRequestsPerHost?.let { maxRequestsPerHost ->
+      dispatcher.maxRequestsPerHost = maxRequestsPerHost
+    }
     builder.dispatcher(dispatcher)
 
     val connectionPool = ConnectionPool(
-        config.clientConfig.maxIdleConnections,
-        config.clientConfig.keepAliveDuration.toMillis(),
+        config.clientConfig.maxIdleConnections ?: Defaults.maxIdleConnections,
+        (config.clientConfig.keepAliveDuration?:Defaults.keepAliveDuration).toMillis(),
         TimeUnit.MILLISECONDS)
     builder.connectionPool(connectionPool)
 
--- a/misk/src/main/kotlin/misk/client/HttpClientsConfig.kt
+++ b/misk/src/main/kotlin/misk/client/HttpClientsConfig.kt
@@ -1,115 +1,75 @@
 package misk.client
 
+import com.fasterxml.jackson.annotation.JsonAlias
 import com.fasterxml.jackson.databind.annotation.JsonDeserialize
-import com.fasterxml.jackson.databind.type.TypeFactory
 import misk.config.Config
+import misk.logging.getLogger
 import misk.security.ssl.CertStoreConfig
 import misk.security.ssl.TrustStoreConfig
 import java.io.File
+import java.net.URL
 import java.time.Duration
 
-data class BackwardsCompatibleEndpointConfig(
-    //Common fields
-  val url: String? = null,
-  val envoy: HttpClientEnvoyConfig? = null,
-
-    //Legacy fields
-  val connectTimeout: Duration? = null,
-  val writeTimeout: Duration? = null,
-  val readTimeout: Duration? = null,
-  val pingInterval: Duration? = null,
-  val callTimeout: Duration? = null,
-  val maxRequests: Int = 128,
-  val maxRequestsPerHost: Int = 32,
-  val maxIdleConnections: Int = 100,
-  val keepAliveDuration: Duration = Duration.ofMinutes(5),
-  val ssl: HttpClientSSLConfig? = null,
-
-    //New fields
-  val clientConfig: HttpClientConfig? = null
-)
-
-class BackwardsCompatibleEndpointConfigConverter : com.fasterxml.jackson.databind.util.Converter<BackwardsCompatibleEndpointConfig, HttpClientEndpointConfig> {
-  override fun getInputType(typeFactory: TypeFactory) =
-      typeFactory.constructType(BackwardsCompatibleEndpointConfig::class.java)
-
-  override fun getOutputType(typeFactory: TypeFactory) =
-      typeFactory.constructType(HttpClientEndpointConfig::class.java)
-
-  override fun convert(value: BackwardsCompatibleEndpointConfig) =
-      //http://%s.%s.gns.square
-      if (value.clientConfig == null)
-        HttpClientEndpointConfig(
-            url = value.url,
-            envoy = value.envoy,
-            clientConfig = HttpClientConfig(
-                value.connectTimeout,
-                value.writeTimeout,
-                value.readTimeout,
-                value.pingInterval,
-                value.callTimeout,
-                value.maxRequests,
-                value.maxRequestsPerHost,
-                value.maxIdleConnections,
-                value.keepAliveDuration,
-                value.ssl
-            )
-        )
-      else
-        HttpClientEndpointConfig(
-            url = value.url,
-            envoy = value.envoy,
-            clientConfig = value.clientConfig
-        )
-}
-
+@JsonDeserialize(converter = BackwardsCompatibleClientsConfigConverter::class)
 data class HttpClientsConfig(
-  val defaultConnectTimeout: Duration? = null,
-  val defaultWriteTimeout: Duration? = null,
-  val defaultReadTimeout: Duration? = null,
-  val ssl: HttpClientSSLConfig? = null,
-  val defaultPingInterval: Duration? = null,
-  val defaultCallTimeout: Duration? = null,
-  @JsonDeserialize(contentConverter = BackwardsCompatibleEndpointConfigConverter::class)
+  @JsonAlias("hosts")
+  //Need to retain ordering, hence LinkedHashMap
+  val hostConfigs: LinkedHashMap<String, HttpClientConfig> = linkedMapOf(),
   val endpoints: Map<String, HttpClientEndpointConfig> = mapOf()
 ) : Config {
-  /** @return The [HttpClientEndpointConfig] for the given client, populated with defaults as needed */
-  operator fun get(clientName: String): HttpClientEndpointConfig {
-    // TODO(mmihic): Cache, proxy, etc
-    val endpointConfig = endpoints[clientName] ?: throw IllegalArgumentException(
-        "no client configuration for endpoint $clientName")
+  init {
+    validatePatterns()
+  }
 
-    val clientConfig = endpointConfig.clientConfig
-    return HttpClientEndpointConfig(
-        url = endpointConfig.url,
-        envoy = endpointConfig.envoy,
-        clientConfig = clientConfig.copy(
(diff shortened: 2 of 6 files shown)


misk/src/main/kotlin/misk/client/HttpClientFactory.kt
```diff
@@ -74,13 +87,17 @@ class HttpClientFactory @Inject constructor(
     }
 
     val dispatcher = Dispatcher()
-    dispatcher.maxRequests = config.clientConfig.maxRequests
-    dispatcher.maxRequestsPerHost = config.clientConfig.maxRequestsPerHost
+    config.clientConfig.maxRequests?.let { maxRequests ->
```
bezmax, line 90: This was not-nullable before. Therefore now requires a `?.let`.

misk/src/main/kotlin/misk/client/HttpClientFactory.kt
```diff
@@ -74,13 +87,17 @@ class HttpClientFactory @Inject constructor(
...
+    }
+    config.clientConfig.maxRequestsPerHost?.let { maxRequestsPerHost ->
+      dispatcher.maxRequestsPerHost = maxRequestsPerHost
+    }
     builder.dispatcher(dispatcher)
 
     val connectionPool = ConnectionPool(
-        config.clientConfig.maxIdleConnections,
-        config.clientConfig.keepAliveDuration.toMillis(),
+        config.clientConfig.maxIdleConnections ?: Defaults.maxIdleConnections,
```
bezmax, line 99: Also was not-nullable.

misk/src/main/kotlin/misk/client/HttpClientsConfig.kt
```diff
@@ -1,115 +1,75 @@
...
-import com.fasterxml.jackson.databind.type.TypeFactory
 import misk.config.Config
+import misk.logging.getLogger
 import misk.security.ssl.CertStoreConfig
 import misk.security.ssl.TrustStoreConfig
 import java.io.File
+import java.net.URL
 import java.time.Duration
 
-data class BackwardsCompatibleEndpointConfig(
```
bezmax, line 11: Extracted most of "backwards compatibility" stuff into separate file, to make it less messy.

misk/src/main/kotlin/misk/client/HttpClientsConfig.kt
```diff
@@ -1,115 +1,75 @@
...
-        )
-      else
-        HttpClientEndpointConfig(
-            url = value.url,
-            envoy = value.envoy,
-            clientConfig = value.clientConfig
-        )
-}
-
+@JsonDeserialize(converter = BackwardsCompatibleClientsConfigConverter::class)
```
bezmax, line 13: The backwards compatibility converter moved from per-endpoint, to `HttpClientsConfig` given I want to also deprecate the `default*` attributes.

misk/src/main/kotlin/misk/client/HttpClientsConfig.kt
```diff
@@ -1,115 +1,75 @@
...
-  val defaultPingInterval: Duration? = null,
-  val defaultCallTimeout: Duration? = null,
-  @JsonDeserialize(contentConverter = BackwardsCompatibleEndpointConfigConverter::class)
+  @JsonAlias("hosts")
+  //Need to retain ordering, hence LinkedHashMap
+  val hostConfigs: LinkedHashMap<String, HttpClientConfig> = linkedMapOf(),
   val endpoints: Map<String, HttpClientEndpointConfig> = mapOf()
 ) : Config {
+  init {
+    validatePatterns()
```
bezmax, line 21: This is to fail fast if configuration has an incorrect regex specified.

misk/src/main/kotlin/misk/client/HttpClientsConfig.kt
```diff
@@ -1,115 +1,75 @@
...
+    val endpointConfig = requireNotNull(endpoints[clientName]) {
+        "no client configuration for endpoint $clientName"
+    }
+
+    val allMatchingConfigs = sequence {
+      yield(httpClientConfigDefaults)
+      yieldAll(
+          endpointConfig.url?.let { url ->
+            findUrlMatchingConfigs(url)
+          } ?: findWildcardConfigs()
```
bezmax, line 46: Wildcard configurations are applied even to `envoy` endpoints.

misk/src/main/kotlin/misk/client/HttpClientsConfig.kt
```diff
@@ -1,115 +1,75 @@
...
-        clientConfig = HttpClientConfig(
-            connectTimeout = defaultConnectTimeout,
-            writeTimeout = defaultWriteTimeout,
-            readTimeout = defaultReadTimeout,
-            pingInterval = defaultPingInterval,
-            callTimeout = defaultCallTimeout,
-            ssl = ssl
-        )
+  private fun findWildcardConfigs() =
+      hostConfigs.filter { (k, _) -> k == ".*" }.values
```
bezmax, line 59: This is strictly for backwards compatibility to support `envoy` endpoints, for which we do not have URL to match against.

misk/src/main/kotlin/misk/client/HttpClientsConfig.kt
```diff
@@ -125,45 +85,30 @@ data class HttpClientConfig(
   val readTimeout: Duration? = null,
   val pingInterval: Duration? = null,
   val callTimeout: Duration? = null,
-  val maxRequests: Int = 128,
```
bezmax, line 128: Given that this data structure is now used for both defaults and overrides - all fields need to be null. Otherwise, merging an override on top of default would always override all non-nullable fields.

misk/src/main/kotlin/misk/client/HttpClientsConfig.kt
```diff
@@ -125,45 +85,30 @@ data class HttpClientConfig(
...
-  val maxRequestsPerHost: Int = 32,
-  val maxIdleConnections: Int = 100,
-  val keepAliveDuration: Duration = Duration.ofMinutes(5),
+  val maxRequests: Int? = null,
+  val maxRequestsPerHost: Int? = null,
+  val maxIdleConnections: Int? = null,
+  val keepAliveDuration: Duration? = null,
   val ssl: HttpClientSSLConfig? = null,
-  val unixSocketFile: File? = null,
+  val unixSocketFile: String? = null,
```
bezmax, line 93: Changed to String, as there's a weird Jackson issue, where it's incorrectly deserializing `\u0000` character into `File`, but correctly deserializes same exact value into `String`. Covered by test.

misk/src/main/kotlin/misk/client/HttpClientsConfig.kt
```diff
@@ -125,45 +85,30 @@ data class HttpClientConfig(
...
-        pingInterval,
-        callTimeout,
-        maxRequests,
-        maxRequestsPerHost,
-        maxIdleConnections,
-        keepAliveDuration,
-        ssl
+fun HttpClientConfig.applyDefaults(other: HttpClientConfig) =
+    HttpClientConfig(
+        connectTimeout = this.connectTimeout ?: other.connectTimeout,
```
bezmax, line 99: Added names to avoid accidental order changes.

misk/src/test/kotlin/misk/config/HttpClientsConfigBackwardsCompatibilityTest.kt
```diff
@@ -23,7 +23,7 @@ class HttpClientsConfigBackwardsCompatibilityTest {
                 connectTimeout = Duration.ofSeconds(41),
                 readTimeout = Duration.ofSeconds(42),
                 writeTimeout = Duration.ofSeconds(43)
-            )
+            ).applyDefaults(HttpClientsConfig.httpClientConfigDefaults)
```
bezmax, line 26: Wasn't sure if it makes sense to just hardcode the default values in the test, or apply global defaults. Might revisit.

misk/src/test/resources/http_clients_config_old-testing.yaml
```diff
@@ -1,3 +1,5 @@
+defaultReadTimeout: PT60S
```
bezmax, line 1: Test backwards compatibility of original `default*` attributes.

misk/src/main/kotlin/misk/client/HttpClientFactory.kt
```diff
@@ -8,11 +8,25 @@ import okhttp3.OkHttpClient
...
 import javax.inject.Inject
 import javax.inject.Singleton
 import javax.net.ssl.X509TrustManager
 
+
+private object Defaults {
+  /*
+    Copied from okhttp3.ConnectionPool, as it does not provide "use default" option
+   */
+  val maxIdleConnections = 5
```
jjestrel, line 22: I'm not sure I follow why we want these to differ from the old defaults in `HttpClientConfig`

misk/src/main/kotlin/misk/client/HttpClientsConfig.kt
```diff
@@ -1,115 +1,75 @@
...
+  @JsonAlias("hosts")
+  //Need to retain ordering, hence LinkedHashMap
+  val hostConfigs: LinkedHashMap<String, HttpClientConfig> = linkedMapOf(),
   val endpoints: Map<String, HttpClientEndpointConfig> = mapOf()
 ) : Config {
+  init {
+    validatePatterns()
+  }
+
+  private fun validatePatterns() = try {
```
jjestrel, line 24: nit: put `validatePatterns` underneath `operator fun get`

misk/src/test/kotlin/misk/config/HttpClientsConfigBackwardsCompatibilityTest.kt
```diff
@@ -45,14 +46,16 @@ class HttpClientsConfigBackwardsCompatibilityTest {
...
-            url = "https://google.com/",
+            url = "https://test.google.com/",
             clientConfig = HttpClientConfig(
                 connectTimeout = Duration.ofSeconds(31),
                 readTimeout = Duration.ofSeconds(32),
                 writeTimeout = Duration.ofSeconds(33),
-                unixSocketFile = File("file.socket"),
-                protocols = listOf("http1", "http2", "http3")
-            )
+                unixSocketFile = "\u0000egress.sock",
```
jjestrel, line 54: nice

misk/src/main/kotlin/misk/client/HttpClientsConfigBackwardsCompatibility.kt
```diff
@@ -0,0 +1,146 @@
+package misk.client
+
+import com.fasterxml.jackson.annotation.JsonAlias
+import com.fasterxml.jackson.databind.type.TypeFactory
+import java.time.Duration
+
+data class BackwardsCompatibleEndpointConfig(
+    //Common fields
```
jjestrel, line 8: nit: spacing and missing ` ` after `//`

misk/src/main/kotlin/misk/client/HttpClientsConfigBackwardsCompatibility.kt
```diff
@@ -0,0 +1,146 @@
...
+
+    //Shared fields
+  val endpoints: Map<String, BackwardsCompatibleEndpointConfig> = mapOf(),
+
+    //New fields
+  @JsonAlias("hosts")
+  val hostConfigs: LinkedHashMap<String, HttpClientConfig> = linkedMapOf()
+)
+
+class BackwardsCompatibleClientsConfigConverter : com.fasterxml.jackson.databind.util.Converter<BackwardsCompatibleClientsConfig, HttpClientsConfig> {
```
jjestrel, line 45: nice

misk/src/main/kotlin/misk/client/HttpClientsConfigBackwardsCompatibility.kt
```diff
@@ -0,0 +1,146 @@
...
+    val endpoints = value.endpoints.mapValues { (_, v) -> convert(v) }
+
+    return HttpClientsConfig(
+        hostConfigs = hostConfigs,
+        endpoints = endpoints
+    )
+  }
+
+  private fun convert(value: BackwardsCompatibleEndpointConfig) =
+      //http://%s.%s.gns.square
```
jjestrel, line 91: do we need this?

misk/src/main/kotlin/misk/client/HttpClientsConfig.kt
```diff
@@ -125,45 +85,30 @@ data class HttpClientConfig(
...
-        writeTimeout,
-        readTimeout,
-        pingInterval,
-        callTimeout,
-        maxRequests,
-        maxRequestsPerHost,
-        maxIdleConnections,
-        keepAliveDuration,
-        ssl
+fun HttpClientConfig.applyDefaults(other: HttpClientConfig) =
```
jjestrel, line 97: Can we make this `internal`?

misk/src/main/kotlin/misk/client/HttpClientFactory.kt
```diff
@@ -8,11 +8,25 @@ import okhttp3.OkHttpClient
...
 import java.net.Proxy
+import java.time.Duration
 import java.util.concurrent.TimeUnit
 import javax.inject.Inject
 import javax.inject.Singleton
 import javax.net.ssl.X509TrustManager
 
+
+private object Defaults {
+  /*
```
jjestrel, line 19: Should these be `/**`?

misk/src/main/kotlin/misk/client/HttpClientFactory.kt
bezmax: This is the defaults that were not present in original, as they were not nullable. I still have original values set up by always prepending `misk.client.HttpClientsConfig.Companion#getHttpClientConfigDefaults`, and this object has all the original defaults. The defaults you see here are kind of lowest-layer, in case the given HttpClientConfig somehow has nulls there (which should never happen).

misk/src/main/kotlin/misk/client/HttpClientsConfig.kt
bezmax: I explicitly did not want to put this into `get`, as I want the class to fail-fast on construction, instead of waiting until somebody actually requests a value with `get` and then gets an error.

jjestrel: Agreed that makes sense. I mean underneath (declared after) not in to `operator fun get`

misk/src/main/kotlin/misk/client/HttpClientsConfigBackwardsCompatibility.kt
bezmax: Fixed the space after `//`, but won't fix the spacing before as that's how our IntelliJ rules reformat it.

misk/src/main/kotlin/misk/client/HttpClientsConfigBackwardsCompatibility.kt
bezmax: No, my mistake, removed.

misk/src/main/kotlin/misk/client/HttpClientsConfig.kt
bezmax: Not sure. I probably prefer not making it internal given that HttpClientConfig is not internal and this function is basically composition of those configs. So in theory, if somebody wants to use HttpClientConfig directly - they will find this function useful.

misk/src/main/kotlin/misk/client/HttpClientFactory.kt
bezmax: Not sure. IMHO `/**` is meant to be a part of Javadoc, and describe the meaning of the field and it's usage. In this case, the comment is more about meta-information, it's almost a TODO or an explanatory note, as in "If by any chance those fields become public in ConnectionPool - remove them from here".

PR #1540, opened 2020-06-16
Support OkHttp interceptors

alecholmes:
Misk `NetworkInterceptor`s end up mapping to OkHttp `NetworkInterceptor` but there is currently no way to add an OkHttp `Interceptor`. This adds support for that for typed HTTP clients. GRPC and untyped clients aren't supported in this PR but I'll draft a possibility in a comment.

### DIFF
--- a/misk/src/main/kotlin/misk/client/HttpClientFactory.kt
+++ b/misk/src/main/kotlin/misk/client/HttpClientFactory.kt
@@ -4,12 +4,13 @@ import misk.security.ssl.SslContextFactory
 import misk.security.ssl.SslLoader
 import okhttp3.ConnectionPool
 import okhttp3.Dispatcher
+import okhttp3.Interceptor
 import okhttp3.OkHttpClient
 import okhttp3.Protocol
-import java.io.File
 import java.net.Proxy
 import java.util.concurrent.TimeUnit
 import javax.inject.Inject
+import javax.inject.Provider
 import javax.inject.Singleton
 import javax.net.ssl.X509TrustManager
 
@@ -20,6 +21,9 @@ class HttpClientFactory @Inject constructor(
 ) {
   @com.google.inject.Inject(optional = true)
   lateinit var envoyClientEndpointProvider: EnvoyClientEndpointProvider
+
+  @com.google.inject.Inject(optional = true)
+  var okhttpInterceptors: Provider<List<Interceptor>>? = null
 
   /** Returns a client initialized based on `config`. */
   fun create(config: HttpClientEndpointConfig): OkHttpClient {
@@ -83,6 +87,10 @@ class HttpClientFactory @Inject constructor(
         config.clientConfig.keepAliveDuration.toMillis(),
         TimeUnit.MILLISECONDS)
     builder.connectionPool(connectionPool)
+
+    okhttpInterceptors?.let {
+      builder.interceptors().addAll(it.get())
+    }
 
     return builder.build()
   }
--- a/misk/src/main/kotlin/misk/client/HttpClientsConfig.kt
+++ b/misk/src/main/kotlin/misk/client/HttpClientsConfig.kt
@@ -112,6 +112,9 @@ data class HttpClientsConfig(
         )
     )
   }
+
+  /** Names of configured endpoints, all of which can be fetched using [get] */
+  fun endpointNames(): Set<String> = endpoints.keys
 }
 
 data class HttpClientSSLConfig(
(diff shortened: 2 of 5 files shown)


misk/src/main/kotlin/misk/client/ClientInterceptor.kt
```diff
@@ -0,0 +1,18 @@
...
+
+  /**
+   * This interface is used with Guice multibindings. Register instances by calling `multibind()`
+   * in a `KAbstractModule`:
+   *
+   * ```
+   * multibind<ClientInterceptors.Factory>().to<MyFactory>()
+   * ```
+   */
+  interface Factory {
```
alecholmes, line 15: If we wanted to support GRPC interceptors too, could have separate factory types:

```
  interface Factory {
    fun create(action: ClientAction): Interceptor?
  }

  interface GRPCFactory {
    fun create(grpcClientClass: KClass<*>): Interceptor?
  }
```

misk/src/main/kotlin/misk/client/HttpClientFactory.kt
```diff
@@ -56,10 +55,12 @@ class HttpClientFactory @Inject constructor(
     }
 
     config.clientConfig.protocols?.let {
-      builder.protocols(
-          config.clientConfig.protocols.map { Protocol.get(it) }
-              ?: listOf(Protocol.HTTP_1_1) //Safe default
```
alecholmes, line 61: This is a bug that already exists. `?:` only evaluates for nullable types, which the result of line 60 is not.

bezmax: I have fixed this in my other PR (currently in review).

bezmax: Although, given that your PR does not have any other changes in this file, I think it's better to not commit this. This is not actually a bug, just a leftover from refactoring that I forgot to remove. The `protocols` will never be empty, it's either null or has something in it.

misk/src/main/kotlin/misk/client/ClientInvocationHandler.kt
```diff
@@ -39,6 +39,7 @@ internal class ClientInvocationHandler(
   private val clientName: String,
   retrofit: Retrofit,
   okHttpTemplate: OkHttpClient,
+  clientInterceptorFactories: Provider<List<ClientInterceptors.Factory>>,
```
bezmax, line 42: Is it enough to do in this class only? I think I have seen places using HttpClientFactory directly, without doing anything with ClientInvocationHandler. Maybe it's better to inject it directly there?

misk/src/main/kotlin/misk/client/ClientInvocationHandler.kt
```diff
@@ -39,6 +39,7 @@ internal class ClientInvocationHandler(
...
   networkInterceptorFactories: Provider<List<ClientNetworkInterceptor.Factory>>,
   applicationInterceptorFactories: Provider<List<ClientApplicationInterceptor.Factory>>,
```
ryanhall07, line 44: I get why you're doing this, but it doesn't seem satisfying that we have `ClientApplicationInterceptor.Factory` for retrofit and `ClientInterceptors.Factory` for okhttp.  I almost would argue we should remove the abstraction and call them `OkHttpInterceptor` and `RetrofitInterceptor`

alecholmes:
I reworked this after @ryanhall07's comment, which keeps things nice and simple. Also added a test.

misk/src/main/kotlin/misk/client/HttpClientsConfig.kt
```diff
@@ -112,6 +112,9 @@ data class HttpClientsConfig(
         )
     )
   }
+
+  /** Names of configured endpoints, all of which can be fetched using [get] */
+  fun endpointNames(): Set<String> = endpoints.keys
```
alecholmes, line 117: Unrelated but useful.

misk/src/main/kotlin/misk/client/HttpClientFactory.kt
```diff
@@ -21,6 +22,8 @@ class HttpClientFactory @Inject constructor(
   @com.google.inject.Inject(optional = true)
   lateinit var envoyClientEndpointProvider: EnvoyClientEndpointProvider
 
+  @Inject lateinit var okhttpInterceptors: Provider<List<Interceptor>>
```
ryanhall07, line 25: multibinder provides a `List`?

alecholmes:
TIL it does.

PR #1541, opened 2020-06-17
Only update connection metrics in a background thread

ryanhall07:
Updating metrics requires Prometheus client object monitors to ensure
synchronized updates to the metrics. For a highly threaded jetty server
(+1000 threads), this can cause a lot of contention.

There is no reason to update metrics immediately, since they are
refreshed in the background anyways and metrics inherently have a lag.
Now we just queue up open/closed changes and process them in the
background thread before updating metrics.

### DIFF
--- a/misk/src/main/kotlin/misk/web/jetty/ConnectionListener.kt
+++ b/misk/src/main/kotlin/misk/web/jetty/ConnectionListener.kt
@@ -4,6 +4,7 @@ import org.eclipse.jetty.io.Connection
 import org.eclipse.jetty.util.component.AbstractLifeCycle
 import java.util.Collections
 import java.util.concurrent.ConcurrentHashMap
+import java.util.concurrent.ConcurrentLinkedQueue
 import java.util.concurrent.atomic.AtomicLong
 
 internal class ConnectionListener(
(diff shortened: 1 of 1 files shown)


ryanhall07:
a thread dump showed a lot of this

```
"jetty-thread-187" #187 prio=5 os_prio=0 cpu=3562.21ms elapsed=1041.14s tid=0x00007f90200ed800 nid=0xe0 waiting for monitor entry  [0x00007f8fe68e7000]
   java.lang.Thread.State: BLOCKED (on object monitor)
	at io.prometheus.client.TimeWindowQuantiles.insert(TimeWindowQuantiles.java:36)
	- waiting to lock <merged>(a io.prometheus.client.TimeWindowQuantiles)
	at io.prometheus.client.Summary$Child.observe(Summary.java:283)
	at misk.prometheus.PrometheusHistogram.record(PrometheusHistogram.kt:10)
	at misk.web.jetty.ConnectionListener.refreshMetrics(ConnectionListener.kt:58)
	at misk.web.jetty.ConnectionListener.onClosed(ConnectionListener.kt:31)
	at org.eclipse.jetty.io.AbstractConnection.onClosed(AbstractConnection.java:237)
	at org.eclipse.jetty.io.AbstractConnection.onClose(AbstractConnection.java:229)
	at org.eclipse.jetty.server.HttpConnection.onClose(HttpConnection.java:519)
	at org.eclipse.jetty.io.ssl.SslConnection.onClose(SslConnection.java:366)
	at org.eclipse.jetty.io.SelectorManager.connectionClosed(SelectorManager.java:345)
	at org.eclipse.jetty.io.ManagedSelector$DestroyEndPoint.run(ManagedSelector.java:967)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(java.base@11.0.6/ThreadPoolExecutor.java:1128)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(java.base@11.0.6/ThreadPoolExecutor.java:628)
	at java.lang.Thread.run(java.base@11.0.6/Thread.java:834)
```

misk/src/main/kotlin/misk/web/jetty/ConnectionListener.kt
```diff
@@ -18,24 +19,28 @@ internal class ConnectionListener(
...
   override fun onOpened(connection: Connection) {
-    connections.add(ConnectionKey(connection))
-    metrics.activeConnections.labels(*labels).inc()
-    metrics.acceptedConnections.labels(*labels).inc()
+    openedConnections.add(ConnectionKey(connection))
   }
 
   override fun onClosed(connection: Connection) {
-    // Force a refresh so that we gather the final stats on the connection
-    refreshMetrics()
```
ryanhall07, line 31: to clarify I don't think the problem is really Prometheus locking to update metrics. The real problem is we were iterating over every connection (maybe 1000s) every time a connection is closed. That's a lot of contention for the same metrics.

ryanhall07:
@swankjesse good catch.

in fact the refresh happens in a guava service background thread, which could easily shutdown if there is some exception thrown by the refresh! 

I think I could just redo this by being smarter on how we update the metrics on connection close. no need to update the metrics for _all_ connections when we remove a single connection

ryanhall07:
reverted and tried again with https://github.com/cashapp/misk/pull/1542

PR #1544, opened 2020-06-27
Use attributes in Fake Feature Flags.

minostro:
Added the attributes to all override value variances

### DIFF
--- a/misk-feature-testing/src/main/kotlin/misk/feature/testing/FakeFeatureFlags.kt
+++ b/misk-feature-testing/src/main/kotlin/misk/feature/testing/FakeFeatureFlags.kt
@@ -10,6 +10,7 @@ import misk.feature.FeatureFlags
 import misk.feature.FeatureService
 import misk.feature.fromSafeJson
 import misk.feature.toSafeJson
+import java.util.PriorityQueue
 import java.util.concurrent.ConcurrentHashMap
 import javax.inject.Inject
 import javax.inject.Provider
@@ -19,26 +20,25 @@ import javax.inject.Singleton
  * In-memory test implementation of [FeatureFlags] that allows flags to be overridden.
  */
 @Singleton
-class FakeFeatureFlags @Inject constructor(private val moshi : Provider<Moshi>) : AbstractIdleService(),
-    FeatureFlags,
-    FeatureService,
-    DynamicConfig {
-
+class FakeFeatureFlags @Inject constructor(
+  val moshi : Provider<Moshi>
+) : AbstractIdleService(), FeatureFlags, FeatureService, DynamicConfig {
   companion object {
     const val KEY = "fake_dynamic_flag"
+    val defaultAttributes = Attributes()
   }
 
   override fun startUp() {}
   override fun shutDown() {}
 
-  private val overrides = ConcurrentHashMap<MapKey, Any>()
+  private val overrides = ConcurrentHashMap<MapKey, PriorityQueue<MapValue>>()
 
   override fun getBoolean(feature: Feature, key: String, attributes: Attributes): Boolean {
     return getOrDefault(feature, key, false)
   }
 
   override fun getInt(feature: Feature, key: String, attributes: Attributes): Int =
-      get(feature, key) as? Int ?: throw IllegalArgumentException(
+      get(feature, key, attributes) as? Int ?: throw IllegalArgumentException(
           "Int flag $feature must be overridden with override() before use")
 
   override fun getString(feature: Feature, key: String, attributes: Attributes): String =
@@ -52,7 +52,7 @@ class FakeFeatureFlags @Inject constructor(private val moshi : Provider<Moshi>) 
     attributes: Attributes
   ): T {
     @Suppress("unchecked_cast")
-    return getOrDefault(feature, key, clazz.enumConstants[0]) as T
+    return getOrDefault(feature, key, clazz.enumConstants[0], attributes) as T
   }
 
   override fun <T> getJson(
@@ -61,7 +61,7 @@ class FakeFeatureFlags @Inject constructor(private val moshi : Provider<Moshi>) 
     clazz: Class<T>,
     attributes: Attributes
   ): T {
-    val jsonFn = get(feature, key) as? Function0<*> ?: throw IllegalArgumentException(
+    val jsonFn = get(feature, key, attributes) as? Function0<*> ?: throw IllegalArgumentException(
         "JSON flag $feature must be overridden with override() before use: ${get(feature, key)}")
     // The JSON is lazily provided to handle the case where the override is provided by the
     // FakeFeatureFlagModule and the Moshi instance cannot be accessed inside the module.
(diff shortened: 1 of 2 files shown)


adrw:
What would be the use case for passing in attributed to a fake? For our usages, we only use overrideBoolean... etc type methods to induce the flag condition we're looking to test a class against.

minostro:
I have a use case where I am using `Attributes` and depending on those values the app behaves differently. The mechanism you describe works, but I think it's just a stop gap because `Attributes` were not considered in the fake lib.

adrw:
My impression is that the attributes are for use in targeting rules in LaunchDarkly, I'm curious what use case your service would have in introspecting the attributes directly. For example, in notifications we include attributes for a given notificationId in the LaunchDarkly flag such that we can turn on traffic on a per notificationId basis. The only place the service code cares about the attributes is in the feature flag call since the targetting rules in LaunchDarkly encompass the decision logic using the attributes.

minostro:
>My impression is that the attributes are for use in targeting rules in LaunchDarkly, I'm curious what use case your service would have in introspecting the attributes directly.

I am using attributes to provide a json payload. This is based on the name of the job that's about to run.

>The only place the service code cares about the attributes is in the feature flag call since the targetting rules in LaunchDarkly encompass the decision logic using the attributes.

I think my case is similar. The code that consumes the json payload from LaunchDarkly is generic and it passes down the json payload to each job.

I think this is similar when you add an override at the key level. You still want to test that your code is passing the key and that it would do something different if launch-darkly returns a different value than the default one. In my case, I am not using `key` but `attributes`.

minostro:
@swankjesse Let me share with you a test case to make sure we are on the same page:

```
    // Can be overridden
    subject.overrideJson(FEATURE, JsonFeature("overrideJson"))
    assertThat(subject.getJson<JsonFeature>(FEATURE, TOKEN)).isEqualTo(JsonFeature("overrideJson"))
```

This must be true because the override works at the `FEATURE` level, since there is not override for the key `TOKEN`.

```
    // Can be overridden with specific keys
    subject.overrideKeyJson(FEATURE, "bad-joker", JsonFeature("overrideKeyJson"), Attributes())
    assertThat(subject.getJson<JsonFeature>(FEATURE, "bad-joker"))
        .isEqualTo(JsonFeature("overrideKeyJson"))
```

This must be true because we are overriding the key, so instead of getting `JsonFeature("overrideJson")` we must get `JsonFeature("overrideKeyJson")`.

```
    assertThat(subject.getJson<JsonFeature>(FEATURE, "bad-joker", Attributes(mapOf("no-override" to "for-me"))))
        .isEqualTo(JsonFeature("overrideKeyJson"))
```
Is the expectation that this call returns `JsonFeature("overrideKeyJson")` and not `JsonFeature("overrideJson")`? The current implementation returns `JsonFeature("overrideJson")`, and I think your point is that that is incorrect.

minostro:
@swankjesse PTAL

misk-feature-testing/src/main/kotlin/misk/feature/testing/FakeFeatureFlags.kt
```diff
@@ -77,76 +77,149 @@ class FakeFeatureFlags @Inject constructor(private val moshi : Provider<Moshi>)
...
+    //
+    // 1. Return feature level override if there is no override at the key level.
+    // 2. if there is an override at the key level, then:
+    //   2.1 Return the override value associated with the attribute override that contains
+    //       all the attributes of the given attributes.
+    //   2.2 If there is no match, return the override value defined at the key level.
+    val overrideMapValues = overrides[MapKey(feature, key)]
+        ?: return overrides[MapKey(feature)]?.let { it.first().value }
+
+    val overrideMapValuesCopy = PriorityQueue(overrideMapValues)
```
swankjesse, line 108: is this intending to defend against races? Or just so you can mutate?

misk-feature-testing/src/main/kotlin/misk/feature/testing/FakeFeatureFlags.kt
```diff
@@ -77,76 +77,149 @@ class FakeFeatureFlags @Inject constructor(private val moshi : Provider<Moshi>)
...
+
+  /**
+   * Data class that holds the override value provided for the given [feature, key, attributes].
+   */
+  private data class MapValue(
+    val order: Int = 1, //Order is used to pick the latest provided map value.
+    val attributes: Attributes = defaultAttributes,
+    val value: Any
+  ) : Comparable<MapValue> {
+    override fun compareTo(other: MapValue) = when (attributes.text.size) {
```
swankjesse, line 220: I find this control flow really awkward to follow because attributes.text.size shows up in 3 places. Simpler to go member-by-member?
```
val attributeSizeCompare = -attributes.text.size.compareTo(other.attributes.text.size)
if (attributeSizeCompare != 0) return attributeSizeCompare

return -order.compareTo(other.order)
```

misk-feature-testing/src/test/kotlin/misk/feature/testing/FakeFeatureFlagsTest.kt
```diff
@@ -89,6 +112,38 @@ internal class FakeFeatureFlagsTest {
...
+        .isEqualTo(JsonFeature("bad-joker"))
+    assertThat(subject.getJson<JsonFeature>(FEATURE, "joker", sleepyBadJokerAttributes))
+        .isEqualTo(JsonFeature("bad-joker"))
+    //Provides the key level override when there is no match on attributes
+    assertThat(subject.getJson<JsonFeature>(FEATURE, "joker", goodJokerAttributes))
+        .isEqualTo(JsonFeature("joker"))
+  }
+
+  @Test
+  fun `provides the latest override in case two or more attributes are a match`() {
```
swankjesse, line 136: nice

misk-feature-testing/src/test/kotlin/misk/feature/testing/FakeFeatureFlagsTest.kt
```diff
@@ -135,6 +190,7 @@ internal class FakeFeatureFlagsTest {
 
   enum class Dinosaur {
     PTERODACTYL,
-    TYRANNOSAURUS
+    TYRANNOSAURUS,
+    TALARURUS
```
swankjesse, line 194: NICE
Screen Shot 2020-07-15 at 14 17 36

misk-feature-testing/src/main/kotlin/misk/feature/testing/FakeFeatureFlags.kt
minostro: the latter, so I can do the `poll`. I didn't want to mutate the original data structure so that's why I am making a copy of it. The other option that I see is to mutate the original data structure, put the elements in a stack, and then when I am done put them back. (too much work without any much benefits).

Thoughts?

misk-feature-testing/src/main/kotlin/misk/feature/testing/FakeFeatureFlags.kt
minostro: will change.

PR #1545, opened 2020-06-30
Bump datadog version to 0.56.0

isafan:
Bumping datadog version to latest 
https://github.com/DataDog/dd-trace-java/releases/tag/v0.56.0


isafan:
Breaking because of TracingLoggingEnhancer, we can no longer cast to DDSpan. See https://github.com/DataDog/dd-trace-java/releases/tag/v0.51.0

PR #1547, opened 2020-07-02
Add LogRateLimiter

isafan:
Added a LogRateLimiter but it is not being used yet. It will be added to the RequestLoggingInterceptor to rate limit on log emission for WebActions.

### DIFF
--- a/misk-actions/src/main/kotlin/misk/web/interceptors/LogRequestResponse.kt
+++ b/misk-actions/src/main/kotlin/misk/web/interceptors/LogRequestResponse.kt
@@ -10,7 +10,22 @@ package misk.web.interceptors
  *
  * If arguments and responses may include sensitive information, it is expected that the toString()
  * methods of these objects will redact it.
+ *
+ * Rate limiting is used to sample the number of requests logged. The value specified is the
+ * log events per sec allotted per caller per action.
+ *
+ * By default we set rate limiting for both successes and errors to 1 log event per sec,
+ * enough to show things are happening without sending too many logs.
+ *
+ * If you would like to turn off rate limiting and emit all logs, set rateLimiting and/or
+ * errorRateLimiting to 0.
  */
 @Retention(AnnotationRetention.RUNTIME)
-@Target(AnnotationTarget.FUNCTION)
-annotation class LogRequestResponse(val sampling: Double, val includeBody: Boolean)
+@Target(AnnotationTarget.FUNCTION, AnnotationTarget.CLASS)
+annotation class LogRequestResponse(
+  // TODO(isabel): Remove sampling and includeBody. Will be replaced with rate limiting
+  val sampling: Double,
+  val includeBody: Boolean,
+  // Currently unused
+  val rateLimiting: Long = 1,
+  val errorRateLimiting: Long = 1)
--- a/misk/src/main/kotlin/misk/ratelimit/RateLimiter.kt
+++ b/misk/src/main/kotlin/misk/ratelimit/RateLimiter.kt
@@ -1,0 +1,121 @@
+package misk.ratelimit
+
+import com.google.common.base.Ticker
+import misk.concurrent.Sleeper
+import java.time.Duration
+import java.util.concurrent.TimeUnit
+import java.util.concurrent.atomic.AtomicLong
+import javax.inject.Inject
+
+/**
+ * A deterministic testable rate limiter that uses two variables:
+ *
+ * * Permits per second. This is the application's configured rate. We express as a per-second rate
+ *   but use it as a time-between-permits. For example, 250 permits per second is a permit every
+ *   4 milliseconds. This may be zero, in which case all acquire attempts will return false.
+ *
+ * * Window size. If the application specified 250 permits per second, that doesn't specify how many
+ *   permits can be returned at once. An implementation could strictly return 1 permit every 4
+ *   milliseconds, or batches of 1000 permits every 4 seconds. This class hard codes the window size
+ *   to 1 second. Small windows shrink batch sizes which is inefficient; large windows grow batch
+ *   sizes which is bursty. This class uses 1 second to balance latency and throughput.
+ *
+ * The implementation tracks a future timestamp that permits are consumed until.
+ *
+ * This class is similar to Guava's rate limiter. Unlike Guava's rate limiter this class is testable
+ * by application code using the rate limiter. It also has very predictable behavior because its
+ * internal mechanisms are simpler than Guava's.
+ */
+class RateLimiter private constructor(
+  private val ticker: Ticker,
+  private val sleeper: Sleeper
+) {
+  @Volatile var permitsPerSecond: Long = 0L
+
+  /**
+   * The nanoTime that we've consumed all permits through. This is at most [windowSizeNs] + current
+   * nanoTime.
+   */
+  private val atomicAllocatedUntil = AtomicLong(ticker.read())
+
+  /** The size of our window where we can borrow bytes from the future. */
+  private var windowSizeNs = TimeUnit.SECONDS.toNanos(1L)
+
+  /**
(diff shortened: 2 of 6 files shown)


misk/src/main/kotlin/misk/logging/RateLimiter.kt
```diff
@@ -0,0 +1,120 @@
+package misk.logging
```
mightyguava, line 1: This is copied from clientsync? Where are the tests?

misk/src/main/kotlin/misk/logging/LogBucketId.kt
```diff
@@ -0,0 +1,13 @@
+package misk.logging
+
+import kotlin.reflect.KClass
+
+data class LogBucketId(
```
mightyguava, line 5: nit: this doesn't need to be its own file

misk-actions/src/main/kotlin/misk/web/interceptors/LogRequestResponse.kt
```diff
@@ -12,5 +12,10 @@ package misk.web.interceptors
  * methods of these objects will redact it.
  */
 @Retention(AnnotationRetention.RUNTIME)
-@Target(AnnotationTarget.FUNCTION)
-annotation class LogRequestResponse(val sampling: Double, val includeBody: Boolean)
+@Target(AnnotationTarget.FUNCTION, AnnotationTarget.CLASS)
+annotation class LogRequestResponse(
+  val rateLimiting: Long = 0,
```
mightyguava, line 17: documentation? and I believe we agreed that `rateLimiting` should be nonzero? I think 10 qps is probably ok

misk/src/main/kotlin/misk/logging/LogRateLimiter.kt
```diff
@@ -0,0 +1,56 @@
...
+ * [LogRequestResponse] annotation, which is the number of buckets per pod per service.
+ */
+@Singleton
+class LogRateLimiter @Inject constructor(
+  private val rateLimiterFactory: RateLimiter.Factory
+) {
+  private val liveConfigRef =
+    AtomicReference(LiveConfig(
+      rateLimiters = ConcurrentHashMap<LogBucketId, RateLimiter>(),
+      errorRateLimiters = ConcurrentHashMap<LogBucketId, RateLimiter>()
```
mightyguava, line 23: Why not just make `error` a boolean field on the `LogBucketId`? Then you can just have a single ConcurrentHashMap. And you just need a single `tryAcquire` method

misk/src/main/kotlin/misk/logging/LogRateLimiter.kt
```diff
@@ -0,0 +1,56 @@
...
+    ))
+
+  fun tryAcquireOnError(bucketId: LogBucketId, permits: Long): Boolean {
```
mightyguava, line 26: Why is `permits` an argument? You return a boolean here, so why would you ever acquire more than 1?

I think there might be performance tradeoffs w.r.t. acquiring a batch of permits, but I don't think that's relevant here. cc: @swankjesse

misk/src/main/kotlin/misk/logging/LogRateLimiter.kt
```diff
@@ -0,0 +1,56 @@
...
+    val liveConfig: LiveConfig = liveConfigRef.get()
+    val rateLimiter = liveConfig.errorRateLimiters.getOrPut(bucketId) {
+      val bucketQps =
+        bucketId.actionClass.java.getAnnotation(LogRequestResponse::class.java).errorRateLimiting
+      if (bucketQps == 0L) {
```
mightyguava, line 31: This check should live outside the `getOrPut()`, since you aren't creating a rate limiter

misk-actions/src/main/kotlin/misk/web/interceptors/LogRequestResponse.kt
isafan: Ah yeah we did, updating

misk/src/main/kotlin/misk/logging/RateLimiter.kt
isafan: Yes I carried it over. Not using Guava RateLimiter because it was easier to test with this, simpler to interpret, and accomplishes what we need. Hope that is alright. 

Just added the tests as well.

misk/src/main/kotlin/misk/logging/LogRateLimiter.kt
isafan: Good point, adding.

misk/src/main/kotlin/misk/logging/LogRateLimiter.kt
isafan: I had passed in more than one permit in LogRateLimiterTest, but I agree for logging we should only ever tryAcquire on 1 bucket

misk/src/main/kotlin/misk/logging/RateLimiter.kt
acostama: this could probably go in its own package or module. This was we can change clientsync to use the misk one and its available for other services as well

misk/src/main/kotlin/misk/logging/LogRateLimiter.kt
```diff
@@ -0,0 +1,48 @@
...
+import javax.inject.Singleton
+import kotlin.reflect.KClass
+
+/**
+ * Holds rate limiters for logging success and error responses. There is a rate limiter for every
+ * action and service that calls it, for both success and error. The rate limiters are created
+ * according to the value set in [LogRequestResponse] annotation.
+ */
+@Singleton
+class LogRateLimiter @Inject constructor(
```
mightyguava, line 18: This should be in the same package (or even same file) as the logging interceptors. The `logging` package shouldn't have a dependency on `web`.

misk-actions/src/main/kotlin/misk/web/interceptors/LogRequestResponse.kt
isafan: Added documentation

misk-actions/src/main/kotlin/misk/web/interceptors/LogRequestResponse.kt
```diff
@@ -10,7 +10,20 @@ package misk.web.interceptors
  *
  * If arguments and responses may include sensitive information, it is expected that the toString()
  * methods of these objects will redact it.
+ *
+ * Rate limiting is used to sample the number of requests logged. The value specified is the
+ * buckets/sec allotted per service per pod. By default we set rate limiting for both successes
```
mightyguava, line 15: I think `log events per sec` makes more sense to say?

misk-actions/src/main/kotlin/misk/web/interceptors/LogRequestResponse.kt
```diff
@@ -10,7 +10,20 @@ package misk.web.interceptors
...
+ * and errors to 10 buckets/sec, enough to show things are happening without sending too many logs.
```
mightyguava, line 16: Maybe let's start off more aggressive and see if people complain. Let's do 1 log event / sec

misk-actions/src/main/kotlin/misk/web/interceptors/LogRequestResponse.kt
mightyguava, line 15: isn't this per caller per action?

isafan: Yeah sure

isafan: Ah yeah it is, per caller (service), per action, per pod

PR #1548, opened 2020-07-06
Extract Metrics module

acostama:
This will allow service modules to include metrics without depending on all of misk. This is useful for testing or fakes in services


swankjesse:
FYI @isafan @mightyguava

PR #1550, opened 2020-07-07
Expose direct access to the JPA CriteriaBuilder

swankjesse:
So it isn't necessary to update Misk whenever a long-tail constraint
isn't built-in.

### DIFF
--- a/misk-hibernate/src/main/kotlin/misk/hibernate/ReflectionQueryFactory.kt
+++ b/misk-hibernate/src/main/kotlin/misk/hibernate/ReflectionQueryFactory.kt
@@ -758,8 +758,12 @@ internal class ReflectionQuery<T : DbEntity<T>>(
     }
   }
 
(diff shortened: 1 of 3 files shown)


misk-hibernate/src/main/kotlin/misk/hibernate/ReflectionQueryFactory.kt
```diff
@@ -758,9 +758,8 @@ internal class ReflectionQuery<T : DbEntity<T>>(
     }
   }
 
-  internal fun addConstraint(predicateFactory: PredicateFactory): ReflectionQuery<T> {
-    constraints.add(predicateFactory)
-    return this
+  override fun addConstraint(block: PredicateFactory) {
```
becassel, line 761: I think this is the culprit for the failing Unit tests?

swankjesse: Yep. Needed to return the right result. Fixed!

PR #1554, opened 2020-07-08
sqs: each job handler gets its own trace

mightyguava:
We want each sqs job handled to have its own trace. This previously
worked properly while we used jaeger, assuming that by submitting tasks
to the handler thread pool does not propagate any existing trace.

However, now that we switched to datadog, it actually automagically uses
code generation to propagate traces across asynchronous execution
boundaries. So now the receiver scheduling thread becomes one giant long
trace that includes every job.

See https://github.com/DataDog/dd-trace-java/tree/master/dd-java-agent/instrumentation/java-concurrent/src/main/java/datadog/trace/instrumentation/java/concurrent

This change forces a new root span for each job.


ryanhall07:
datadog scares me

PR #1555, opened 2020-07-08
Add misk-embedded with Embeddedmisk

acostama:
From the misk-embedded README:
Sometimes we want to embed a Misk Service in another JVM project (such as another Misk service,
but not necessarily) while avoiding classpath collisions on dependencies. For example, some services
may use Hibernate 4 and want to embed Misk applications that use Hibernate 5.

Misk embedded encapsulates a Misk service and its classpath for use in another service.


misk-embedded-sample/embedded/build.gradle
```diff
@@ -0,0 +1,32 @@
...
+  dependencies {
+    classpath dep.shadowJarPlugin
+  }
+}
+
+apply plugin: 'com.github.johnrengelman.shadow'
+
+dependencies {
+  implementation project(':misk-embedded-sample:service')
+  // somehow exclude this from the above dependencies
```
zcross, line 11: Does this need to be addressed now?

misk-embedded-sample/embedded/build.gradle
```diff
@@ -0,0 +1,32 @@
...
+  // compile project(':misk-embedded-sample:client')
+}
+
+shadowJar {
+  baseName = 'misk.embedded.sample.embedded.EmbeddedSampleService'
+  classifier = ''
+  archiveVersion = ''
+  dependencies {
+    exclude(project(':misk-embedded-sample:client'))
+    // TODO: implicitly exclude transitive runtime deps of client
```
zcross, line 21: Is this a TODO (of some explicit action to take), or an explanation of something implicit in this current configuration?

misk-embedded/README.md
```diff
@@ -0,0 +1,66 @@
...
+
+
+Requirements
+------------
+
+The embedded project should have a client/ and service/ submodule: 
+ * The service/ submodule should depend on the client/ submodule
+ * The client/ submodule MUST NOT depend on the service/ submodule.
+
+The consuming project's depends on the embedded project's client/ submodule.
```
zcross, line 18: s/project's/project

misk-embedded/README.md
```diff
@@ -0,0 +1,66 @@
...
+    ```
+    val myService = EmbeddedMisk.create<MyService>()
+    ```
+
+   
+Warning!
+--------
+
+The embedded service will be loaded in an isolated class loader. This means it won't collide with
+classes in the consuming service, even if they share the same name. Static singletons will not be
```
zcross, line 47: Good call out re: singletons

misk-embedded/src/test/kotlin/misk/embedded/EmbeddedMiskTest.kt
```diff
@@ -0,0 +1,67 @@
+package misk.embedded
+
+import misk.embedded.sample.embedded.EmbeddedSampleService
+import misk.embedded.sample.embedded.SampleEvent
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.jupiter.api.Test
+
+class EmbeddedMiskTest {
```
zcross, line 8: It would be nice to see what a service to service action invocation looks like when it's from one test service to an embedded test service. I'm wondering if we're going to end up using test HTTP clients and exercising our stack of request interceptors, or just bypass that with direct invocations to action classes in the embedded service's dependency graph.

misk-embedded/src/main/kotlin/misk/embedded/EmbeddedMisk.kt
```diff
@@ -0,0 +1,82 @@
...
+   * This is the Application Loader and loads all host project classes and classes depended on by
+   * the embedding interface.
+   */
+  private var hostLoader: ClassLoader = EmbeddedMisk::class.java.classLoader
+
+  inline fun <reified T : Any> create() = create(T::class.java)
+
+  fun <T : Any> create(embeddingInterfaceType: Class<T>): T {
+    val classLoader = createIsolatedLoader(embeddingInterfaceType)
+    val injector = createInjector(classLoader, "${embeddingInterfaceType.name}InjectorFactory")
```
zcross, line 34: Seems like the injectorFactoryClassName construction can be done in createInjector if this just takes `embeddingInterfaceType`?

misk-embedded/src/main/kotlin/misk/embedded/EmbeddedMisk.kt
```diff
@@ -0,0 +1,82 @@
...
+
+    return injector.getInstance(embeddingInterfaceType)!!
+  }
+
+  private fun <T : Any> copyAndExtractJar(embeddingInterfaceType: Class<T>): File {
+    val extractedJarPath = File.createTempFile(embeddingInterfaceType.name, "jar")
+    extractedJarPath.deleteOnExit()
```
zcross, line 41: If `deleteOnExit` only deletes when the JVM is terminated, do we want to memoize all of this work in case multiple instances of an `EmbeddedMisk` service are spun up sequentially in the lifetime of one JVM (e.g., in a series of integration tests)?

zcross: Or is that taken care of by this resolution existing somewhere in an eagerly loaded singleton?

misk-embedded-sample/client/build.gradle
```diff
@@ -0,0 +1,18 @@
...
+  compile dep.guice
+
+  testCompile dep.assertj
+  testCompile dep.junitApi
+  testCompile dep.junitEngine
+  testCompile dep.junitParams
+  testCompile project(':misk-testing')
+}
+
+afterEvaluate { project ->
```
swankjesse, line 13: delete?

misk-embedded-sample/embedded/build.gradle
swankjesse: In follow-up!

misk-embedded-sample/embedded/build.gradle
swankjesse: Something we need to implement. Might get hairy because we might have to traverse transitive dependencies.

misk-embedded-sample/service/build.gradle
```diff
@@ -0,0 +1,21 @@
...
+  implementation project(':misk-inject')
+
+  testImplementation dep.assertj
+  testImplementation dep.junitApi
+  testImplementation dep.junitEngine
+  testImplementation dep.junitParams
+  testImplementation project(':misk-testing')
+}
+
+afterEvaluate { project ->
```
swankjesse, line 16: delete

misk-embedded-sample/service/src/main/kotlin/misk/embedded/sample/embedded/RealEmbeddedSampleService.kt
```diff
@@ -0,0 +1,40 @@
...
+    hibernate.save(event.data)
+  }
+
+  override fun loadPersistedValues() = hibernate.loadAll()
+
+  companion object {
+    val globalCounter = AtomicInteger()
+  }
+
+  /** Local to the service submodule. */
```
swankjesse, line 25: oh cool!

misk-embedded/README.md
```diff
@@ -0,0 +1,66 @@
...
+
+
+Usage
+-----
+
+1. Create an embedding interface in your service's client/ submodule.  
+
+2. Implement that embedding interface in your service's service/ submodule.  
+
+3. Implement an `InjectorFactory` in your service's service/ submodule. The InjectorFactory class
```
swankjesse, line 28: One more constraint here . . . the InjectorFactory must bind the embedding interface.

misk-embedded/src/main/kotlin/misk/embedded/EmbeddedMisk.kt
swankjesse: We should extract to a deterministic location (sha256 of .jar contents?), and avoid the work of double-extraction. This is difficult!
 - need to extract to a different place, then move. Otherwise concurrently-running tests will fail because they’ll see a partially-extracted .jar
 - need to garbage collect the old extracted objects somehow.

Perhaps this as the target name:
```
  /tmp/embedded-misk/20200708-<sha>.jar
```

Then we delete `.jar` files whose file name is less than yesterday?

```
val yesterdayPrefix = formatDate(today - 48 hours)
for (file in embeddedMiskDir.list()) {
  if (file.name < yesterdayPrefix) file.delete()
}
```

Not worth doing until we’ve proven that this thing is useful for NewsWriter. But then it’ll definitely be worth doing!

misk-embedded/src/main/kotlin/misk/embedded/RoutingClassLoader.kt
```diff
@@ -0,0 +1,25 @@
+package misk.embedded
+
+/**
+ * When a child of [RoutingClassLoader] is loading a class, [RoutingClassLoader] will be tried
+ * first. If the class should be handled by the child, [RoutingClassLoader] delegates down. On the
+ * other hand, if the class should not be handled by the child, [RoutingClassLoader] delegates the
+ * call to its parent loader.
+ */
+abstract class RoutingClassLoader(
```
swankjesse, line 9: This is great!

misk-embedded/src/test/kotlin/misk/embedded/EmbeddedMiskTest.kt
swankjesse: I hope we can make direct invocations work. Step debugging is great!

(And if we need interceptors for stuff like auth, we don’t need HTTP for that. We can use decorator patterns.)

misk-embedded/README.md
swankjesse: (please add a note on the docs!)

misk-embedded-sample/embedded/build.gradle
acostama: Real todo for us, will clarify comment

PR #1558, opened 2020-07-15
Don't set gRPC response headers until the action completes

swankjesse:
That way if the action throws an exception we don't send any gRPC
response headers.

### DIFF
--- a/misk/src/test/kotlin/misk/grpc/GrpcConnectivityTest.kt
+++ b/misk/src/test/kotlin/misk/grpc/GrpcConnectivityTest.kt
@@ -5,6 +5,7 @@ import com.squareup.protos.test.grpc.HelloReply
 import com.squareup.protos.test.grpc.HelloRequest
 import com.squareup.wire.Service
 import com.squareup.wire.WireRpc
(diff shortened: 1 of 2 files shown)


misk/src/test/kotlin/misk/grpc/GrpcConnectivityTest.kt
```diff
@@ -67,21 +72,62 @@ class GrpcConnectivityTest {
...
+      assertThat(response.headers["grpc-encoding"]).isEqualTo("identity")
+      assertThat(response.body!!.contentType()).isEqualTo("application/grpc".toMediaType())
 
-    for (i in 0 until response.headers.size) {
-      println("${response.headers.name(i)}: ${response.headers.value(i)}")
+      val reader = GrpcMessageSource(
+          response.body!!.source(), HelloReply.ADAPTER,
+          response.header("grpc-encoding")
+      )
+      assertThat(reader.read()).isEqualTo(HelloReply("howdy, jesse!"))
```
mpawliszyn, line 85: 👋

PR #1565, opened 2020-07-17
Use attributes in Fake Feature Flags

minostro:
This PR contains two commits:

1. Incorporate original logic that was reverted because it was broken.
1. Added the public methods deleted by mistake, and also add methods for all the public methods exposed by the Feature Flags module.

### DIFF
--- a/misk-feature-testing/src/main/kotlin/misk/feature/testing/FakeFeatureFlags.kt
+++ b/misk-feature-testing/src/main/kotlin/misk/feature/testing/FakeFeatureFlags.kt
@@ -10,6 +10,7 @@ import misk.feature.FeatureFlags
 import misk.feature.FeatureService
 import misk.feature.fromSafeJson
 import misk.feature.toSafeJson
+import java.util.PriorityQueue
 import java.util.concurrent.ConcurrentHashMap
 import javax.inject.Inject
 import javax.inject.Provider
@@ -19,26 +20,25 @@ import javax.inject.Singleton
  * In-memory test implementation of [FeatureFlags] that allows flags to be overridden.
  */
 @Singleton
-class FakeFeatureFlags @Inject constructor(private val moshi : Provider<Moshi>) : AbstractIdleService(),
-    FeatureFlags,
-    FeatureService,
-    DynamicConfig {
-
+class FakeFeatureFlags @Inject constructor(
+  val moshi : Provider<Moshi>
+) : AbstractIdleService(), FeatureFlags, FeatureService, DynamicConfig {
   companion object {
     const val KEY = "fake_dynamic_flag"
+    val defaultAttributes = Attributes()
   }
 
   override fun startUp() {}
   override fun shutDown() {}
 
-  private val overrides = ConcurrentHashMap<MapKey, Any>()
+  private val overrides = ConcurrentHashMap<MapKey, PriorityQueue<MapValue>>()
 
   override fun getBoolean(feature: Feature, key: String, attributes: Attributes): Boolean {
     return getOrDefault(feature, key, false)
   }
 
   override fun getInt(feature: Feature, key: String, attributes: Attributes): Int =
-      get(feature, key) as? Int ?: throw IllegalArgumentException(
+      get(feature, key, attributes) as? Int ?: throw IllegalArgumentException(
           "Int flag $feature must be overridden with override() before use")
 
   override fun getString(feature: Feature, key: String, attributes: Attributes): String =
@@ -52,7 +52,7 @@ class FakeFeatureFlags @Inject constructor(private val moshi : Provider<Moshi>) 
     attributes: Attributes
   ): T {
     @Suppress("unchecked_cast")
-    return getOrDefault(feature, key, clazz.enumConstants[0]) as T
+    return getOrDefault(feature, key, clazz.enumConstants[0], attributes) as T
   }
 
   override fun <T> getJson(
@@ -61,7 +61,7 @@ class FakeFeatureFlags @Inject constructor(private val moshi : Provider<Moshi>) 
     clazz: Class<T>,
     attributes: Attributes
   ): T {
-    val jsonFn = get(feature, key) as? Function0<*> ?: throw IllegalArgumentException(
+    val jsonFn = get(feature, key, attributes) as? Function0<*> ?: throw IllegalArgumentException(
         "JSON flag $feature must be overridden with override() before use: ${get(feature, key)}")
     // The JSON is lazily provided to handle the case where the override is provided by the
     // FakeFeatureFlagModule and the Moshi instance cannot be accessed inside the module.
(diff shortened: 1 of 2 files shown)


minostro:
@adrw @swankjesse could you review this?

misk-feature-testing/src/main/kotlin/misk/feature/testing/FakeFeatureFlags.kt
```diff
@@ -77,76 +77,171 @@ class FakeFeatureFlags @Inject constructor(private val moshi : Provider<Moshi>)
...
+    //
+    // 1. Return feature level override if there is no override at the key level.
+    // 2. if there is an override at the key level, then:
+    //   2.1 Return the override value associated with the attribute override that contains
+    //       all the attributes of the given attributes.
+    //   2.2 If there is no match, return the override value defined at the key level.
+    val overrideMapValues = overrides[MapKey(feature, key)]
+        ?: return overrides[MapKey(feature)]?.let { it.first().value }
+
+    val overrideMapValuesCopy = PriorityQueue(overrideMapValues)
```
swankjesse, line 108: does this crash if overrideMapValues is null?

misk-feature-testing/src/main/kotlin/misk/feature/testing/FakeFeatureFlags.kt
```diff
@@ -77,76 +77,171 @@ class FakeFeatureFlags @Inject constructor(private val moshi : Provider<Moshi>)
...
+    var currentMapValue = overrideMapValuesCopy.poll()
+    while (overrideMapValuesCopy.isNotEmpty() || currentMapValue != null) {
```
swankjesse, line 110: I find this loop hard to understand. I think the `isNotEmpty()` part of the condition can be safely removed?

swankjesse: (because it its not empty, then we know currentMapValue will not be null)

misk-feature-testing/src/main/kotlin/misk/feature/testing/FakeFeatureFlags.kt
```diff
@@ -77,76 +77,171 @@ class FakeFeatureFlags @Inject constructor(private val moshi : Provider<Moshi>)
...
+    return currentMapValue.value
   }
 
-  fun override(feature: Feature, value: Boolean) {
-    overrides[MapKey(feature)] = value
-  }
+  fun override(
+    feature: Feature,
+    value: Boolean
+  ) = override<Boolean>(feature, value)
```
swankjesse, line 120: Is this a behavior change that causes all overrides to be stored as JSON?

swankjesse: Previously they weren’t!

misk-feature-testing/src/main/kotlin/misk/feature/testing/FakeFeatureFlags.kt
```diff
@@ -77,76 +77,171 @@ class FakeFeatureFlags @Inject constructor(private val moshi : Provider<Moshi>)
...
-  inline fun <reified T> overrideKeyJson(feature: Feature, key: String, value: T) {
-    overrideKey(feature, key, value, T::class.java)
+  fun <T> overrideKey(
+    feature: Feature,
+    key: String,
+    value: T,
+    attributes: Attributes = Attributes()
+  ) {
+    val mapKey = MapKey(feature, key)
+    overrides[mapKey] = overrides[mapKey] ?: PriorityQueue()
```
swankjesse, line 204: I think you want `computeIfAbsent()`

misk-feature-testing/src/main/kotlin/misk/feature/testing/FakeFeatureFlags.kt
minostro: if we get here it is because `overrideMapValues` is not null. I am using the elvis operator to return from the method (Line 106) in the case `overrideMapValues` is null.

misk-feature-testing/src/main/kotlin/misk/feature/testing/FakeFeatureFlags.kt
minostro: You are right. `currentMapValue` will never be null at this point. At this point we want to "test" the overrides we have stored against the `attributes` provided. So we have to iterate over all the overrides until we get a match, or until we exhausted all the overrides. Not sure how to express without a `while` loop.

I have updated the while loop in a new commit.

misk-feature-testing/src/main/kotlin/misk/feature/testing/FakeFeatureFlags.kt
minostro: nope, the execution flow will be:

1. `subject.override(Feature(name="foo"), true)`
1. `FakeFeatureFlags#override(feature: Feature, value: Boolean)` method Line 121
1. `FakeFeatureFlags#<T>override(feature: Feature, value: T)` method Line 141
1. `FakeFeatureFlags#<T>overrideKey(feature: Feature, key: String, value: T, attributes: Attributes = defaultAttributes)` method Line 204

One thing I could do to simplify the implementation is to get rid off `3.` and call `4.` directly in the implementation of methods `2.`

misk-feature-testing/src/main/kotlin/misk/feature/testing/FakeFeatureFlags.kt
minostro: nice. I have updated the code accordingly.

minostro:
@swankjesse I have addressed your comments in a new commit so you can see the diff more easily. will squash into `cee28f3 ` once I get approval.

misk-feature-testing/src/main/kotlin/misk/feature/testing/FakeFeatureFlags.kt
swankjesse: I missed the `return ` there. Yep, you’re right.

misk-feature-testing/src/test/kotlin/misk/feature/testing/FakeFeatureFlagsTest.kt
```diff
@@ -89,6 +112,101 @@ internal class FakeFeatureFlagsTest {
...
+    subject.override(OTHER_FEATURE, "other-feature")
+    assertThat(subject.getString(FEATURE, TOKEN)).isEqualTo("feature")
+    assertThat(subject.getString(OTHER_FEATURE, TOKEN)).isEqualTo("other-feature")
+
+    // Can override with specific keys
+    subject.overrideKey(FEATURE, "joker", "feature-joker")
+    assertThat(subject.getString(FEATURE, TOKEN)).isEqualTo("feature")
+    assertThat(subject.getString(FEATURE, "joker")).isEqualTo("feature-joker")
+
+    // Can override with specific keys and attributes
```
swankjesse, line 186: nit: is it useful to test the override precedence for every single type? This test might have been simpler to have a bunch of tests for override precedence, and a bunch of tests for types, but the cross-product is not very useful.

swankjesse: (this lousy structure was here when you got here, so you can continue it, but I think it’s bad)

misk-feature-testing/src/test/kotlin/misk/feature/testing/FakeFeatureFlagsTest.kt
```diff
@@ -89,6 +112,101 @@ internal class FakeFeatureFlagsTest {
...
+    assertThat(subject.getString(FEATURE, "joker")).isEqualTo("feature-joker")
+    assertThat(subject.getString(FEATURE, "joker", attributes))
+        .isEqualTo("feature-joker-with-attrs")
+    //Provides the key level override when there is no match on attributes
+    assertThat(subject.getString(FEATURE, "joker", Attributes(mapOf("don't" to "exist"))))
+        .isEqualTo("feature-joker")
+  }
+
+  @Test
+  fun `provides the latest override in case two or more attributes are a match`() {
```
swankjesse, line 199: nice

misk-feature-testing/src/main/kotlin/misk/feature/testing/FakeFeatureFlags.kt
```diff
@@ -77,76 +77,173 @@ class FakeFeatureFlags @Inject constructor(private val moshi : Provider<Moshi>)
...
+        ?: return overrides[MapKey(feature)]?.let { it.first().value }
+
+    val overrideMapValuesCopy = PriorityQueue(overrideMapValues)
+
+    var currentMapValue = overrideMapValuesCopy.poll()
+    while(overrideMapValuesCopy.isNotEmpty()) {
+      if (attributes.text.entries.containsAll(currentMapValue.attributes.text.entries)) { break }
+      currentMapValue = overrideMapValuesCopy.poll()
+    }
+    return currentMapValue.value
```
swankjesse, line 115: is there a test that exercises this line? there are overrides for attributes, and none match?

swankjesse: this is probably the line I understand the least . . . the attributes in this result won’t match, right?

minostro: > is there a test that exercises this line? there are overrides for attributes, and none match?

Yes, I defined the following test:

```
    // Override key level
    subject.overrideKeyJson(FEATURE, "joker", JsonFeature("joker"))

    //override key and attributes level
    val badJokerAttributes = Attributes(mapOf("type" to "bad"))
    subject.overrideKeyJson(FEATURE, "joker", JsonFeature("bad-joker"), badJokerAttributes)
```

On exact attributes match `badJokerAttributes`, return `JsonFeature("bad-joker")`:
```
    assertThat(subject.getJson<JsonFeature>(FEATURE, "joker", badJokerAttributes))
        .isEqualTo(JsonFeature("bad-joker"))
```

On subtype attribute match, notice that `sleepyBadJokerAttributes` has one more attribute than `badJokerAttributes`, return `JsonFeature("bad-joker")`:

```
    val sleepyBadJokerAttributes = Attributes(mapOf("type" to "bad", "state" to "sleepy"))
    assertThat(subject.getJson<JsonFeature>(FEATURE, "joker", sleepyBadJokerAttributes))
        .isEqualTo(JsonFeature("bad-joker"))
```

On none attribute match, return `JsonFeature("joker")` which is the key level override:

```
    val goodJokerAttributes = Attributes(mapOf("type" to "good"))
    assertThat(subject.getJson<JsonFeature>(FEATURE, "joker", goodJokerAttributes))
        .isEqualTo(JsonFeature("joker"))
```

This test is defined here: https://github.com/cashapp/misk/pull/1565/files#diff-e2b08dff1b3751d173f3a21aa99d695aR116-R132

minostro: > this is probably the line I understand the least . . . the attributes in this result won’t match, right?

So the main idea is that you loop through all the attributes we have in the overrides. If a match is found, the while loop breaks and the match is returned. There is the case that the while loop won't ever break but it will end when all the overrides have been consumed. If you ever reached that case, then you are in the situation where none of the attributes were a match, so the key level override will be returned. The logic is commented here in the code: https://github.com/cashapp/misk/pull/1565/files#diff-89dcbd7609483933353dac8a90bd2ca4R97-R104 (see the line with `2.2` in the front).

misk-feature-testing/src/test/kotlin/misk/feature/testing/FakeFeatureFlagsTest.kt
minostro: > nit: is it useful to test the override precedence for every single type?

it was useful from the test coverage perspective but also to uncover certain bugs when I did the refactoring.

> This test might have been simpler to have a bunch of tests for override precedence, and a bunch of tests for types, but the cross-product is not very useful.

Are you saying that we should organize the tests by having `override`, `overrideKey`, and `overrideKeyWithArguments`? each one of these methods will implement the logic for every single type?, something like:

```
fun override() {

  //override for Int
  //override for String
  //override for Enum
  //override for Boolean
  //override for Json
}
```

Happy to improve the test organization.

minostro:
@swankjesse I replied to your questions. I would like to get another round of feedback before merging. Thanks!

PR #1568, opened 2020-07-21
Add sampling logger

isafan:
Added `KLogger.sample` which uses a RateLimitingSampler. Moved around some existing classes to prevent circular dependencies. The RateLimitingSampler uses misk.RateLimiter and not Guava's to test and for consistency. 

It would be better to create a SamplingLogger that inherits from KLogger but this is a phase 1 approach.


mightyguava:
I'm worried this interface will be confusing to use. The `Sampler` must be a global, but since this requires you to pass the sampler into the log call, it makes it tempting to instantiate the sampler in the call, like `logger.info(RateLimitSampler(...)) { "hello"}`, which wouldn't work. How can we prevent improper usage of this? Or make it obvious how to use it correctly?

If we went directly to phase 2, where we have a sampled logger, it becomes simpler.

```
companion object {
  sampledLogger = logger.sampled(RateLimitSampler(...))
}
```

and then you use the sampled logger.

mightyguava:
Yeah, in the long run, I don't think we would have a use for this current interface of passing in a `Sampler()`. We can merge this, but we should do the work of creating a rate limiting logger

isafan:
> Yeah, in the long run, I don't think we would have a use for this current interface of passing in a `Sampler()`. We can merge this, but we should do the work of creating a rate limiting logger

Yeah I agree. I implemented this as a first pass per the design doc. I can follow this up with a `SamplingLogger` that can take in `PercentSampler`, `AlwaysSampler`, or `RateLimitingSampler`

PR #1570, opened 2020-07-21
RealDynamoDbModule supports specification of ClientConfiguration

zcross:
Expose AWS' ClientConfiguration as optional constructor argument to enable embedding services to override configuration for their use case.

### DIFF
--- a/misk-aws/src/main/kotlin/misk/dynamodb/RealDynamoDbModule.kt
+++ b/misk-aws/src/main/kotlin/misk/dynamodb/RealDynamoDbModule.kt
@@ -1,10 +1,9 @@
 package misk.dynamodb
 
+import com.amazonaws.ClientConfiguration
 import com.amazonaws.auth.AWSCredentialsProvider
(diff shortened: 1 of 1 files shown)


misk-aws/src/main/kotlin/misk/dynamodb/RealDynamoDbModule.kt
```diff
@@ -14,7 +13,8 @@ import javax.inject.Singleton
  * Install this module to have access to an AmazonDynamoDB client. This can be
  * used to create a DynamoDbMapper for querying of a DynamoDb table.
  */
-class RealDynamoDbModule : KAbstractModule() {
+class RealDynamoDbModule constructor(private val clientConfig: ClientConfiguration? = null) :
```
zhxnlai, line 16: Maybe make this non null and default to `ClientConfiguration()`? No need for the null check below

zcross: Good call, I see that its default constructor invocation just sets every option to a default.. works for me!

PR #1573, opened 2020-07-27
Make concurrency limits on-by-default

swankjesse:
Opt-out with `@AvailableWhenDegraded`. Most actions should not need
to do this.


shellderp:
would be nice to add a metric # requests shed

PR #1575, opened 2020-07-29
The SocketAddress.Network host field is now properly populated using the remote IP address.

janzantinge:
The remote IP is that of the client that is directly connected to a service (a proxy or service) and not the origin of the request. This will be used to help identify whether the request is from a trusted proxy.

### DIFF
--- a/misk/src/main/kotlin/misk/web/jetty/WebActionsServlet.kt
+++ b/misk/src/main/kotlin/misk/web/jetty/WebActionsServlet.kt
@@ -1,5 +1,6 @@
 package misk.web.jetty
 
+import jnr.unixsocket.UnixSocket
 import misk.exceptions.StatusCode
 import misk.logging.getLogger
 import misk.web.BoundAction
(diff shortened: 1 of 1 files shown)


misk/src/main/kotlin/misk/web/jetty/WebActionsServlet.kt
```diff
@@ -95,10 +96,15 @@ internal class WebActionsServlet @Inject constructor(
...
-            when (this) {
-              is UnixSocketConnector -> SocketAddress.Unix(this.unixSocket)
-              is ServerConnector -> SocketAddress.Network(this.host ?: "0.0.0.0", this.localPort)
+          linkLayerLocalAddress = with((request as? Request)?.httpChannel) {
+            when (this?.connector) {
+              is UnixSocketConnector -> SocketAddress.Unix(
+                  (this.connector as UnixSocketConnector).unixSocket
+              )
+              is ServerConnector -> SocketAddress.Network(
+                  this.endPoint.remoteAddress.address.hostAddress,
```
janzantinge, line 105: I'm pretty sure this is the right address, testing shows it does the job.

PR #1580, opened 2020-08-04
Optimize BulkShardMigrator for TiDB

kalfonso:
Added support for mutations except a mutation for a value mapper which does not make sense in the context of a bulk update. Further mutations to apply a function to a column, etc, can be added if required


misk-hibernate/src/main/kotlin/misk/hibernate/migrate/BulkShardMigrator.kt
```diff
@@ -389,6 +424,14 @@ class BulkShardMigrator<R : DbRoot<R>, C : DbChild<R, C>> private constructor(
       insert.setObject(parameterIndex, value)
       return 1
     }
+
+    open fun bindUpdate(update: PreparedStatement, parameterIndex: Int): Int {
+      return 0
```
tirsen, line 429: Should this be abstract or throw `UnsupportedOperationException`?

misk-hibernate/src/main/kotlin/misk/hibernate/migrate/BulkShardMigrator.kt
```diff
@@ -417,6 +465,10 @@ class BulkShardMigrator<R : DbRoot<R>, C : DbChild<R, C>> private constructor(
       insert.setObject(parameterIndex, valueMapper.apply(value))
       return 1
     }
+
+    override fun bindUpdate(update: PreparedStatement, parameterIndex: Int): Int {
+      throw UnsupportedOperationException("Cannot apply updates using SetMappingMutation")
```
tirsen, line 470: I guess this is sort of doable just pretty hard :-) Should the error message say that we haven't implemented this because it's hard? :-) You could imagine you pull up the values with a select before or you specify a SQL function as well.

misk-hibernate/src/test/kotlin/misk/hibernate/migrate/BulkShardMigratorTest.kt
```diff
@@ -52,43 +43,90 @@ class BulkShardMigratorTest {
...
+              .size
+        }
+    )
+  }
+}
+
+@MiskTest(startService = true)
+class BulkShardMigratorVitessMySqlTest: BulkShardMigratorTest() {
+  @MiskTestModule
+  val module = MoviesTestModule()
```
tirsen, line 121: Maybe specify `VITESS_MYSQL` here explicitly?

misk-hibernate/src/main/kotlin/misk/hibernate/migrate/BulkShardMigrator.kt
kalfonso: I wanted to avoid doing a select to update the entities as those are exactly the queries that are impacting more (in volume) TiDB. I think also the Jdbc ```PreparedStatement``` doesn't allow to set a value based on the existing value of a column.

misk-hibernate/src/main/kotlin/misk/hibernate/migrate/BulkShardMigrator.kt
kalfonso: It was used as default behavior, used by ```NowMutation```. Although thinking more about it maybe is better to define it as abstract and force all mutations to implement it.

misk-hibernate/src/test/kotlin/misk/hibernate/migrate/BulkShardMigratorTest.kt
kalfonso: Done

misk-hibernate/src/main/kotlin/misk/hibernate/migrate/BulkShardMigrator.kt
kalfonso: Done

misk-hibernate/src/main/kotlin/misk/hibernate/migrate/BulkShardMigrator.kt
```diff
@@ -154,6 +156,10 @@ class BulkShardMigrator<R : DbRoot<R>, C : DbChild<R, C>> private constructor(
     checkNotNull(targetRoot) { "You have to specify entity root target" }
     checkNotNull(sourceRoot) { "You have to specify entity root source" }
 
+    if (transacter.config().type == DataSourceType.TIDB) {
```
tirsen, line 159: I think we can do this on plain MYSQL as well btw? Right? Should optimize things nicely on Aurora as well! Do you want to try that in a follow up?

misk-hibernate/src/main/kotlin/misk/hibernate/migrate/BulkShardMigrator.kt
tirsen: Yeah I totally wouldn't do a select! We can hopefully avoid implementing this. :-) If we do want to implement it I think we should force the user to specify a SQL function expression.

PR #1581, opened 2020-08-06
Fix BulkshardMigrator under TiDB: add extra predicates to where clause

kalfonso:
It keeps it consistent with existing functionality

### DIFF
--- a/misk-hibernate/src/main/kotlin/misk/database/DockerTidbCluster.kt
+++ b/misk-hibernate/src/main/kotlin/misk/database/DockerTidbCluster.kt
@@ -96,7 +96,8 @@ class DockerTidbCluster(
   companion object {
     val logger = KotlinLogging.logger {}
 
-    const val SHA = "5d4eef28ae1024712015a846db9312d8f477f8bed40f7f9a410c62def8342f96"
+    // TiDB version 4.0.4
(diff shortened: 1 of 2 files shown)


misk-hibernate/src/main/kotlin/misk/hibernate/migrate/BulkShardMigrator.kt
```diff
@@ -186,7 +186,7 @@ class BulkShardMigrator<R : DbRoot<R>, C : DbChild<R, C>> private constructor(
           val update = """
           UPDATE ${tableName()}
           SET $setColumns
-          WHERE $rootColumnName = ?
+          WHERE $where
```
kalfonso, line 189: ```where``` clause already contains the predicate to filter child entities from the source root. This is done in ```BulkShardMigrator.source(...)```

misk-hibernate/src/main/kotlin/misk/database/DockerTidbCluster.kt
```diff
@@ -96,6 +96,7 @@ class DockerTidbCluster(
   companion object {
     val logger = KotlinLogging.logger {}
 
+    // TiDB version 4.0.3
```
mightyguava, line 99: 4.0.4 is out

kalfonso: oh tru, updated

PR #1584, opened 2020-08-10
Exposing some properties of DynamoDB configuration for testing

miladghaznavi:
To facilitate testing, expose some properties to provide more information about dynamodb docker


misk-aws-dynamodb-testing/src/main/kotlin/misk/aws/dynamodb/testing/DockerDynamoDb.kt
```diff
@@ -70,19 +83,15 @@ object DockerDynamoDb : ExternalDependency {
 
   override fun shutdown() {
     composer.stop()
+    endpointConfiguration.serviceEndpoint
```
zhxnlai, line 86: What does this line do?

miladghaznavi: It left by mistake

PR #1585, opened 2020-08-10
Log fewer errors when shedding requests

swankjesse:


misk/src/main/kotlin/misk/web/interceptors/ConcurrencyLimitsInterceptor.kt
```diff
@@ -24,14 +27,21 @@ import kotlin.reflect.full.findAnnotation
  */
 internal class ConcurrencyLimitsInterceptor internal constructor(
   private val actionName: String,
-  private val limiter: Limiter<String>
+  private val limiter: Limiter<String>,
+  private val clock: Clock
 ) : NetworkInterceptor {
+  /**
+   * When this fails, it fails a lot. Only log one exception per minute per node and let the
```
mpawliszyn, line 34: log at most one exception per minute?

misk/src/main/kotlin/misk/web/interceptors/ConcurrencyLimitsInterceptor.kt
```diff
@@ -59,6 +69,17 @@ internal class ConcurrencyLimitsInterceptor internal constructor(
     }
   }
 
+  private fun logShedRequest() {
+    val nowMs = clock.millis()
+    val durationSinceLastErrorMs = nowMs - lastErrorLoggedAtMs
```
mpawliszyn, line 74: Do we want to keep track and log how many have happened since the last log statement? Might be useful. But I guess you are right in that we can rely on metrics.

swankjesse: Yeah I think that’s easier in metrics anyway.

misk/src/main/kotlin/misk/web/interceptors/ConcurrencyLimitsInterceptor.kt
swankjesse: Fixed!

PR #1586, opened 2020-08-11
Add github action to create date+sha releases

shellderp:
Publishes snapshots to maven with date/sha added so snapshots aren't overwritten. I want to pull these in in backfila.
Already added this in backfila, check it out:
image

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,37 @@
+name: build
+
+on:
+  push:
+    branches:
+      - master
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set up JDK 11
+        uses: actions/setup-java@v1
+        with:
+          java-version: 11
+      - name: install misk-web
+        run: |
+          mkdir ~/.npm-global
+          npm config set prefix '~/.npm-global'
+          PATH=~/.npm-global/bin:$PATH
+          npm install -g @misk/cli
+          miskweb ci-build -e
+      - name: Cache Gradle packages
+        uses: actions/cache@v1
+        with:
+          path: ~/.gradle/caches
+          key: ${{ runner.os }}-gradle-${{ hashFiles('**/*.gradle') }}
+          restore-keys: ${{ runner.os }}-gradle
+      - name: Set gradle version
+        run: echo "version=String.valueOf(project.version).replace(\"-SNAPSHOT\", \"\") + '-$(date +%Y%m%d)-$(git rev-parse --short HEAD)-SNAPSHOT'" > hooks.gradle
+      - name: Publish packages
+        run: ./gradlew clean uploadArchives
+        env:
+          ORG_GRADLE_PROJECT_SONATYPE_NEXUS_USERNAME: ${{ secrets.SONATYPE_NEXUS_USERNAME }}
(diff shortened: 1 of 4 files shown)


shellderp:
@swankjesse is this silly?

.github/workflows/build.yml
```diff
@@ -0,0 +1,37 @@
...
+          npm install -g @misk/cli
+          miskweb ci-build -e
+      - name: Cache Gradle packages
+        uses: actions/cache@v1
+        with:
+          path: ~/.gradle/caches
+          key: ${{ runner.os }}-gradle-${{ hashFiles('**/*.gradle') }}
+          restore-keys: ${{ runner.os }}-gradle
+      - name: Set gradle version
+        run: echo "version=String.valueOf(project.version).replace(\"-SNAPSHOT\", \"\") + '-$(date +%Y%m%d)-$(git rev-parse --short HEAD)-SNAPSHOT'" > hooks.gradle
```
swankjesse, line 32: wanna do hours+minutes so if we do 2 releases in a day the versions always go forward?

.github/workflows/build.yml
```diff
@@ -0,0 +1,37 @@
...
+      - name: Publish packages
+        run: ./gradlew clean uploadArchives
+        env:
+          ORG_GRADLE_PROJECT_SONATYPE_NEXUS_USERNAME: ${{ secrets.SONATYPE_NEXUS_USERNAME }}
```
swankjesse, line 36: might also need our signing key?
`$secrets.ARTIFACT_SIGNING_PRIVATE_KEY` has it

https://github.com/cashapp/certifikit/blob/master/.github/workflows/Release.yml

swankjesse: (this one is triggered by a tag being pushed)

.github/workflows/build.yml
swankjesse: presumably you use `-SNAPSHOT` so there’s no promotion step?

shellderp: re: hours, sure why not

re snapshot: I figured it was the less offensive / easier to convince option, but if you think normal release is fine I can try to get that working. There's a promote target in the latest maven plugin

.github/workflows/build.yml
shellderp: somehow I had no problem without it, but I should add it

.github/workflows/build.yml
shellderp: Added hours+mins and it's not pushing snapshots anymore

shellderp:
@swankjesse PTAL

shellderp:
@swankjesse Update: I think I have it signing, but now it's getting
```
   > Could not PUT 'https://oss.sonatype.org/service/local/staging/deploy/maven2/com/squareup/misk/misk/0.14.0-20200813.0143-2a633e0/misk-0.14.0-20200813.0143-2a633e0.jar'. Received status code 403 from server: Forbidden```

does the cashapp org-level account not have misk publishing perms? had the same issue when trying to upload a snapshot, so I don't think it's a signing problem

File: build.gradle [hunk 3]
```diff
@@ -108,6 +105,17 @@ subprojects { subProject ->
 
   if (!path.startsWith(":samples") && !path.startsWith(":misk-embedded-sample")) {
     apply plugin: 'com.vanniktech.maven.publish'
+
+    // Override signing plugin details that maven-publish will use.
+    signing {
```
shellderp, line 110: this breaks local publishing, so I guess I can put this part in the `hooks.gradle`?

chris-ryan-square:

chris-ryan-square:

PR #1587, opened 2020-08-11
Kill TiDB images

tirsen:


tirsen:
I wonder if we should do this or just kill the containers using a trap in the Kochiku bash script?

mightyguava:
I think I liked the old old behavior best.

If a container was not running when tests started, stop it when you finish.
If a container was already running when tests started, leave it running when you finish.

PR #1588, opened 2020-08-11
Add BigDecimal Moshi Adapter

minostro:
BigDecimal Moshi Adapter is needed for multiple projects. It's time to promote it to misk.

### DIFF
--- a/misk/src/test/kotlin/misk/moshi/BigDecimalAdapterTest.kt
+++ b/misk/src/test/kotlin/misk/moshi/BigDecimalAdapterTest.kt
@@ -1,0 +1,37 @@
+package misk.moshi
+
+import com.squareup.moshi.Moshi
+import misk.MiskTestingServiceModule
+import misk.inject.KAbstractModule
+import misk.moshi.adapters.BigDecimalAdapter
+import misk.testing.MiskTest
+import misk.testing.MiskTestModule
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.jupiter.api.Test
+import java.math.BigDecimal
+import javax.inject.Inject
+
+@MiskTest(startService = false)
+internal class BigDecimalAdapterTest {
+  @MiskTestModule
+  val module = TestModule()
+
+  @Inject
+  lateinit var moshi: Moshi
+
+  @Test
+  fun `adapter converts from and to json`() {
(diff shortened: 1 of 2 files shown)


misk/src/main/kotlin/misk/moshi/BigDecimalAdapter.kt
```diff
@@ -0,0 +1,13 @@
+package misk.moshi
+
+import com.squareup.moshi.FromJson
+import com.squareup.moshi.ToJson
+import java.math.BigDecimal
+
+object BigDecimalAdapter {
```
minostro, line 7: not sure if we should create a `adapters` package or if this is good for the time been.

misk/src/test/kotlin/misk/moshi/BigDecimalAdapterTest.kt
```diff
@@ -0,0 +1,36 @@
...
+    val value = BigDecimal("5.5")
+    val jsonAdapter = moshi.adapter<BigDecimal>()
+    assertThat(jsonAdapter.toJson(value)).isEqualTo(json)
+    assertThat(jsonAdapter.fromJson(json)).isEqualTo(value)
+  }
+
+  class TestModule : KAbstractModule() {
+    override fun configure() {
+      install(MiskTestingServiceModule())
+      install(MoshiAdapterModule(BigDecimalAdapter))
```
minostro, line 33: Didn't want to add the BigDecimalAdapter as a default adapter, but open to put it in the default adapters.

misk/src/main/kotlin/misk/moshi/BigDecimalAdapter.kt
adrw: Let's create the package and put this adapter there but skip on consolidating the other adapters into it since it will be a main to migrate all of the imports.

minostro: this is done.

misk/src/test/kotlin/misk/moshi/BigDecimalAdapterTest.kt
swankjesse: Why not? I think it’d be great!

PR #1589, opened 2020-08-11
Install BigDecimal Adapter by default

minostro:


minostro:
@swankjesse BigDecimal Adapter is installed by default

PR #1590, opened 2020-08-12
ReflectionQueryFactory supports query hint specification

zcross:
This adds a pass-through for Hibernate's `Query#addQueryHint` in Misk's `ReflectionQueryFactory`-produced `ReflectionQuery`. [See: Query#addQueryHint](https://docs.jboss.org/hibernate/orm/5.0/javadocs/org/hibernate/Query.html#addQueryHint-java.lang.String-)

I'm aware that hints should be used sparingly, but I have a use case (at least with TiDB) where specifying a hint (e.g., for `INDEX_MERGE` on two same-table indices) makes the difference between a table scan (or making two sequential queries to the two independent indices that are suitable for `INDEX_MERGE`.

### DIFF
--- a/misk-hibernate/src/main/kotlin/misk/hibernate/ReflectionQueryFactory.kt
+++ b/misk-hibernate/src/main/kotlin/misk/hibernate/ReflectionQueryFactory.kt
@@ -17,6 +17,7 @@ import java.util.EnumSet
 import javax.inject.Inject
 import javax.inject.Singleton
 import javax.persistence.criteria.CriteriaBuilder
+import javax.persistence.criteria.CriteriaQuery
 import javax.persistence.criteria.Path
 import javax.persistence.criteria.Predicate
 import javax.persistence.criteria.Root
@@ -63,9 +64,14 @@ internal class ReflectionQuery<T : DbEntity<T>>(
   private val orderFactories = mutableListOf<OrderFactory>()
 
   private val disabledChecks: EnumSet<Check> = EnumSet.noneOf(Check::class.java)
+  private val queryHints: MutableSet<String> = mutableSetOf()
 
   override fun disableCheck(check: Check) {
     disabledChecks += check
+  }
+
+  override fun addQueryHint(hintString: String) {
+    queryHints.add(hintString)
   }
 
   override fun dynamicAddConstraint(path: String, operator: Operator, value: Any?) {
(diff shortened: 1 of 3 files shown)


misk-hibernate/src/main/kotlin/misk/hibernate/ReflectionQueryFactory.kt
```diff
@@ -252,7 +259,7 @@ internal class ReflectionQuery<T : DbEntity<T>>(
 
     query.orderBy(buildOrderBys(queryRoot, criteriaBuilder))
 
-    val typedQuery = session.hibernateSession.createQuery(query)
+    val typedQuery = createTypedQueryWithHint(session, query)
```
mightyguava, line 262: nit: the `withHint` is misleading, makes it sound like you always add a hint. Could just drop it from the name.

misk-hibernate/src/test/kotlin/misk/hibernate/ReflectionQueryFactoryTest.kt
```diff
@@ -910,6 +910,29 @@ class ReflectionQueryFactoryTest {
...
+      val jp = DbMovie("Jurassic Park", LocalDate.of(1993, 6, 9))
+      session.save(jp)
+      val jg = DbActor("Jeff Goldblum")
+      session.save(jg)
+      session.save(DbCharacter("Ian Malcolm", jp, jg))
+
+      jp.id
+    }
+
+    val selectResult = transacter.readOnly().transaction { session ->
```
mightyguava, line 925: is there a way to test that the hint was actually applied?

zcross: I wish – I ran into issues with MySQL/Vitess/TiDB treating the inputs slightly differently.
One thing I had in mind was using `IGNORE INDEX` to force a failure when table scans are not allowed on the query. WDYT?

mightyguava: Yeah, I think that could be a good test. How are table scans not allowed though?

zcross:
Because of the imperfect (incomplete) test and a workaround that I've found that obviates using hints at all, I'm going to close this.

PR #1595, opened 2020-08-27
Running dynamoDB with sharedDb option allowing developers to enable dynamoDB streams

miladghaznavi:
…ynamoDB streams


misk-aws-dynamodb-testing/src/main/kotlin/misk/aws/dynamodb/testing/DockerDynamoDb.kt
```diff
@@ -52,6 +52,7 @@ object DockerDynamoDb : ExternalDependency {
     withImage("amazon/dynamodb-local")
         .withName(id)
         .withExposedPorts(exposedClientPort)
+        .withCmd("-jar", "DynamoDBLocal.jar", "-sharedDb")
```
zhxnlai, line 55: could you add a comment, perhaps linking to documentation, on why KCL depends on this option?

miladghaznavi: I can add a comment, but AFAIK there is no AWS documentation around this. We figure this out when I was discussing this with platform team and from some try and error.

PR #1598, opened 2020-09-04
Add a configurable cors filter to misk

dau-square:
This can be used to configure Misk to allow cross origin requests.

### DIFF
--- a/misk/src/main/kotlin/misk/web/WebConfig.kt
+++ b/misk/src/main/kotlin/misk/web/WebConfig.kt
@@ -4,6 +4,7 @@ import misk.config.Config
 import misk.security.ssl.CertStoreConfig
 import misk.security.ssl.TrustStoreConfig
 import misk.web.exceptions.ActionExceptionLogLevelConfig
+import javax.servlet.FilterConfig
 
 data class WebConfig(
   /** HTTP port to listen on, or 0 for any available port. */
@@ -78,7 +79,9 @@ data class WebConfig(
   val gzip: Boolean = true,
 
   /** The minimum size in bytes before the response body will be compressed. */
-  val minGzipSize: Int = 1024
+  val minGzipSize: Int = 1024,
+
+  val cors: Map<String, CorsConfig> = mapOf()
 ) : Config
 
 data class WebSslConfig(
@@ -100,4 +103,31 @@ data class WebUnixDomainSocketConfig(
   val path: String,
   /** If true, the listener will support H2C. */
   val h2c: Boolean? = true
+)
+
+data class CorsConfig(
+    /** A comma separated list of origins that are allowed to access the resources. */
+    val allowedOrigins: Array<String> = arrayOf("*"),
+    /**
+     * A comma separated list of HTTP methods that are allowed to be used when
+     * accessing the resources.
+     */
+    val allowedMethods: Array<String> = arrayOf("GET", "POST", "HEAD"),
+    /**
+     * A comma separated list of HTTP headers that are allowed to be specified when
+     * accessing the resources.
+     */
+    val allowedHeaders: Array<String> = arrayOf("X-Requested-With", "Content-Type", "Accept",
+        "Origin"),
+    /** A boolean indicating if the resource allows requests with credentials. */
+    val allowCredentials: Boolean = true,
+    /** The number of seconds that preflight requests can be cached by the client. */
+    val preflightMaxAge: String = "1800",
+    /**
+     * True if preflight requests are chained to their target resource for normal handling
+     * (as an OPTION request).
+     */
+    val chainPreflight: Boolean = true,
+    /** A comma separated list of HTTP headers that are allowed to be exposed on the client. */
+    val exposedHeaders: Array<String> = arrayOf()
 )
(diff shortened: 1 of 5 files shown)


misk/src/main/kotlin/misk/web/WebConfig.kt
```diff
@@ -101,3 +104,30 @@ data class WebUnixDomainSocketConfig(
   /** If true, the listener will support H2C. */
   val h2c: Boolean? = true
 )
+
+data class CorsConfig(
```
dau-square, line 108: Defaults as defined here: https://www.eclipse.org/jetty/documentation/current/cross-origin-filter.html

I don't agree with all of them, but I'm possibly just being biased.

swankjesse: What changes would you make?!

misk/src/main/kotlin/misk/web/WebConfig.kt
```diff
@@ -101,3 +104,30 @@ data class WebUnixDomainSocketConfig(
...
+    val allowedMethods: Array<String> = arrayOf("GET", "POST", "HEAD"),
+    /**
+     * A comma separated list of HTTP headers that are allowed to be specified when
+     * accessing the resources.
+     */
+    val allowedHeaders: Array<String> = arrayOf("X-Requested-With", "Content-Type", "Accept",
+        "Origin"),
+    /** A boolean indicating if the resource allows requests with credentials. */
+    val allowCredentials: Boolean = true,
+    /** The number of seconds that preflight requests can be cached by the client. */
```
swankjesse, line 124: These docs are handy

misk/src/main/kotlin/misk/web/WebConfig.kt
```diff
@@ -101,3 +104,30 @@ data class WebUnixDomainSocketConfig(
...
+    val preflightMaxAge: String = "1800",
+    /**
+     * If true preflight requests are chained to their target resource for normal handling
```
swankjesse, line 127: supernit: I think this might read better as “True if preflight requests..." rather than “If true preflight requests”

misk/src/main/kotlin/misk/web/jetty/JettyService.kt
```diff
@@ -243,6 +248,25 @@ class JettyService @Inject internal constructor(
 
     server.handler = statisticsHandler
 
+    webConfig.cors.forEach { (path, corsConfig) ->
+      val holder = FilterHolder(CrossOriginFilter::class.java)
+      holder.setInitParameter(CrossOriginFilter.ALLOWED_ORIGINS_PARAM,
+          corsConfig.allowedOrigins.joinToString(","))
+      holder.setInitParameter(CrossOriginFilter.ALLOWED_HEADERS_PARAM,
+          corsConfig.allowedHeaders.joinToString(","))
+      holder.setInitParameter(CrossOriginFilter.ALLOWED_METHODS_PARAM,
```
swankjesse, line 257: mind ordering these statements the same as they’re declared in the config object? This one has [origins, headers, methods...] and the config has [origins, methods, headers...] 

It’s easier to review & maintain if the orders are consistent! (Though I have no preference of order)

misk/src/test/kotlin/misk/web/CorsFilterTest.kt
```diff
@@ -0,0 +1,109 @@
...
+  @Inject lateinit var moshi: Moshi
+
+  @Test fun `Preflight with cors allowed origin`() {
+    val response = preflight("/miskhype/16")
+
+    assertThat(response.headers.get("Access-Control-Allow-Origin"))
+        .isEqualTo("https://localhost:8080")
+  }
+
+  @Test fun `Preflight with cors to restricted domain`() {
```
swankjesse, line 36: what makes it restricted?

misk/src/test/kotlin/misk/web/CorsFilterTest.kt
```diff
@@ -0,0 +1,109 @@
...
+  }
+
+  @Test fun `Preflight to path without cors`() {
+    val response = preflight("/count")
+
+    assertThat(response.headers.get("Access-Control-Allow-Origin"))
+        .isNull()
+  }
+
+  class MiskHypeGetAction @Inject constructor() : WebAction {
```
swankjesse, line 55: Is there any meaningful difference in behavior between the three actions here? Ie. is /restricted special somehow?

swankjesse (approved):

misk/src/test/kotlin/misk/web/CorsFilterTest.kt
dau-square: I'll rename this test, the endpoint called has a more restrictive policy defined that only allows localhost to make cross domain requests - not that the domain itself is restricted.

misk/src/test/kotlin/misk/web/CorsFilterTest.kt
dau-square:
```
install(WebTestingModule(webConfig = WebTestingModule.TESTING_WEB_CONFIG.copy(
          cors = mapOf(
              Pair("/miskhype/*", CorsConfig()),
              Pair("/restricted", CorsConfig(allowedOrigins = arrayOf("https://localhost:*")
          ))))))
```
The module installation has a different config defined for each endpoint, I'll rename the paths to be more reflective of the applied config

misk/src/main/kotlin/misk/web/WebConfig.kt
dau-square: val allowCredentials: Boolean = true
and 
val allowedOrigins: Array<String> = arrayOf("*")

seems a little bit too lenient for a feature that effectively allows opening up a security policy

dau-square: (Going with defaults for now)

PR #1601, opened 2020-09-09
Support loading hybrid public keys not encrypted by a KMS

yoavamit:
This change enables misk-crypto to load Hybrid public keys (for encryption only) that are not encrypted by a KMS.

### DIFF
--- a/misk-crypto/src/main/kotlin/misk/crypto/KeyProviders.kt
+++ b/misk-crypto/src/main/kotlin/misk/crypto/KeyProviders.kt
@@ -27,6 +27,7 @@ import com.google.crypto.tink.streamingaead.StreamingAeadFactory
 import com.google.inject.Inject
 import com.google.inject.Provider
 import misk.logging.getLogger
+import okio.IOException
 import java.security.GeneralSecurityException
 
 open class KeyReader {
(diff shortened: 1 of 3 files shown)


misk-crypto/src/main/kotlin/misk/crypto/KeyProviders.kt
```diff
@@ -153,7 +154,11 @@ internal class HybridEncryptProvider(
   @Inject lateinit var kmsClient: KmsClient
 
   override fun get(): HybridEncrypt {
-    val keysetHandle = readKey(key, kmsUri, kmsClient)
+    val keysetHandle = try {
+      readKey(key, kmsUri, kmsClient)
+    } catch (e: IOException) {
+      readKey(key, kmsUri = null, kmsClient =  null)
+    }
```
yan, line 161: This feels like we're forcing the use of `readEncryptedKey`/`readCleartextKey` and not using the public API. What if we fall back to reading cleartext if the kms uri is `""` as well as null, and leave kmsClient use as is?

yoavamit: I thought about that, but there's a couple of setbacks to doing that:
- In the configuration, `kmsUri` is not nullable, so you have to specify something.
- Playing with the `kmsUri` value might be tricky because there might be services that use both aead keys for example, and unencrypted hybrid public keys, so you don't want to change the `kmsUri` in that case in the configuration.

Another option would be to make `readCleartextKey()` public, but I'm hesitant to do that.

I think that explicitly specifying those null values make it clearer that key is not encrypted and the `readKey` function should call `readCleartextKey` ini turn.

yan (approved):

PR #1605, opened 2020-09-15
Extending SpooledProducer to accept a java.sql.Connection

frojasg:
Extending this interface will allows us to use Hibernate EventListener
to publish event to an unsharded spool.


jjestrel:
Moving from Slack, does `event.getPersister().getFactory().getCurrentSession()` solve your use case?

frojasg:
The suggested API doesn't solve the problem. However, I will close this PR as the cashpub team is rethinking on how to support these features in the future.

PR #1607, opened 2020-09-16
Improve mapping of TiDB write-write conflict in optimistic transaction mode

kalfonso:
```OptimisticLockingException``` is the more intuitive exception to map to which is aso already retryable

### CODE REVIEW

**Review by mightyguava — APPROVED** [review]
Is this something the transacter knows to retry?

File: misk-hibernate/src/main/kotlin/misk/hibernate/RealTransacter.kt [hunk 1]
```diff
@@ -367,8 +367,7 @@ internal class RealTransacter private constructor(
   }
 
   private fun isMessageRetryable(th: SQLException) =
-      isConnectionClosed(th) || isVitessTransactionNotFound(th) ||
-          isCockroachRestartTransaction(th) || isTidbWriteConflict(th)
```
mpawliszyn, line 371: So write conflicts aren't retryable?

misk-hibernate/src/main/kotlin/misk/hibernate/RealTransacter.kt
```diff
@@ -267,7 +267,7 @@ internal class RealTransacter private constructor(
       // write-write conflicts fail at COMMIT rather than waiting on a lock.
       e.cause is SQLException && isTidbWriteConflict(e.cause as SQLException) -> {
         val sqlException = e.cause as SQLException
-        return ConstraintViolationException(sqlException.message, sqlException, "")
```
mpawliszyn, line 270: This might break apps that are relying on this?

misk-hibernate/src/main/kotlin/misk/hibernate/RealTransacter.kt
```diff
@@ -267,7 +267,7 @@ internal class RealTransacter private constructor(
...
+        return OptimisticLockException(sqlException.message, sqlException, "")
```
bezmax, line 270: I am kind of wary about `tidbWriteConflict` being converted into `OptimisticLockException`. Technically, it is  an optimistic lock exception from perspective of TiDB, but from perspective of the app it's entirely different from what `javax.persistence.OptimisticLockException` is meant to be.

mightyguava: Different how?

> This exception may be thrown as part of an API call, a flush or at commit time. The current transaction, if one is active, will be marked for rollback.

We are throwing this at `commit` time for an optimistic lock conflict

bezmax: There's "an" optimistic lock, "JPA's optimistic lock", and "TiDB's optimistic lock". In my opinion, this exception is meant to represent specifically JPA's optimistic lock failures (`@Version` annotation, other implementations that are DB agnostic, etc.). 

But yeah, given that TiDB makes Hibernate's own optimistic locking unreliable, it is reasonable to expect it to unpack the `OptimisticLockException` from a different source, like this error from TiDB.

Yeah, I guess don't mind my original comment, it kind of seems reasonable now that I think about it.

chris-ryan-square:
If this PR is still relevant, then please address comments, rebase and get a green CI build.

chris-ryan-square:

PR #1609, opened 2020-09-24
Add retries to FakeTransactionalJobQueue

drashkov:
Adds the ability to test jobs failing and restarting.

### DIFF
--- a/misk-jobqueue-testing/src/main/kotlin/misk/jobqueue/FakeTransactionalJobQueue.kt
+++ b/misk-jobqueue-testing/src/main/kotlin/misk/jobqueue/FakeTransactionalJobQueue.kt
@@ -1,5 +1,7 @@
 package misk.jobqueue
 
+import misk.backoff.FlatBackoff
+import misk.backoff.retry
 import misk.hibernate.Gid
 import misk.hibernate.Session
 import misk.tokens.TokenGenerator
@@ -73,7 +75,8 @@ class FakeTransactionalJobQueue @Inject constructor(
   /** Returns all jobs that were handled. */
   fun handleJobs(
     queueName: QueueName,
-    assertAcknowledged: Boolean = true
+    assertAcknowledged: Boolean = true,
+    retries: Int = 1
   ): List<FakeJob> {
     val jobHandler = jobHandlers.get()[queueName]!!
     val jobs = jobQueues[queueName] ?: return listOf()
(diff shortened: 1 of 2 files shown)


misk-jobqueue-testing/src/main/kotlin/misk/jobqueue/FakeTransactionalJobQueue.kt
```diff
@@ -73,15 +75,16 @@ class FakeTransactionalJobQueue @Inject constructor(
...
+    retries: Int = 1
   ): List<FakeJob> {
     val jobHandler = jobHandlers.get()[queueName]!!
     val jobs = jobQueues[queueName] ?: return listOf()
 
     val result = mutableListOf<FakeJob>()
     while (true) {
       val job = jobs.poll() ?: break
-      jobHandler.handleJob(job)
+      retry(retries,  FlatBackoff(Duration.ofMillis(100))) { jobHandler.handleJob(job) }
```
alanpaulin, line 87: ultranit: there are 2 spaces after `retries,`

alanpaulin, line 87: Just curious, where did the constant `100` come from? Is that what is used elsewhere?

misk-jobqueue-testing/src/test/kotlin/misk/jobqueue/FakeTransactionalJobQueueTest.kt
```diff
@@ -215,6 +240,17 @@ private class Starport @Inject private constructor(moshi: Moshi) : JobHandler {
   }
 }
 
+private class Crasher @Inject private constructor(val tokenGenerator: TokenGenerator) : JobHandler {
+
+  override fun handleJob(job: Job) {
+    // Tokens are generated once per test run, starting from 1.
+    if (tokenGenerator.generate().endsWith("1")) {
+      throw UnitException()
```
alanpaulin, line 248: Just curious: Is there a significance to `UnitException`, or could it be any `Exception` class?

alanpaulin (approved):
LGTM

misk-jobqueue-testing/src/test/kotlin/misk/jobqueue/FakeTransactionalJobQueueTest.kt
drashkov: Could be any checked exception, this one is keeping with the starcraft theme of this test class.

misk-jobqueue-testing/src/main/kotlin/misk/jobqueue/FakeTransactionalJobQueue.kt
drashkov: Pretty arbitrary, but small timeouts like this are used in the other tests.

PR #1612, opened 2020-09-25
Adds PGP support via BouncyCastle

tso:
For context almost all of the actual PGP encoding/decoding code is from https://github.com/bcgit/bc-java/tree/master/pg/src/main/java/org/bouncycastle/openpgp/examples with cleanup + some changes to support non-file inputs.


misk-crypto/src/main/kotlin/misk/crypto/pgp/PgpEncoderProvider.kt
```diff
@@ -0,0 +1,59 @@
...
+      }
+    }
+
+    val pgpSec = PGPSecretKeyRingCollection(
+        PGPUtil.getDecoderStream(secretKeyStream),
+        JcaKeyFingerprintCalculator()
+    )
+
+    val secretKey = pgpSec.keyRings.next().secretKey
+        .extractPrivateKey(JcePBESecretKeyDecryptorBuilder().setProvider("BC").build(null))
```
tso, line 46: I wonder if this will be upset in tests since i only do ```  init {
    Security.addProvider(BouncyCastleProvider())
  }``` in `RealPgpEncoder`

misk-crypto/src/main/kotlin/misk/crypto/pgp/PublicKeyProvider.kt
```diff
@@ -0,0 +1,19 @@
...
+import org.bouncycastle.openpgp.PGPUtil
+import org.bouncycastle.openpgp.operator.jcajce.JcaKeyFingerprintCalculator
+import java.io.BufferedInputStream
+
+internal class PgpPublicKeyProvider(val key: Key) : Provider<PGPPublicKey>, KeyReader() {
+  override fun get(): PGPPublicKey {
+    val keyIn = BufferedInputStream(key.encrypted_key.value.byteInputStream())
+    val pgpPub =
+        PGPPublicKeyRingCollection(PGPUtil.getDecoderStream(keyIn), JcaKeyFingerprintCalculator())
+    return pgpPub.keyRings.next().publicKey
```
alanpaulin, line 17: this is creating a new public key?

misk-crypto/src/main/kotlin/misk/crypto/pgp/RealPgpEncoder.kt
```diff
@@ -0,0 +1,133 @@
...
+  ): ByteArray {
+    val buffer = okio.Buffer()
+
+    val byteArrayOutputStream = ByteArrayOutputStream()
+    val pgpCompressedDataGenerator = PGPCompressedDataGenerator(PGPCompressedData.ZIP)
+    val pgpLiteralDataGenerator = PGPLiteralDataGenerator()
+    val pgpLiteralDataStream = pgpLiteralDataGenerator.open(
+        pgpCompressedDataGenerator.open(byteArrayOutputStream),
+        PGPLiteralData.BINARY,
+        "", // Name of the file encoded in the data. Just boring metadata.
```
alanpaulin, line 48: Is empty string a legal value for this file name?

misk-crypto/src/main/kotlin/misk/crypto/pgp/RealPgpEncoder.kt
```diff
@@ -0,0 +1,133 @@
...
+        .setProvider(BouncyCastleProvider())
+        .setSecureRandom(SecureRandom())
+    pgpEncryptedDataGenerator.addMethod(keyEncryptionMethodGenerator)
+
+    val bytes = byteArrayOutputStream.toByteArray()
+    val encryptedOut = pgpEncryptedDataGenerator.open(out, bytes.size.toLong())
+    encryptedOut.write(bytes)
+    encryptedOut.close()
+
+    if (armored) {
```
alanpaulin, line 79: What does armored mean in this context?

misk-crypto/src/main/kotlin/misk/crypto/pgp/RealPgpEncoder.kt
```diff
@@ -0,0 +1,133 @@
...
+    pgpCompressedDataGenerator.close()
+
+    val outputStream = buffer.outputStream()
+    val out = when (armored) {
+      true -> ArmoredOutputStream(outputStream)
+      else -> outputStream
+    }
+
+    val jcePGPDataEncryptorBuilder = JcePGPDataEncryptorBuilder(PGPEncryptedData.CAST5)
+        .setWithIntegrityPacket(true)
```
alanpaulin, line 64: What does this do?

alanpaulin (approved):
LGTM with some questions. 
👍👍👍

misk-crypto/src/main/kotlin/misk/crypto/pgp/FakePgpEncoder.kt
```diff
@@ -0,0 +1,23 @@
...
+import javax.inject.Singleton
+
+/**
+ * FakePgpEncoder simply encodes or decodes via base64.
+ */
+@Singleton
+internal class FakePgpEncoder : PgpEncoder {
+  override fun encrypt(
+    byteArray: ByteArray,
+    encryptionKey: PGPPublicKey,
```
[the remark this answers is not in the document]
tso: I think it's okay since it's already validated via the PublicKeyProvider, so it should at least be well-formed

misk-crypto/src/main/kotlin/misk/crypto/pgp/PublicKeyProvider.kt
tso: nah, it just reads weird. I believe the idea is that you might have many public keys in your input stream, but in our case we'll really only ever have one so you can just call `keyRings.next().publicKey`

misk-crypto/src/main/kotlin/misk/crypto/pgp/RealPgpEncoder.kt
tso: it's legal in the sense that it encodes/decodes fine as per the test, but I felt iffy about it as well. any thoughts?

misk-crypto/src/main/kotlin/misk/crypto/pgp/RealPgpEncoder.kt
tso: "Sets whether or not the resulting encrypted data will be protected using an integrity packet."

From https://tools.ietf.org/html/rfc4880#section-5.13: "An implementation MUST support decrypting these packets and SHOULD prefer generating them to the older Symmetrically Encrypted Data packet when possible."

misk-crypto/src/main/kotlin/misk/crypto/pgp/RealPgpEncoder.kt
tso: armored is a PGP specific term for whether we use a binary output or an emailable-text form like ```-----BEGIN PGP MESSAGE-----
Version: BCPG C# v1.6.1.0

hQIMA0tM4ZUzpKCkAQ//ZO8hVhp6LMxshCLqcqgPkXcU1kBHPQjiUBs6QPxaQFcg
PIXJEdZWQ1RkMUKdCfzjvevizi/2NWAwZ797fELdbpVyH6JI2Me1Ov10f8qgYvbG
V0GJ3loRt84KjKdVocKguJVfMvOeMVJanPpxfPt3Ak+Nt/voZHmJdTkV4cjWTmcU
x31xqc5LZcvgF4Lv+ZcMw090b281yMENwALnvQJ9FQg5WTfZ+YautUMwWGk8W8XM
H3tYkdh3jKizy8YTw+zfVDn1yhw2BgSFp0QIgHIQQ2Q2gyknRfj4zdfs/4bZYcu1
KNgj9op/IejZNstP9JWw4labC78nTIW5s1f0LWF5hhZUkxeAOEyVluUVJgEw44B4
LKb0p/8jb4J/dVrsmaTDOONZ48wtNjZOTV61DhCnn+UiNVSfQe6I7XMtFpEH0PN4
WzermmeEgjw1MADeh0jg/9wMM/p9UnTxihGQrPRgHYUu45jI3ys3qArsD2GvAi9y
QYqsuRJNB7EEu2clBRASA3zoWkpqbs3H9S2fB9Sf9ZzZTPCzowOdK1oe0vlMVt3A
mCILe05s2nowwBrBWu4PHUTkllsnLliHhrihu9sqXCHy6u1pdNQMvJgY6kbgQmsG
bVY+brdkLdQ1BxbTT1YixzLm//zqMqGC1ZJUsr5PAIbNtyKK6u6+mCkZj0Y3fU3J
Il0x5V09QcC9AHyxJ8VMOxzhQZjwVYNtD6SGeY8jOyfUScY=
=DHKq
-----END PGP MESSAGE-----
```

File: misk-crypto/src/main/kotlin/misk/crypto/pgp/PgpEncoder.kt [hunk 7]
```diff
@@ -0,0 +1,27 @@
+package misk.crypto.pgp
+
+import org.bouncycastle.openpgp.PGPPublicKey
+
+interface PgpEncoder {
```
tso, line 5: @yoavamit brought up a good point on my branch that perhaps this should really be two different interfaces: an encryptor and an encoder.

I guess in this case we'd always create an encoder for a public key, and always create a decryptor for a private key? Does that seem legit?

tso: I implemented it as suggested. I think it's better this way.

misk-crypto/src/main/kotlin/misk/crypto/pgp/RealPgpEncoder.kt
alanpaulin: I'd say leave it as you have it. We can change it if we discover an issue.

misk-crypto/src/main/kotlin/misk/crypto/pgp/RealPgpEncoder.kt
alanpaulin: 👍

misk-crypto/src/main/kotlin/misk/crypto/CryptoConfig.kt
```diff
@@ -50,5 +51,7 @@ enum class KeyType {
   DIGITAL_SIGNATURE,
   HYBRID_ENCRYPT,
   HYBRID_ENCRYPT_DECRYPT,
-  STREAMING_AEAD
+  STREAMING_AEAD,
+  PGP_DECRYPT,
```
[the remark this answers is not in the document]
yoavamit: I commented about this earlier suggesting to use verbs like PGP_ENCRYPT and PGP_DECRYPT to refer to the operation, not the implementation ;)

misk-crypto/src/main/kotlin/misk/crypto/pgp/FakePgpEncrypterDecrypter.kt
```diff
@@ -0,0 +1,20 @@
+package misk.crypto.pgp
+
+import java.util.Base64
+
+/**
+ * [FakePgpEncrypterDecrypter] uses base64 rather than any PGP encryption/decryption. Since
+ * our private PGP keys are encrypted there's no way to use them in our fakes/tests.
```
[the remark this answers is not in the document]
yoavamit: We have a fake implementation of the AWS KMS API for testing, which "encrypts" and "decrypts" a key.
That way, you can use a real PGP key and test everything full cycle

misk-crypto/src/main/kotlin/misk/crypto/pgp/PgpDecrypterProvider.kt
```diff
@@ -0,0 +1,59 @@
...
+  @Inject private lateinit var kmsClient: KmsClient
+
+  override fun get(): PgpDecrypter {
+    val jsonAdapter = moshi.adapter<PgpKeyJsonFile>()
+    val pgpKeyJsonFile =
+        jsonAdapter.fromJson(key.encrypted_key.value) ?: error("could not deserialize json file")
+
+    val decoded = Base64.getDecoder().decode(pgpKeyJsonFile.encrypted_private_key)
+
+    val secretKeyStream = when (kmsUri) {
```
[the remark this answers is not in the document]
yoavamit: Yeah, this condition shouldn't be here, the provider should assume the KMS URI is never null.
Otherwise, we're introducing here a way to load unencrypted keys which we don't want to.

If this was intended for testing purposes, check out the `FakeKmsClient` class - it should be injected instead of the real client in the tests.

misk-crypto/src/main/kotlin/misk/crypto/pgp/PgpEncrypterProvider.kt
```diff
@@ -0,0 +1,25 @@
...
+internal class PgpEncrypterProvider(
+  private val key: Key,
+  private val isReal: Boolean
+) : Provider<PgpEncrypter>, KeyReader() {
+  override fun get(): PgpEncrypter {
+    val keyIn = BufferedInputStream(key.encrypted_key.value.byteInputStream())
+    val pgpPub =
+        PGPPublicKeyRingCollection(PGPUtil.getDecoderStream(keyIn), JcaKeyFingerprintCalculator())
+    val publicKey = pgpPub.keyRings.next().publicKey
+    return when (isReal) {
```
yoavamit, line 20: This should be taken care of in the testing module, the provider shouldn't have a `isReal` variable

misk-crypto/src/main/kotlin/misk/crypto/pgp/RealPgpDecrypter.kt
```diff
@@ -0,0 +1,70 @@
...
+import org.bouncycastle.openpgp.PGPPublicKeyEncryptedData
+import org.bouncycastle.openpgp.PGPUtil
+import org.bouncycastle.openpgp.jcajce.JcaPGPObjectFactory
+import org.bouncycastle.openpgp.operator.jcajce.JcePublicKeyDataDecryptorFactoryBuilder
+import org.bouncycastle.util.io.Streams
+import java.security.Security
+
+internal class RealPgpDecrypter(private val privateKey: PGPPrivateKey) : PgpDecrypter {
+  init {
+    Security.addProvider(BouncyCastleProvider())
```
yoavamit, line 19: I'd move this statement to `MiskCryptoModule` where the rest of the Tink primitives are initialized.

misk-crypto/src/main/kotlin/misk/crypto/pgp/RealPgpDecrypter.kt
```diff
@@ -0,0 +1,70 @@
...
+  }
+
+  override fun decrypt(byteArray: ByteArray): ByteArray {
```
yoavamit, line 22: nit: rename `byteArray` to `ciphertext`

misk-crypto/src/main/kotlin/misk/crypto/pgp/PgpEncoder.kt
```diff
@@ -0,0 +1,57 @@
...
+   * Without we would get binary data which isn't viewable in text format.
+   */
+  fun encrypt(byteArray: ByteArray, armored: Boolean): ByteArray
+}
+
+/**
+ * A [PgpDecrypter] is tied to a private PGP key which it uses to decrypt
+ * messages.
+ */
+interface PgpDecrypter {
```
yoavamit, line 41: super nit: is it `Decrypter` or `Decryptor`?
I've seen people write it both ways... 🤷

misk-crypto/src/main/kotlin/misk/crypto/pgp/RealPgpDecrypter.kt
```diff
@@ -0,0 +1,70 @@
...
+      is PGPOnePassSignatureList ->
+        throw PGPException("encrypted message contains a signed message - not literal data.")
+
+      else -> throw PGPException("message is not a simple encrypted file - type unknown.")
+    }
+    if (pgpPublicKeyEncryptedData.isIntegrityProtected && !pgpPublicKeyEncryptedData.verify()) {
+      error("message failed integrity check")
+    }
+
+    return buffer.readByteArray()
```
yoavamit, line 68: Please make sure all streams used in this method are properly closed and that buffers that might contain plaintext are overwritten (for example using `bytes.fill(0)`

misk-crypto/src/main/kotlin/misk/crypto/pgp/RealPgpEncrypter.kt
```diff
@@ -0,0 +1,70 @@
...
+import java.security.SecureRandom
+import java.security.Security
+import java.util.Date
+
+internal class RealPgpEncrypter(private val encryptionKey: PGPPublicKey) : PgpEncrypter {
+  init {
+    Security.addProvider(BouncyCastleProvider())
+  }
+
+  override fun encrypt(byteArray: ByteArray, armored: Boolean): ByteArray {
```
yoavamit, line 24: nit: rename `byteArray` to `cleartext`/`plaintext`

misk-crypto/src/test/kotlin/misk/crypto/PgpKeyTest.kt
```diff
@@ -0,0 +1,92 @@
...
+
+    // The fake pgp encrypter/decrypter only uses base64.
+    val encryptedBytes = pgpEncrypter.encrypt(secretMessage.toByteArray(), true)
+    assertThat(encryptedBytes.decodeToString()).isEqualTo(secretMessageBase64)
+
+    val decryptedBytes = pgpDecrypter.decrypt(encryptedBytes)
+    assertThat(decryptedBytes.decodeToString()).isEqualTo(secretMessage)
+  }
+
+  private fun getPgpDecrypter(filePath: String, real: Boolean): PgpDecrypter {
```
yoavamit, line 64: Checkout the MiskTestModule.

When you initialize the keys like this, you don't actually use the service modules to test this.
Instead, the keys should be provided and loaded via the configuration, so you'll actually test the entire flow (loading the key from the configuration, "decrypting" it, getting it from the provider and using it)

misk-crypto/src/main/kotlin/misk/crypto/pgp/RealPgpDecrypter.kt
```diff
@@ -0,0 +1,70 @@
...
+
+  override fun decrypt(byteArray: ByteArray): ByteArray {
+    val buffer = okio.Buffer()
+    val outputStream = buffer.outputStream()
+    val decoderStream = PGPUtil.getDecoderStream(byteArray.inputStream())
+    val jcaPGPObjectFactory = JcaPGPObjectFactory(decoderStream)
+
+    val pgpEncryptedDataList = when (val nextObject = jcaPGPObjectFactory.nextObject()) {
+      is PGPEncryptedDataList -> nextObject
+      else -> jcaPGPObjectFactory.nextObject() as PGPEncryptedDataList
```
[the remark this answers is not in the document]
tso: no idea, bouncy castle is obscure and ultimately copying the examples they have seems to be the best way

misk-crypto/src/main/kotlin/misk/crypto/pgp/RealPgpDecrypter.kt
```diff
@@ -0,0 +1,70 @@
...
+
+    val publicKeyDataDecryptorFactory = JcePublicKeyDataDecryptorFactoryBuilder()
+        .setProvider("BC")
+        .setContentProvider("BC")
+        .build(privateKey)
+
+    val clear = pgpPublicKeyEncryptedData.getDataStream(publicKeyDataDecryptorFactory)
+    val plainFactory = JcaPGPObjectFactory(clear)
+    var message = plainFactory.nextObject()
+    if (message is PGPCompressedData) {
```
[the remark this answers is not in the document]
tso: i believe we support both in this implementation. i don't know why the code is like this, see the other comment about bouncy castle being bizarre

misk-crypto/src/main/kotlin/misk/crypto/pgp/FakePgpEncrypterDecrypter.kt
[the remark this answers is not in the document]
tso: i don't agree with this approach as it makes testing harder. ultimately we have tests for the full flow already inside of this package so i don't see why services outside of misk-crypto should need to verify that the PGP implementation functions properly.

misk-crypto/src/main/kotlin/misk/crypto/pgp/PgpDecrypterProvider.kt
```diff
@@ -0,0 +1,59 @@
...
+  private val isReal: Boolean
+) : Provider<PgpDecrypter>, KeyReader() {
+  @Inject private lateinit var kmsClient: KmsClient
+
+  override fun get(): PgpDecrypter {
+    val jsonAdapter = moshi.adapter<PgpKeyJsonFile>()
+    val pgpKeyJsonFile =
+        jsonAdapter.fromJson(key.encrypted_key.value) ?: error("could not deserialize json file")
+
+    val decoded = Base64.getDecoder().decode(pgpKeyJsonFile.encrypted_private_key)
```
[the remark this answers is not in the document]
tso: i could do `pgpKeyJsonFile.encrypted_private_key.decodeBase64()!!.toByteArray()` i guess, though it doesn't really read much better

misk-crypto/src/main/kotlin/misk/crypto/pgp/PgpEncrypterProvider.kt
tso: well, whether or not i make a `RealPgpEncrypterProvider.kt` and `FakePgpEncrypterProvider.kt` or use a boolean inside a single provider seems kind of irrelevant to me?

can you explain more as to what you're envisioning here?

misk-crypto/src/test/kotlin/misk/crypto/PgpKeyTest.kt
tso: well there's no tests around that anywhere in misk-crypto as far as i can tell? so i don't know why we need to add them specifically for this pr

tso: i guess there's marginally more testing with ```private fun getInjector(keyMap: List<Pair<String, KeysetHandle>>): Injector {
    val keys = keyMap.map {
      var keyType = KeyType.AEAD
      val keyTypeUrl = it.second.keysetInfo.getKeyInfo(0).typeUrl
      if (keyTypeUrl.endsWith("HmacKey")) {
        keyType = KeyType.MAC
      } else if (keyTypeUrl.endsWith("Ed25519PrivateKey")) {
        keyType = KeyType.DIGITAL_SIGNATURE
      } else if (keyTypeUrl.endsWith("EciesAeadHkdfPrivateKey")) {
        keyType = KeyType.HYBRID_ENCRYPT_DECRYPT
      } else if (keyTypeUrl.endsWith("EciesAeadHkdfPublicKey")) {
        keyType = KeyType.HYBRID_ENCRYPT
      } else if (keyTypeUrl.endsWith("AesSivKey")) {
        keyType = KeyType.DAEAD
      } else if (keyTypeUrl.endsWith("AesGcmHkdfStreamingKey")) {
        keyType = KeyType.STREAMING_AEAD
      }
      Key(it.first, keyType, generateEncryptedKey(it.second))
    }
    val config = CryptoConfig(keys, "test_master_key")
    return Guice.createInjector(CryptoTestModule(), CryptoModule(config), DeploymentModule.forTesting())
  }``` but like....not really

misk-crypto/src/main/kotlin/misk/crypto/pgp/PgpDecrypterProvider.kt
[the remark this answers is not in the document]
AlexSzlavik: I guess you still need it as a byte array in the end, so not really saving anything. Unclear to me whether this would provide any cleaner error handling.

misk-crypto/src/main/kotlin/misk/crypto/pgp/RealPgpDecrypter.kt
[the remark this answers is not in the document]
AlexSzlavik: I'm not a big fan of the argument "It's hard to reason about, so let's assume it's fine". Have you tried this code with any GPG generated payloads? What options have you tried? I'll try to read a bit about BCs implementation tomorrow.

misk-crypto/src/main/kotlin/misk/crypto/pgp/RealPgpDecrypter.kt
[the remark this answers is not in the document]
tso: no, we don't only support compressed data. this covers the case of compressed data, but you can see below that we carry on in the case of uncompressed as well.

Issue #1613, opened 2020-09-28
QueryPager adds duplicated constraints and order


mightyguava:
Query pagination adds an additional constraint and an additional ORDER term for every page. We end up building up monster queries like

`
select dbbankingt0_.id, dbbankingt0_.customer_id, dbbankingt0_.created_at, dbbankingt0_.display_date, dbbankingt0_.merchant_id, dbbankingt0_.token, dbbankingt0_.transaction_data, dbbankingt0_.updated_at, dbbankingt0_.version 
from banking_transactions dbbankingt0_ 
where dbbankingt0_.customer_id = ? and dbbankingt0_.id > ? and dbbankingt0_.id > ? and dbbankingt0_.id > ? and dbbankingt0_.id > ? and dbbankingt0_.id > ? and dbbankingt0_.id > ? and dbbankingt0_.id > ? and dbbankingt0_.id > ? and dbbankingt0_.id > ? and dbbankingt0_.id > ? and dbbankingt0_.id > ? and dbbankingt0_.id > ? and dbbankingt0_.id > ? and dbbankingt0_.id > ? and dbbankingt0_.id > ? and dbbankingt0_.id > ? and dbbankingt0_.id > ? and dbbankingt0_.id > ? and dbbankingt0_.id > ? and dbbankingt0_.id > ? and dbbankingt0_.id > ? and dbbankingt0_.id > ? and dbbankingt0_.id > ? and dbbankingt0_.id > ? and dbbankingt0_.id > ? and dbbankingt0_.id > ? and dbbankingt0_.id > ? and dbbankingt0_.id > ? and dbbankingt0_.id > ? and dbbankingt0_.id > ? and dbbankingt0_.id > ? and dbbankingt0_.id > ? and dbbankingt0_.id > ? and dbbankingt0_.id > ? and dbbankingt0_.id > ? and dbbankingt0_.id > ? and dbbankingt0_.id > ? and dbbankingt0_.id > ? and dbbankingt0_.id > ? and dbbankingt0_.id > ? and dbbankingt0_.id > ? and dbbankingt0_.id > ? and dbbankingt0_.id > ? and dbbankingt0_.id > ? and dbbankingt0_.id > ? and dbbankingt0_.id > ? and dbbankingt0_.id > ? and dbbankingt0_.id > ? and dbbankingt0_.id > ? 
[...]

PR #1614, opened 2020-09-28
Pager does not generate many duplicate where conditions and order by …

mightyguava:
…columns

Adds a `clone()` method to the Query so that the Pager can modify a copy
of the query instead of the original.

Fixes https://github.com/cashapp/misk/issues/1613

### DIFF
--- a/misk-hibernate/src/main/kotlin/misk/hibernate/ReflectionQueryFactory.kt
+++ b/misk-hibernate/src/main/kotlin/misk/hibernate/ReflectionQueryFactory.kt
@@ -68,6 +68,28 @@ internal class ReflectionQuery<T : DbEntity<T>>(
     disabledChecks += check
   }
 
+  override fun <Q: Query<*>> clone(): Q {
+    val copy = ReflectionQuery(
+        this.queryClass,
+        this.rootEntityType,
+        this.queryMethodHandlers,
+        this.tracer,
+        this.queryLimitsConfig
+    )
+    if (copy.maxRows != -1) {
+      copy.maxRows = this.maxRows
+    }
+    if (copy.firstResult != 0) {
+      copy.firstResult = this.firstResult
+    }
+    copy.constraints.addAll(this.constraints)
+    copy.orderFactories.addAll(this.orderFactories)
+    copy.disabledChecks.addAll(this.disabledChecks)
+
+    @Suppress("UNCHECKED_CAST")
+    return copy.toProxy() as Q
+  }
+
   override fun dynamicAddConstraint(path: String, operator: Operator, value: Any?) {
     val pathList = path.split('.')
     addConstraint(pathList, operator, value)
--- a/misk-hibernate/src/main/kotlin/misk/hibernate/pagination/RealPager.kt
+++ b/misk-hibernate/src/main/kotlin/misk/hibernate/pagination/RealPager.kt
@@ -2,6 +2,7 @@ package misk.hibernate.pagination
 
 import misk.hibernate.DbEntity
 import misk.hibernate.Query
+import misk.hibernate.ReflectionQuery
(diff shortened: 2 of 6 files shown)


misk-hibernate/src/test/kotlin/misk/hibernate/pagination/RealPagerTest.kt
```diff
@@ -10,19 +10,23 @@ import misk.hibernate.Query
...
+import java.time.Instant
 import java.time.LocalDate
 import javax.inject.Inject
 
 @MiskTest(startService = true)
 class RealPagerTest {
 
   @MiskTestModule
-  val module = MoviesTestModule()
+  val module = MoviesTestModule(DataSourceType.MYSQL)
```
mightyguava, line 29: The default `VITESS_MYSQL` data source doesn't like `SET GLOBAL`, used to enable the general query log

misk-hibernate/src/main/kotlin/misk/hibernate/ReflectionQueryFactory.kt
```diff
@@ -68,6 +68,28 @@ internal class ReflectionQuery<T : DbEntity<T>>(
     disabledChecks += check
   }
 
+  override fun <Q: Query<*>> clone(): Q {
```
mightyguava, line 71: This is a scary hack but it works and I couldn't think of a better way of doing thiis

zcross: Mind adding a simple regression test?

misk-hibernate/src/test/kotlin/misk/hibernate/pagination/RealPagerTest.kt
```diff
@@ -120,6 +124,62 @@ class RealPagerTest {
...
+      }
+    }
+
+    val pageQueries = queries.reversed().filter { it.contains("select") }.map {
+      Regex("""/\* (.*) \*/""").find(it)!!.groups[1]!!.value
+    }
+    assertThat(pageQueries).hasSize(3)
+    // The first page doesn't have an (id > ?) condition
+    assertThat(countClauses(pageQueries[0])).isEqualTo(Clauses(1, 1))
+    // The second page each have exactly one (id > ?) condition and one ORDER BY column
```
zcross, line 163: Excellent test coverage

misk-hibernate/src/main/kotlin/misk/hibernate/pagination/RealPager.kt
```diff
@@ -19,6 +20,8 @@ internal class RealPager<T : DbEntity<T>, Q: Query<T>>(
   }
 
   override fun nextPage(session: Session): Page<T>? {
+    @Suppress("UNCHECKED_CAST")
+    val query = query.clone<Q>()
```
zcross, line 24: WDYT about renaming either the local val or the "original" query val?

zcross (approved):
LGTM, thanks for fixing so this soon after spotting the symptoms

misk-hibernate/src/main/kotlin/misk/hibernate/pagination/RealPager.kt
mightyguava: I did this to purposefully shadow the "original" query val, so you can't accidentally touch it in this function

misk-hibernate/src/main/kotlin/misk/hibernate/ReflectionQueryFactory.kt
mightyguava: Yeah good idea, added a test

swankjesse (commented):
Nice fix. Thank you

PR #1615, opened 2020-09-30
Removing duplicate BouncyCastleProvider instance, adding assumption d…

AlexSzlavik:
…ocumentation

### DIFF
--- a/misk-crypto/src/main/kotlin/misk/crypto/pgp/RealPgpDecrypter.kt
+++ b/misk-crypto/src/main/kotlin/misk/crypto/pgp/RealPgpDecrypter.kt
@@ -12,6 +12,14 @@ import org.bouncycastle.openpgp.jcajce.JcaPGPObjectFactory
 import org.bouncycastle.openpgp.operator.jcajce.JcePublicKeyDataDecryptorFactoryBuilder
 import org.bouncycastle.util.io.Streams
 
+/**
+ * RealPgpDecrypter is a simple PGP decryption helper.
+ * The implementation makes a  couple key assumptions:
+ * - The input data is public key encrypted.
+ * - The input data contains only ONE encrypted payload.
+ * - The first OR second packet in the input must be the encryption method packet.
+ * - No signature verification.
+ */
 internal class RealPgpDecrypter(
   private val privateKeys: Map<Long, PGPPrivateKey>
 ) : PgpDecrypter {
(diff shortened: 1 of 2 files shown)


misk-crypto/src/main/kotlin/misk/crypto/pgp/RealPgpDecrypter.kt
```diff
@@ -12,6 +12,14 @@ import org.bouncycastle.openpgp.jcajce.JcaPGPObjectFactory
 import org.bouncycastle.openpgp.operator.jcajce.JcePublicKeyDataDecryptorFactoryBuilder
 import org.bouncycastle.util.io.Streams
 
+/**
+ * RealPgpDecrypter is a simple PGP decryption helper.
+ * The implementation makes a  couple key assumptions:
+ * - The input data is public key encrypted.
+ * - The input data contains only ONE encrypted payload.
```
tso, line 19: I'm not sure this is true. We do however assume that _we_ are the only recipient currently.

Specifically we would need to modify this to search through each until we find data encrypted via a key we have the private key for:
```    val pgpPublicKeyEncryptedData =
        pgpEncryptedDataList.encryptedDataObjects.next() as PGPPublicKeyEncryptedData

    val pgpSecretKey = privateKeys[pgpPublicKeyEncryptedData.keyID]
        ?: error("no private key able to decrypt this message")```

File: misk-crypto/src/main/kotlin/misk/crypto/pgp/RealPgpDecrypter.kt [hunk 2]
```diff
@@ -12,6 +12,14 @@ import org.bouncycastle.openpgp.jcajce.JcaPGPObjectFactory
...
+ * - The first OR second packet in the input must be the encryption method packet.
```
tso, line 20: Is this any different than "the input data is PGP encrypted"? I would imagine that this is literally required by the format

AlexSzlavik: I mean, it's dumb, but the spec doesn't point it out. Depending on the PGP client used to generate the payload, you could have none encryption packets before the actual data. In that case, instead of ignoring them, we'd fail ungracefully (since the cast would fail)

tso (approved):

PR #1621, opened 2020-10-13
Add basic CompressionCodec abstraction and a few impls

zcross:
When I have some downtime, I want to write something something to support compression on our Hibernate (and other persistence integrations) field definitions... Something like:

```
@Column(nullable = false)
@ProtoColumn
@Compressed(codec = SnappyCompressionCodec::class)
var some_cool_data: LargeCoolDataType
```

But also something like this, where the compression is applied to the plaintext (and not counterproductively to the ciphertext):
```
@Column(nullable = false)
@Compressed(codec = SnappyCompressionCodec::class)
@SecretColumn(key_name = "super-secure-key")
var large_sensitive_data: ByteArray // Nice-to-have: proto-aware SecretColumn
```

Here's the first step of just cribbing some existing compression codec wrappers we have that should play nice with our Hibernate user types.

### DIFF
--- a/misk-core/build.gradle
+++ b/misk-core/build.gradle
@@ -15,6 +15,7 @@ dependencies {
   compile dep.okio
   compile dep.okHttp
   compile dep.slf4jApi
+  compile dep.snappy
 
   compile project(':misk-inject')
   compile project(':misk-service')
--- a/misk-core/src/main/kotlin/misk/compress/DeflateCompressionCodec.kt
+++ b/misk-core/src/main/kotlin/misk/compress/DeflateCompressionCodec.kt
@@ -1,0 +1,52 @@
+package com.squareup.misk.compress
+
+import java.io.ByteArrayOutputStream
+import java.io.IOException
+import java.util.zip.Deflater
+import java.util.zip.Deflater.DEFAULT_COMPRESSION
+import java.util.zip.DeflaterOutputStream
+import java.util.zip.InflaterOutputStream
+
+/**
+ * A {@link CompressionCodec} implementation for the deflate compression format. It uses the
+ * implementation provided in the standard {@code java.util.zip} package.
+ *
+ * Cribbed from SQ/repos/java/browse/compression.
+ */
+class DeflateCompressionCodec(private val compressionLevel: Int = DEFAULT_COMPRESSION) :
+    CompressionCodec {
+  init {
+    check(-1 <= compressionLevel && compressionLevel <= 9) {
+      "Invalid compression level $compressionLevel"
+    }
+  }
+
+  override fun encode(input: ByteArray): ByteArray {
+    val out = ByteArrayOutputStream()
+    val deflater = Deflater(compressionLevel)
+    try {
+      DeflaterOutputStream(out, deflater).use { compressed -> compressed.write(input) }
+    } catch (e: IOException) {
+      throw RuntimeException(e)
+    } finally {
+      deflater.end()
+    }
+    return out.toByteArray()
+  }
+
+  /**
+   * Decompresses `input` bytes using a [InflaterOutputStream].
+   *
+   * @param input bytes to be decoded
+   * @return decoded bytes
+   */
+  override fun decode(input: ByteArray): ByteArray {
+    val out = ByteArrayOutputStream()
+    try {
+      InflaterOutputStream(out).use { inflate -> inflate.write(input) }
+    } catch (e: IOException) {
+      throw RuntimeException(e)
+    }
+    return out.toByteArray()
+  }
+}
--- a/misk-core/src/test/kotlin/misk/compress/CompressionCodecTest.kt
+++ b/misk-core/src/test/kotlin/misk/compress/CompressionCodecTest.kt
@@ -1,0 +1,28 @@
+package com.squareup.misk.compress
+
+import org.assertj.core.api.Assertions.assertThat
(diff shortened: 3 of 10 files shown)


misk-core/build.gradle
```diff
@@ -15,6 +15,7 @@ dependencies {
   compile dep.okio
   compile dep.okHttp
   compile dep.slf4jApi
+  compile dep.snappy
```
swankjesse, line 18: I’d love to figure out a way to do this outside of misk-core. I’d like for that module to get smaller, not bigger

misk-core/src/main/kotlin/misk/compress/DeflateCompressionCodec.kt
```diff
@@ -0,0 +1,52 @@
...
+import java.util.zip.Deflater
+import java.util.zip.Deflater.DEFAULT_COMPRESSION
+import java.util.zip.DeflaterOutputStream
+import java.util.zip.InflaterOutputStream
+
+/**
+ * A {@link CompressionCodec} implementation for the deflate compression format. It uses the
+ * implementation provided in the standard {@code java.util.zip} package.
+ *
+ * Cribbed from SQ/repos/java/browse/compression.
```
swankjesse, line 14: Not particularly useful for OSS

misk-core/src/main/kotlin/misk/compress/DeflateCompressionCodec.kt
```diff
@@ -0,0 +1,52 @@
...
+ */
+class DeflateCompressionCodec(private val compressionLevel: Int = DEFAULT_COMPRESSION) :
+    CompressionCodec {
+  init {
+    check(-1 <= compressionLevel && compressionLevel <= 9) {
```
swankjesse, line 19: require –> you passed a bad argument
check –> you can’t call this now

misk-core/src/main/kotlin/misk/compress/DeflateCompressionCodec.kt
```diff
@@ -0,0 +1,52 @@
...
+      "Invalid compression level $compressionLevel"
+    }
+  }
+
+  override fun encode(input: ByteArray): ByteArray {
+    val out = ByteArrayOutputStream()
```
swankjesse, line 25: if you want better performance, consider `Buffer` and `DeflaterSink`. ([details](https://publicobject.com/2020/07/26/optimizing-new-byte/))

misk-core/src/test/kotlin/misk/compress/CompressionCodecTest.kt
```diff
@@ -0,0 +1,28 @@
...
+      "0f6a64771ec115039851676c6541d24b8758684a803ae9aff45a0bd6cf5f7039",
+      "9eee9f73ae587affa120bcc89a998a0e10d29581762177b5ac93f560ae829a48efc9fdb01a2f60a956ce6f2ec64a66d1db79abd38d85c39b07652c9f5e9459b3",
+      "repetition repetition repetition repetition",
+      "mostly distinct letters"
+  ).map { it.toByteArray() }
+
+  @TestFactory
+  fun `test reversibility of small input`(): List<DynamicTest> {
+    return fixtureValues.mapIndexed { index, testFixture ->
+      DynamicTest.dynamicTest("test fixture $index") {
```
swankjesse, line 22: neato

swankjesse (commented):
I think I’d prefer `@SnappyCompressed` or `@DeflateCompressed` rather than fully pluggable compression codecs. The world market for compression algorithms is near to 1. We probably want Deflate and perhaps Brotli.

swankjesse:
oooh neat https://quixdb.github.io/squash-benchmark/#results

misk-core/src/main/kotlin/misk/compress/DeflateCompressionCodec.kt
zcross: Thanks!

misk-core/src/main/kotlin/misk/compress/DeflateCompressionCodec.kt
zcross: Whoopsies, caught in the act. This was a skim PR first :)

misk-core/build.gradle
zcross: Do we have a tiny utility module yet?

misk-core/src/main/kotlin/misk/compress/DeflateCompressionCodec.kt
zcross: > I think I’d prefer `@SnappyCompressed` or `@DeflateCompressed` rather than fully pluggable compression codecs. The world market for compression algorithms is near to 1. We probably want Deflate and perhaps Brotli.

That seems reasonable – note that I've not really locked in the annotations yet

zcross:
> I think I’d prefer `@SnappyCompressed` or `@DeflateCompressed` rather than fully pluggable compression codecs. The world market for compression algorithms is near to 1. We probably want Deflate and perhaps Brotli.

Works for me – fortunately I haven't added anything annotation-like in this PR, just an interface and implementations

zcross:
This PR is such a thin wrapper around other code that I think it's useless outside of its intended context. I'll reopen something when I have the time to actually integrate it with our DB related utilities which is my actual end goal.

PR #1628, opened 2020-10-26
Combines FakeJobQueue and FakeTransactionalJobQueue to make testing easier

tso:


misk-jobqueue-testing/src/main/kotlin/misk/jobqueue/FakeTransactionalJobQueue.kt
```diff
@@ -28,12 +22,10 @@ import javax.inject.Singleton
  * ```
  */
 @Singleton
+@Deprecated("Please use FakeJobQueue instead")
 class FakeTransactionalJobQueue @Inject constructor(
-  private val jobHandlers: Provider<Map<QueueName, JobHandler>>,
-  private val tokenGenerator: TokenGenerator
+  private val fakeJobQueue: FakeJobQueue
 ) : TransactionalJobQueue {
```
swankjesse, line 28:
```
) : TransactionalJobQueue by fakeJobQueue {
}
```

swankjesse (approved):
Nice.

misk-jobqueue-testing/src/main/kotlin/misk/jobqueue/FakeTransactionalJobQueue.kt
tso: oh that's really cool! but doesn't work for everything since we have fake specific methods that aren't on `TransactionalJobQueue`. I removed the enqueue stuff using that though

PR #1629, opened 2020-10-27
Don't retry on HTTP connection failures

swankjesse:
This is not that useful in server apps where there are already
facilities to retry on connectivity failures.


davidgarth (approved):
this makes sense to me, most of our services retry anyway

PR #1630, opened 2020-10-27
Use service instead of principal in MetricsInterceptor

sfavron:
When recording the caller for http_request_latency_ms, we should only be
recording the calling service name, if it exists, not a username. Since
`caller.get()?.principal` will return service or username, instead use
just `.service`. Having the username tagged on these metrics causes
cardinality explosion.

### DIFF
--- a/misk/src/main/kotlin/misk/web/interceptors/MetricsInterceptor.kt
+++ b/misk/src/main/kotlin/misk/web/interceptors/MetricsInterceptor.kt
@@ -22,7 +22,7 @@ internal class MetricsInterceptor internal constructor(
     val elapsedTimeMillis = elapsedTime.toMillis().toDouble()
     val callingPrincipal = when {
       caller.get()?.service != null -> caller.get()?.service!!
-      caller.get()?.user != null -> "user"
+      caller.get()?.user != null -> "<user>"
       else -> "unknown"
     }
 
--- a/misk/src/main/kotlin/misk/web/interceptors/MetricsInterceptor.kt
+++ b/misk/src/main/kotlin/misk/web/interceptors/MetricsInterceptor.kt
@@ -20,10 +20,14 @@ internal class MetricsInterceptor internal constructor(
     val (elapsedTime, result) = timed { chain.proceed(chain.httpCall) }
 
     val elapsedTimeMillis = elapsedTime.toMillis().toDouble()
-    val callingService = caller.get()?.service ?: "unknown"
+    val callingPrincipal = when {
+      caller.get()?.service != null -> caller.get()?.service!!
+      caller.get()?.user != null -> "user"
+      else -> "unknown"
+    }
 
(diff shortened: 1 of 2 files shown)


misk/src/main/kotlin/misk/web/interceptors/MetricsInterceptor.kt
```diff
@@ -20,10 +20,10 @@ internal class MetricsInterceptor internal constructor(
     val (elapsedTime, result) = timed { chain.proceed(chain.httpCall) }
 
     val elapsedTimeMillis = elapsedTime.toMillis().toDouble()
-    val callingPrincipal = caller.get()?.principal ?: "unknown"
+    val callingService = caller.get()?.service ?: "unknown"
```
mightyguava, line 23: we should still tag user as something, like `<user>`

clarkds14, line 23: add a test that creates a request with `USER_HEADER` set and show that unknown is set in the metric?

clarkds14 (changes requested):

misk/src/main/kotlin/misk/web/interceptors/MetricsInterceptor.kt
```diff
@@ -20,10 +20,10 @@ internal class MetricsInterceptor internal constructor(
...
 
     val statusCode = chain.httpCall.statusCode
-    requestDuration.record(elapsedTimeMillis, actionName, callingPrincipal, statusCode.toString())
+    requestDuration.record(elapsedTimeMillis, actionName, callingService, statusCode.toString())
```
clarkds14, line 26: add a test that creates a request with USER_HEADER set and show that unknown is set in the metric?

misk/src/main/kotlin/misk/web/interceptors/MetricsInterceptor.kt
```diff
@@ -20,7 +20,11 @@ internal class MetricsInterceptor internal constructor(
     val (elapsedTime, result) = timed { chain.proceed(chain.httpCall) }
 
     val elapsedTimeMillis = elapsedTime.toMillis().toDouble()
-    val callingPrincipal = caller.get()?.principal ?: "unknown"
+    val callingPrincipal = when {
+      caller.get()?.service != null -> caller.get()?.service!!
+      caller.get()?.user != null -> "user"
```
mightyguava, line 25: This should have some special characters in there to indicate it's "special" and not conflict with a service potentially named "user", hence why I suggested `<user>`

sfavron: 👍🏻 will update

clarkds14 (approved):

mightyguava (approved):

PR #1633, opened 2020-11-02
Move TokenGenerator to misk-core

swankjesse:
It introduces no additional dependencies.


szabado:
Will this be a breaking change to services that depend on misk but not misk-core?

szabado (approved):
LGTM, I trust that this won't be a breaking change

swankjesse:
Not a breaking change because Misk has a `compile` dependency on `misk-core`.

PR #1638, opened 2020-11-09
Update DatabaseQueryMetadata to match tab

adrw:


misk/src/main/kotlin/misk/web/metadata/DatabaseQueryMetadata.kt
```diff
@@ -2,12 +2,14 @@ package misk.web.metadata
 
 /** Metadata front end model for Database Query Misk-Web Tab */
 data class DatabaseQueryMetadata(
-  val capabilities: Set<String> = setOf(),
-  val services: Set<String> = setOf(),
+  val allowedCapabilities: Set<String> = setOf(),
```
adrw, line 5: This matches existing Misk-Web patterns

misk/src/main/kotlin/misk/web/metadata/DatabaseQueryMetadata.kt
```diff
@@ -2,12 +2,14 @@ package misk.web.metadata
...
+  val allowedServices: Set<String> = setOf(),
   /** SQL table name */
   val table: String,
   /** DbTable entity class */
   val entityClass: String,
+  /** Query class */
+  val queryClass: String,
```
adrw, line 12: This will provide the Query class (ie. `MovieQueries`) that should be shown in the UI

acostama (approved):

PR #1640, opened 2020-11-10
Update DatabaseQueryMetadata interface

adrw:
Better match Misk-Web Form Builder and provide more information for Constraints, Orders, Selects


misk/src/main/kotlin/misk/web/metadata/DatabaseQueryMetadata.kt
```diff
@@ -40,6 +43,28 @@ data class DatabaseQueryMetadata(
...
+    override val parametersType: String,
+    val path: String,
+    val operator: String
+  ) : FunctionMetadata
+
+  data class OrderMetadata(
+    override val name: String,
+    override val parametersType: String,
+    val path: String,
+    val operator: String
```
acostama, line 61: what is operator for?

adrw: Example constraint

```
@Constraint(path = "release_date", operator = Operator.EQ)
 fun releaseDateEqualTo(upperBound: LocalDate?): OperatorsMovieQuery
```

acostama (approved):

misk/src/main/kotlin/misk/web/metadata/DatabaseQueryMetadata.kt
```diff
@@ -40,6 +43,28 @@ data class DatabaseQueryMetadata(
...
       constraints = constraints,
       orders = orders,
-      selects = selects
+      selects = selects,
+      types = types
   )
+
+  data class ConstraintMetadata(
+    override val name: String,
+    override val parametersType: String,
```
acostama, line 52: should we continue to use field for the parameter? 
the other options is to include a separate field for repeated constraints

PR #1641, opened 2020-11-12
Remove custom trace instrumentation for hibernate

clarkds14:
A number of "screwed up" Hibernate related spans is because of the mix of auto and custom instrumentation done for Hibernate that don't quite work well together. Since we auto-instrument Hibernate using the dd-java-tracer, let's remove the custom instrumentation.


mightyguava:
Tear it all down!

mightyguava (approved):

PR #1642, opened 2020-11-12
split misk-jdbc from misk-hibernate

mightyguava:
The great split is finally happening! The hibernate package is a convoluted tangled mess, so I had to move quite a few things around.

Summary of changes

* The SchemaMigrator, PingDatabaseService, and StartDatabaseService are moved into the `jdbc` package
* A `JdbcModule` is added to install just the basic necessities for to get a migrated database and a `DataSource`
* The entire `jdbc` package is moved into the `misk-jdbc` module, which `misk-hibernate` depends on and exports
* SchemaMigrator no longer uses a Transacter. Since we default `autocommit = false`, SchemaMigrator commits its own changes without a transaction. This is probably fine since it only does DDL in tests
* Vitess-specific things are really tangled up so I ended up splitting the stuff that `jdbc` needs to depend on into the `jdbc` package, and leaving the rest.
* Logic to query shards for Vitess has been duplicated from the Transacter to SchemaMigrator. Aside from that there is no support in the `jdbc` package for querying Vitess shards directly.

There are no breaking API changes, but some classes have been moved which may cause breakages. Fortunately these are all classes applications should never have to depend on, so it's probably fine.


.circleci/config.yml
```diff
@@ -74,7 +74,7 @@ jobs:
           name: "Build and test"
           command: |
             if [ "$CIRCLE_NODE_INDEX" -eq 1 ]; then
-              ./gradlew testShardHibernate -i --scan --parallel --build-cache
```
mightyguava, line 77: we can no longer run the tests in parallel because now more than one package depends on the docker containers and they don't know how to coordinate across separate jvms

.circleci/config.yml
```diff
@@ -74,7 +74,7 @@ jobs:
...
+              ./gradlew testShardHibernate -i --scan --build-cache
```
mightyguava, line 77: we can no longer run the tests in parallel because now more than one package depends on the docker containers and they don't know how to coordinate across separate jvms. Separate runs will clobber each other

misk-jdbc/src/main/kotlin/misk/jdbc/JdbcModule.kt
```diff
@@ -0,0 +1,177 @@
...
+import javax.sql.DataSource
+import kotlin.reflect.KClass
+
+/**
+ * Binds database connectivity for a qualified data source. This binds the following public types:
+ *
+ *  * @Qualifier [misk.jdbc.DataSourceConfig]
+ *
+ * This also registers services to connect to the database ([SessionFactoryService]) and to verify
+ * that the schema is up-to-date ([SchemaMigratorService]).
```
swankjesse, line 26: nice docs

misk-jdbc/src/main/kotlin/misk/jdbc/JdbcModule.kt
```diff
@@ -0,0 +1,177 @@
...
+ */
```
swankjesse, line 27: if you move this documentation to immediately precede JdbcModule, then it’ll be ‘attached’ to that in Javadoc and IDE integration. As-is it’s attached to `MAX_MAX_ROWS`.

swankjesse (approved):
Love it.

misk-jdbc/src/main/kotlin/misk/jdbc/JdbcModule.kt
mightyguava: 😂these are lies! I forgot to update them after copying from HibernateModule

tirsen:
omg what a heroic effort!

.circleci/config.yml
tirsen: Yeah I've been thinking about this. In another project I've been running database docker containers without explicitly specified exported ports. (You export the port but don't tell docker what port number to use.) In that case docker will allocate randomly selected ports which you can look up using the API. Based on that we should be able to run many tests in parallel.

This is obviously a change that should be done outside this PR. :-) It could be something for next hack week! :-)

PR #1664, opened 2020-11-24
Dependency updates via polyrepo dep-upgrade

chris-ryan-square:


chris-ryan-square:
Too many problems with updating all the dependencies - going to close this and just add the new dependency by itself.

PR #1665, opened 2020-11-25
Add specific dynamodb dependency and usages

chris-ryan-square:
Dependency added via polyrepo; manually applying to dependencies.gradle (and build.gradle) for usage - to avoid other updates to dependencies being done at the same time.


adrw (approved):
LG

PR #1666, opened 2020-11-25
Binds other Transacter instance using reader qualifier and datasource

frojasg:


misk-hibernate/src/test/kotlin/misk/hibernate/TransacterTest.kt
```diff
@@ -76,6 +77,20 @@ abstract class TransacterTest {
         assertThat(query.uniqueResult(session)!!.actor?.name).isEqualTo("Jeff Goldblum")
       }
     }
+    readerTransacter.transaction { session ->
```
ChunYin, line 80: Could you add a test that you try to do an update in this session and should throw some sort of ReadOnly exception?

misk-hibernate/src/main/kotlin/misk/hibernate/HibernateModule.kt
```diff
@@ -106,6 +106,20 @@ class HibernateModule(
...
     }).asSingleton()
 
+    if (readerQualifier != null) {
+      val readerTransacterKey = Transacter::class.toKey(readerQualifier)
+      bind(readerTransacterKey).toProvider(object : Provider<Transacter> {
+        @Inject lateinit var executorServiceFactory: ExecutorServiceFactory
+        override fun get(): RealTransacter = RealTransacter(
+            qualifier = readerQualifier,
+            sessionFactoryProvider = readerSessionFactoryProvider!!,
+            readerSessionFactoryProvider = readerSessionFactoryProvider,
```
ChunYin, line 116: ...I wish we could do away with this entirely, but too many codebases are now using replicaRead. We now have two official ways to access the read only replica and one is going through what you think is a writer annotated DB.

Could you update the comment on replicaRead{} to steer people away from the leaky abstraction?

misk-hibernate/src/test/kotlin/misk/hibernate/TransacterTest.kt
frojasg: added. I'm reusing the `readOnly` behavior in the transacter.

misk-hibernate/src/main/kotlin/misk/hibernate/HibernateModule.kt
frojasg: How strongly we want to steep people away from `readReplica`?

ChunYin: I feel pretty strongly about it because I think of a transacter as a map 1:1 to a particular DB state. Having a transacter map to either a RO or a RW DB instance based on a method has a strong code smell.

misk-hibernate/src/main/kotlin/misk/hibernate/Transacter.kt
```diff
@@ -49,6 +49,8 @@ interface Transacter {
    *   replicas.
    * * If no reader is configured for replica reads when installing the [HibernateModule], this
    *   method will throw an [IllegalStateException].
+   * * There's another instance of [Transacter] with the reader qualifier that uses the read
```
ChunYin, line 52: `Note: You can do it another way, where you annotate the [Transacter] with the readerQualifer defined by [SkimHibernateModule], which will use the read only replica as the datasource.`

ChunYin (approved):

misk-hibernate/src/main/kotlin/misk/hibernate/Transacter.kt
frojasg: updated

misk-hibernate/src/main/kotlin/misk/hibernate/HibernateModule.kt
```diff
@@ -106,6 +108,40 @@ class HibernateModule(
...
+    if (readerQualifier != null) {
+      val readerTransacterKey = Transacter::class.toKey(readerQualifier)
+      bind(readerTransacterKey).toProvider(object : Provider<Transacter> {
+        @Inject lateinit var executorServiceFactory: ExecutorServiceFactory
+        override fun get(): Transacter = RealTransacter(
+            qualifier = readerQualifier,
+            sessionFactoryProvider = readerSessionFactoryProvider!!,
+            readerSessionFactoryProvider = readerSessionFactoryProvider,
+            config = config,
+            executorServiceFactory = object : ExecutorServiceFactory {
```
frojasg, line 120: I wonder if we need this abstraction somewhere else.

misk-hibernate/src/main/kotlin/misk/hibernate/HibernateModule.kt
adrw: If we feel strongly about it, we should add a Deprecated annotation with a message on `readReplica` that suggests the path forward using this less leaky approach.

misk-hibernate/src/test/kotlin/misk/hibernate/TransacterTest.kt
```diff
@@ -23,6 +23,7 @@ import kotlin.test.assertFailsWith
 
 abstract class TransacterTest {
   @Inject @Movies lateinit var transacter: Transacter
+  @Inject @MoviesReader lateinit var readerTransacter: Transacter
```
adrw, line 26: I like having this `*Reader` qualifier annotation 👍

adrw (approved):

misk-hibernate/src/main/kotlin/misk/hibernate/HibernateModule.kt
mightyguava: Let's bind this transacter only for MySQL. TiDB and Vitess replicas both work differently and this is not the correct configuration. That will also avoid your problem with failing Vitess tests

mightyguava (changes requested):

misk-hibernate/src/main/kotlin/misk/hibernate/HibernateModule.kt
frojasg: If we do this I'm going to need to skip the Reader Transacter test for vitess somehow.

mightyguava (approved):

misk-hibernate/src/main/kotlin/misk/hibernate/HibernateModule.kt
frojasg: We need to do some additional work to make this work for TiDB, and Vitess. I'm not going to add the `Deprecated` annotation for now.

misk-hibernate/src/test/kotlin/misk/hibernate/ReaderTransacterTest.kt
```diff
@@ -0,0 +1,95 @@
...
+      // Shard targeting works.
+      val shard = ianMalcolm.rootId.shard(session)
+      session.target(shard) {
+        assertThat(query.uniqueResult(session)!!.actor?.name).isEqualTo("Jeff Goldblum")
+      }
+    }
+  }
+
+  @Test
+  fun readerTransacterWontSave() {
```
swankjesse, line 48: nice

swankjesse (commented):
Nice tests

misk-hibernate/src/main/kotlin/misk/hibernate/HibernateModule.kt
tirsen: The reason we have a `replicaRead` method is that on TiDB and Vitess it's possible to do replica reads with the same connection so there's no need to maintain separate connection pools for writes and replica reads. I also made it possible to use the `replicaRead` syntax on MySQL.

I agree 100% that we should settle on a single way to do this and then support that on both Aurora and TiDB. I have a preference but I don't feel strongly which approach we pick as long as there is one way to do it, not two. :-)

(Personally I prefer the succinctness and explicitness of the `replicaRead` syntax since it behaves differently from a `transaction` method. I prefer if semantic differences has a syntactical representation. Using a `transaction` method for both doesn't seem to fulfill that. But again, I don't feel strongly, either way is fine as long as there is just one.)

PR #1672, opened 2020-11-27
Load classpath resources lazily

swankjesse:
Previously we used Guava's classpath scanner to load all resources
eagerly. This was easy but it also required us to load a lot of
resources that we might not need, potentially slowing app launch.

Also this should make it possible to make further improvements to
load resources in more complex classloader situations.

### DIFF
--- a/misk-core/src/main/kotlin/misk/resources/ResourceLoader.kt
+++ b/misk-core/src/main/kotlin/misk/resources/ResourceLoader.kt
@@ -2,16 +2,16 @@ package misk.resources
 
 import com.google.common.collect.ImmutableList
 import com.google.common.collect.ImmutableSet
+import java.nio.file.Files
+import java.nio.file.Path
+import javax.inject.Inject
+import javax.inject.Singleton
 import misk.resources.ResourceLoader.Backend
 import okio.BufferedSource
 import okio.ByteString
 import okio.ByteString.Companion.encodeUtf8
 import okio.buffer
 import okio.sink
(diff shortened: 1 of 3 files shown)


misk-core/src/main/kotlin/misk/resources/ResourceLoader.kt
```diff
@@ -141,12 +141,10 @@ class ResourceLoader @Inject constructor(
    * Copies all resources with [root] as a prefix to the directory [dir].
    */
   fun copyTo(root: String, dir: Path) {
-    val (scheme, path) = parseAddress(root)
-    val prefix = if (path.endsWith("/")) path else "$path/"
-    for (resource in backends[scheme].all()) {
```
swankjesse, line 146: needed to avoid `all()`

misk-core/src/test/kotlin/misk/resources/ResourceLoaderTest.kt
```diff
@@ -44,6 +44,12 @@ class ResourceLoaderTest {
         .doesNotContain("classpath:/misk/resources/ResourceLoaderTest.txt")
   }
 
+  @Test
+  fun listContainsResourcesFromJar() {
+    assertThat(resourceLoader.list("classpath:/META-INF/"))
+        .contains("classpath:/META-INF/MANIFEST.MF")
```
swankjesse, line 50: this is unlikely to be on the filesystem!

acostama (approved):

PR #1673, opened 2020-11-28
Support deliveryDelay and processing of individual jobs by FakeJobQueue

alatyshau:


misk-jobqueue-testing/src/main/kotlin/misk/jobqueue/FakeJobQueue.kt
```diff
@@ -55,15 +58,7 @@ class FakeJobQueue @Inject constructor(
     attributes: Map<String, String>
   ) {
     session.onPostCommit {
-      val id = tokenGenerator.generate("fakeJobQueue")
-      val job = FakeJob(
-          queueName = queueName,
-          id = id,
-          idempotenceKey = idempotenceKey,
-          body = body,
-          attributes = attributes
```
alatyshau, line 64: There was a minor bug: deliveryDelay argument was missed

misk-jobqueue-testing/src/main/kotlin/misk/jobqueue/FakeJobQueue.kt
```diff
@@ -150,14 +169,30 @@ class FakeJobQueue @Inject constructor(
...
+  private fun pollNextJob(jobs: ConcurrentLinkedDeque<FakeJob>, considerDelays: Boolean): FakeJob? {
+    if (!considerDelays) {
+      return jobs.poll()
+    }
+    val now = clock.instant()!!
+    while (true) {
+      val job = jobs.firstOrNull {
+        it.deliveryDelay == null || !it.enqueuedAt.plus(it.deliveryDelay).isAfter(now)
+      }
+      // If removeFirstOccurrence() is false, then we lost race to another worker and should retry.
```
alatyshau, line 190: Looks like `ConcurrentLinkedDeque` does not provide an atomic `find and remove`.

misk-jobqueue-testing/src/main/kotlin/misk/jobqueue/FakeJobQueue.kt
```diff
@@ -28,10 +32,11 @@ import javax.inject.Singleton
  */
 @Singleton
 class FakeJobQueue @Inject constructor(
+  private val clock: Clock,
   private val jobHandlers: Provider<Map<QueueName, JobHandler>>,
   private val tokenGenerator: TokenGenerator
 ) : JobQueue, TransactionalJobQueue {
-  private val jobQueues = ConcurrentHashMap<QueueName, ConcurrentLinkedDeque<FakeJob>>()
+  private val jobQueues = ConcurrentHashMap<QueueName, PriorityBlockingQueue<FakeJob>>()
```
alatyshau, line 39: This is not strictly backward-compatible and might break unit tests of services that use FakeJobQueue and make assumptions about the order of messages.
A safer approach would be to isolate this change using a property in FakeJobQueueModule. But from the other side, not ordering by `deliveryAt` feels fragile because it can execute jobs in an absurd oder (something that is delayed to happen in 5m can precede something that happens now). Therefore, if we do not share `misk` with parties outside Cash yet, I find it fine to rather break those tests and fix them.

misk-jobqueue-testing/src/main/kotlin/misk/jobqueue/FakeJobQueue.kt
```diff
@@ -75,13 +72,14 @@ class FakeJobQueue @Inject constructor(
...
-    jobQueues.getOrPut(queueName, ::ConcurrentLinkedDeque).add(job)
+    val job =
+        FakeJob(queueName, id, idempotenceKey, body, attributes, clock.instant(), deliveryDelay)
+    jobQueues.getOrPut(queueName, ::PriorityBlockingQueue).add(job)
   }
 
   fun peekJobs(queueName: QueueName): List<Job> {
     val jobs = jobQueues[queueName]
-    return jobs?.toList() ?: listOf()
+    return jobs?.sortedBy { it.id } ?: listOf()
```
alatyshau, line 82: When we peek, we typically want to peek in the order the job was enqueued rather than in order of execution.

misk-jobqueue-testing/src/test/kotlin/misk/jobqueue/FakeJobQueueTest.kt
```diff
@@ -209,22 +425,34 @@ private enum class Color {
...
 private class ColorException : Throwable()
 
-private data class ExampleJob(val color: Color, val message: String)
+private data class ExampleJob(
+  val color: Color,
+  val message: String,
+  val hint: ExampleJobHint? = null
+)
+
+private enum class ExampleJobHint {
```
alatyshau, line 434: I've added also THROW_ONCE to support tests in `FakeTransactionalJobQueueTest`. I saw transactional queue is deprecated, so I anticipate all tests might be moved here eventually.

misk-jobqueue-testing/src/test/kotlin/misk/jobqueue/FakeTransactionalJobQueueTest.kt
```diff
@@ -195,6 +195,9 @@ internal class FakeTransactionalJobQueueTest {
     assertThat(fakeTransactionalJobQueue.peekJobs(DEADLETTER_QUEUE)).isEmpty()
     assertThat(fakeTransactionalJobQueue.peekDeadlettered(DEADLETTER_QUEUE)).hasSize(1)
     assertThat(fakeTransactionalJobQueue.reprocessDeadlettered(DEADLETTER_QUEUE, false)).hasSize(1)
+
+    // After reprocessing, all jobs returned to the deadletter queue again.
+    assertThat(fakeTransactionalJobQueue.peekDeadlettered(DEADLETTER_QUEUE)).hasSize(1)
```
alatyshau, line 200: This is essential because there could be a potential infinite loop in processing deadletter queue, so we should test this case explicitly.

misk-jobqueue-testing/src/main/kotlin/misk/jobqueue/FakeJobQueue.kt
```diff
@@ -150,14 +182,30 @@ class FakeJobQueue @Inject constructor(
...
+  private fun pollNextJob(jobs: Queue<FakeJob>, considerDelays: Boolean): FakeJob? {
+    if (!considerDelays) {
+      return jobs.poll()
+    }
+    val now = clock.instant()!!
+    while (true) {
+      val job = jobs.firstOrNull {
+        it.deliveryDelay == null || !it.enqueuedAt.plus(it.deliveryDelay).isAfter(now)
+      }
+      // If remove() is false, then we lost race to another worker and should retry.
```
alatyshau, line 203: I haven't find an atomic `find and remove`.

misk-jobqueue-testing/src/main/kotlin/misk/jobqueue/FakeJobQueue.kt
```diff
@@ -93,54 +91,88 @@ class FakeJobQueue @Inject constructor(
   fun handleJobs(
     queueName: QueueName,
     assertAcknowledged: Boolean = true,
-    retries: Int = 1
+    retries: Int = 1,
+    considerDelays: Boolean = false
```
swankjesse, line 95: I’d love to flip this default, or retire this option. You’re right that delays should just work.

misk-jobqueue-testing/src/main/kotlin/misk/jobqueue/FakeJobQueue.kt
```diff
@@ -93,54 +91,88 @@ class FakeJobQueue @Inject constructor(
...
   ): List<FakeJob> {
-    val jobHandler = jobHandlers.get()[queueName]!!
-    val jobs = when (deadletter) {
-      true -> deadletteredJobs[queueName] ?: return listOf()
-      else -> jobQueues[queueName] ?: return listOf()
-    }
-
+    val jobHandlers = jobHandlers.get()
     val result = mutableListOf<FakeJob>()
+    // Used to prevent an infinite loop by mistake in supplier.
```
swankjesse, line 158: Nice

misk-jobqueue-testing/src/main/kotlin/misk/jobqueue/FakeJobQueue.kt
swankjesse: Callers _probably_ shouldn’t be doing this concurrently?

misk-jobqueue-testing/src/main/kotlin/misk/jobqueue/FakeJobQueue.kt
```diff
@@ -150,14 +182,30 @@ class FakeJobQueue @Inject constructor(
...
     return result
   }
+
+  private fun pollNextJob(jobs: Queue<FakeJob>, considerDelays: Boolean): FakeJob? {
+    if (!considerDelays) {
+      return jobs.poll()
+    }
+    val now = clock.instant()!!
+    while (true) {
+      val job = jobs.firstOrNull {
```
swankjesse, line 200: firstOrNull will traverse the entire queue. I think you probably want to just `peek()` and check that

misk-jobqueue-testing/src/main/kotlin/misk/jobqueue/FakeJobQueue.kt
```diff
@@ -180,4 +234,28 @@ data class FakeJob(
   override fun deadLetter() {
     deadLettered = true
   }
+
+  override fun equals(other: Any?): Boolean {
```
swankjesse, line 238: I’m anxious about any object where the equals() state is different from the actual state. For example, tests would pass if you assertEquals() on two jobs with the same ID and different bodies.

Rollback this part of the change?

misk-jobqueue-testing/src/main/kotlin/misk/jobqueue/FakeJobQueue.kt
```diff
@@ -180,4 +234,28 @@ data class FakeJob(
...
+    if (javaClass != other?.javaClass) return false
+
+    other as FakeJob
+
+    if (id != other.id) return false
+
+    return true
+  }
+
+  override fun hashCode(): Int {
```
swankjesse, line 249: (and this too)

misk-jobqueue-testing/src/test/kotlin/misk/jobqueue/FakeJobQueueTest.kt
```diff
@@ -185,6 +204,203 @@ internal class FakeJobQueueTest {
     assertThat(fakeJobQueue.peekJobs(GREEN_QUEUE)).isEmpty()
     assertThat(fakeJobQueue.peekJobs(RED_QUEUE)).isEmpty()
   }
+
+  @Test
+  fun considerDeliveryDelaysInFakeJob() {
```
swankjesse, line 209: there’s a lot happening in this test. Would you consider separate tests for each of the interesting assertions here?

jobsDoNotStartUntilDeliveryDelayElapses()
multipleJobsMayBeReadyConcurrently()

This test is exercising a lot of code but not without much direction. I’d prefer more small tests that each describe an invariant and test only that.

swankjesse (approved):

misk-jobqueue-testing/src/main/kotlin/misk/jobqueue/FakeJobQueue.kt
alatyshau: The reason I added this override are two vars:
```
  var acknowledged: Boolean = false
  var deadLettered: Boolean = false
```
FakeJob is used in `queue.remove()` so this is important. Well, we basically don't use FakeJobQueue in tests with concurrency involved, so I'd not care, but misk is a very shared library and it uses ConcurrentHashMap etc., so to stay multi-thread safe, we need this override here.

But I can include more immutable fields: all of vals. Would that work?

misk-jobqueue-testing/src/main/kotlin/misk/jobqueue/FakeJobQueue.kt
alatyshau: I think tests should avoid using concurrency at all, because it is not deterministic. But then  I'd replace ConcurrentHashMap by usual HashMap, etc. Otherwise using `ConcurrentHashMap` and then not caring about the rest is confusing. Can I remove all concurrency related stuff from this class? :)

misk-jobqueue-testing/src/main/kotlin/misk/jobqueue/FakeJobQueue.kt
alatyshau: You right, since queue is sorted by deliveryAt, using `peek()` will work here. I will fix that.
In my first version I reused `ConcurrentLinkedDeque` which was not sorted by deliveryAt, and later fixed to PriorityBlockingQueue.
I will also then make sure that dead letter queue never uses this branch (it is still on ConcurrentLinkedDeque)

misk-jobqueue-testing/src/main/kotlin/misk/jobqueue/FakeJobQueue.kt
alatyshau: This will very probably break a bunch of tests in Franklin :)) 
It is better to do in a separate PR:
1) Introduce an option in FakeJobQueueModule
2) Use this option in all services and fix the tests
3) Then make this option default

misk-jobqueue-testing/src/main/kotlin/misk/jobqueue/FakeJobQueue.kt
alatyshau: So equals and hashCode become:
```


  override fun equals(other: Any?): Boolean {
    return this === other
  }

  override fun hashCode(): Int {
    return System.identityHashCode(this)
  }
```
Does this look good?

alatyshau: Or should I better make FakeJob not a data class: `class FakeJob(` and then skip overriding hashCode/equals?

alatyshau: (Dropped my previous comments: nevermind)
The only place where it was critical is  when we `find and remove`, and this would be the same instance. So even if FakeJob was concurrently modified (it has two vars), the instance would be the same.. I have just overthought this.. So Im rolling this back.

misk-jobqueue-testing/src/test/kotlin/misk/jobqueue/FakeJobQueueTest.kt
alatyshau: Broke down on three tests:
jobsDoNotStartUntilDeliveryDelayElapses
jobsStartOnDeliveryDelayPassed
jobsStartAfterDeliveryDelayPassed

swankjesse (approved):

PR #1679, opened 2020-12-08
Make MoshiModule public

acostama:
Making this available publicly, so that lightweight tests that only need moshi set up and don't rely on more misk modules can depend on it

### DIFF
--- a/misk/src/main/kotlin/misk/moshi/MoshiModule.kt
+++ b/misk/src/main/kotlin/misk/moshi/MoshiModule.kt
@@ -15,7 +15,11 @@ import javax.inject.Singleton
 import com.squareup.wire.WireJsonAdapterFactory as WireOnlyJsonAdapterFactory
 import misk.moshi.wire.WireMessageAdapter as MiskOnlyMessageAdapter
 
-internal class MoshiModule(
+/**
+ * For service setup, prefer to install [misk.MiskCommonServiceModule] over installing [MoshiModule]
+ * directly.
+ */
+class MoshiModule(
   private val useWireToRead: Boolean = false,
   private val useWireToWrite: Boolean = false
 ) : KAbstractModule() {


misk/src/main/kotlin/misk/moshi/MoshiModule.kt
```diff
@@ -13,7 +13,11 @@ import misk.moshi.wire.WireMessageAdapter
 import java.util.Date
 import javax.inject.Singleton
 
-internal class MoshiModule : KAbstractModule() {
+/**
+ * For service setup, prefer to install [misk.MiskCommonServiceModule] than installing [MoshiModule]
+ * directly.
+ */
+class MoshiModule : KAbstractModule() {
```
swankjesse, line 20: change this to `object MoshiModule` ? It’s unlikely to ever need to be stateful

swankjesse: And then you can install it 2x and it doesn’t hurt you!

swankjesse (approved):

misk/src/main/kotlin/misk/moshi/MoshiModule.kt
acostama: changed!

swankjesse (approved):

swankjesse:
```

e: /home/circleci/misk/misk/src/test/kotlin/misk/moshi/MoshiAdapterPrecedenceTest.kt: (37, 7): Unresolved reference. None of the following candidates is applicable because of receiver type mismatch: 
public operator fun <T, R> DeepRecursiveFunction<TypeVariable(T), TypeVariable(R)>.invoke(value: TypeVariable(T)): TypeVariable(R) defined in kotlin
:misk:compileTestKotlin (Thread[Execution worker for ':' Thread 2,5,main]) completed. Took 1 mins 28.469 secs.
```

chris-ryan-square:
@acostama, if you still want this PR, please fix the problems and resubmit.

chris-ryan-square:
Looking at the conflicts/master, I think this is no longer required.  Closing.

acostama:
going to reopen this and try to land it

PR #1680, opened 2020-12-08
Make kotlin-test a test-only dependency

r3mariano:
I tracked dependencies using: `ls | grep misk- | xargs sh -c 'for arg do echo $arg; ./gradlew :$arg:dependencies --configuration compileClasspath | grep test; done'`

After this change, only the -testing libraries depend on kotlin-test.

### DIFF
--- a/misk-redis/src/main/kotlin/misk/redis/FakeRedis.kt
+++ b/misk-redis/src/main/kotlin/misk/redis/FakeRedis.kt
@@ -1,8 +1,7 @@
 package misk.redis
 
-import misk.time.FakeClock
 import okio.ByteString
-import okio.ByteString.Companion.encodeUtf8
+import java.time.Clock
 import java.time.Duration
 import java.time.Instant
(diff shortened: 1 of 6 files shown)


misk-redis/src/main/kotlin/misk/redis/FakeRedis.kt
```diff
@@ -1,16 +1,15 @@
...
+import java.time.Clock
 import java.time.Duration
 import java.time.Instant
 import java.util.concurrent.ConcurrentHashMap
 import javax.inject.Inject
 
 /** Mimics a Redis instance for testing. */
 class FakeRedis : Redis {
-  @Inject lateinit var clock: FakeClock
+  @Inject lateinit var clock: Clock
```
r3mariano, line 12: Because this file is in implementation classes we have no access to FakeClock. In tests, this should bind to one.

chris-ryan-square (approved):

PR #1683, opened 2020-12-09
Delete the obsolete VITESS datasource type

swankjesse:
It's replaced with VITESS_MYSQL. The VITESS type no longer works now that we've
split up misk-jdbc and misk-hibernate.


mpawliszyn (approved):
LGTM

mpawliszyn (approved):
Still looks good.

PR #1685, opened 2020-12-10
New misk-events-testing package and FakeEventProcessor

swankjesse:
This is a completely in-memory system for producing and consuming
events. It is suitable for tests only. Unlike our other testing
systems this doesn't use any background threads for processing.


misk-events-testing/src/main/kotlin/misk/events/FakeEventProcessor.kt
```diff
@@ -0,0 +1,106 @@
...
+ *
+ * To publish, inject a [Producer] and call [publish].
+ *
+ * To consume all published events, call [deliverAll]. All events will be delivered events to the
+ * corresponding consumers. Note that no call to [Consumer.subscribe] is necessary.
+ */
+@Singleton
+class FakeEventProcessor @Inject constructor(
+  private val consumers: Map<Topic, Consumer.Handler>
+) : Producer {
```
zhxnlai, line 32: We are removing the usages of the `misk-events` package and will deprecate it very soon. I would try not write new code on top of these abstractions

zhxnlai: What would you like to test in your service?

biasedbit: Agree w/ @zhxnlai, I'd love if we didn't add anything else to the old events package so as to make the migration to eventing as smooth as possible — we're currently in the process of converting all producer and consumer apps.

zhxnlai (approved):

swankjesse:
Gonna merge this now, and I volunteer to migrate anything to new systems.

PR #1687, opened 2020-12-10
Add onlyRetry list to misk.backoff.retry

cashguy:


mightyguava (commented):
I think this is retry package has a bad interface. Let's deprecate and abandon it.

This is much better https://github.com/michaelbull/kotlin-retry

PR #1689, opened 2020-12-10
Mark misk event as deprecated

zhxnlai:


biasedbit (approved):

hamdanjaveed (approved):

misk-events-core/src/main/kotlin/misk/events/Consumer.kt
```diff
@@ -1,8 +1,10 @@
 package misk.events
 
 /** A [Consumer] allows applications to receive events from a source */
+@Deprecated("This API is no longer supported and replaced by go/cash-eventing's client library")
```
adrw, line 4: Should this message better obfuscate Cash specific implementation details before landing in public Misk?

PR #1693, opened 2020-12-11
Migrate Misk to Github Actions and automated Sonatype Releases

adrw:
This brings Misk up to speed with the similar patterns now across many Cash App Gradle based repos including Tempest, Barber, and SQLDelight.


gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,5 +1,5 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-6.2-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-6.5-bin.zip
```
adrw, line 3: Is Misk okay to move to 6.5? The build was failing before with it at 6.2. @lkerford @swankjesse

lkerford: Moving to 6.5 is good.

lkerford (approved):

PR #1697, opened 2020-12-14
Cleanup Database Misk Web tab, add LocalDate adapter

adrw:
Unless otherwise configured, we attempt to mimic `select *` by including all DbEntity declared member properties as the select paths. Some of these aren't actually queryable and so we've added a deny list of paths to make the default unconfigured query experience in the dashboard tab more intuitive. Additionally, the lack of a LocalDate adapter caused this default query to fail.

Motivation: I had experienced this myself but it was also quickly experienced by folks trying out the new tab.

### DIFF
--- a/misk-hibernate/src/test/kotlin/misk/hibernate/actions/HibernateDatabaseQueryDynamicActionTest.kt
+++ b/misk-hibernate/src/test/kotlin/misk/hibernate/actions/HibernateDatabaseQueryDynamicActionTest.kt
@@ -1,11 +1,13 @@
 package misk.hibernate.actions
 
+import misk.exceptions.BadRequestException
 import misk.exceptions.UnauthorizedException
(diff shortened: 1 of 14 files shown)


misk-hibernate/src/test/kotlin/misk/hibernate/actions/HibernateDatabaseQueryDynamicActionTest.kt
```diff
@@ -65,8 +68,6 @@ class HibernateDatabaseQueryDynamicActionTest {
     }
   }
 
-  // TODO (adrw) re-enable once LocalDate Moshi adapter is written/bound to support DbMovie.release_date
-  @Disabled
```
lkerford, line 69: nice

lkerford (approved):

PR #1700, opened 2020-12-16
expose hibernateCore dep in misk-hibernate

szabado:
This patches two breaks introduced in #1699. Consumers of this library rely on it to publish the `javax.persistence:javax.persistence-api:2.2` artifact, which this library gets from hibernateCore. Consumers also rely on `misk-jdbc`. Changing `implementation` to `api` means that the dependencies are transitively published, like they were before #1699

### DIFF
--- a/misk-hibernate/build.gradle
+++ b/misk-hibernate/build.gradle
@@ -15,12 +15,12 @@ allOpen {
 }
 
 dependencies {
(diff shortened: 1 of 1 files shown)


adrw (approved):
Thanks for catching this!

swankjesse:
Thank you!

PR #1702, opened 2020-12-16
Remove get*OrNull calls from FeatureFlags

r3mariano:
This wasn't a good idea - the naming implied a default value, and the case I
initially catered for in LD (off, no default value) should have been
exceptional.


swankjesse (approved):

mightyguava (approved):
Thanks, can you add a PR description for future readers?

PR #1706, opened 2020-12-17
PoC: Netty-based Google gRPC server using Wire for the encoding format

mightyguava:
This is a rough PoC of using the official Google Java gRPC implementation instead of Wire's custom implementation. There's some rough MiSK integration in `RouteGuideMiskServiceModule` to show that it can be done, but a lot more work will need to go into this to get to a production-ready state.

This PR contains working implementations of unary, client streaming, server streaming, and bi-directional streaming RPCs.


misk-grpc-vanilla/src/main/kotlin/misk/grpc/vanilla/wireserver/RouteGuideWireGrpc.kt
```diff
@@ -0,0 +1,323 @@
+package misk.grpc.vanilla.wireserver
```
mightyguava, line 1: This entire file should be generated. We can either build this into (or fork) the Wire compiler, fork the Google Java compiler, or build it as a `protoc` plugin

misk-grpc-vanilla/build.gradle
```diff
@@ -0,0 +1,74 @@
...
+  compile dep.junitApi
+  compile dep.docker
+  compile dep.grpcNetty
+  compile dep.grpcProtobuf
+  compile dep.grpcStub
+  compile dep.kotlinxCoroutines
+  compile dep.wireGrpcClient
+  compile dep.wireRuntime
+  compile project(':misk')
+  compile project(':misk-testing')
```
chris-ryan-square, line 66: nit: should be in the test compile path, not normal compile

misk-grpc-vanilla/src/main/kotlin/misk/grpc/vanilla/wireserver/RouteGuideWireGrpc.kt
r3mariano: I think this is the key piece

mightyguava: It is! Defining what should get generated is step 1. Generating it should be fairly easy.

Would you like to try doing it? I'm happy to get it started.

misk-grpc-vanilla/src/main/kotlin/misk/grpc/vanilla/wireserver/RouteGuideWireGrpc.kt
```diff
@@ -0,0 +1,323 @@
...
+            .addMethod(getListFeaturesMethod())
+            .addMethod(getRecordRouteMethod())
+            .addMethod(getRouteChatMethod())
+            .build()
+          serviceDescriptor = result
+        }
+      }
+    }
+    return result
+  }
```
r3mariano, line 58: I'm curious why this is the pattern vs something like
```kotlin
  val serviceDescriptor: ServiceDescriptor
    get() { ... }
```

r3mariano: I see, it makes the thing get instantiated once?

r3mariano: It's also the what grpc uses in the codegen'd java code...

r3mariano:
I'll close this. There's some work on this merged into https://github.com/square/wire now, though a bit unproven.

PR #1709, opened 2020-12-17
Add missing line to autorelease all subprojects

adrw:


misk-actions/build.gradle
```diff
@@ -3,3 +3,12 @@ dependencies {
   implementation dep.okHttp
   implementation dep.okio
 }
+
+afterEvaluate { project ->
```
adrw, line 7: This was missing strangely

ejao1 (approved):

misk/build.gradle
```diff
@@ -71,3 +71,5 @@ afterEvaluate { project ->
     outputFormat = 'gfm'
   }
 }
+
+apply from: "$rootDir/gradle-mvn-publish.gradle"
```
swankjesse, line 75: would it be simpler to add this to the allprojects clause in the root project? (And probably remove the same for GFM config)

PR #1710, opened 2020-12-18
Add dynamoDbRetry

szabado:


misk-aws/src/main/kotlin/misk/dynamodb/DynamoDbHelpers.kt
```diff
@@ -0,0 +1,40 @@
...
+
+import com.amazonaws.services.dynamodbv2.model.AmazonDynamoDBException
+import java.time.Duration
+import misk.backoff.Backoff
+import misk.backoff.DontRetryException
+import misk.backoff.ExponentialBackoff
+import misk.backoff.retry
+
+object DynamoDbHelpers {
+  val DEFAULT_BACKOFF = ExponentialBackoff(
```
zhxnlai, line 11: Unfortunately this object is [stateful](https://github.com/cashapp/misk/blob/master/misk-core/src/main/kotlin/misk/backoff/ExponentialBackoff.kt#L17). You probably want to instantiate a new one in `dynamoDbRetry`

szabado: ... I really should have seen that coming. My bad

mightyguava (commented):
I'm pretty sure the AWS SDK has built-in retries for DynamoDB.

Also, I actually think that this is something you should define in your application, versus putting in MiSK. How retries are done is an application-level concern. The small amount of code reuse probably doesn't make it worthwhile to stick in the framework.

misk-aws/src/main/kotlin/misk/dynamodb/DynamoDbHelpers.kt
zhxnlai: nah I think it was just bad API design

misk-aws/src/main/kotlin/misk/dynamodb/DynamoDbHelpers.kt
```diff
@@ -0,0 +1,42 @@
...
+import java.time.Duration
+import misk.backoff.Backoff
+import misk.backoff.DontRetryException
+import misk.backoff.ExponentialBackoff
+import misk.backoff.retry
+
+object DynamoDbHelpers {
+  fun getDefaultBackoff(): ExponentialBackoff {
+    return ExponentialBackoff(
+        baseDelay = Duration.ofMillis(10),
```
szabado, line 13: Do these seem reasonable? They're not based on any hard numbers

misk-aws/src/main/kotlin/misk/dynamodb/DynamoDbHelpers.kt
```diff
@@ -0,0 +1,40 @@
...
+import misk.backoff.Backoff
+import misk.backoff.DontRetryException
+import misk.backoff.ExponentialBackoff
+import misk.backoff.retry
+
+object DynamoDbHelpers {
+  val DEFAULT_BACKOFF = ExponentialBackoff(
+      baseDelay = Duration.ofMillis(10),
+      maxDelay = Duration.ofMillis(500),
+      jitter = Duration.ofMillis(5)
```
miladghaznavi: We may need larger ``jitter`` to decrease the chance of time conflicts, how about a ``jitter`` of 10?

misk-aws/src/main/kotlin/misk/dynamodb/DynamoDbHelpers.kt
```diff
@@ -0,0 +1,40 @@
...
+  )
+
+  /**
+   * Retries any [AmazonDynamoDBException] or [PleaseRetryException]. [DontRetryException] will
+   * not be retried.
+   */
+  fun <T> dynamoDbRetry(
+    retries: Int = 3,
```
miladghaznavi, line 24: why 3?

misk-aws/src/main/kotlin/misk/dynamodb/DynamoDbHelpers.kt
```diff
@@ -0,0 +1,42 @@
...
+    )
+  }
+
+  /**
+   * Retries any [AmazonDynamoDBException] or [PleaseRetryException]. [DontRetryException] will
+   * not be retried.
+   */
+  fun <T> dynamoDbRetry(
+    retries: Int = 3,
+    backoff: Backoff = getDefaultBackoff(),
```
zhxnlai, line 25: I think this is still incorrect: `getDefaultBackoff()` gets evaluated once statically and is shared as a global variable

misk-aws/src/main/kotlin/misk/dynamodb/DynamoDbHelpers.kt
```diff
@@ -0,0 +1,42 @@
...
+import misk.backoff.DontRetryException
+import misk.backoff.ExponentialBackoff
+import misk.backoff.retry
+
+object DynamoDbHelpers {
+  fun getDefaultBackoff(): ExponentialBackoff {
+    return ExponentialBackoff(
+        baseDelay = Duration.ofMillis(10),
+        maxDelay = Duration.ofMillis(500),
+        jitter = Duration.ofMillis(5)
```
miladghaznavi, line 15: We might need larger jitters to reduce the chance of conflicts. Maybe ``10``?

szabado:
@mightyguava the thing that this is designed to help is optimistic locking. The SDK doesn't allow you to retry everything (i.e. fetch the new version and then retry the write).

That said, I totally understand on the code reuse front. I didn't realize `retry` existed until I wrote this, that covers the hard part of this and the rest isn't the most reusable block of code

misk-aws/src/main/kotlin/misk/dynamodb/DynamoDbHelpers.kt
zhxnlai: I would put `100ms`

misk-aws/src/main/kotlin/misk/dynamodb/DynamoDbHelpers.kt
szabado: That's the default max in the hibernate transacter

misk-aws/src/main/kotlin/misk/dynamodb/DynamoDbHelpers.kt
miladghaznavi: Not sure how trivial is coming up with a good number.

misk-aws/src/main/kotlin/misk/dynamodb/DynamoDbHelpers.kt
zhxnlai, line 15: I would put a larger value here like 200ms. See https://aws.amazon.com/blogs/architecture/exponential-backoff-and-jitter/

misk-aws/src/main/kotlin/misk/dynamodb/DynamoDbHelpers.kt
szabado: I don't follow. Isn't `getDefaultBackoff()` called every time `dynamoDbRetry` is called? Are the default value of parameters static?

szabado:
If there's interest in merging this I'll write tests, if the sentiment from the other maintainers is aligned with Yunchi's then I'll close this

mightyguava:
> That said, I totally understand on the code reuse front. I didn't realize `retry` existed until I wrote this, that covers the hard part of this and the rest isn't the most reusable block of code

Ack. Please also use https://github.com/michaelbull/kotlin-retry instead of misk retry

misk-aws/src/main/kotlin/misk/dynamodb/DynamoDbHelpers.kt
```diff
@@ -0,0 +1,42 @@
...
+import misk.backoff.DontRetryException
+import misk.backoff.ExponentialBackoff
+import misk.backoff.retry
+
+object DynamoDbHelpers {
+  fun getDefaultBackoff(): ExponentialBackoff {
+    return ExponentialBackoff(
+        baseDelay = Duration.ofMillis(100),
+        maxDelay = Duration.ofSeconds(10),
+        jitter = Duration.ofMillis(200)
```
miladghaznavi, line 15: These numbers might lead to connection timeout. All said, I am not sure how can we judge good numbers here. Maybe using default values from other places might make sense

misk-aws/src/main/kotlin/misk/dynamodb/DynamoDbHelpers.kt
miladghaznavi: sounds reasonable

szabado:
I'm going to close this based on the feedback.

lukeSquared:
Just to summarize, I think the takeaway here is to use [kotlin-retry](https://github.com/michaelbull/kotlin-retry) to wrap the `save` call? @mightyguava do you have any recommendations for the values to use for the retry count, delay, etc.? For context, this PR was prompted by us failing to retry save operations that fail due to optimistic locking exceptions.

PR #1712, opened 2020-12-18
Remove the HTTP boundary in the in-process DynamoDB

swankjesse:
We had an unnecessary HTTP client and server in each call to DynamoDB. This didn't add
any value to our tests!

### DIFF
--- a/misk-aws-dynamodb-testing/src/main/kotlin/misk/aws/dynamodb/testing/InProcessDynamoDbModule.kt
+++ b/misk-aws-dynamodb-testing/src/main/kotlin/misk/aws/dynamodb/testing/InProcessDynamoDbModule.kt
@@ -1,9 +1,21 @@
 package misk.aws.dynamodb.testing
 
+import com.amazonaws.services.dynamodbv2.AmazonDynamoDB
+import com.amazonaws.services.dynamodbv2.AmazonDynamoDBStreams
+import com.amazonaws.services.dynamodbv2.local.shared.access.AmazonDynamoDBLocal
+import com.amazonaws.services.dynamodbv2.local.shared.access.LocalDBClient
+import com.amazonaws.services.dynamodbv2.local.shared.access.sqlite.SQLiteDBAccess
+import com.amazonaws.services.dynamodbv2.local.shared.jobs.JobsRegister
+import com.google.inject.Provides
+import java.io.File
+import java.util.concurrent.ExecutorService
+import javax.inject.Named
+import javax.inject.Singleton
+import kotlin.reflect.KClass
 import misk.ServiceModule
+import misk.concurrent.ExecutorServiceFactory
 import misk.inject.KAbstractModule
 import misk.inject.toKey
-import kotlin.reflect.KClass
 
 /**
  * Executes a DynamoDB service in-process per test. It clears the table content before each test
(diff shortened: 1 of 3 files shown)


swankjesse:
BEFORE
Screen Shot 2020-12-18 at 15 27 23

AFTER
Screen Shot 2020-12-18 at 15 26 57

misk-aws-dynamodb-testing/src/main/kotlin/misk/aws/dynamodb/testing/InProcessDynamoDbModule.kt
```diff
@@ -1,9 +1,21 @@
...
+import javax.inject.Singleton
+import kotlin.reflect.KClass
 import misk.ServiceModule
+import misk.concurrent.ExecutorServiceFactory
 import misk.inject.KAbstractModule
 import misk.inject.toKey
-import kotlin.reflect.KClass
 
 /**
  * Executes a DynamoDB service in-process per test. It clears the table content before each test
```
zcross, line 21: Does this comment cover how it isolates side effects for concurrent test execution?

misk-aws-dynamodb-testing/src/main/kotlin/misk/aws/dynamodb/testing/InProcessDynamoDbModule.kt
```diff
@@ -21,9 +33,48 @@ class InProcessDynamoDbModule(
...
+  fun provideSqliteDbAccess(): SQLiteDBAccess {
+    Libsqlite4JavaLibraryPathInitializer.init()
+    return SQLiteDBAccess(null as File?)
+  }
+
+  @Provides @Singleton @Named("InProcessDynamoDb")
+  fun provideExecutorService(
+    executorServiceFactory: ExecutorServiceFactory
+  ): ExecutorService {
+    return executorServiceFactory.fixed("InProcessDynamoDb", 10)
```
zcross, line 54: 10 -> constant, maybe comment if it is significant or interesting to change?

misk-aws-dynamodb-testing/src/main/kotlin/misk/aws/dynamodb/testing/InProcessDynamoDbModule.kt
```diff
@@ -21,9 +33,48 @@ class InProcessDynamoDbModule(
...
+      multibind<DynamoDbTable>().toInstance(table)
+    }
     install(ServiceModule<InProcessDynamoDbService>())
     install(ServiceModule<CreateTablesService>()
         .dependsOn(InProcessDynamoDbService::class.toKey()))
   }
+
+  @Provides @Singleton
+  fun provideSqliteDbAccess(): SQLiteDBAccess {
+    Libsqlite4JavaLibraryPathInitializer.init()
```
zcross, line 46: works for me

zcross (approved):
LGTM, agree that the

misk-aws-dynamodb-testing/src/main/kotlin/misk/aws/dynamodb/testing/InProcessDynamoDbModule.kt
swankjesse: oooh, good question. With this change we’re a bit more shielded from concurrent execution because we’re not binding ports or anything.

misk-aws-dynamodb-testing/src/main/kotlin/misk/aws/dynamodb/testing/InProcessDynamoDbModule.kt
swankjesse: I used 10 threads ’cause it’s what DynamoDB uses as their default. It’s probably more than we need. I don’t want to put too much justification into that number because I think even they pulled it from thin air.

zhxnlai:
I believe removing the proxy server caused `ConditionalCheckFailedException` to become `DynamoDBLocalServiceException`

PR #1718, opened 2021-01-12
Fix the re-running of the tests in misk-hibernate so they succeed aft…

chris-ryan-square:
…er the first attempt (ensures age column alter is reverted after test)


r3mariano (approved):
Good catch.

I feel like something's been broken with branch builds in misk, but that's not this PR.

Issue #1514, opened 2020-05-27
Threads leak with @MiskTest(startService = false)


swankjesse:
We allow threads to be created when the `ExecutorServiceFactory` is used, even if services aren’t running in the current test. These threads are not released when the test completes because there’s no service to shut down.

r3mariano:
I tried to shim together an alternate shutdown hook for `ExecutorServiceFactory`, but it didn't seem to alleviate memory problems.

I've also tried analysing a project when it runs tests; something else is causing thread locals to leak even with startService = true and reduced  threads (as in https://github.com/cashapp/misk/pull/1785).

r3mariano:
Fixed the rest in https://github.com/cashapp/misk/pull/1797.

PR #1724, opened 2021-01-18
WIP: misktest start service true always

chris-ryan-square:
This PR is for issue: https://github.com/cashapp/misk/issues/1514

This is deprecating the startService value in MiskTest.

Tests in Misk that are annotated with @MiskTest(startService = X) have not been altered.  That can be done in a followup PR to avoid further obscuring this one.


misk-aws/src/test/kotlin/misk/jobqueue/sqs/SqsJobQueueServiceTest.kt
```diff
@@ -46,6 +47,7 @@ internal class SqsJobQueueServiceTest {
     serviceManager.awaitStopped(20, TimeUnit.SECONDS)
   }
 
+  @Disabled("Are manual services real?")
```
chris-ryan-square, line 50: Disabled because the test was trying to run a service manually.

misk-jdbc/src/test/kotlin/misk/jdbc/SchemaMigratorTest.kt
```diff
@@ -237,6 +242,7 @@ internal abstract class SchemaMigratorTest(val type: DataSourceType) {
     }).hasMessageContaining("Duplicate migrations found")
   }
 
+  @Disabled("Service is already started...")
```
chris-ryan-square, line 245: Disabled, because the test was expecting to start the service itself, but the service is already running.

swankjesse (approved):

chris-ryan-square:
Not going to take this one through - the impact downstream is too much.

PR #1727, opened 2021-01-18
Add flag to disable concurrency limiting

shellderp:
Adds a way to tune and also turn it off, so we can turn it off until we're happy with its behaviour

### DIFF
--- a/misk/src/test/kotlin/misk/web/interceptors/ConcurrencyLimitsInterceptorTest.kt
+++ b/misk/src/test/kotlin/misk/web/interceptors/ConcurrencyLimitsInterceptorTest.kt
@@ -1,6 +1,7 @@
 package misk.web.interceptors
 
 import ch.qos.logback.classic.Level
+import com.netflix.concurrency.limits.Limiter
 import com.netflix.concurrency.limits.limit.SettableLimit
 import com.netflix.concurrency.limits.limiter.SimpleLimiter
 import misk.Action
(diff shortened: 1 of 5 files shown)


swankjesse (commented):
Do you mind inverting the boolean?

I’d prefer to see this in code which looks appropriately dangerous

    concurrency_limiter_disabled: true

shellderp:
> Do you mind inverting the boolean?
> 
> I’d prefer to see this in code which looks appropriately dangerous
> 
> ```
> concurrency_limiter_disabled: true
> ```

I wasn't worried because of kotlin default values, but I can totally do that

shellderp:
Also, do you have any suggestions on how to be able to override the `Limiter` used?

swankjesse:
Yep, optional injection sucks. Maybe multibind something like this?

```
interface LimiterFactory {
  fun create(action: Action): Limiter<String>?
}
```

The first one to return non-null is used?

shellderp:
> Yep, optional injection sucks. Maybe multibind something like this?
> 
> ```
> interface LimiterFactory {
>   fun create(action: Action): Limiter<String>?
> }
> ```
> 
> The first one to return non-null is used?

ah k, so by default none are bound and in that case we create a SimpleLimiter?

misk/src/test/kotlin/misk/web/interceptors/ConcurrencyLimitsInterceptorCustomLimiterTest.kt
```diff
@@ -0,0 +1,105 @@
...
+import misk.web.RealNetworkChain
+import misk.web.actions.WebAction
+import okhttp3.HttpUrl.Companion.toHttpUrl
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.jupiter.api.Test
+import java.time.Duration
+import javax.inject.Inject
+
+@MiskTest(startService = true)
+class ConcurrencyLimitsInterceptorCustomLimiterTest {
```
shellderp, line 28: I had to basically copy the existing test to install a different module

misk/src/test/kotlin/misk/web/interceptors/ConcurrencyLimitsInterceptorTest.kt
```diff
@@ -131,6 +141,19 @@ class ConcurrencyLimitsInterceptorTest {
...
       install(LogCollectorModule())
       install(MiskTestingServiceModule())
+
+      multibind<ConcurrencyLimiterFactory>().toInstance(CustomLimiterFactory())
+    }
+  }
+
+  class CustomLimiterFactory : ConcurrencyLimiterFactory {
+    override fun create(action: Action): Limiter<String>? {
+      if (action.function == CustomLimiterAction::call) {
```
swankjesse, line 151: oh cool

swankjesse (approved):

PR #1732, opened 2021-01-19
Disable github actions java build

r3mariano:
This part of the build seems to consistently break in PRs. Disable it for now.

### DIFF
--- a/.github/workflows/Test.yml
+++ b/.github/workflows/Test.yml
@@ -26,8 +26,9 @@ jobs:
     strategy:
       matrix:
         include:
-          - shard: ./gradlew testShardHibernate -i --scan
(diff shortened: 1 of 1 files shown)


chris-ryan-square (approved):

adrw:
I hope we can fix the flakes and turn GH Actions back on and finish migrating from CircleCI, I got it 90% there but got pulled into other things the past few weeks and haven't had a chance to iron out the flakes. My guess is that it seems like the Gradle build ends up starting before the DB is ready, which wasn't an issue with CircleCI but is with GH Actions.

r3mariano:
One of the errors is a Bad Request one, the DB tests look fine but it's maybe something with the ports or system setup (or something else entirely)

PR #1734, opened 2021-01-20
Add read write retry helper

szabado:
The name of this helper is pretty sucky, if you have a better idea please lmk. 

This type of functionality has started to become necessary in a couple of services in two places:
- retrying DynamoDb optimistic locking (the state needs to be reloaded and have the operation re-applied on top of it)
- re-establishing a connection to S3 when it dies unexpectedly (this is admittedly a hacky workaround, but being able to leverage this function would make it less ugly)

If you have a suggestion for a different place for this to live let me know, misk was my go-to because spinning out a new library repo for this seemed like overkill.

### DIFF
--- a/dependencies.gradle
+++ b/dependencies.gradle
@@ -59,6 +59,7 @@ ext.dep = [
   "kotlinGradlePlugin": "org.jetbrains.kotlin:kotlin-gradle-plugin:1.4.0",
   "kotlinNoArgPlugin": "org.jetbrains.kotlin:kotlin-noarg:1.4.0",
   "kotlinReflection": "org.jetbrains.kotlin:kotlin-reflect:1.4.0",
+  "kotlinRetry": "com.michael-bull.kotlin-retry:kotlin-retry:1.0.6",
   "kotlinStdLibJdk8": "org.jetbrains.kotlin:kotlin-stdlib-jdk8:1.4.0",
   "kotlinTest": "org.jetbrains.kotlin:kotlin-test:1.4.0",
   "kotlinxCoroutines": "org.jetbrains.kotlinx:kotlinx-coroutines-core:1.3.5",
--- a/misk-core/build.gradle
+++ b/misk-core/build.gradle
@@ -22,6 +22,7 @@ dependencies {
 
   testImplementation dep.assertj
   testImplementation dep.kotlinTest
+  testImplementation dep.kotlinxCoroutines
   testImplementation project(':misk-testing')
 }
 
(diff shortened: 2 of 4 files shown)


dnagler (approved):

dnagler:
Maybe `readWriteWithRetry` or `readWriteWithRetryPolicy`?

mightyguava:
This `reader` and `writer` split seems kinda confusing. I might be missing some context. What about retrying the whole computation? E.g.

For DDB

```kotlin
retry(policy) {
  entity = loadEntity(id)
  entity.someValue++
  saveEntity(id)
}
```

For S3

```kotlin
retry(policy) {
  s3 = getConnection()
  s3.uploadObject(foo)
}
```

szabado:
The reason for the `reader`/`writer` split is to decouple them. It means that instead of having the retry at the highest level, you can push it down as far as possible. This is advantageous in batched operations.

Let's say you need to cancel all the orders for a customer (for example). Pseudocode is below.

```kotlin
class CustomerOrderManager {
  fun cancelOrdersForCustomer(customer: CustomerToken) {
    customerOrderEntityOperator.getOrdersForCustomer(customer).forEach { it.cancel() }
  }

  fun getOrdersForCustomer(customerToken: CustomerToken): List<DyCustomerOrder> {
    return TODO("return a list of orders")
  }

  fun cancelOrder(order: DyCustomerOrder) {
    var latestOrderVersion = order
    readWriteRetry(limitAttempts(5) + fullJitterBackoff(...),
        reader = {
          latestOrderVersion = dynamoDbMapper.load(latestOrderVersion)
        },
        writer = {
          if (order.canBeCancelled()) {
            order.cancelled = true
            dynamoDbMapper.save(order)
          }
        }
    )
  }
}
```

If you wanted to retry the entire operation, I see ways of doing it:
- you pull it up to the `cancelOrdersForCustomer` level and then you have to retry everything every time
- you push it down to `cancelOrder`, which then has to re-read the database state every time before it writes, meaning you end up doing a lot of unnecessary reads assuming you have relatively low contention

I showed it in one class for ease of reading/writing but in the relevant services it was pushed down into the lowest layer possible, which let the top level `Manager` classes not care about database retries at all

mightyguava:
I see, that makes sense. A small suggestion on the interface then, not sure about the exact lambda syntax

```kotlin
    retry(limitAttempts(5) + fullJitterBackoff(...) + beforeRetryHook({
          latestOrderVersion = dynamoDbMapper.load(latestOrderVersion)
    }) {
        if (order.canBeCancelled()) {
          order.cancelled = true
          dynamoDbMapper.save(order)
        }
    )
```

Since you just want to add a hook before each subsequent retry attempt.

Not directly related, if you can implement `canBeCancelled()` as dynamodb condition expressions on the `UpdateItem` call, that might be a more efficient way to do it than implementing a read/check/write on the clientside

szabado:
Genius name, thanks yunchi. I'll rename it

mightyguava (approved):

PR #1739, opened 2021-01-22
Expose guice deps in misk-inject

szabado:
If these aren't included then the consumer needs to also import them, which isn't clean

### DIFF
--- a/misk-inject/build.gradle
+++ b/misk-inject/build.gradle
@@ -5,8 +5,8 @@ buildscript {
 }
 
 dependencies {
-  implementation dep.guice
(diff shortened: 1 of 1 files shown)


misk-inject/build.gradle
```diff
@@ -5,8 +5,8 @@ buildscript {
 }
 
 dependencies {
-  implementation dep.guice
-  implementation dep.guiceMultibindings
+  api dep.guice
```
szabado, line 8: This is needed otherwise you can't implement `KAbstractModule`

swankjesse (approved):

PR #1741, opened 2021-01-25
Fix auto release versioning

shellderp:
Snapshots are useless IMO, they aren't consistent images (replaced every build) and aren't signed.

I made this signed releases in the format `0.17.0-20210128.0016-d60234a`, but manual trigger so we don't clutter maven.

Added miskweb build step


adrw (approved):
Great contribution!

PR #1742, opened 2021-01-25
Add onlyRetry retry policy

szabado:


mightyguava (approved):
This LGTM, though personally I think both `doNotRetry` and `onlyRetry` are unnecessary. The kotlin-retry library already provides a `retryIf` that is reusable and is pretty concise.

https://github.com/michaelbull/kotlin-retry/blob/master/src/test/kotlin/com/github/michaelbull/retry/policy/RetryIfTest.kt

I prefer to layer on helper functions only if they add significant abstraction or reusability, but I think in this case it's not too different from using what the library provides.

szabado:
Super good point, didn't realize that `retryIf` existed. I opened a ticket downstream to see if they'll release a new version for us to use. https://github.com/michaelbull/kotlin-retry/issues/15

szabado:
Closing this, I'll deprecate the existing function once we update our kotlinretry version

PR #1743, opened 2021-01-27
Allow @Nested MiskTest

acostama:
Strictly we only allow `@MiskTest`, `@MiskTestModule` and `@MiskExternalDependencies` on the outermost / root test class in Misk test. This is easy to reason about: nested classes inherit what ever the root class specifies. We inject members for each test, nested or not.

The main motivation here is to used `@Nested` classes to provide namespacing withing large tests suites. This means that related tests methods can share variables without putting awkward contraints on tests that don't need them. For example, since `@Nested` classes inherit `@BeforeEach` setup, we can use sibling nested classes to isolate different setups.

For this purpose its enough to limit a global misk/guice definition rather than allowing custome modules / injections for each nested class.
The injector will not inject `@Nested` classes' members, injections are only for the root class

### DIFF
--- a/misk-testing/README.md
+++ b/misk-testing/README.md
@@ -1,0 +1,19 @@
+# misk-testing
+
+Tests annotated with `@MiskTest` are misk tests. Misk tests automate injector setup and Guava 
+`Service` lifecycle.
+
+A misk test requires the annotated class to have a field of type `Module` annotated 
+`@MiskTestModule`. This module will be used to create the injector and inject any other fields.
+
+Guava `Service` lifecycle is opt-in; do this by specifying `@MiskTest(startService = true)`.
+
+Note that injection and `Service` start/stop occur for each test method. This provides a fresh setup 
+for each test method for state changed by injection/services. 
+
+Misk tests are also compatible with JUnit's `@Nested` classes. Any `@Nested` classes inherit all the
+misk test setup from their outermost class. This means that only the outermost test class can have 
+`@MiskTest`, `@MiskTestModule`, `@MiskExternalDependency` annotations. They are ignored on `@Nested` 
+classes. Additionally, the injector only injects members into the outermost class. Any injected 
+fields of inner classes will also be ignored.  
+
--- a/misk-testing/src/main/kotlin/misk/testing/ExtensionContextExtensions.kt
+++ b/misk-testing/src/main/kotlin/misk/testing/ExtensionContextExtensions.kt
@@ -2,6 +2,12 @@ package misk.testing
 
 import org.junit.jupiter.api.extension.ExtensionContext
 
+internal val ExtensionContext.rootRequiredTestInstance: Any
+  get() = requiredTestInstances.allInstances.first()
+
+internal val ExtensionContext.rootRequiredTestClass: Class<*>
+  get() = rootRequiredTestInstance.javaClass
+
 /** Stores an object scoped to the test class on the context */
 fun <T> ExtensionContext.store(name: String, value: T) {
   val namespace = ExtensionContext.Namespace.create(requiredTestClass)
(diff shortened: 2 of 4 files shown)


misk-testing/README.md
```diff
@@ -0,0 +1,19 @@
+# misk-testing
+
+Tests annotated with `@MiskTest` are misk tests. Misk tests automate injector setup and Guava 
+`Service` lifecycle.
+
+A misk test requires the annotated class to have a field of type `Module` annotated 
+`@MiskTestModule`. This module will be used to create the injector and inject any other fields.
+
+Guava `Service` lifecycle is opt-in; do this by specifying `@MiskTest(startService = true)`.
```
adrw, line 9: Thanks for adding these docs!

misk-testing/src/test/kotlin/misk/testing/NestedTestsTest.kt
```diff
@@ -0,0 +1,52 @@
+package misk.testing
+
+import misk.inject.KAbstractModule
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.jupiter.api.Nested
+import org.junit.jupiter.api.Test
+import javax.inject.Inject
+import javax.inject.Singleton
+
+@MiskTest(startService = false)
```
adrw, line 10: Might be good to have a test or update this one to `true` to ensure this is still working

adrw (approved):

misk-testing/src/test/kotlin/misk/testing/NestedTestsTest.kt
acostama: updated to include a service ‼️

misk-testing/src/main/kotlin/misk/testing/ExtensionContextExtensions.kt
```diff
@@ -11,5 +17,45 @@ fun <T> ExtensionContext.store(name: String, value: T) {
...
+internal inline fun <reified A : Annotation, T> ExtensionContext.fieldsAnnotatedBy(): Iterable<T> =
+    fieldsAnnotatedBy(A::class.java)
+
+/** Find [annotation]-annotated [T]s on the outermost test class and recursively on base classes. */
+private fun <T> ExtensionContext.fieldsAnnotatedBy(
+  annotation: Class<out Annotation>
+): Iterable<T> {
+  val rootRequiredTestInstance = rootRequiredTestInstance
+  @Suppress("UNCHECKED_CAST")
+  return generateSequence(rootRequiredTestInstance.javaClass) { c -> c.superclass }
```
swankjesse, line 51: ooooh very nice

misk-testing/src/main/kotlin/misk/testing/ExtensionContextExtensions.kt
```diff
@@ -11,5 +17,45 @@ fun <T> ExtensionContext.store(name: String, value: T) {
...
+  val namespace = ExtensionContext.Namespace.create(rootRequiredTestClass)
+  return getStore(namespace).getOrComputeIfAbsent(storeKey, { this.creator() }, typeClass)
+}
+
+/** Find [A]-annotated [T]s on the outermost test class and recursively on base classes. */
+internal inline fun <reified A : Annotation, T> ExtensionContext.fieldsAnnotatedBy(): Iterable<T> =
+    fieldsAnnotatedBy(A::class.java)
+
+/** Find [annotation]-annotated [T]s on the outermost test class and recursively on base classes. */
+private fun <T> ExtensionContext.fieldsAnnotatedBy(
```
swankjesse, line 46: This says `fields` but it returns the fields’ values?

swankjesse: Ahh, I see this is just a refactor. Keep it as-is.

swankjesse (approved):

swankjesse:
An interesting next step – Injected fields on the nested classes.

PR #1750, opened 2021-02-02
Filter headers being captured in request/response logging interceptor

r3mariano:
For requests it keeps a shorter list; for responses it keeps none.

### DIFF
--- a/misk/src/test/kotlin/misk/web/interceptors/RequestLoggingInterceptorTest.kt
+++ b/misk/src/test/kotlin/misk/web/interceptors/RequestLoggingInterceptorTest.kt
@@ -13,13 +13,19 @@ import misk.testing.MiskTest
 import misk.testing.MiskTestModule
 import misk.web.Get
 import misk.web.PathParam
+import misk.web.Post
+import misk.web.RequestBody
+import misk.web.RequestHeaders
 import misk.web.ResponseContentType
 import misk.web.WebActionModule
+import misk.web.WebTestClient
 import misk.web.WebTestingModule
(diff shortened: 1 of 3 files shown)


misk/src/test/kotlin/misk/web/interceptors/RequestLoggingInterceptorTest.kt
```diff
@@ -133,6 +140,19 @@ internal class RequestLoggingInterceptorTest {
...
 
+  @Test
+  fun capturesSubsetOfHeaders() {
+    assertThat(webTestClient.call("/call/withHeaders") {
+      post("hello".toRequestBody(MediaTypes.APPLICATION_JSON_MEDIA_TYPE))
+      addHeader(FakeCallerAuthenticator.SERVICE_HEADER, "some-value")
+    }.response.isSuccessful)
+      .isTrue()
+    val messages = logCollector.takeMessages(RequestLoggingInterceptor::class)
+    assertThat(messages).containsExactly(
```
chris-ryan-square, line 151: Because we're wanting to ensure non-approved headers aren't in the message, maybe add an assertion that those header(s) are NOT in the message explicitly?

chris-ryan-square (approved):

misk/src/test/kotlin/misk/web/interceptors/RequestLoggingInterceptorTest.kt
r3mariano: That's fair, will add it to the test.

PR #1753, opened 2021-02-03
Use AmazonSQSBufferedAsyncClient for sends

biasedbit:
Experiment with SQS buffering behind feature flag. Flag (disabled by default)
can toggle (real-time, per service; no restart required) whether to use
buffering or not — i.e. switches between using `AmazonSQSBufferedAsyncClient`
or a non-buffering impl (status quo).

`AmazonSQSBufferedAsyncClient` attempts to group similar `SendMessage` and
`DeleteMessage` requests into batch requests. Each request will wait up to 200ms
for similar requests before being sent. Batches are immediately sent once full.

The trade-off is fewer network calls (which costs less $ since SQS cost is
proportional to API calls) at a slightly higher individual call duration.

This buffered client is only used for sending/deleting messages (not receiving)
but it's configured with no pre-fetching, meaning any receives are on-demand
(i.e. just like a unbuffered client) and don't spawn any extra background
threads for pre-fetching.

---

This is an optimization that had been planned since we swapped Jobsy with SQS but wasn't
picked up because we needed all hands on deck to replace Evently and weren't aware of
`AmazonSQSBufferedAsyncClient` (i.e. thought we'd have to build buffering ourselves).

Turns out it's pretty trivial.

By pooling sends and deletes, we'll reduce the number of outbound SQS calls and put less
strain on our proxy, which should help alleviate SQS request/response times under periods of
intense load.

Feature flag:
* staging: https://app.launchdarkly.com/cash/staging/features/jobqueue-buffered-sqs-client/targeting
* prod: https://app.launchdarkly.com/cash/production/features/jobqueue-buffered-sqs-client/targeting

### DIFF
--- a/misk-aws/src/main/kotlin/misk/jobqueue/sqs/AwsSqsJobQueueModule.kt
+++ b/misk-aws/src/main/kotlin/misk/jobqueue/sqs/AwsSqsJobQueueModule.kt
@@ -3,7 +3,10 @@ package misk.jobqueue.sqs
 import com.amazonaws.ClientConfiguration
 import com.amazonaws.auth.AWSCredentialsProvider
 import com.amazonaws.services.sqs.AmazonSQS
+import com.amazonaws.services.sqs.AmazonSQSAsyncClientBuilder
 import com.amazonaws.services.sqs.AmazonSQSClientBuilder
+import com.amazonaws.services.sqs.buffered.AmazonSQSBufferedAsyncClient
+import com.amazonaws.services.sqs.buffered.QueueBufferConfig
 import com.google.inject.Provider
 import com.google.inject.Provides
 import com.google.inject.Singleton
@@ -11,6 +14,7 @@ import misk.ServiceModule
 import misk.cloud.aws.AwsRegion
 import misk.clustering.lease.LeaseManager
 import misk.concurrent.ExecutorServiceModule
+import misk.config.AppName
 import misk.feature.FeatureFlags
 import misk.inject.KAbstractModule
 import misk.inject.keyOf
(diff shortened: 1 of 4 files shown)


misk-aws/src/main/kotlin/misk/jobqueue/sqs/AwsSqsJobQueueModule.kt
```diff
@@ -102,38 +111,84 @@ class AwsSqsJobQueueModule(
...
 
   private class AmazonSQSProvider(
-    val config: AwsSqsJobQueueConfig, val region: AwsRegion, val forSqsReceiving: Boolean
+    val config: AwsSqsJobQueueConfig,
+    val region: AwsRegion,
+    val forSqsReceiving: Boolean
   ) : Provider<AmazonSQS> {
     @Inject lateinit var credentials: AWSCredentialsProvider
+    @Inject lateinit var features: FeatureFlags
+    @Inject lateinit var appName: String
```
zhxnlai, line 120: does this need to be qualified with an `@AppName` annotation?

biasedbit: aw crap, you're absolutely right!

- [ ] annotate `appName` with `@AppName`

biasedbit: Done; good catch!

impossible catch

misk-aws/src/main/kotlin/misk/jobqueue/sqs/AwsSqsJobQueueModule.kt
```diff
@@ -102,38 +111,84 @@ class AwsSqsJobQueueModule(
...
+        buildReceivingClient(credentials, region)
+      } else {
+        buildClient(appName, config, credentials, region, features)
+      }
+    }
   }
 
   companion object {
+    /** Build an unbuffered [AmazonSQS] client for receiving messages. */
+    private fun buildReceivingClient(
```
zhxnlai, line 133: I am tempted to convert this companion object into a Singleton factory whose constructor takes these arguments
```
class AmazonSQSFactory(
    @AppName appName: String,
    region: AwsRegion,
    credentials: AWSCredentialsProvider,
    features: FeatureFlags
) {
  fun create() ...
  fun createReceiving() ...
}
```
We can inject this in `provideSQSClient`, `provideSQSClientForReceiving`, and `AmazonSQSProvider`

This way we pass less arguments around

zhxnlai (approved):

misk-aws/src/main/kotlin/misk/jobqueue/sqs/AwsSqsJobQueueModule.kt
biasedbit: Sounds like a good legibility improvement. I'll clean this up when we drop the feature-flag gated behavior.

PR #1756, opened 2021-02-05
Update Exemplar service to have Admin Dashboard for development

adrw:
This will let us better test and develop the dashboard all within the Misk repo instead of having to use a downstream service to build tabs against.

### DIFF
--- a/samples/exemplar/src/main/kotlin/com/squareup/exemplar/ExemplarAccessModule.kt
+++ b/samples/exemplar/src/main/kotlin/com/squareup/exemplar/ExemplarAccessModule.kt
@@ -1,0 +1,23 @@
+package com.squareup.exemplar
+
+import misk.MiskCaller
+import misk.inject.KAbstractModule
+import misk.security.authz.*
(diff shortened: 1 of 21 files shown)


samples/exemplar/src/main/kotlin/com/squareup/exemplar/ExemplarAccessModule.kt
```diff
@@ -0,0 +1,23 @@
...
+
+class ExemplarAccessModule : KAbstractModule() {
+  override fun configure() {
+    install(AccessControlModule())
+    multibind<MiskCallerAuthenticator>().to<FakeCallerAuthenticator>()
+
+    // Give engineers access to the admin dashboard for Exemplar service
+    multibind<AccessAnnotationEntry>().toInstance(
+      AccessAnnotationEntry<AdminDashboardAccess>(
+      capabilities = listOf("admin_console"))
```
AlexSzlavik, line 16: Thanks for adding this example module

AlexSzlavik (approved):

PR #1757, opened 2021-02-05
Use static keys in tests

yoavamit:
When testing a misk service that uses `misk-crypto` you'd have to provide a plaintext tink keysetin the test configuration which is annoying and doesn't actually help asserting the keys are configured as expected.

This PR loads a static constant key so that instead of providing a separate configuration files for the `test`, `development`, `staging` and `production` environments, it's possible to write a single block in the common configuration file.

Since the keys are static, it also means that the encrypted data the services writes and reads in development stays consistent.


misk-crypto/src/main/kotlin/misk/crypto/TestKeysets.kt
```diff
@@ -0,0 +1,194 @@
...
+      val plaintextKey = CleartextKeysetHandle.read(JsonKeysetReader.withString(key.value))
+      plaintextKey.write(JsonKeysetWriter.withOutputStream(keyOutput), kek)
+      return MiskConfig.RealSecret(keyOutput.toString())
+    }
+    /**
+     * Created with tikney:
+     * `tinkey create-keyset --key-template AES256_GCM`
+     */
+    val AEAD = MiskConfig.RealSecret("""{
+    "primaryKeyId": 287541552,
```
AlexSzlavik, line 28: nit: Would be nice to have these all as blobs in a yaml so they are easiest to swap.

misk-crypto/src/main/kotlin/misk/crypto/TestKeysets.kt
```diff
@@ -0,0 +1,194 @@
...
...WdJXYCZUkCqNqQl8yAEqKKOwBMhbcAx0s6Uw9UFt9podZa4GZBa9wdQcEfULISgJ3TS5GoAMdgVh3Ia2vVI29u4/vur3ucN1wQcJo2h3OiIAZDKgOHiH+B/d2r2yt1oPx2qIr2qY73so4FvLMxGOaM0Ep5mLp7TnDEtbGMIP/ig9l0Tbl1NwB0BGkLkK1JRK0+MFnBT3i+3LnWOZCy7xj9fIjleIVMNGFaEUwKi1Je29jyKLvs8SCUYUiJU4dBXTZGTZOzPT5DpuD9nUoc5Bps/OA972FOVBUxxWNCmrayB5SOAVblEjUGpWEaE6hRby8E1AjNPrmvsGoYphU3gHxKMMDFcntvG4MM9TFSf2bYjSZ1jbS7PyeDQTowLgJBM2FuQPMkAv4H5JIStQEGC2G3aG4wy/AVaS66WCjVBgHAVX8ddcm1YTXYAvWG0tIhEjVlO5fvtw5Hdwl+09S5mHwSjkrQPkQsdBR2POP+gk7KDR7I7FC7wRva56Xjpj4uo0P/Zprb2KCIiTOwcA4Fulx+gERz4X1UlODFxWKYpwdHFYKhbBJhCaISieTazh8flCCMLg7wMD79SPA8zJtg5BeBoofYPCPZM1e1faf+D4blKOreOZsXvXhgAzOg6HXW6WkR+Dhn4UJGk5Zd0y6ctfeRpR7zm87f6oIXtfYCje85DgeQoPlr8Ja61hk3SCs4s7HFUWojCytCYNzfDKYUiiw57CXVSvswN2y1QsjR/DAicKhYUkw842McpdzrKheaMaYl6DMWC7yutiZUdoBpUM622NPJUJTLAcacdISInTO/3y1Eed35m2ifFUcRm30VEoz4L4/WJlDLln4bDZC3QJofHekuRVUnY0xqx4BjVX1xMvao1HGwF3U4sg/Ocn8rnNYOho0WRnpWwlTTtnLvX9pNVxfi3SW3ppCeEKHb+Jf2fhCGrxAUH68gLyOXjwlXGGma/MoM5e49altIFFqohMj+mlLKtNQHmrWM80DoGE+EarACd5j+Jx9RLmdNds0j2HxKPawYZet3hgytck93gA6J9xFTc2pxpTuJJJV2dd5AiQhp8oyviit7dYzNQZae5hKmuLf6uxytS73z5YYoc6tRhFpbWO7EaV8T5lo08q3O04WZ6nEHaX48zkmyRLuwfCcnE2/mRJDIIN6ZvaB76TODAImkQ/4223yWJJ++uZxzQcIq",
```
AlexSzlavik, line 133: lol

AlexSzlavik (approved):
This is only for test runs and not for staging. Staging will still require the devs to setup keys (which I don't think we want to remove since it actually puts them through the motions of setting things up)

PR #1759, opened 2021-02-11
add setnx to redis

hdou731:


misk-redis/src/main/kotlin/misk/redis/FakeRedis.kt
```diff
@@ -71,4 +71,14 @@ class FakeRedis : Redis {
       expiryInstant = clock.instant().plusSeconds(expiryDuration.seconds)
     )
   }
+
+  override fun setnx(key: String, value: ByteString) {
+    // If the key already exists don't overwrite it
+    if (get(key) != null) return
+    // Set the key to expire at the latest possible instant
+    keyValueStore[key] = Value(
```
frojasg, line 79: nit: you can use `keyValueStore .putIfAbsent`

misk-redis/src/main/kotlin/misk/redis/RealRedis.kt
```diff
@@ -60,6 +60,13 @@ class RealRedis(private val jedisPool: JedisPool) : Redis {
     }
   }
 
+  // Set a ByteArray value if it doesn't already exist
+  override fun setnx(key: String, value: ByteString) {
```
frojasg, line 64: could you add this method accepting a TTL value
```
override fun setnx(key: String, expiryDuration: Duration, value: ByteString)
```

misk-redis/src/main/kotlin/misk/redis/FakeRedis.kt
hdou731: done

misk-redis/src/main/kotlin/misk/redis/RealRedis.kt
hdou731: so the setnx function doesn't have an expiration field

hdou731: at least the one in jedis doesn't

frojasg (approved):

Issue #1556, opened 2020-07-13
RepeatedTaskQueue has memory leak when lots of tests use it


adrw:
`terminated` won’t be called until `stopping` returns, and `stopping` won’t return until `terminated` is called.

```kotlin
 /**
   * Builds a new instance of a [RepeatedTaskQueue] for testing
   */
  fun forTesting(name: String, backingStorage: ExplicitReleaseDelayQueue<DelayedTask>):
      RepeatedTaskQueue {
    val queue = RepeatedTaskQueue(
        name,
        clock,
        newDirectExecutorService(),
        executorServiceFactory.single("$name-%d"),
        backingStorage,
        metrics,
        RepeatedTaskQueueConfig()
    )
    // Install a status listener that will explicitly release all of the tasks from the
    // underlying delay queue at shutdown, ensuring that the termination action runs and
    // allowing the task queue itself to shutdown
    val fullyTerminated = AtomicBoolean(false)
    queue.addListener(object : Service.Listener() {
      override fun stopping(from: Service.State) {
        // Keep kicking the storage until the task queue finally shuts down
        while (!fullyTerminated.get()) {
          backingStorage.releaseAll()
          Thread.sleep(500)
        }
      }
      override fun terminated(from: Service.State) {
        fullyTerminated.set(true)
      }
    }, newSingleThreadExecutor())
    return queue
  }
```

Originally found by @swankjesse

adrw:
Write a test case that when a service is told to shutdown, it actually does.

Fix separate listeners for the following

stopping = newSingleThreadExecutor
terminated = newDirectExecutorService

chris-ryan-square:
I have had a look into this issue.

I worked out that in RepeatedTaskQueue the pendingTasks BlockingQueue can't be reused between instances because of the triggerShutdown method (it clears the queue and then adds a no reschedule task).  If it's shared, it's indeterminate which of the queues will pick it up and stop correctly.  The test referred to in the PR 1557, does a stopAsync on all the queues created, which goes through that triggerShutdown process, so pendingTasks can get modified each time.

The easiest fix for the test is to pass in separate BlockingQueues to each test queue.  Doing that fixes the problem.  This requires any apps doing tests to update their code to not do this.

This will never be an issue in production, because each created RepeatedTaskQueue is given a new instance (DelayQueue)

[...]

PR #1760, opened 2021-02-12
Repeated task queue - fix for test issues wrt not terminating correctly.

chris-ryan-square:
For Issue: https://github.com/cashapp/misk/issues/1556


misk/src/main/kotlin/misk/tasks/RepeatedTaskQueue.kt
```diff
@@ -68,10 +68,15 @@ class RepeatedTaskQueue @VisibleForTesting internal constructor(
     }, executor())
   }
 
+  override fun startUp() {
+    if (!running.compareAndSet(false, true)) return
```
r3mariano, line 72: why have an if here?

r3mariano (approved):

misk/src/main/kotlin/misk/tasks/RepeatedTaskQueue.kt
chris-ryan-square: Good question - it was already there, I think it's only in the diff because of formatting.

tristandl: i imagine it was copied from the `triggerShutdown()` below, or there was previously a log/other code in this method - i'm pretty certain you could replace with `running.compareAndSet(false, true)` but if someone did add code below, they might not remember to put it back

PR #1764, opened 2021-02-16
Drop support for ClientApplicationInterceptors

swankjesse:
They're unused in practice.


misk/src/main/kotlin/misk/client/ClientInvocationHandler.kt
```diff
@@ -117,84 +106,11 @@ internal class ClientInvocationHandler(
...
-    private val interceptors: List<ClientApplicationInterceptor>,
-    private val args: List<*>,
-    private val wrapped: Call<Any>
-  ) : Call<Any> {
-    override fun enqueue(callback: Callback<Any>) {
-      val allInterceptors = interceptors + RetrofitRequestInterceptor(wrapped)
-      RealClientChain(action, args, wrapped, callback, allInterceptors).proceed(args, callback)
-    }
-
-    override fun execute(): Response<Any> {
```
swankjesse, line 149: # ⚠️ Behavior change!

Previously this abstraction caused all network calls to use a thread from OkHttp’s dispatcher thread pool. With this change synchronous calls won’t use that extra thread.

r3mariano (approved):

PR #1770, opened 2021-02-18
Some attempts to fix flakey test - can't replicate locally :(

chris-ryan-square:


misk/src/main/kotlin/misk/web/jetty/JettyService.kt
```diff
@@ -282,7 +302,9 @@ class JettyService @Inject internal constructor(
     val stopwatch = Stopwatch.createStarted()
     logger.info("Stopping Jetty")
 
-    server.stop()
+    if (server.isRunning) {
+      server.stop()
+    }
```
chris-ryan-square, line 307: This is the real change in the file - don't try to stop if we're not running

r3mariano (approved):
nit: This has quite a few formatting changes that make it harder to read, but overall looks good.

mightyguava:
You can actually ssh into the node running the tests in circleci and reproduce the failure

PR #1776, opened 2021-02-22
use client config when configuring grpc client provider

podlech:


swankjesse (approved):

swankjesse:
Nice find.

misk/src/main/kotlin/misk/client/OkHttpClientCommonConfigurator.kt
```diff
@@ -0,0 +1,66 @@
...
+import com.google.inject.Singleton
+import okhttp3.ConnectionPool
+import okhttp3.Dispatcher
+import okhttp3.OkHttpClient.Builder
+import java.time.Duration
+import java.util.concurrent.TimeUnit
+
+@Singleton
+class OkHttpClientCommonConfigurator @Inject internal constructor() {
+  fun configure(builder: Builder, config: HttpClientEndpointConfig): Builder {
```
frojasg, line 13: Should we inject `Provider<HttpClientsConfig>` so we don't need to pass `HttpClientEndpointConfig` as an argument?

frojasg (approved):
LGTM!

misk/src/main/kotlin/misk/client/OkHttpClientCommonConfigurator.kt
podlech: Unfortunately [HttpClientFactory.get](https://github.com/podlech/misk/blob/ce105346cf587e473ef5e46c3335ad3235e2cab5/misk/src/main/kotlin/misk/client/HttpClientFactory.kt#L28) receives the config as a parameter, so I'll keep it as is.

swankjesse (approved):

PR #1777, opened 2021-02-24
[NO TICKET] Add feature flag key validation

angles-n-daemons:
Adds simple validation rules on feature flag keys that map to LaunchDarkly, Optimizely, Unleash, Splitio and likely other feature flagging system's constraints.

This is done to prevent incorrectly named feature flag keys from passing builds - and providing engineers with more insight as to why the build will have failed.


misk-feature/src/main/kotlin/misk/feature/FeatureFlags.kt
```diff
@@ -109,7 +109,18 @@ inline fun <reified T> FeatureFlags.getJson(
 /**
  * Typed feature string.
  */
-data class Feature(val name: String)
+data class Feature(val name: String) {
+  init {
+    if (name.length == 0) {
+      throw RuntimeException("feature name must not be empty")
+    }
+    if (!name.contains(Regex("^[0-9A-Za-z-_.]+$"))) {
```
angles-n-daemons, line 117: TIL that `[A-z]` captures more than just letters 😭

misk-feature-testing/src/test/kotlin/misk/feature/testing/FakeFeatureFlagsTest.kt
```diff
@@ -281,6 +281,41 @@ internal class FakeFeatureFlagsTest {
...
+    Feature("abc123")
+    Feature("abc.123")
+    Feature("abc.123-doeraemi")
+    Feature("abc.123-doeraemi_YOUME")
+  }
+
+  @Test
+  fun inValidFeatureKey() {
+    assertThrows<RuntimeException> {
+      println(Feature(""))
```
r3mariano, line 297: nit: no need for println?

r3mariano (approved):
Looks great! Thanks for jumping onto this.

misk-feature-testing/src/test/kotlin/misk/feature/testing/FakeFeatureFlagsTest.kt
angles-n-daemons: whoops 😅

PR #1779, opened 2021-02-25
Add a Pager.listAll overload that takes a Session

staktrace:
This is mostly for convenience; sometimes it's useful to use the listAll
function from inside an existing transaction.


staktrace:
The test failure seems unrelated to the patch, AFAICT

staktrace:
Per discussion with @swankjesse I'm going to close this PR. The rationale is that this introduces a footgun-ny API which makes it easy to accidentally create giant transactions. The existing `listAll` function reads one page per transaction and therefore uses small transactions.

PR #1780, opened 2021-02-25
Adds access to the fakeClock from static context.

xethorn:
For unit-tests as well as areas where bindings are not always available, we need to be able to access directly the fakeclock instance.

### DIFF
--- a/misk-testing/src/main/kotlin/misk/time/FakeClockModule.kt
+++ b/misk-testing/src/main/kotlin/misk/time/FakeClockModule.kt
@@ -6,6 +6,10 @@ import java.time.Clock
 class FakeClockModule : KAbstractModule() {
   override fun configure() {
     bind<Clock>().to<FakeClock>()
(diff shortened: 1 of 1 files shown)


Kuresov (approved):

swankjesse:
Why aren’t bindings available?

This is broken for parallel test execution: two injectors shouldn’t share the same clock! Also it changes the behavior; previously each test got a fresh fake clock; now the fake clock inherits state from the previous test.

Would you consider putting the static fakeClock in your application code? Perhaps this?

```
class MyTest {
  @Inject lateinit var fakeClock

  @Before
  fun setUp() {
    sharedFakeClock = fakeClock
  }
  ...
}
```

```
var sharedFakeClock = FakeClock()
```

xethorn:
The issue comes from the inability to `@Inject lateinit var fakeClock`. Since the code is written in a way that aims to establish the clock as a singleton, I was hopeful we could assess it without having to go through the dependency injection.

PR #1781, opened 2021-02-25
Use wire to write json

r3mariano:
This is a breaking change: Misk will now encode json from protos with wire,
which changes how it's encoded for kotlin keywords like "default".


swankjesse (approved):

swankjesse:
Yes please.

PR #1792, opened 2021-03-08
Allow better configuration over the set of SSL protocols used

staktrace:


misk/src/main/kotlin/misk/security/ssl/TlsProtocols.kt
```diff
@@ -3,4 +3,6 @@ package misk.security.ssl
 object TlsProtocols {
   /** List of safe to use TLS protocols, in preferred order */
   val safe = arrayOf("TLSv1.2", "TLSv1.1")
+  /** List of modern TLS protocols for extra security, in preferred order */
+  val modern = arrayOf("TLSv1.3")
```
swankjesse, line 7: I really like Google Cloud’s terms for these. In their terminology, `safe` would be `modern`. And maybe rename this `restricted` and add TLSv1.2 ? 

https://cloud.google.com/load-balancing/docs/ssl-policies-concepts

misk/src/main/kotlin/misk/web/WebConfig.kt
```diff
@@ -91,7 +91,9 @@ data class WebSslConfig(
   val port: Int,
   val cert_store: CertStoreConfig,
   val trust_store: TrustStoreConfig? = null,
-  val mutual_auth: MutualAuth = MutualAuth.REQUIRED
+  val mutual_auth: MutualAuth = MutualAuth.REQUIRED,
+  /** When set to true, only allow modern ciphers and protocols (as of this writing, TLS v1.3+). */
+  val modern_protocols_only: Boolean = false,
```
swankjesse, line 96: I like the term modern, but only if it’s consistent with `modern` from the Google Cloud site.

swankjesse: (Their definition of modern will change over time, as should ours!)

misk/src/main/kotlin/misk/security/ssl/TlsProtocols.kt
```diff
@@ -2,5 +2,7 @@ package misk.security.ssl
 
 object TlsProtocols {
   /** List of safe to use TLS protocols, in preferred order */
-  val safe = arrayOf("TLSv1.2", "TLSv1.1")
+  val compatible = arrayOf("TLSv1.3", "TLSv1.2", "TLSv1.1")
```
swankjesse, line 5: love it

swankjesse (approved):

PR #1797, opened 2021-03-10
Don't track the previous session

swankjesse:
This was a source of complexity and potential memory leaks for very little value.

### DIFF
--- a/misk-hibernate/src/main/kotlin/misk/hibernate/RealTransacter.kt
+++ b/misk-hibernate/src/main/kotlin/misk/hibernate/RealTransacter.kt
@@ -32,7 +32,6 @@ import java.util.concurrent.Callable
 import java.util.concurrent.CompletableFuture
 import java.util.concurrent.Future
 import java.util.concurrent.TimeUnit
-import javax.inject.Provider
 import javax.persistence.OptimisticLockException
 import kotlin.reflect.KClass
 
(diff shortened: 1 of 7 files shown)


misk-hibernate/src/main/kotlin/misk/hibernate/RealTransacter.kt
```diff
@@ -223,17 +210,6 @@ internal class RealTransacter private constructor(
...
-  /**
-   * Returns a string describing the most recent attempt. This includes whether the attempt used the
-   * same connection which might help in diagnosing stale data problems.
-   */
-  private fun attemptNote(attempt: Int): String {
-    if (attempt == 1) return "attempt 1"
-    val latestSession = threadLatestSession.get()!!
-    if (latestSession.sameConnection) return "attempt $attempt, same connection"
-    return "attempt $attempt, different connection"
-  }
```
swankjesse, line 235: This function required us to remember the previous session, which I think was a pretty sketchy thing for us to do. Not remembering it makes so many other things simpler!

misk-hibernate/src/main/kotlin/misk/hibernate/SessionFactoryService.kt
```diff
@@ -46,6 +46,10 @@ internal class SessionFactoryService(
 ) : AbstractIdleService(), Provider<SessionFactory>, TransacterService {
   private var sessionFactory: SessionFactory? = null
 
+  val threadInTransaction = object : ThreadLocal<Boolean>() {
+    override fun initialValue() = false
+  }
```
swankjesse, line 51: Much of the other changes in this PR are so we can put the in-transaction ThreadLocal here on the SessionFactoryService. It’s 1:1 with the database!

chris-ryan-square (approved):

PR #1799, opened 2021-03-12
Simple cron runner

scotte-square:
Easily configure a Runnable instance to run on a crontab pattern.

It is expected that each pod will have its own instances of CronManager and that the CronTask lease will ensure only one of these gets called each minute.  If the CronManager's last run time is more than 3 minutes ago, it assumes that its queue is out of date and rebuilds with new execution times based on the current time.  Thus, if the responsibility for execution switches pods, then it is possible that an execution may get missed.


alanpaulin:
> Thus, if the responsibility for execution switches pods, then it is possible that an execution may get missed.
Is it better to try to run the task at least once, than to skip an execution?

misk-cron/src/main/kotlin/misk/cron/CronManager.kt
```diff
@@ -0,0 +1,136 @@
...
+import java.time.ZonedDateTime
+import java.util.PriorityQueue
+import javax.inject.Inject
+import javax.inject.Singleton
+
+@Singleton
+class CronManager @Inject constructor(moshi: Moshi) {
+  @Inject private lateinit var clock: Clock
+  @Inject private lateinit var jobQueue: JobQueue
+  @Inject @ForMiskCron private lateinit var queueName: QueueName
```
alanpaulin, line 24: For a service to use a cron in Misk, this impliues that they need to setup a job queue with the name that gets injected here?

misk-cron/src/main/kotlin/misk/cron/CronManager.kt
```diff
@@ -0,0 +1,136 @@
...
+      logger.info { "Rebuilding cron queue required" }
+      rebuildQueue()
+    }
+
+    logger.info {
+      "Looking for ready crons - ${cronQueue.size} entries to scan, current time=$now"
+    }
+
+    while (cronQueue.isNotEmpty()) {
+      val cronQueueEntry = cronQueue.peek()
```
alanpaulin, line 92: Is `cronQueue` sorted by date that the entry was added (like a FIFO queue), or by the entry's `executeAt`?

misk-cron/src/main/kotlin/misk/cron/CronManager.kt
```diff
@@ -0,0 +1,136 @@
...
+      if (cronQueueEntry.executeAt > now) break
+
+      cronQueue.remove()
+      val cronEntry = cronQueueEntry.cronEntry
+      logger.info {
+        "CronJob ${cronEntry.name} was ready at " +
+          "${ZonedDateTime.ofInstant(cronQueueEntry.executeAt, zoneId)}"
+      }
+
+      enqueueCronJob(cronEntry)
```
alanpaulin, line 102: I see, so when the entry from the cronQueue is ready, we enqueue it in a job queue and process it on a handler thread. Nice.

misk-cron/src/main/kotlin/misk/cron/CronTask.kt
```diff
@@ -0,0 +1,42 @@
...
+  }
+
+  override fun shutDown() {
+    logger.info { "Stopping CronTask" }
+    cronClient.removeAllCrons()
+  }
+
+  companion object {
+    private const val CLUSTER_LEASE_NAME = "misk.cron.crontask-cluster"
+    private val INTERVAL = Duration.ofSeconds(60L)
```
alanpaulin, line 38: Is it typical for a cron job to have an error of up to 60 seconds?

misk-cron/src/test/kotlin/misk/cron/CronTest.kt
```diff
@@ -0,0 +1,117 @@
...
+import misk.testing.MiskTestModule
+import misk.time.FakeClock
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.jupiter.api.Test
+import java.time.Duration
+import javax.inject.Inject
+import javax.inject.Singleton
+
+@MiskTest(startService = true)
+class CronTest {
```
alanpaulin, line 15: Nice tests

alanpaulin (approved):
LGTM, though someone more familiar with Misk should also review.

scotte-square:
@alanpaulin re the switching of pods, another way it could be done is to have all the pods runReadyCrons but only one of them actually does run them (the one with the lease).  That way they all stay in sync and there would be no need to rebuild the queue on a switch over.

misk-cron/src/main/kotlin/misk/cron/CronTask.kt
scotte-square: I think for the granularity we're doing that would be fine.  I don't think we would have any tasks that would run every 60 seconds.

misk-cron/src/main/kotlin/misk/cron/CronManager.kt
scotte-square: CronQueue is ordered by next execution so it just has to go through until one in the future is found and then it's done.

misk-cron/src/main/kotlin/misk/cron/CronManager.kt
scotte-square: Correct - the CronJobHandler needs to be registered and the cron service registered plus individual crons registered as needed.

scotte-square:
Switched to each CronManager checking the lease to run or queue a ready entry

misk-cron/src/main/kotlin/misk/cron/CronTask.kt
```diff
@@ -0,0 +1,47 @@
...
+import misk.tasks.Status
+import java.time.Clock
+import java.time.Duration
+import javax.inject.Inject
+import javax.inject.Singleton
+
+@Singleton
+internal class CronTask @Inject constructor() : AbstractIdleService() {
+  @Inject private lateinit var clock: Clock
+  @Inject private lateinit var cronClient: CronManager
```
r3mariano, line 16: nit: `cronManager: CronManager`?

misk-cron/src/main/kotlin/misk/cron/CronManager.kt
```diff
@@ -0,0 +1,99 @@
...
+    cronEntries.values.forEach { cronEntry ->
+      val nextExecutionTime = cronEntry.executionTime.nextExecution(previousTime).orElseThrow()
+        .withSecond(0)
+        .withNano(0)
+
+      if (nextExecutionTime.toInstant() <= now) {
+        logger.info {
+          "CronJob ${cronEntry.name} was ready at $nextExecutionTime"
+        }
+        enqueueCronJob(cronEntry)
```
r3mariano, line 70: Why enqueue jobs rather than putting them in a different thread pool?

r3mariano (approved):

misk-cron/src/main/kotlin/misk/cron/CronManager.kt
scotte-square: Tell me more about this and how that would work?  My thinking is that the work to be done shouldn't be done on the CronTask thread since we don't know how long that would take - hence the job queue.

scotte-square: Ah, I see and I think that might be a better pattern

scotte-square: @r3mariano PTAL

r3mariano: LGTM!

PR #1803, opened 2021-03-18
Introduce statusCode-compatible call in WebActionException

r3mariano:
...to make migration to WebActionException easier. This should remove the breaking change in #1801

### DIFF
--- a/misk-actions/src/main/kotlin/misk/exceptions/Exceptions.kt
+++ b/misk-actions/src/main/kotlin/misk/exceptions/Exceptions.kt
@@ -56,14 +56,9 @@ open class WebActionException(
   val isServerError = code in (500..599)
 
   @Deprecated("Use code instead")
-  val statusCode: StatusCode = run {
-    for (value in StatusCode.values()) {
-      if (value.code == code) {
-        return@run value
-      }
-    }
-    throw IllegalStateException("Response code is not in StatusCode. Use `code` instead")
-  }
+  val statusCode: StatusCode
+    get() = StatusCode.values().find { it.code == code }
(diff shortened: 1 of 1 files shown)


misk-actions/src/main/kotlin/misk/exceptions/Exceptions.kt
```diff
@@ -54,6 +54,16 @@ open class WebActionException(
 ) : Exception(message, cause) {
   val isClientError = code in (400..499)
   val isServerError = code in (500..599)
+
+  @Deprecated("Use code instead")
+  val statusCode: StatusCode = run {
```
tristandl, line 59: what's the advantage of using `run` over a `get()`?

tristandl (approved):

misk-actions/src/main/kotlin/misk/exceptions/Exceptions.kt
```diff
@@ -54,6 +54,16 @@ open class WebActionException(
...
+    for (value in StatusCode.values()) {
+      if (value.code == code) {
+        return@run value
+      }
+    }
+    throw IllegalStateException("Response code is not in StatusCode. Use `code` instead")
+  }
```
tristandl, line 66:
```suggestion
  val statusCode: StatusCode
    get() = StatusCode.values().find { it.code == code }
      ?: throw IllegalStateException("Response code is not in StatusCode. Use `code` instead")
```

chris-ryan-square (approved):

misk-actions/src/main/kotlin/misk/exceptions/Exceptions.kt
r3mariano: Nothing really, just wanted to have it throw here.

PR #1806, opened 2021-03-20
Add mechanism to add additional MDC context to web requests

frojasg:
Description
-----------
Applications owners want to be able to add additional contextual
information in the MDC in such a way that gets included into the
`ExceptionHandlingInterceptor` and `RequestLoggingInterceptor`.

There are two possible ways to implements this:
* Add a way to enrich the `RequestLogContextInterceptor`. This approach
  is less flexible but way simpler to understand.
* Add a way to change the order in which web interceptors are executed,
  so application owners can put their interceptors before
  `RequestLoggingInterceptor`. However, this approach could be error-prone
  as interceptor ordering is complicated.


misk/src/main/kotlin/misk/web/interceptors/RequestLogContextInterceptor.kt
```diff
@@ -17,31 +18,37 @@ import javax.servlet.http.HttpServletRequest
...
+      logContextProviders.keySet().forEach { MDC.remove(it) }
     }
   }
 
   @Singleton
   class Factory @Inject internal constructor(
     private val currentCaller: @JvmSuppressWildcards ActionScoped<MiskCaller?>,
-    private val currentRequest: @JvmSuppressWildcards ActionScoped<HttpServletRequest>
+    private val currentRequest: @JvmSuppressWildcards ActionScoped<HttpServletRequest>,
+    @Suppress("PLATFORM_CLASS_MAPPED_TO_KOTLIN")
```
swankjesse, line 47: alternately, `MutableMap`

swankjesse (approved):
Test case please!

frojasg:
@swankjesse [RequestLogContextInterceptorTest](https://github.com/cashapp/misk/blob/master/misk/src/test/kotlin/misk/web/interceptors/RequestLogContextInterceptorTest.kt) is testing the behavior still works

clarkds14:
LGTM @frojasg

PR #1807, opened 2021-03-22
Start on splitting up misk code to non-DI dependant code - 'wisp' mod…

chris-ryan-square:
…ules

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,3 +120,11 @@ Misk-Web powers the Misk Admin Dashboard with modular Typescript + React powered
 [misk]: https://cashapp.github.io/misk/
 [miskweb]: https://cashapp.github.io/misk-web/
 [snap]: https://oss.sonatype.org/content/repositories/snapshots/
+
+## What are the wisp* modules?
+
+The wisp* modules contain no Dependency Injection based code (i.e. no Guice, etc).  These modules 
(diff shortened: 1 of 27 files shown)


r3mariano:
lgtm

misk-core/src/main/kotlin/misk/logging/Logging.kt
```diff
@@ -2,14 +2,27 @@ package misk.logging
...
+@Deprecated(
+  "Use wisp-logging",
+  ReplaceWith(
+    "getLogger<T>()",
+    "wisp.logging.getLogger",
+  )
+)
 inline fun <reified T> getLogger(): KLogger {
-  return KotlinLogging.logger(T::class.qualifiedName!!)
+  return wisp.logging.getLogger<T>()
```
r3mariano, line 25: I realise I've taken a different approach locally - instead of deprecating and calling them `wisp.logging.*` I've kept the same package name and moved them to a different module. I think this will reduce disruption. Thoughts on changing package vs keeping it the same?

chris-ryan-square: Yes.  I want to keep the wisp structure clean from any misk modules.  So for most things I've been wrapping the wisp mosules in the misk packages.  The deprecations are because I could not do it here.

r3mariano (approved):

PR #1814, opened 2021-03-23
Added ExternalDataKeys annotation

yoavamit:


misk-crypto/src/main/kotlin/misk/crypto/ExternalDataKeys.kt
```diff
@@ -0,0 +1,18 @@
...
+ *
+ * This annotation can be useful for cases where one would need access to all external data
+ * keys available to the service.
+ */
+@Qualifier
+@Target(
+  AnnotationTarget.FIELD,
+  AnnotationTarget.VALUE_PARAMETER
+)
+annotation class ExternalDataKeys()
```
yan, line 18: Should this be `public` or `internal`? I assume we'd export primitives/key managers, not raw keys.

yoavamit: This is a first step towards exposing a more convenient interface and for now, I just want to be able to expose the map of key name -> key type for external keys.

I think it's useful to have it as a public annotation so that other consumers could use it as well.
It doesn't expose any key material so it should be safe to do.

In other words - this is just a convenient way to get to the external keys block specified in the configuration.

natvyas (approved):

PR #1824, opened 2021-03-25
Use all outer class names for nested actions

r3mariano:
The typical use case for this is when people write nested gRPC actions, e.g.

```
class MyAction {
  class GrpcAction {
    // ...
  }
}
```

This changes the action name for eg metrics purposes from `GrpcAction` to `MyAction.GrpcAction`.


chris-ryan-square (approved):

misk/src/main/kotlin/misk/Actions.kt
```diff
@@ -21,7 +21,9 @@ fun KFunction<*>.asAction(
   val actualParameters = parameters.drop(1)
   val actionName = instanceParameter.type.classifier?.let {
     when (it) {
-      is KClass<*> -> it.simpleName
+      // Use the class name including all outer classes, but excluding the package (assuming a
+      // packages start with a lowercase character and classes start with an uppercase one).
+      is KClass<*> -> it.qualifiedName?.replace(Regex("^[a-z.]*"), "")
```
quad, line 26: Would `it.java.package?.name` work to get the package name?

quad (approved):

misk/src/main/kotlin/misk/Actions.kt
r3mariano: Oh, I like that a lot more. Will change up.

PR #1830, opened 2021-03-26
Support non-standard key bucket region

yan:
Allow for the ability to refer to keys in a non-local region


misk-crypto/src/main/kotlin/misk/crypto/S3ExternalKeyManager.kt
```diff
@@ -32,21 +35,31 @@ import misk.logging.getLogger
...
+    logger.info("creating S3ExternalKeyManager S3 client for $region")
+    AmazonS3ClientBuilder
+      .standard()
+      .withRegion(region)
+      .withCredentials(awsCredentials)
+      .build()
+  } ?: defaultS3
 
-  private val metadataKeyKeyType = "key-type"
+  private val logger by lazy { getLogger<S3ExternalKeyManager>() }
```
yoavamit, line 62: super nit: I think we usually prefer to initialize the logger object inside a `companion object {}` to make it static

yoavamit (approved):

Issue #1145, opened 2019-08-12
Allow Operator.EQ to match nulls for Hiberate Queries


keeferrourke:
e.g. if we want to persist an object with a `String?` property and run a query that matches on that property, right now you have to do something cumbersome like this:

```kotlin
interface MaybeStringQuery: Query<DbMaybeString> {
  @Constraint(path = "maybe_string", operator = Operator.EQ)
  hasString(string: String?): MaybeStringQuery
  
  @Constraint(path = "maybe_string", operator = Operator.IS_NULL)
  nullString(): MaybeStringQuery
}

data class MaybeString(val string: String?)

...

transacter.transaction { session ->
  queryFactory.newQuery<MaybeStringQuery>()
      .apply {  // Annoying part!
        if (maybeString.string == null) nullString()
        else hasString(maybeString.string)
      }.list(session)
}
```

It would be nice if `Operator.EQ` or a new operator captured null equality as well, so we could just do `hasString(null)` to match nullable fields.

amitdev:
Is there any objection in fixing this? We also hit similar problem in cash-idv. I think it makes sense for the `EQ` operator to generate `isNull` if the value is null. I can raise a PR for this if it sounds good.

PR #1833, opened 2021-03-30
Generate is null condition for nulls in Operator.EQ

amitdev:
This is a potential fix for issue #1145. Another option is to introduce a new operator that also handles null, but I think it is intuitive for `EQ` to behave this way. This would avoid adding multiple `@Constraint` just for handing nulls.


misk-hibernate/src/main/kotlin/misk/hibernate/ReflectionQueryFactory.kt
```diff
@@ -94,8 +94,14 @@ internal class ReflectionQuery<T : DbEntity<T>>(
   fun addConstraint(path: List<String>, operator: Operator, value: Any?) {
     when (operator) {
       Operator.EQ -> {
```
r3mariano, line 96: It might be better to use a new operator for this, e.g. `EQ_OR_IS_NULL`. I don't think this should break anything in `EQ`, but just want to be sure.

r3mariano (approved):

r3mariano:
changes look good to me!

PR #1845, opened 2021-04-09
Stop SQS receiver if the service is shutting down

frojasg:
Description
-----------
This is causing scary logs when the service is being deploy (old pods
are being shutdown)


tso (approved):
LGTM! Unsure if there's any negative effects that I don't know about. Maybe @swankjesse could take a look, given the expertise on guice/guava services/etc?

PR #1846, opened 2021-04-13
Remove MiskTestingServiceModule from WebTestingModule

frojasg:
Most misk tests already requires the inclusion of `MiskTestingServiceModule`, and if there's more than one module importing you will get the following error
```
A binding to com.google.common.util.concurrent.ServiceManager was already configured at misk.ServiceManagerModule.provideServiceManager$misk_service() 
```

### DIFF
--- a/misk-testing/src/main/kotlin/misk/web/WebTestingModule.kt
+++ b/misk-testing/src/main/kotlin/misk/web/WebTestingModule.kt
@@ -1,5 +1,6 @@
 package misk.web
 
+import misk.MiskTestingServiceModule
 import misk.environment.Environment
 import misk.environment.EnvironmentModule
 import misk.inject.KAbstractModule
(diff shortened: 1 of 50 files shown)


r3mariano (approved):

misk-testing/src/main/kotlin/misk/web/WebTestingModule.kt
```diff
@@ -11,12 +11,27 @@ import misk.security.ssl.SslLoader
  * A module that starts an embedded Jetty web server configured for testing. The server supports
  * both plaintext and TLS.
  */
+@Deprecated(
+  message = "This modules incorrectly imports MiskTestingServiceModule",
```
r3mariano, line 15: Better to say to "install both WebServerTestingModule and MiskTestingServiceModule instead" but also should have more reasoning here than that just that it's incorrect.

misk-testing/src/main/kotlin/misk/web/WebTestingModule.kt
```diff
@@ -11,12 +11,27 @@ import misk.security.ssl.SslLoader
...
-    install(EnvironmentModule(Environment.TESTING))
+    install(WebServerTestingModule(webConfig))
     install(MiskTestingServiceModule())
+  }
+  companion object {
+    val TESTING_WEB_CONFIG =  WebServerTestingModule.TESTING_WEB_CONFIG
+  }
+}
+
+class WebServerTestingModule(
```
r3mariano, line 30: Should copy the Javadoc for WebTestingModule here.

r3mariano (approved):
Nice!

PR #1850, opened 2021-04-14
Additional information on web actions tab

benaskins:
In an effort to make our server APIs self-documenting and more useful to our client engineers, we've made a few changes to the admin web actions tab.

* show request parameters and response type clearly on the web action card
* group action cards by package type so that it's clear which actions are implemented by the service, and which are included by Misk
* added a new annotation for documenting an action

To demonstrate how this works for an action that supports protobufs we've also added a new action to the exemplar app.

# How it looks

# Next steps

One additional step we'd like to take is provide a config option to support linking to a service that hosts protobuf definitions, but wanted to understand the appetite for these changes first.


samples/exemplar/src/main/kotlin/com/squareup/exemplar/actions/HelloWebProtoAction.kt
```diff
@@ -0,0 +1,29 @@
...
+import misk.web.actions.WebAction
+import misk.web.mediatype.MediaTypes
+import javax.inject.Inject
+
+class HelloWebProtoAction @Inject constructor() : WebAction {
+  @Post("/hello/proto/")
+  @Unauthenticated
+  @RequestContentType(MediaTypes.APPLICATION_PROTOBUF)
+  @ResponseContentType(MediaTypes.APPLICATION_PROTOBUF)
+  @Documentation("""
```
r3mariano, line 21: `@Description` might work better.

I wonder as well if this can work for javadoc comments rather than introducing an annotation...

r3mariano (commented):
I haven't tried locally, but looking good so far! Let us know when it's ready to fully review.

samples/exemplar/src/main/kotlin/com/squareup/exemplar/actions/HelloWebProtoAction.kt
benaskins: I explored javadoc comments but couldn't find a way to reflect on those from within misk. We could perhaps add javadoc generation to the build process and link to the generated documentation from another admin tab?

misk/src/main/kotlin/misk/web/metadata/webaction/WebActionMetadata.kt
```diff
@@ -46,10 +52,19 @@ data class WebActionMetadata(
   ) : this(
     name = name,
     function = function.toString(),
+    packageName = packageName(function.toString()),
```
ThinTim, line 55: The `function` string is created from a `KFunction` :
https://github.com/cashapp/misk/blob/9593078083385e235dda816af539bd48bae616c3/misk/src/main/kotlin/misk/Action.kt#L14

So we might be able to get the package name easier by calling [javaMethod](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.reflect.jvm/java-method.html).[getDeclaringClass()](https://docs.oracle.com/javase/8/docs/api/java/lang/reflect/Method.html#getDeclaringClass--).[getPackage()](https://docs.oracle.com/javase/8/docs/api/java/lang/Class.html#getPackage--) and adding that as another string parameter to `WebActionMetadata`.

samples/exemplar/src/main/kotlin/com/squareup/exemplar/actions/HelloWebProtoAction.kt
ThinTim: Unfortunately comments [don't seem to be included in the compiled class](https://stackoverflow.com/a/4135365) so annotations are the only way to go.

tristandl: i was looking at this a few weeks ago too - i don't think reflection can do it. I was looking at dokka (kotlin codedoc tool) and was imagining something like writing code doc to /resources or /build or wherever is appropriate

misk/src/main/kotlin/misk/web/metadata/webaction/WebActionMetadata.kt
benaskins: Thanks Tim! Addressed in: https://github.com/cashapp/misk/pull/1850/commits/e5bfa6f5c5c70522b57ded6a029e42636d41b2d0

ThinTim:
@benaskins the "send a request" widget already seems to have access to the parameters needed to make a request to the action. Where is it pulling it's information from?

misk/web/tabs/web-actions/src/rewrite/WebActionParameters.tsx
```diff
@@ -0,0 +1,25 @@
...
+
+export default function WebActionParameters({ parameters }: Props) {
+  if (parameters.length == 0) {
+    return null
+  }
+
+  return (
+    <Card style={{ marginTop: "12px", padding: "12px" }}>
+      <H5> Request Parameters </H5>
+      <HTMLTable>
```
tso, line 18: Is there any way to get the space below here to be less big? It looks kinda bleh to have
image

tso (approved):
other than the minor visual nit lgtm for react stuff

adrw (approved):
Might be a personal preference, but I'd be inclined to wrap the new data in full width `MetadataCollapse` blocks so that if I'm scrolling to find an action, I don't have to scroll through tons of request definitions and documentation I'm not interested in.

misk/web/tabs/web-actions/src/rewrite/WebActionParameters.tsx
benaskins: Addressed in: https://github.com/cashapp/misk/pull/1850/commits/c710c167ddcebbf73c5b752e103294c98ed6c3c5

benaskins:
> @benaskins the "send a request" widget already seems to have access to the parameters needed to make a request to the action. Where is it pulling its information from?

@ThinTim this is not looking at each request parameter, but is looking at whichever parameter is annotated with `@RequestBody` and then deconstructing the proto to build the form.

benaskins:
> Might be a personal preference, but I'd be inclined to wrap the new data in full width `MetadataCollapse` blocks so that if I'm scrolling to find an action, I don't have to scroll through tons of request definitions and documentation I'm not interested in.

Address in: https://github.com/cashapp/misk/pull/1850/commits/4f988fffebcef20caf74bc063437b17772765691

aaron-edwards (approved):

PR #1858, opened 2021-04-19
Use safe call on RequestResponseCapture.get

podlech:
bodyCapture.get() sometimes throws exceptions, adding try/catch to
capture them.

### DIFF
--- a/misk/src/main/kotlin/misk/web/interceptors/RequestBodyLoggingInterceptor.kt
+++ b/misk/src/main/kotlin/misk/web/interceptors/RequestBodyLoggingInterceptor.kt
@@ -90,9 +90,7 @@ internal class RequestResponseCapture @Inject constructor() {
     private val capture = ThreadLocal<RequestResponseBody>()
   }
 
-  fun get(): RequestResponseBody {
-    return capture.get()
-  }
+  fun get(): RequestResponseBody? = capture.get()
 
   fun set(value: RequestResponseBody) {
     capture.set(value)


misk/src/main/kotlin/misk/web/interceptors/RequestLoggingInterceptor.kt
```diff
@@ -122,7 +122,11 @@ class RequestLoggingInterceptor internal constructor(
     val sampling = if (isError) errorBodySampling else bodySampling
     val randomDouble = random.current().nextDouble()
     if (randomDouble < sampling) {
-      val requestResponseBody = bodyCapture.get()
+      val requestResponseBody = try {
+        bodyCapture.get()
+      } catch (e: Throwable) {
```
r3mariano, line 127: I think it would be a lot more obvious to do an isPresent() check rather than catching an NPE.

adrw: +1, exception based control is definitely less preferable if there's an available method we can use instead

podlech: I've changed the code to use safe calls instead, please review again.

misk/src/main/kotlin/misk/web/interceptors/RequestBodyLoggingInterceptor.kt
```diff
@@ -90,9 +90,7 @@ internal class RequestResponseCapture @Inject constructor() {
     private val capture = ThreadLocal<RequestResponseBody>()
   }
 
-  fun get(): RequestResponseBody {
-    return capture.get()
-  }
+  fun get(): RequestResponseBody? = capture?.run { get() }
```
r3mariano, line 93: Something like this:
```suggestion
  fun get(): RequestResponseBody? {
    if (capture.isPresent()) {
      return capture.get()
    }
    return null;
  }
```

r3mariano: Ah, no, this isn't quite right either.

r3mariano, line 93: capture is never null, but capture.get() can be. All you need is to change the output type.

```suggestion
  fun get(): RequestResponseBody? = capture.get()
```

r3mariano (approved):
Nice, thanks!

Also would be good to write a test for this, but not a blocker for merging.

misk/src/main/kotlin/misk/web/interceptors/RequestBodyLoggingInterceptor.kt
podlech: Got it! Changed to return a null-safe type.

PR #1861, opened 2021-04-22
Show proto fields for protobuf based web actions

benaskins:
When an action's request type or response type is a proto show the field definitions for that proto.

Before:

Misk Web Actions - Google Docs 2021-04-22 11-52-37

After:

Misk Web Actions - Google Docs 2021-04-22 11-53-10

### DIFF
--- a/misk/src/main/kotlin/misk/web/metadata/webaction/WebActionMetadata.kt
+++ b/misk/src/main/kotlin/misk/web/metadata/webaction/WebActionMetadata.kt
@@ -26,7 +26,9 @@ data class WebActionMetadata(
   val parameters: List<ParameterMetaData>,
   val requestType: String?,
   val returnType: String,
+  val responseType: String,
   val types: Map<String, MiskWebFormBuilder.Type>,
+  val responseTypes: Map<String, MiskWebFormBuilder.Type>,
   val pathPattern: String,
   val applicationInterceptors: List<String>,
   val networkInterceptors: List<String>,
@@ -45,6 +47,7 @@ data class WebActionMetadata(
     parameters: List<KParameter>,
     requestType: KType?,
     returnType: KType,
+    responseType: KType?,
     pathPattern: PathPattern,
     applicationInterceptors: List<ApplicationInterceptor>,
     networkInterceptors: List<NetworkInterceptor>,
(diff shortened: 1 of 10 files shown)


r3mariano (approved):

misk/web/tabs/web-actions/src/rewrite/WebActionCardBody.tsx
```diff
@@ -3,8 +3,9 @@ import { Menu, H5, UL } from "@blueprintjs/core"
 import { WebActionMetadata } from "./types"
 import WebActionCollapse from "./WebActionCollapse"
 import WebActionSendRequest from "./WebActionSendRequest"
-import WebActionParameters from "./WebActionParameters";
-import WebActionDescription from "./WebActionDescription";
+import WebActionRequestParameters from "./WebActionRequestParameters"
+import WebActionDescription from "./WebActionDescription"
+import WebActionResponseType from "./WebActionResponseType";
```
adrw, line 8: nit: rm semicolon

misk/src/main/kotlin/misk/web/metadata/webaction/WebActionMetadata.kt
```diff
@@ -69,7 +72,9 @@ data class WebActionMetadata(
     },
     requestType = requestType.toString(),
     returnType = returnType.toString(),
```
adrw, line 74: Could be wrong, but is return type == response type? And I think the return type is already included in the `types` map so the `responseTypes` map would be duplicative

adrw (commented):
Curious if this can be done with the existing `types` map instead of adding a duplicate `responseTypes`.

misk/web/tabs/web-actions/src/rewrite/WebActionCardBody.tsx
benaskins: I'm finding that IntelliJ doesn't format the code in the way I'd expect when automatically importing. I probably need to set some options here. Do you have any tips?

misk/src/main/kotlin/misk/web/metadata/webaction/WebActionMetadata.kt
benaskins: Unfortunately not. `returnType` will be set to the literal Kotlin type return by the function that implements the action. For example `misk.web.Response<com.squareup.exemplar.protos.HelloWebResponse>`. For the response type I'm reflecting on the `KType` to determine the classified class, in this case `com.squareup.exemplar.protos.HelloWebResponse`.

Currently `types` only includes the types for the request parameter annotated with `@RequestBody` - it doesn't include types for the response.

misk/web/tabs/web-actions/src/rewrite/WebActionCardBody.tsx
benaskins: I think I've got this sorted now. Thanks for the pickup.

PR #1876, opened 2021-04-29
ConcurrencyLimitsInterceptor only uses rate limited ERROR-level logging

zcross:
IMO, the rate limited error logging in this is enough. As commented previously, folks can reliably use RPC response code metrics to understand how often load shedding is occurring.

Unconditional `INFO`-level logging just fills high traffic services' logs with an unactionable message. Even if logs are cheap, this feels like a waste of resources in addition to filling logs with noise.


swankjesse (approved):

zcross:
Tests are pretty flaky (unrelated to change), working on getting build passing

PR #1880, opened 2021-04-30
Change TruncateTableService to enhance SchemaMigratorService

swankjesse:
Otherwise services that depend on TransacterService will run in parallel
with TruncateTableService, and that is not good at all.


adrw (approved):
:bless:

PR #1881, opened 2021-05-03
Add admin dashboard config to web config

benaskins:
This PR adds an admin dashboard config to misk's web config. It will be used to render links to protobuf definitions from the misk web action dashboard.

For the exemplar app with the following configuration:

```
web:
  port: 8080
  idle_timeout: 30000
  admin_dashboard:
    protobuf_definition_url: "https://example.com"
```

When rendering the web actions tab for the action: `HelloWebProtoAction` we would render a link to `https://example.com/com.squareup.exemplar.protos.HelloWebRequest` for the request parameter.

This PR does not make any changes to web action dashboard, as a follow up misk-web change is required to provide this context.

### DIFF
--- a/misk/src/main/kotlin/misk/web/WebConfig.kt
+++ b/misk/src/main/kotlin/misk/web/WebConfig.kt
@@ -85,7 +85,7 @@ data class WebConfig(
   /** If true, disables automatic load shedding when degraded. */
   val concurrency_limiter_disabled: Boolean = false,
 
-  val adminDashboard: AdminDashboardConfig
+  val admin_dashboard: AdminDashboardConfig = AdminDashboardConfig()
 ) : Config
 
 data class WebSslConfig(
@@ -156,5 +156,5 @@ data class AdminDashboardConfig(
   /**
    * URL that can be used by the admin dashboard to link to protobuf definitions
    */
-  val protobufDefinitionUrl: String? = null,
+  val protobuf_definition_url: String? = null,
 )
--- a/misk/src/main/kotlin/misk/web/WebConfig.kt
+++ b/misk/src/main/kotlin/misk/web/WebConfig.kt
@@ -84,6 +84,8 @@ data class WebConfig(
 
   /** If true, disables automatic load shedding when degraded. */
   val concurrency_limiter_disabled: Boolean = false,
+
+  val adminDashboard: AdminDashboardConfig
 ) : Config
 
 data class WebSslConfig(
@@ -148,4 +150,11 @@ data class CorsConfig(
   val chainPreflight: Boolean = true,
   /** A comma separated list of HTTP headers that are allowed to be exposed on the client. */
   val exposedHeaders: Array<String> = arrayOf()
+)
+
+data class AdminDashboardConfig(
+  /**
+   * URL that can be used by the admin dashboard to link to protobuf definitions
+   */
+  val protobufDefinitionUrl: String? = null,
 )
--- a/misk/src/main/kotlin/misk/web/metadata/DashboardMetadataAction.kt
+++ b/misk/src/main/kotlin/misk/web/metadata/DashboardMetadataAction.kt
@@ -3,10 +3,12 @@ package misk.web.metadata
 import misk.MiskCaller
 import misk.scope.ActionScoped
 import misk.security.authz.Unauthenticated
+import misk.web.AdminDashboardConfig
 import misk.web.Get
 import misk.web.PathParam
 import misk.web.RequestContentType
 import misk.web.ResponseContentType
+import misk.web.WebConfig
 import misk.web.actions.WebAction
 import misk.web.dashboard.DashboardHomeUrl
 import misk.web.dashboard.DashboardNavbarItem
@@ -35,6 +37,7 @@ class DashboardMetadataAction @Inject constructor() : WebAction {
   @Inject private lateinit var allNavbarStatus: List<DashboardNavbarStatus>
   @Inject private lateinit var allHomeUrls: List<DashboardHomeUrl>
   @Inject lateinit var callerProvider: @JvmSuppressWildcards ActionScoped<MiskCaller?>
+  @Inject lateinit var webConfig: WebConfig
 
   @Get("/api/dashboard/{dashboard_slug}/metadata")
   @RequestContentType(MediaTypes.APPLICATION_JSON)
@@ -64,7 +67,8 @@ class DashboardMetadataAction @Inject constructor() : WebAction {
       home_url = homeUrl,
       navbar_items = navbarItems,
       navbar_status = navbarStatus,
-      tabs = authorizedDashboardTabs
+      tabs = authorizedDashboardTabs,
+      admin_dashboard = webConfig.adminDashboard
     )
     return Response(
       dashboardMetadata = dashboardMetadata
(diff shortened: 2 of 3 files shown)


misk/src/main/kotlin/misk/web/metadata/DashboardMetadataAction.kt
```diff
@@ -75,7 +79,8 @@ class DashboardMetadataAction @Inject constructor() : WebAction {
     val home_url: String = "",
     val navbar_items: List<String> = listOf(),
     val navbar_status: String = "",
-    val tabs: List<DashboardTab> = listOf()
+    val tabs: List<DashboardTab> = listOf(),
+    val admin_dashboard: AdminDashboardConfig = AdminDashboardConfig()
```
r3mariano, line 83: Should this match https://github.com/cashapp/misk-web/pull/1442/files#diff-c27559eac873e2f6fac51c844fe9fbbce4ead534bc37584157fba842244c9f8dR29 (i.e. `admin_dashboard_config`)?

Also, it shouldn't really cause problems in this case, but I'm a little reluctant to couple the config object and this action directly because it makes it easy to inadvertently output all parts of this config object (though really, I'm mostly just thinking about separation of concerns).

samples/exemplar/src/main/resources/exemplar-common.yaml
```diff
@@ -1,6 +1,8 @@
 web:
   port: 8080
   idle_timeout: 30000
+  admin_dashboard:
+    protobuf_definition_url: "https://example.com"
```
r3mariano, line 5: Is this a URL prefix/would you derive a specific definition URL given this one, and how?

r3mariano (approved):

misk/src/main/kotlin/misk/web/metadata/DashboardMetadataAction.kt
benaskins: > Should this match https://github.com/cashapp/misk-web/pull/1442/files#diff-c27559eac873e2f6fac51c844fe9fbbce4ead534bc37584157fba842244c9f8dR29 (i.e. admin_dashboard_config)?

Yep, thanks @r3mariano - I'll change misk-web to match.

> Also, it shouldn't really cause problems in this case, but I'm a little reluctant to couple the config object and this action directly because it makes it easy to inadvertently output all parts of this config object (though really, I'm mostly just thinking about separation of concerns).

What would be the alternative? Introduce a new endpoint to return this data and make another API call?

samples/exemplar/src/main/resources/exemplar-common.yaml
benaskins: Yep, a prefix and we'll append the fully qualified proto name, e.g:

`https://example.com/com.squareup.exemplar.protos.HelloWebRequest`

misk/src/main/kotlin/misk/web/metadata/DashboardMetadataAction.kt
r3mariano: Could introduce an object that can be created from the config, e.g. 
```kotlin
data class AdminDashboardConfigResponse(...) { 
  constructor(config: Config) {
    // ... 
  }
}
```

To get to that object, a pattern I usually see is to have the contents of the config provided at the module (this also makes testing easier at the module level). Though, you can still inject WebConfig directly, I'm not too fussed about it either way.

That reminds me, you should probably add a test that this field is outputted.

samples/exemplar/src/main/resources/exemplar-common.yaml
r3mariano: Suggest adding code-level docs to that effect (and maybe adding a `_prefix` to this field).

misk/src/main/kotlin/misk/web/metadata/DashboardMetadataAction.kt
benaskins: > What would be the alternative? Introduce a new endpoint to return this data and make another API call?

Thinking this through, if we went down this path it could really simplify the solution. There'd be no need to modify misk-web at all if we introduced an additional API call from the web-actions `TabContainer.tsx`. WDYT?

r3mariano: I don't follow, but if it makes it simpler that sounds good!

misk/src/main/kotlin/misk/web/WebConfig.kt
```diff
@@ -149,3 +151,10 @@ data class CorsConfig(
   /** A comma separated list of HTTP headers that are allowed to be exposed on the client. */
   val exposedHeaders: Array<String> = arrayOf()
 )
+
+data class AdminDashboardConfig(
```
r3mariano, line 155: I think this should be a completely separate set of config in `misk.web.metadata`, since `WebConfig` deals with the more base-level web server config.

Alternatively, thinking about this a bit more, this could really just be configured at the module, skipping the yaml reading altogether.

benaskins: Do you have an example of what you mean by "configured at the module"?

samples/exemplar/src/main/resources/exemplar-common.yaml
benaskins: There are some comments here, I'll clarify and provide an example: https://github.com/cashapp/misk/pull/1881/files#diff-3c3e885f03acdd9f9b5dda55fe404d5d77314b4dcc68dcb2db089c62a9f8ffb9R156-R158

misk/src/main/kotlin/misk/web/metadata/DashboardMetadataAction.kt
benaskins: Well, the approach in this PR is to extend the DashboardMetadata to include configuration for the admin dashboard. Because the call to retrieve the DashboardMetadata is made from within `misk-web` it results in us needing to use a react context provider there to provide the context to tabs rendered by misk. To make this simpler we could make an additional API call [when we mount the TabContainer](https://github.com/cashapp/misk/blob/7feb5edc7a55e72a1314070cfbae44561b9a8df7/misk/web/tabs/web-actions/src/rewrite/TabContainer.tsx#L14-L36) to retrieve the admin config.

misk/src/main/kotlin/misk/web/WebConfig.kt
r3mariano: https://github.com/cashapp/misk/blob/master/misk/src/test/kotlin/misk/web/metadata/MetadataTestingModule.kt, used in the dashboard tests, binds a DashboardHomeUrl among other data classes at that point. Following that pattern, something like:
```kotlin
bind<ProtobufDefinitionUrlPrefix>().toInstance(ProtobufDefinitionUrlPrefix("http://example.com")
```

misk/src/main/kotlin/misk/web/metadata/DashboardMetadataAction.kt
r3mariano: Do you need this data for all tabs or just web actions? If the latter, you might get an even simpler solution adding it to `WebActionMetadataAction`?

benaskins: Just for web actions. It's not something that will change per action though, just a single config value that the client can then use to construct a URL.

r3mariano: Yes, then the response object can contain a new field that's not the list of actions

benaskins: Ok, I think I'll close this PR and start again. Thanks Ric 😄

benaskins:
Closing this in favour of a simpler approach. See discussion here: https://github.com/cashapp/misk/pull/1881#discussion_r624882620

misk/src/main/kotlin/misk/web/metadata/DashboardMetadataAction.kt
adrw: Agreed, `WebActionMetadataAction` is a more fitting spot to put this, good catch @r3mariano

PR #1888, opened 2021-05-05
Misk to Wisp for Containers Testing code

chris-ryan-square:


misk-testing/src/main/kotlin/misk/containers/Containers.kt
```diff
@@ -66,154 +61,19 @@ data class Container(
...
-  private class GracefulWaitContainerResultCallback : WaitContainerResultCallback() {
-    override fun onError(throwable: Throwable?) {
-      // this is ok, just meant that the container already terminated before we tried to wait
-      if (throwable is NotFoundException) {
-        return
-      }
-      super.onError(throwable)
-    }
-  }
+  var delegate: wisp.containers.Composer
```
r3mariano, line 64: nit: Something similar to this so `delegate` is a `val`? (Also, `private/internal val`?)
```kotlin
val delegate: Composer = run {
  // ... init code
}
```
?

wisp-containers-testing/src/main/kotlin/wisp/containers/Containers.kt
```diff
@@ -0,0 +1,220 @@
...
+ *         .withExposedPorts(ExposedPort.tcp(port))
+ *         .withPortBindings(Ports().apply {
+ *           bind(ExposedPort.tcp(9102), Ports.Binding.bindPort(9102))
+ *         })
+ *         .withEnv(
+ *           "KAFKA_ZOOKEEPER_CONNECT=zookeeper:2181",
+ *           "KAFKA_ADVERTISED_LISTENERS=PLAINTEXT://localhost:9102")
+ *         }
+ *     val composer = Composer("e-kafka", zkContainer, kafka)
+ *     composer.start()
```
r3mariano, line 64: Is this a new example? It's possible to do more syntax sugar with the existing code:
```kotlin
val zkContainer = Container {
    withImage("confluentinc/cp-zookeeper")
    withName("zookeeper")
    withEnv("ZOOKEEPER_CLIENT_PORT=2181")
 }
```
or an infix equivalent, but this seems fine.

r3mariano (approved):

misk-testing/src/main/kotlin/misk/containers/Containers.kt
chris-ryan-square: Yeah, should have changed that - will fix.

wisp-containers-testing/src/main/kotlin/wisp/containers/Containers.kt
chris-ryan-square: Existing comments, just transferred, but that's better, will fix

PR #1895, opened 2021-05-08
OPA module

AlexSzlavik:
Adding support scaffold for MISK + OPA. 
The current implementation only support querying documents. The interface is type safe and marshalled via retrofit / okhttp under the hood.


misk-policy/src/main/kotlin/misk/policy/opa/OpaPolicyEngine.kt
```diff
@@ -0,0 +1,93 @@
...
+  private fun queryOpa(
+    document: String,
+    inputString: String? = null
+  ): retrofit2.Response<ResponseBody> {
+    val response = when (inputString) {
+      null -> opaApi.queryDocument(document).execute()
+      else -> opaApi.queryDocument(document, inputString).execute()
+    }
+    if (!response.isSuccessful) {
+      throw RuntimeException("[${response.code()}]: ${response.errorBody()?.string()}")
```
yan, line 65: There's a few `RuntimeException`s here, we can create a `PolicyEvaluationException` to ease integrating it.

misk-policy/src/main/kotlin/misk/policy/opa/OpaPolicyEngine.kt
```diff
@@ -0,0 +1,93 @@
...
+  }
+
+  /**
+   * Evaluate / Query a document with no additional input.
+   * This will connect to OPA via a retrofit interface and perform a /v1/data/{document} POST.
+   *
+   * @throws RuntimeException if the request to OPA failed or the response shape didn't match R.
+   * @return Response shape R from OPA.
+   */
+  fun <R> evaluateInternal(document: String, returnType: Class<R>): R {
```
yan, line 51: can be `private`

AlexSzlavik: actually I tried to make that happen, but because the helper routines are inlined they would be exposing these private functions. I think I'm missing something here on how to hide the internals away properly using generics / inlines

misk-policy/build.gradle.kts
```diff
@@ -0,0 +1,29 @@
...
+  implementation(project(":misk"))
+  implementation(project(":misk-core"))
+  implementation(project(":misk-inject"))
+  implementation(project(":misk-actions"))
+  api(project(":wisp-config"))
+  api(project(":wisp-logging"))
+
+  testImplementation(Dependencies.assertj)
+  testImplementation(Dependencies.logbackClassic)
+  testImplementation(project(":wisp-logging"))
```
chris-ryan-square, line 16: wisp-logging is included already (line 12)

misk-policy/src/main/kotlin/misk/policy/opa/OpaPolicyEngine.kt
```diff
@@ -0,0 +1,106 @@
...
+ * Support for the Open Policy Engine (OPA).
+ * OPA provides a means to decouple policy from business logic. The resulting query response may
+ * have arbitrary shapes.
+ */
+class OpaPolicyEngine @Inject constructor(
+  private val opaApi: OpaApi,
+  @Named("opa-moshi") private val moshi: Moshi
+) {
+
+  /**
```
yoavamit, line 19: nit: move the doc to the public method

misk-policy/src/main/kotlin/misk/policy/opa/OpaModule.kt
```diff
@@ -0,0 +1,35 @@
...
+import com.squareup.moshi.Moshi
+import com.squareup.moshi.kotlin.reflect.KotlinJsonAdapterFactory
+import misk.client.HttpClientFactory
+import misk.inject.KAbstractModule
+import retrofit2.converter.scalars.ScalarsConverterFactory
+import javax.inject.Named
+import javax.inject.Singleton
+
+class OpaModule : KAbstractModule() {
+  override fun configure() {
```
yoavamit, line 13: optional: instead of injecting the objects below using `@Provider` methods, you could make the configuration object a constructor parameter and then bind these objects in the `configure` method.

That will allow you to also check if the config is empty and then exit early if no config was provided.
It might be useful later, because then you could include this module in the `MiskCommonModule` and not worry about whether configuration was provided on not

misk-policy/src/main/kotlin/misk/policy/opa/OpaApi.kt
```diff
@@ -0,0 +1,23 @@
...
+  @POST("/v1/data/{documentPath}")
+  fun queryDocument(
+    @Path("documentPath") documentPath: String,
+  ): Call<ResponseBody>
+
+  @Headers("Content-Type: application/json")
+  @POST("/v1/data/{documentPath}")
+  fun queryDocument(
+    @Path("documentPath") documentPath: String,
+    @Body input: String
```
yoavamit, line 21: out of curiosity, is it possible to change `input` to `@Body input: String?` and then combine these 2 methods?

yoavamit (approved):

misk-policy/src/main/kotlin/misk/policy/opa/OpaApi.kt
AlexSzlavik: Retrofit @Body parameters must be non-null. I think I tried changing this to `@Body input: String = ""`, but I don't remember if that worked of if OPA had issues with the resulting request. I'll try it again and see if I can collapse them.

misk-policy/src/main/kotlin/misk/policy/opa/OpaModule.kt
AlexSzlavik: yea good call, also matches other module patterns.

misk-policy/src/main/kotlin/misk/policy/opa/OpaApi.kt
AlexSzlavik: I went with a single retrofit API and pass in an empty body in the case where no input is required

PR #1911, opened 2021-05-14
Expose tracking API throught DynamicConfig/FeatureFlags

bezmax:
It's frequently useful to be able to listen for feature flag changes instead of polling them. Some libraries (LD) support that natively, while for others a polling adapter with a separate thread can be built. 

This PR adds a set of "tracker" methods both on FeatureFlags and DynamicConfig interfaces. It also adds an LD implementation forwarding those methods to `getTracker`, etc.


tirsen (approved):

escardin (approved):
LGTM! Bring on the Sevs! :D

PR #1928, opened 2021-05-26
Use an older version of java in tests

r3mariano:
Also reinstates the mysql service since the in-built one seems to not be documented anymore. (EDIT: it is documented, just in a different place, and with a different version of mysql. This change should still remain.)

This resolves CI issues with running Vitess tests.

### DIFF
--- a/.github/workflows/Test.yml
+++ b/.github/workflows/Test.yml
@@ -36,6 +36,14 @@ jobs:
   jvm:
     runs-on: ${{ matrix.os }}
 
+    services:
+      mysql:
+        image: mysql:5.7
+        ports:
+          - 3306:3306
+        env:
+          MYSQL_ALLOW_EMPTY_PASSWORD: 1
+
     strategy:
       matrix:
         os: [ubuntu-latest]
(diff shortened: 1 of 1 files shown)


.github/workflows/Test.yml
```diff
@@ -36,6 +36,14 @@ jobs:
   jvm:
     runs-on: ${{ matrix.os }}
 
+    services:
+      mysql:
```
chris-ryan-square, line 40: Do we really want to go back to pulling down an image instead of using the existing mysql in the image?

chris-ryan-square (approved):

.github/workflows/Test.yml
r3mariano: We could, but we should:
a) use 5.7 (or update that separately rather than implying a version based on the runner), otherwise we cause a mysql update breaking builds: https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-software
b) use an action to set it up: https://github.com/mirromutth/mysql-action#the-default-mysql

PR #1931, opened 2021-05-26
Add the stacktrace when logging the error to create datasource

frojasg:
It would help to debug transient issues


minostro (approved):

misk-jdbc/src/main/kotlin/misk/jdbc/DataSourceService.kt
```diff
@@ -44,7 +44,7 @@ class DataSourceService(
     try {
       createDataSource(baseConfig)
     } catch (e: Exception) {
-      logger.warn("Fail to start the data source to master tablet, trying to do it with replica")
+      logger.error(e) {"Fail to start the data source to master tablet, trying to do it with replica"}
```
prudhvi, line 47: Why does this use the word tablet? I thought tablet word is related to vitess? Also if this code path is used by regular MySQL, does having `baseConfig.asReplica()` still make sense?

prudhvi: or may check that it's vitess before trying to connect to replica?

PR #1932, opened 2021-05-27
Hook up metrics to the concurrency limiter

swankjesse:


misk/src/test/kotlin/misk/tracing/ClientServerTraceTest.kt
```diff
@@ -68,7 +67,7 @@ internal class ClientServerTraceTest {
     val client = clientInjector.getInstance<ReturnADinosaur>(Names.named("dinosaur"))
     client.getDinosaur(dinosaurRequest).execute()
 
-    val serverSpan = serverTracer.take()
+    val serverSpan = serverTracer.take("http.action")
```
swankjesse, line 70: I needed to add these parameters because I introduced a ScheduledExecutorService above, and that is automatically traced, and those traces fight with this test.

jdm-square (approved):
Nice! The netflix adapter seems like a reasonable way to bridge the two models.

PR #1937, opened 2021-05-27
Add a health check for DynamoDB

swankjesse:
This will make sure Misk services can successfully connect to DynamoDB before
self-reporting as healthy.

This may have a small availability benefit: connecting to DynamoDB before we
serve live traffic will mean the DynamoDB connection will be initialized when
traffic arrives.

Note that this PR also promotes HealthCheck from misk/ to misk-core/.

PR #1941, opened 2021-05-28
Add a health check for DynamoDB

swankjesse:
This restores reverted PR #1937, using describeTable() instead
of listTables(). That endpoint requires a lot more configuration,
including wiring the table names all the way through.

This reverts commit c23012833a5d29ab515cd32c842d9e21064596f9.

### DIFF
--- a/misk-aws-dynamodb/src/main/kotlin/misk/dynamodb/DynamoDbHealthCheck.kt
+++ b/misk-aws-dynamodb/src/main/kotlin/misk/dynamodb/DynamoDbHealthCheck.kt
@@ -1,0 +1,30 @@
+package misk.dynamodb
+
+import com.amazonaws.services.dynamodbv2.AmazonDynamoDB
+import javax.inject.Inject
+import javax.inject.Singleton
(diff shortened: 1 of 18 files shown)


misk-aws-dynamodb/src/main/kotlin/misk/dynamodb/DynamoDbHealthCheck.kt
```diff
@@ -0,0 +1,30 @@
...
+import misk.healthchecks.HealthStatus
+import wisp.logging.getLogger
+
+@Singleton
+class DynamoDbHealthCheck @Inject constructor(
+  private val dynamoDb: AmazonDynamoDB,
+  private val requiredTables: List<RequiredDynamoDbTable>,
+) : HealthCheck {
+  override fun status(): HealthStatus {
+    for (table in requiredTables) {
```
gothamgeek, line 16: nit:  should we require at least one table? I assume this list could, in theory, be empty.

gothamgeek (approved):

PR #1942, opened 2021-05-28
FakeMetrics

swankjesse:
The Prometheus metrics API we're using isn't particularly test-friendly,
but we can get at most of its insides. Most of the value in this class
is convenience functions to extract the state of the Metrics object for
making assertions.


frojasg (approved):
Thanks for adding this!

PR #1944, opened 2021-05-31
Add an API to accept required tables by type

swankjesse:
Previously these were passed by name which causes callers to have
to duplicate table names in code.


acostama (approved):
👍

PR #1948, opened 2021-05-31
Include a default value for the ClientConfiguration parameter

swankjesse:
Omitting this was an oversight.

### DIFF
--- a/misk-aws-dynamodb/src/main/kotlin/misk/dynamodb/RealDynamoDbModule.kt
+++ b/misk-aws-dynamodb/src/main/kotlin/misk/dynamodb/RealDynamoDbModule.kt
@@ -18,23 +18,18 @@ import misk.inject.KAbstractModule
 /**
  * Install this module to have access to an AmazonDynamoDB client. This can be
  * used to create a DynamoDbMapper for querying of a DynamoDb table.
+ *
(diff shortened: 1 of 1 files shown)


acostama (approved):

misk-aws-dynamodb/src/main/kotlin/misk/dynamodb/RealDynamoDbModule.kt
```diff
@@ -25,7 +25,7 @@ class RealDynamoDbModule constructor(
 ) : KAbstractModule() {
   /** @param requiredTableTypes a list of mapper classes annotated [DynamoDBTable]. */
   constructor(
-    clientConfig: ClientConfiguration,
+    clientConfig: ClientConfiguration = ClientConfiguration(),
```
acostama, line 28: do you want the default on both constructors? (should the other one be private?)

swankjesse: Yeah good point. I’ve been struggling with this. I supported both constructors to retain the option to add more features to RequiredDynamoDbTable. But that’s too speculative! Lemme make the other constructor private.

PR #1958, opened 2021-06-01
Defer creating Service instances until dependencies are started

swankjesse:
This makes it easier to directly inject upstream dependencies without Providers
or other indirections.

### DIFF
--- a/misk-service/src/main/kotlin/misk/CoordinatedService.kt
+++ b/misk-service/src/main/kotlin/misk/CoordinatedService.kt
@@ -6,6 +6,7 @@ import com.google.common.util.concurrent.Service
 import com.google.common.util.concurrent.Service.Listener
 import com.google.common.util.concurrent.Service.State
 import javax.inject.Provider
+import misk.CoordinatedService.Companion.checkNew
 
 @Suppress("UnstableApiUsage") // Guava's Service is @Beta.
 internal class CoordinatedService(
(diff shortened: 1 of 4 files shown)


keeferrourke (approved):
Nice 👍

misk-service/src/main/kotlin/misk/CoordinatedService.kt
```diff
@@ -14,6 +15,31 @@ internal class CoordinatedService(
 
   override val service: Service by lazy {
     serviceProvider.get()
+      .also { created ->
+        created.checkNew("$created must be NEW for it to be coordinated")
+        created.addListener(
+          object : Listener() {
+            override fun running() {
+              synchronized(this@CoordinatedService) {
```
swankjesse, line 23: I added the `@CoordinatedService` bit, which I think was always intended but not included.

keeferrourke: yup, that makes sense to me! :)

misk-service/src/test/kotlin/misk/ServiceManagerModuleTest.kt
```diff
@@ -140,4 +142,71 @@ internal class ServiceManagerModuleTest {
...
+  ) : AbstractService() {
+    init {
+      log.append("ConsumerService.init\n")
+    }
+    override fun doStart() {
+      log.append("ConsumerService.startUp\n")
+      notifyStarted()
+    }
+
+    override fun doStop() {
```
swankjesse, line 176: Just changed this to use AbstractService instead of AbstractIdleService because AbstractIdleService is concurrent and was flaking in CI.

PR #1959, opened 2021-06-01
Further defer creating services

swankjesse:
I had a logic bug where I was resolving the lazy before checking
that all of its upstreams were ready.


misk-service/src/main/kotlin/misk/CoordinatedService.kt
```diff
@@ -135,11 +135,13 @@ internal class CoordinatedService(
 
   private fun startIfReady() {
     synchronized(this) {
-      if (state() != State.STARTING || service.state() != State.NEW) return
```
swankjesse, line 138: This call to service.state() causes the service to be created, before doing the upstream services check below

keeferrourke (approved):

PR #1960, opened 2021-06-02
Add commit hooks to JDBC and Jooq modules

Sethuraman:
Commit hooks are at the moment provided as a feature of the misk-hibernate module alone and is missing in the misk-jdbc and the recently merged in misk-jooq module. 

Commit hooks are being used in a bunch of places specifically in places where we commit a transaction to the DB and immediately afterwards send a event to kafka for example.

This PR does these things:

1. Create a Session interface in misk-jdbc that the Session in misk-hibernate extends from.
2. Provides commit hooks to both the misk-jdbc Transacter and the misk-jooq Transacter

The hope is any libraries / common services that are built that used rely on misk.hibernate.Session should instead rely on misk.jdbc.Session


misk-jooq/src/main/kotlin/misk/jooq/JooqTransacter.kt
```diff
@@ -54,16 +54,25 @@ class JooqTransacter(
...
     }
   }
 
   private fun <RETURN_TYPE> createDSLContextAndCallback(
-    callback: (dslContext: DSLContext) -> RETURN_TYPE
+    callback: (jooqSession: JooqSession) -> RETURN_TYPE
   ): RETURN_TYPE {
-    return dslContext.transactionResult { configuration ->
-      callback(DSL.using(configuration))
+    var jooqSession: JooqSession? = null
```
Sethuraman, line 65: @davidgarth Does this function read well to you?

misk-jdbc/src/main/kotlin/misk/jdbc/Transacter.kt
```diff
@@ -14,19 +14,41 @@ interface Transacter {
...
+  /**
+   * Starts a transaction on the current thread, executes [work], and commits the transaction.
+   * If the work raises an exception the transaction will be rolled back instead of committed.
+   *
+   * This session object passed in wraps a connection and provides a way to add pre and post commit
+   * hooks that execute before and after a transaction is committed.
+   *
+   * It is an error to start a transaction if another transaction is already in progress.
+   */
+  fun <T> transactionWithSession(work: (session: JDBCSession) -> T): T
```
Sethuraman, line 35: I ideally wanted to call this transaction() as well, but kotlin java interop doesn't work well with the same function name.

r3mariano (approved):

PR #1969, opened 2021-06-04
Add BatchEnqueue to JobQueue

halakaraki:
- Add BatchEnqueue to JobQueue
- Implement BatchEnqueue in SqsQueue
- Implement BatchEnqueue in FakeJobQueue

### DIFF
--- a/misk-aws/src/main/kotlin/misk/jobqueue/sqs/SqsJobQueue.kt
+++ b/misk-aws/src/main/kotlin/misk/jobqueue/sqs/SqsJobQueue.kt
@@ -1,14 +1,13 @@
 package misk.jobqueue.sqs
 
-import com.amazonaws.services.sqs.model.BatchResultErrorEntry
 import com.amazonaws.services.sqs.model.MessageAttributeValue
 import com.amazonaws.services.sqs.model.SendMessageBatchRequest
 import com.amazonaws.services.sqs.model.SendMessageBatchRequestEntry
 import com.amazonaws.services.sqs.model.SendMessageRequest
 import com.squareup.moshi.Moshi
 import datadog.trace.core.DDSpan
+import io.opentracing.Span
 import io.opentracing.Tracer
-import misk.jobqueue.Job
 import misk.jobqueue.JobQueue
 import misk.jobqueue.QueueName
 import misk.moshi.adapter
@@ -26,6 +25,16 @@ internal class SqsJobQueue @Inject internal constructor(
   private val moshi: Moshi,
   private val tracer: Tracer
 ) : JobQueue {
+
+  private fun checkAttributeSize(attributes: Map<String, String>) {
+    // Ensure there are at most 9 attributes; AWS SQS enforces a limit of 10 custom attributes
+    // per message, 1 of which is reserved for this library (jobqueue metadata).
+    // https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-message-attributes.html
+    check(attributes.size <= 9) {
+      "a maximum of 9 attributes are supported (got ${attributes.size})"
+    }
+  }
+
   override fun enqueue(
     queueName: QueueName,
     body: String,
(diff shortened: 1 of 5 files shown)


misk-aws/src/main/kotlin/misk/jobqueue/sqs/SqsJobQueue.kt
```diff
@@ -97,6 +103,102 @@ internal class SqsJobQueue @Inject internal constructor(
...
+        }
+
+        val queue = queues.getForSending(queueName)
+
+        val (sendDuration, result) = queue.call { client ->
+          val messageEntries = jobs.map { message ->
+            // Ensure there are at most 8 attributes; AWS SQS enforces a limit of 10 custom attributes
+            // per message, 2 of which are reserved for this library (trace id and jobqueue metadata).
+            // https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-message-attributes.html
+            // TODO(bruno): change to 9 once we drop trace id
```
halakaraki, line 124: is trace id dropped?

misk-aws/src/main/kotlin/misk/jobqueue/sqs/SqsJobQueue.kt
```diff
@@ -97,6 +103,102 @@ internal class SqsJobQueue @Inject internal constructor(
...
+              val metadata = mutableMapOf(
+                SqsJob.JOBQUEUE_METADATA_ORIGIN_QUEUE to queueName.parentQueue.value,
+                SqsJob.JOBQUEUE_METADATA_IDEMPOTENCE_KEY to message.idempotenceKey
+              )
+
+              // Preserve original trace id, if available.
+              (span as? DDSpan)?.let {
+                val traceId = it.context().traceId.toString()
+                metadata[SqsJob.JOBQUEUE_METADATA_ORIGINAL_TRACE_ID] = traceId
+                // TODO(bruno): drop this attribute after rollout; moved to metadata
```
halakaraki, line 153: did rollout happen?

misk-aws/src/main/kotlin/misk/jobqueue/sqs/SqsJobQueue.kt
```diff
@@ -97,6 +103,102 @@ internal class SqsJobQueue @Inject internal constructor(
     }
   }
 
+  override fun batchEnqueue(
```
swankjesse, line 106: does it make sense to make single enqueue just delegate to this?

misk-aws/src/main/kotlin/misk/jobqueue/sqs/SqsJobQueue.kt
```diff
@@ -97,6 +103,102 @@ internal class SqsJobQueue @Inject internal constructor(
...
+          )
+        })
+
+        if(batchResult.failed.isNotEmpty()){
+          log.error { "failed to batch enqueue to ${queueName.value}" }
+          metrics.jobEnqueueFailures.labels(queueName.value, queueName.value).inc(batchResult.failed.size.toDouble())
+        }
+        batchResult
+      } catch (th: Throwable) {
+        log.error(th) { "failed to enqueue to ${queueName.value}" }
```
swankjesse, line 195: nit: this is the log and throw antipattern. better to rethrow with more context than to end up with two identical stacktraces in the logs.

misk-aws/src/main/kotlin/misk/jobqueue/sqs/SqsJobQueue.kt
```diff
@@ -97,6 +103,102 @@ internal class SqsJobQueue @Inject internal constructor(
...
+          JobQueue.EnqueueErrorResult(
+            it.id,
+            it.isSenderFault,
+            it.code,
+            it.message
+          )
+        })
+
+        if(batchResult.failed.isNotEmpty()){
+          log.error { "failed to batch enqueue to ${queueName.value}" }
```
swankjesse, line 190: is this the right place to log an error? If it’s truly an error should this function throw instead of returning normally?

misk-aws/src/main/kotlin/misk/jobqueue/sqs/SqsJobQueue.kt
```diff
@@ -97,6 +103,102 @@ internal class SqsJobQueue @Inject internal constructor(
...
   }
 
+  override fun batchEnqueue(
+    queueName: QueueName,
+    jobs: List<JobQueue.JobRequest>,
+  ): JobQueue.BatchEnqueueResult {
+    return tracer.traceWithSpan("batch-enqueue-job-${queueName.value}") { span ->
+      metrics.jobsEnqueued.labels(queueName.value, queueName.value).inc(jobs.size.toDouble())
+      try {
+        check(jobs.size <= 10) {
```
swankjesse, line 113: should this limit be in a constant somewhere so that callers can batch on it?

misk-jobqueue/src/main/kotlin/misk/jobqueue/JobQueue.kt
```diff
@@ -41,4 +41,69 @@ interface JobQueue {
     idempotenceKey: String = UUID.randomUUID().toString(),
     deliveryDelay: Duration? = null,
     attributes: Map<String, String> = mapOf())
+
+  /**
+   * Batch Enqueue a set of jobs onto the given queue, along with a set of job attributes.
+   *
+   * @param queueName The name of the queue on which to place the job.
```
swankjesse, line 48: https://publicobject.com/2007/08/05/stop_writing_redundant_javadoc/

swankjesse (approved):

misk-aws/src/main/kotlin/misk/jobqueue/sqs/SqsJobQueue.kt
biasedbit: yup

misk-aws/src/main/kotlin/misk/jobqueue/sqs/SqsJobQueue.kt
biasedbit: yup

misk-aws/src/main/kotlin/misk/jobqueue/sqs/SqsJobQueue.kt
halakaraki: fair point, I was following the pattern of the single enqueue. In this case since the client gets the successful & failed messages they can choose to log

misk-aws/src/main/kotlin/misk/jobqueue/sqs/SqsJobQueue.kt
halakaraki: thought of that.. but this is technically and SQS limitation. Should we enforce it at the Interface level? I think SQS is the only implementation, but if there were others this number would be implementation dependent.

halakaraki: thoughts?

misk-aws/src/main/kotlin/misk/jobqueue/sqs/SqsJobQueue.kt
halakaraki: Possibly, but the batch api has a different contract than the single one. It doesn't throw. Not sure how to make it look the same (or if we should), in case callers of this were catching specific dynamo errors..

halakaraki: discussed offline: Batch enqueue will throw an exception that contains the list of failed items

misk-aws/src/main/kotlin/misk/jobqueue/sqs/SqsJobQueue.kt
halakaraki: removed the logging.
Removed it for the single enqueue case as well. Let me know if there's a reason not to

misk-jobqueue/src/main/kotlin/misk/jobqueue/JobQueue.kt
halakaraki: removed what it think is redundant

misk-aws/src/main/kotlin/misk/jobqueue/sqs/SqsJobQueue.kt
halakaraki: added it to the interface prefixed with SQS. It's implementation specific & the sqs implementation class is internal. Not sure of a good way to do this

misk-aws/src/test/kotlin/misk/jobqueue/sqs/SqsJobQueueTest.kt
```diff
@@ -421,6 +423,129 @@ internal class SqsJobQueueTest {
...
+  fun batchEnqueueBatchLimit() {
+    val handledJobs = CopyOnWriteArrayList<Job>()
+    val allJobsComplete = CountDownLatch(10)
+    consumer.subscribe(queueName) {
+      handledJobs.add(it)
+      it.acknowledge()
+      allJobsComplete.countDown()
+    }
+
+    assertFailsWith<java.lang.IllegalStateException> {
```
swankjesse, line 534: nice

swankjesse (approved):

misk-aws/src/main/kotlin/misk/jobqueue/sqs/SqsJobQueue.kt
swankjesse: I’d prefer we write code assuming exactly two implementations, SQS and a fake. If ever we add an alternate to SQS, I don’t think it’ll be a simple implementation change.

misk-aws/src/main/kotlin/misk/jobqueue/sqs/SqsJobQueue.kt
```diff
@@ -22,82 +26,151 @@ internal class SqsJobQueue @Inject internal constructor(
...
+      return@executeWithTracingAndErrorHandling sendDuration
+    }
+  }
+
+  private fun String.toMessageAttributeValue(): MessageAttributeValue = MessageAttributeValue()
+    .withDataType("String")
+    .withStringValue(this)
+
+  private fun getDelaySeconds(deliveryDelay: Duration?): Int {
+    return (deliveryDelay?.toMillis() ?: 0 / 1000).toInt()
```
biasedbit, line 132: @halakaraki what's the precedence order here? Do we want `(deliveryDelay?.toMillis() ?: 0) / 1000` or `(deliveryDelay?.toMillis() ?: (0 / 1000))`?

biasedbit: Tangent: why `toMillis()` and then `/1000`? Could we just use `getSeconds()`?

halakaraki:
I think this is slightly different. In our particular case we actually want to make sure the event gets published before succeeding the call.

halakaraki:
Discussed more offline. The experimental feature is synchronous, however it blocks for 200ms to wait for the batch size to get to 10. If the batch size is fixed at 10, then it's great. For variable batch sizes, this would mean a min sla of 200ms. For our use case, they tend to average around 2-3

misk-aws/src/main/kotlin/misk/jobqueue/sqs/SqsJobQueue.kt
halakaraki: working on it

PR #1972, opened 2021-06-10
jobqueue bugfix: change getDelaySeconds to return seconds instead of …

mhickman:
…milliseconds

### DIFF
--- a/misk-aws/src/main/kotlin/misk/jobqueue/sqs/SqsJobQueue.kt
+++ b/misk-aws/src/main/kotlin/misk/jobqueue/sqs/SqsJobQueue.kt
@@ -129,7 +129,8 @@ internal class SqsJobQueue @Inject internal constructor(
     .withStringValue(this)
 
   private fun getDelaySeconds(deliveryDelay: Duration?): Int {
-    return (deliveryDelay?.toMillis() ?: 0 / 1000).toInt()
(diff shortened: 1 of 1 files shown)


misk-aws/src/main/kotlin/misk/jobqueue/sqs/SqsJobQueue.kt
```diff
@@ -129,7 +129,8 @@ internal class SqsJobQueue @Inject internal constructor(
     .withStringValue(this)
 
   private fun getDelaySeconds(deliveryDelay: Duration?): Int {
-    return (deliveryDelay?.toMillis() ?: 0 / 1000).toInt()
+    val delayMillis = deliveryDelay?.toMillis() ?: 0
```
surrahman, line 132: Should we have a test for this?

surrahman (approved):

misk-aws/src/main/kotlin/misk/jobqueue/sqs/SqsJobQueue.kt
mhickman: I'm sure one is useful - I was just biasing to getting something fixed asap.

biasedbit: Agree, let's get this out there and then reach out to the author of the breaking change to add a test.

PR #1976, opened 2021-06-14
Revert part #1958 to try and fix shutdown deadlock

### LINKED ISSUE #1975: Service shutdown deadlock in CoordinatedService
This is a regression from #1958 / #1959 and is causing tests on various misk repos to time out intermittently. The deadlock can be seen in the attached [stacks.txt](https://github.com/cashapp/misk/files/6649420/stacks.txt) file (generated by `jstack` in an instance where this was reproduced locally).

It looks like the change in lock object used in the synchronized statement ended up causing a deadlock when shutting down multiple services.

staktrace:
The change to the synchronization lock object in #1958 appears
to have resulted in an intermittent deadlock. Tests that exercise
this code heavily frequently hit the deadlock and time out.
Details of the thread stacks during the deadlock can be found
in issue #1975.

This patch reverts the change to the synchronization lock object,
which appears incidental to the main change that was being made
in #1958. Hopefully this will fix the deadlock while retaining
the intended functional change. Follow-up PRs can be used to
implement a better solution.

Fixes #1975.

### DIFF
--- a/misk-service/src/main/kotlin/misk/CoordinatedService.kt
+++ b/misk-service/src/main/kotlin/misk/CoordinatedService.kt
@@ -20,14 +20,14 @@ internal class CoordinatedService(
         created.addListener(
           object : Listener() {
             override fun running() {
-              synchronized(this@CoordinatedService) {
+              synchronized(this) {
                 notifyStarted()
               }
               downstreamServices.forEach { it.startIfReady() }
             }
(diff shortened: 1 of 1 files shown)


breeseAtSquare (approved):

keeferrourke (approved):
LGTM

Issue #1853, opened 2021-04-14
GrpcClientProviderTest is flaky


swankjesse:
```
GrpcClientProviderTest > happyPath() FAILED
    java.io.IOException: gRPC transport failure (HTTP status=200, grpc-status=null, grpc-message=null)
        at com.squareup.wire.internal.GrpcKt.grpcResponseToException(grpc.kt:166)
        at com.squareup.wire.internal.RealGrpcCall.readExactlyOneAndClose(RealGrpcCall.kt:104)
        at com.squareup.wire.internal.RealGrpcCall.executeBlocking(RealGrpcCall.kt:77)
        at misk.client.GrpcClientProviderTest.happyPath(GrpcClientProviderTest.kt:80)

        Caused by:
        okhttp3.internal.http2.StreamResetException: stream was reset: CANCEL
            at okhttp3.internal.http2.Http2Stream$FramingSource.read(Http2Stream.kt:358)
            at okhttp3.internal.connection.Exchange$ResponseBodySource.read(Exchange.kt:276)
            at okio.RealBufferedSource.exhausted(RealBufferedSource.kt:197)
            at com.squareup.wire.internal.GrpcMessageSource.read(GrpcMessageSource.kt:37)
            at com.squareup.wire.internal.GrpcMessageSource.readExactlyOneAndClose(GrpcMessageSource.kt:66)
            at com.squareup.wire.internal.RealGrpcCall.readExactlyOneAndClose(RealGrpcCall.kt:102)
            ... 2 more
```

carterhudson:
I'm encountering this in a real environment; any ideas what could be causing it? My current theory is that it's a keep-alive configuration issue.

`java.io.IOException: gRPC transport failure (HTTP status=200, grpc-status=null, grpc-message=null)`

r3mariano:
Some more logs from the CI build with extra okhttp logging added:
```
INFO: [0 ms] callStart: Request{method=POST, url=https://127.0.0.1:39961/test/GetDestinationWarehouse, headers=[te:trailers, grpc-trace-bin:, grpc-accept-encoding:gzip, grpc-encoding:gzip], tags={class com.squareup.wire.GrpcMethod=com.squareup.wire.GrpcMethod@20376e8d}}
INFO: [0 ms] proxySelectStart: https://127.0.0.1:39961/
INFO: [0 ms] proxySelectEnd: [DIRECT]
INFO: [0 ms] dnsStart: 127.0.0.1
[...]

PR #1983, opened 2021-06-18
Send gRPC errors properly

### LINKED ISSUE #1933: Handle errors with gRPC
It is throwing HTTP status codes like 429 or 503 when a gRPC action throws. It instead needs to return HTTP 200 always and instead set the grpc-status header. This should let me have more detailed error messaging, instead of a generic IOException

ewolak-sq:
According to the [gRPC spec](https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md), all
gRPC responses, including errors, should have an HTTP status of 200. gRPC errors are signaled using
the `grpc-status` header. This brings Misk in compliance with the spec, translating HTTP errors into
properly formatted gRPC errors.

Fixes #1933


misk/src/main/kotlin/misk/web/HttpCall.kt
```diff
@@ -31,7 +31,12 @@ interface HttpCall {
   val requestHeaders: Headers
 
   /** The HTTP response under construction. */
+  /** Meaningful HTTP status about what actually happened */
```
r3mariano, line 34: nit: remove old javadoc comment, and add comment about how this may not be the code actually sent over the network and why this is the case. I see it's commented elsewhere, but adding the comment here too will help reduce confusion

misk-grpc-tests/src/test/kotlin/misk/grpc/MiskClientMiskServerTest.kt
```diff
@@ -105,4 +110,55 @@ class MiskClientMiskServerTest {
...
+
+    runBlocking {
+      val e = assertFailsWith<GrpcException> {
+        routeGuide.GetFeature().execute(point)
+      }
+      assertThat(e.grpcMessage).isEqualTo("unexpected latitude error!")
+      assertThat(e.grpcStatus).isEqualTo(GrpcStatus.UNKNOWN)
+
+      assertResponseCount(200, 0)
+      assertResponseCount(500, 1)
```
r3mariano, line 129: This test is broken; will approve when it passes.

Also, nit: add code comment about why these are the expected metrics

misk/src/main/kotlin/misk/web/exceptions/ExceptionHandlingInterceptor.kt
```diff
@@ -37,13 +45,67 @@ class ExceptionHandlingInterceptor(
...
+    }
+  }
+
+  /**
+   * Borrow behavior from [GrpcFeatureBinding] to send a gRPC error with an HTTP 200 status code.
+   * This is weird but it's how gRPC clients work.
+   *
+   * One thing to note is for our metrics we want to pretend that the HTTP code is what we sent.
+   * Otherwise gRPC requests that crashed and yielded an HTTP 200 code will confuse operators.
+   */
```
r3mariano, line 69: I see that this will reduce disruption with metrics, but I feel like having two sets of HTTP status codes is confusing in the long run, and it adds a layer of indirection with gRPC statuses. Why not have a metric for different gRPC statuses instead?

To provide a good migration path, we can probably emit both HTTP and gRPC status code metrics, then turn off HTTP status code metrics for gRPC endpoints later.

misk-grpc-tests/src/main/kotlin/misk/grpc/miskserver/GetFeatureGrpcAction.kt
```diff
@@ -10,6 +11,9 @@ import javax.inject.Inject
 class GetFeatureGrpcAction @Inject constructor() : WebAction, RouteGuideGetFeatureBlockingServer {
   @LogRequestResponse(bodySampling = 1.0, errorBodySampling = 1.0)
   override fun GetFeature(request: Point): Feature {
+    if (request.latitude == -1) {
+      throw WebActionException(request.longitude ?: 500, "unexpected latitude error!")
+    }
```
r3mariano, line 16: This feels so hacky lol. How about a separate RPC that always fails?

misk/src/main/kotlin/misk/web/exceptions/ExceptionHandlingInterceptor.kt
r3mariano: Suggest new metrics that look like these, but no real need to make them identical I think? https://github.com/grpc-ecosystem/java-grpc-prometheus

r3mariano: (fwiw, this is fine to keep as is in this PR)

misk/src/main/kotlin/misk/web/HttpCall.kt
```diff
@@ -31,7 +31,12 @@ interface HttpCall {
...
   var statusCode: Int
+
+  /** The HTTP status code actually sent over the network. For gRPC, this is *always* 200, even
+   * for errors. */
+  var networkStatusCode: Int
```
swankjesse, line 39: # 👍🏻

misk/src/main/kotlin/misk/web/ServletHttpCall.kt
```diff
@@ -25,8 +25,16 @@ internal data class ServletHttpCall(
...
+    get() = _actualStatusCode ?: upstreamResponse.statusCode
+    set(value) {
+      _actualStatusCode = value
+      upstreamResponse.statusCode = value
+    }
+
+  override var networkStatusCode: Int
     get() = upstreamResponse.statusCode
     set(value) {
       upstreamResponse.statusCode = value
```
swankjesse, line 40: there’s a bit of a hazard here where you gotta set this one last. Ugh!

Maybe add a function and make `networkStatusCode` a val?

```
fun setStatusCodes(statusCode: Int, networkStatusCode: Int) { ... }
```

misk/src/main/kotlin/misk/web/exceptions/ExceptionHandlingInterceptor.kt
swankjesse: gRPC is always 200 OK, even when it’s not okay, and that is awful and a guarantee that there will be confusion. 

My preference is that we try to hide the insincere 200 OKs in implementation details. Metrics and alerting and even just developer expectations are pretty invested in 200/400/500 meaning different things. The 200 is a lie, and we want to quietly send it over the Wire without spreading that lie to other systems.

I think trying to get all those downstream tools to have independent HTTP and gRPC modes only spreads the confusion.

swankjesse (approved):

misk/src/main/kotlin/misk/web/exceptions/ExceptionHandlingInterceptor.kt
r3mariano: I get that gRPC errors being HTTP 200 is confusing. Thinking about it more, my concerns are a) [HTTP 4xx/5xx and gRPC have a non-1:1 mapping with different semantics](https://grpc.github.io/grpc/core/md_doc_http-grpc-status-mapping.html), and b) hiding the HTTP and gRPC status makes it a bit more difficult to trace requests when debugging (e.g. the server said 502 in its logs, but the client got a 200).

I see what you mean with metrics though, and am ok with having a mapping for metrics purposes.

misk-grpc-tests/src/test/kotlin/misk/grpc/MiskClientMiskServerTest.kt
ewolak-sq: Can you be more specific about what you mean by "broken" here? The assertion is about what's counted in metrics, which we very much do want to be what's asserted here. See the discussion in https://github.com/cashapp/misk/pull/1983#discussion_r655746280

misk-grpc-tests/src/main/kotlin/misk/grpc/miskserver/GetFeatureGrpcAction.kt
ewolak-sq: I'm pretty sure that the `RouteGuide` service is borrowed from the gRPC reference implementation, so I'm loathe to extend that with another method just for this sort of test. Given that it's strictly test code, I'm not sure I agree that it's worth doing in a perfect way.

@swankjesse what do you think?

misk/src/main/kotlin/misk/web/ServletHttpCall.kt
ewolak-sq: Yeah, good call

ewolak-sq:
I'm fairly certain that the test failure on `GrpcConnectivityTest` is a flake. I have it passing and failing locally with the same sets of tests before it, so it's definitely flaking. See #1853.

misk-grpc-tests/src/test/kotlin/misk/grpc/MiskClientMiskServerTest.kt
r3mariano: Ah I saw a test failure here, but it looks like it's working now.

r3mariano (approved):

misk/src/main/kotlin/misk/web/exceptions/ExceptionHandlingInterceptor.kt
ewolak-sq: I agree we should emit some gRPC-specific metrics to improve observability, but I think we can do that as a follow-up PR (#1992). This PR strictly improves the situation from "Misk counts 4xx/5xx metrics but breaks gRPC clients" to "Misk counts 4xx/5xx metrics and doesn't break gRPC clients". The third option, of "Misk fails to count 4xx/5xx metrics for gRPC but clients get the right error responses" seems strictly worse than either of those to me.

ewolak-sq:
@r3mariano @swankjesse 😠  I'm getting a _consistent_ failure with `okhttp3.internal.http2.StreamResetException: stream was reset: CANCEL` in CI, but it won't fail locally. I think it's just #1853 rearing its head. What should I do to move forward?

r3mariano:
> @r3mariano @swankjesse 😠 I'm getting a _consistent_ failure with `okhttp3.internal.http2.StreamResetException: stream was reset: CANCEL` in CI, but it won't fail locally. I think it's just #1853 rearing its head. What should I do to move forward?

I've disabled the offending test in #2004.

ewolak-sq:
Ugh. I think all these tests are flaky in different and interesting ways. `MiskClientMiskServerTest.serverFailureNotFound()` sometimes gets `java.io.IOException: gRPC transport failure (HTTP status=200, grpc-status=null, grpc-message=null)`, which I'm guessing is the same flakiness as in `GrpcConnectivityTest`. Then the metrics counts seem to have a race condition, so I probably need to await for them.

misk-grpc-tests/src/main/kotlin/misk/grpc/miskserver/GetFeatureGrpcAction.kt
swankjesse: Happy as is.

PR #1990, opened 2021-06-22
Check http2 is enabled if gRPC services are bound

r3mariano:
Also enables http2 by default in tests.


misk/src/test/kotlin/misk/web/InvalidActionsTest.kt
```diff
@@ -50,4 +51,52 @@ class InvalidActionsTest {
...
+    }
+  }
+
+  @Singleton
+  class HelloRpcAction @Inject constructor() : WebAction, GreeterSayHello {
+    var failNextRequest = false
+    var sleep: Duration = Duration.ofMillis(0)
+
+    override fun sayHello(request: HelloRequest): HelloReply {
+      Thread.sleep(sleep.toMillis())
```
chris-ryan-square, line 85: This is really needed?

chris-ryan-square (approved):

misk/src/test/kotlin/misk/web/InvalidActionsTest.kt
r3mariano: No this was copy-paste. Can simplify this a bit more.

PR #1993, opened 2021-06-23
Add OPA Policy gradle testing plugin

AlexSzlavik:
Gradle plugin which is distributed with misk-policy-testing. The plugin will launch an opa docker container, make policy from the src/policy folder available and run `opa test` on the contents.

The point of this is to make CI testing of OPA policy as easy as possible for service owners. This allows us to simply add a dependency to on misk-policy-testing as part of the services buildscripts configuration and apply the plugin for the service.

### DIFF
--- a/misk-policy-testing/src/main/kotlin/misk/policy/opa/OpaTestPlugin.kt
+++ b/misk-policy-testing/src/main/kotlin/misk/policy/opa/OpaTestPlugin.kt
@@ -1,0 +1,81 @@
+package misk.policy.opa
+
+import com.github.dockerjava.api.model.Bind
+import com.github.dockerjava.api.model.Binds
+import com.github.dockerjava.api.model.Frame
+import com.github.dockerjava.api.model.HostConfig
+import com.github.dockerjava.api.model.Volume
+import com.github.dockerjava.core.DockerClientBuilder
+import com.github.dockerjava.core.async.ResultCallbackTemplate
+import com.github.dockerjava.core.command.WaitContainerResultCallback
+import com.github.dockerjava.netty.NettyDockerCmdExecFactory
+import okio.Buffer
+import org.gradle.api.Plugin
+import org.gradle.api.Project
+import org.gradle.api.provider.Property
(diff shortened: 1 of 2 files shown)


misk-policy-testing/src/main/kotlin/misk/policy/opa/OpaTestPlugin.kt
```diff
@@ -0,0 +1,81 @@
...
+    val opaTestWithDocker = project.tasks.create("opaTestWithDocker") {
+      group = "OPA"
+      description = "Evaluate OPA tests via docker OPA instance"
+
+      doLast {
+        val dockerClient = DockerClientBuilder.getInstance()
+          .withDockerCmdExecFactory(NettyDockerCmdExecFactory())
+          .build()
+        val policyDir = project.projectDir.absolutePath + "/" + config.policyDir.get()
+        val containerId = dockerClient.createContainerCmd("openpolicyagent/opa")
```
AlexSzlavik, line 35: Should probably extract this into a variable.

misk-policy-testing/src/main/kotlin/misk/policy/opa/OpaTestPlugin.kt
```diff
@@ -0,0 +1,81 @@
...
+          .withHostConfig(
+            HostConfig.newHostConfig()
+              .withAutoRemove(true)
+              .withBinds(Binds(Bind(policyDir, Volume("/repo"))))
+          )
+          .withName("opa_tester")
+          .withTty(true)
+          .exec().id
+
+        dockerClient.startContainerCmd(containerId)
```
AlexSzlavik, line 46: Should also wait here until the docker has downloaded the docker image. Possible also consider pinning the image version

staktrace (approved):
Reviewing per request on Slack. The patch itself looks reasonable to me, although I don't understand the larger context surrounding the OPA code and how this plugin is going to be used. Looks like it's a new module added to misk recently and still being built out so I don't have concerns about this breaking anything existing.

PR #2003, opened 2021-06-24
Add hget, hset, and hgetAll to misk-redis

ddiekroeger:
hget, hgetall, and hset are useful redis operators that let you add a `field` in addition to a `key`. We would like to use this in our service, so I am proposing that we add them here to misk redis.


misk-redis/src/main/kotlin/misk/redis/RealRedis.kt
```diff
@@ -47,6 +47,24 @@ class RealRedis(private val jedisPool: JedisPool) : Redis {
     }
   }
 
+  // Get a map of field -> value pairs for the given key.
```
adrw, line 50: Could these be proper kotlindoc comments?

```
/** Get a map ... */
```

misk-redis/src/main/kotlin/misk/redis/RealRedis.kt
```diff
@@ -76,6 +94,13 @@ class RealRedis(private val jedisPool: JedisPool) : Redis {
     }
   }
 
+  // Set a ByteArray value for the given key and field.
```
adrw, line 97: kotlin doc style?

misk-redis/src/main/kotlin/misk/redis/RealRedis.kt
```diff
@@ -47,6 +47,24 @@ class RealRedis(private val jedisPool: JedisPool) : Redis {
...
+    jedisPool.resource.use { jedis ->
+      return jedis.hgetAll(key.toByteArray(charset))?.mapKeys {
+        it.key.toString(charset)
+      }?.mapValues {
+        it.value.toByteString()
+      }
+    }
+  }
+
+  // Get a ByteString value for the given key and field
```
adrw, line 61: kotlin doc style?

adrw (approved):

misk-redis/src/main/kotlin/misk/redis/RealRedis.kt
ddiekroeger: good call, updated :)

PR #2005, opened 2021-06-24
Move testing code (Fake) to testing jar; clean up deps; remove need f…

chris-ryan-square:
…or guava


wisp-resource-loader/src/main/kotlin/wisp/resources/ResourceLoader.kt
```diff
@@ -91,15 +90,15 @@ class ResourceLoader(
...
+    val resourcesResult = mutableListOf<String>()
     for (resource in list(address)) {
       if (list(resource).isEmpty()) {
         resourcesResult.add(resource)
       } else {
         resourcesResult.addAll(walk(resource))
       }
     }
-    return resourcesResult.build()
+    return Collections.unmodifiableList(resourcesResult)
```
r3mariano, line 101: nit: resourcesResult.toList() should be enough, kotlin lists are read-only

r3mariano (approved):

PR #2012, opened 2021-06-28
WarmupTask, a mechanism to improve request latency

swankjesse:
We've been having with newly-started services having unacceptable
latency. This is a new mechanism to plug-in warmup tasks.


swankjesse:
I was reading this Tweet repeatedly while writing this!

https://twitter.com/APStylebook/status/588063260726972416
> Warmup is one word when it's a noun, warm up is two words as a verb. #APStyleChat 
@TheSocialBlitz

misk-warmup/src/test/kotlin/misk/warmup/WarmupTest.kt
```diff
@@ -0,0 +1,144 @@
...
+      "LoggingService startUp",
+      "LoggingWarmupTask created on warmup-0",
+      "LoggingWarmupTask warming on warmup-0",
+      "HealthChecks all passed",
+      "LoggingService shutDown",
+    )
+  }
+
+  @Test
+  fun `service is healthy even after warmup task throws`() {
```
zcross, line 45: nice behavior, and nice test

misk-warmup/src/test/kotlin/misk/warmup/WarmupTest.kt
```diff
@@ -0,0 +1,144 @@
...
+    startUpAndShutDown(
+      ServiceModule<LoggingService>(),
+      WarmupModule<ThrowingWarmupTask>(),
+    )
+
+    assertThat(events).containsExactly(
```
zcross, line 51: nit: assert `ERROR` level?

zcross (approved):
Love it, thanks for making it reusable up front; as soon as we share example impacts of using this in some of our services, I suspect others may follow suit

adrw (approved):
LG

misk-warmup/src/test/kotlin/misk/warmup/WarmupTest.kt
swankjesse: Done!

PR #1967, opened 2021-06-04
Environment destruction

chris-ryan-square:
This removes Environment, EnvironmentModule and misk's Deployment class (use wisp's instead).

DO NOT MERGE!  Let me handle the merge process as it will require updating breaking changes to downstream users.

The Env class has also been removed.  There were major areas of impact - config loading (MiskConfig) and bucket names (BucketNameSource).  Note the use of deployment.mapToEnvironmentName() to map the deployment to a name, e.g. if isProduction is set, the name returned is "production".  These values are used in the config loading and bucket names.

This change is related to issue: https://github.com/cashapp/misk/issues/1416

PR #2013, opened 2021-06-28
Ensure service metadata environment is uppercased

keeferrourke:
Misk Web expects an upper-cased environment from its service metadata ([ref](https://github.com/cashapp/misk-web/blob/f122ea6a19ee6ee16ca1a0f19929ccda53103335/packages/@misk/core/src/utilities/interfaces.ts#L39-L44)). 

For example, a response from `/api/service/metadata` should look like the following:

```
{"serviceMetadata":{"app_name":"example","environment":"PRODUCTION"}}
```

The recent removal of Environment (see https://github.com/cashapp/misk/pull/1967) caused this endpoint to start returning lowercase environment names. As a result, services with Misk Web UIs that choose to render their environment names could no longer do so.

This change fixes that regression and ensures compatibility with present versions of misk-web.

### DIFF
--- a/misk/src/main/kotlin/misk/web/metadata/ServiceMetadataAction.kt
+++ b/misk/src/main/kotlin/misk/web/metadata/ServiceMetadataAction.kt
@@ -7,7 +7,7 @@ import misk.web.RequestContentType
 import misk.web.ResponseContentType
 import misk.web.actions.WebAction
 import misk.web.mediatype.MediaTypes
-import wisp.deployment.getDeploymentFromEnvironmentVariable
+import wisp.deployment.Deployment
 import javax.inject.Inject
 import javax.inject.Singleton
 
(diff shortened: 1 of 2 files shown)


misk/src/main/kotlin/misk/web/metadata/ServiceMetadataAction.kt
```diff
@@ -39,9 +39,8 @@ class ServiceMetadataAction @Inject constructor(
    * https://github.com/google/guice/wiki/FrequentlyAskedQuestions#how-can-i-inject-optional-parameters-into-a-constructor
    */
   @Singleton
-  class OptionalBinder @Inject constructor(@AppName val appName: String) {
+  class OptionalBinder @Inject constructor(@AppName val appName: String, deployment: Deployment) {
```
keeferrourke, line 42: While I was at it -- we can inject a deployment here, since web actions are wired up by guice. No need to get a new Deployment from an env check.

misk/src/test/kotlin/misk/web/metadata/ServiceMetadataActionTest.kt
```diff
@@ -0,0 +1,62 @@
...
+import misk.testing.MiskTestModule
+import misk.web.jetty.JettyService
+import okhttp3.OkHttpClient
+import okhttp3.Request
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.jupiter.api.Test
+import javax.inject.Inject
+
+@MiskTest(startService = true)
+class ServiceMetadataActionTest {
```
keeferrourke, line 18: The body of this test is shamelessly copied from here: https://github.com/cashapp/misk/blob/fe47b15b4ac8734e810d74dc2ff1cf6fa90580bf/misk/src/test/kotlin/misk/web/metadata/DashboardMetadataActionTest.kt#L24

adrw: Surprised there wasn't already one, would have caught this regression!

adrw (approved):

PR #2026, opened 2021-07-01
Support Customization of gRPC status by mappers

ewolak-sq:
Instead of always mapping an HTTP status code to a gRPC status, give ExceptionMappers a chance to
explicitly produce a gRPC status and message. This gives us more fine-grained control over exactly
what gets returned in :grpc-message headers.

In particular, we want to return the raw exception message in gRPC responses while still masking it
for HTTP responses. This is because gRPC clients are generally internal, while HTTP clients may be
external organizations that we don't want to leak internal information to.

### DIFF
--- a/misk/src/main/kotlin/misk/web/exceptions/ExceptionHandlingInterceptor.kt
+++ b/misk/src/main/kotlin/misk/web/exceptions/ExceptionHandlingInterceptor.kt
@@ -44,10 +44,10 @@ class ExceptionHandlingInterceptor(
       chain.proceed(chain.httpCall)
     } catch (th: Throwable) {
       val response = toResponse(th)
-      chain.httpCall.statusCode = response.statusCode
       if (chain.httpCall.dispatchMechanism == DispatchMechanism.GRPC) {
-        sendGrpcFailure(chain.httpCall, response)
+        sendGrpcFailure(chain.httpCall, response.statusCode, toGrpcResponse(th))
       } else {
+        chain.httpCall.statusCode = response.statusCode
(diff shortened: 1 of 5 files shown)


misk/src/main/kotlin/misk/web/exceptions/ExceptionHandlingInterceptor.kt
```diff
@@ -117,6 +105,23 @@ class ExceptionHandlingInterceptor(
     } ?: toInternalServerError(th)
   }
 
+  private fun toGrpcResponse(th: Throwable): GrpcErrorResponse = when (th) {
+    is UnauthenticatedException -> GrpcErrorResponse(GrpcStatus.UNAUTHENTICATED, th.message)
+    is UnauthorizedException -> GrpcErrorResponse(GrpcStatus.PERMISSION_DENIED, th.message)
+    is InvocationTargetException -> toGrpcResponse(th.targetException)
+    is UncheckedExecutionException -> toGrpcResponse(th.cause!!)
```
r3mariano, line 112: What are these two about?

r3mariano (approved):
lgtm!

misk/src/main/kotlin/misk/web/exceptions/ExceptionHandlingInterceptor.kt
ewolak-sq: These are copied from `toResponse()` above. AIUI, the idea is to unwrap exceptions that aren't helpful to get the meaningful exception within. For example, I think `InvocationTargetException` gets thrown when wrapping an exception thrown by a Java reflection call.

PR #2029, opened 2021-07-06
Add support for double feature flags

ThePumpingLemma:
While here, make some minor cleanups + reorganization for boolean
feature flags


r3mariano (commented):
I'm wary of potential floating-point precision losses with `Double` and the potential for confusion there, e.g. `featureFlags.getDouble(...) != 0.3` despite it being set on the feature flag UI. (I see that the LD SDK uses a double too, and that uses Gson under the hood. See for instance https://github.com/google/gson/issues/1127)

escardin (approved):

r3mariano:
> Sure, but why do we care? Every flag I've looked at in older systems that does this is using it for something like a percentage rollout. The difference between 0.3 and 0.299999999999999988897769753748434595763683319091796875 for all practical purposes are the same. We also offer floats, and that's fine too.

This isn't a percentage rollout, it's a decimal number flag value. I get that they are practically the same, but I'm not sure if a computer will always think so too, hence my previous example. In any case, happy to try it out.

> IMO if you need that level of precision in your decimal number... don't use a double, and you should already know that, because you learned not to use floating point for decimals/money already.

Misk's a pretty "guard-rails"-heavy framework IMO. Even if parts of it do right now, we shouldn't rely on "you should know that".

ThePumpingLemma:
@r3mariano are you okay with me merging this?

PR #2031, opened 2021-07-06
misk-policy-testing: Add OpaDevelopmentModule

AlexSzlavik:
The OpaDevelopmentModule will stand up a docker container, running OPA as a side service. With this, local development may run against the real service policy.


amanymous-net (approved):

misk-policy-testing/src/main/kotlin/misk/policy/opa/LocalOpaService.kt
```diff
@@ -103,4 +109,30 @@ class LocalOpaService(
...
+      retry(
+        5, ExponentialBackoff(
+        Duration.ofSeconds(1),
+        Duration.ofSeconds(5)
+      )
+      )
+      {
+        val client = OkHttpClient()
+        val request = Request.Builder()
+          .url("http://localhost:8181/health")
```
AlexSzlavik, line 127: This should use the configured EXPOSED_OPA_PORT

Issue #1997, opened 2021-06-23
Remove use of junit 4


r3mariano:
We already use junit 5.

r3mariano:
Unless we remove MockWebServer, this isn't possible because Misk uses it, and this issue isn't resolved: https://github.com/square/okhttp/issues/4667

PR #2049, opened 2021-07-14
Remove junit4 dep - not used

chris-ryan-square:


chris-ryan-square:
See https://github.com/cashapp/misk/issues/1997

PR #2090, opened 2021-07-29
Make WebActionException a RuntimeException

Sethuraman:
Is there a reason why we have WebActionException an Exception instead of a RuntimeException. This works okay in kotlin as all exceptions are treated by default as runtime exceptions. But the issue comes with the inter-op with java libraries. 

For my specific case, when a checked exception is thrown inside a transaction in jooq, it gets wrapped into a DataAccessException whereas a RuntimeException is rethrown as is.


Sethuraman:
Will this break something?

misk-actions/src/main/kotlin/misk/exceptions/Exceptions.kt
```diff
@@ -23,7 +23,7 @@ open class WebActionException(
    */
   message: String,
   cause: Throwable? = null
-) : Exception(message, cause) {
+) : RuntimeException(message, cause) {
```
chris-ryan-square, line 26: Comment the reason for this - i.e. java inter-op - otherwise it's likely someone in the future will change it not realising why.

chris-ryan-square (approved):

PR #2122, opened 2021-08-27
Add expire, expireAt, pExpire and pExpireAt

szabado:


misk-redis/src/test/kotlin/misk/redis/FakeRedisTest.kt
```diff
@@ -288,4 +288,86 @@ class FakeRedisTest {
...
+    clock.add(Duration.ofSeconds(1))
+    assertNull(redis["foo"])
+  }
+
+  @Test fun expireInOneSecondTimestamp() {
+    // Setup
+    redis["foo"] = "baz".encodeUtf8()
+
+    // Exercise
+    // Expire immediately
```
r3mariano, line 337: nit: 
```suggestion
    // Expire in one second
```

misk-redis/src/test/kotlin/misk/redis/FakeRedisTest.kt
```diff
@@ -288,4 +288,86 @@ class FakeRedisTest {
...
+    assertNull(redis.hget("foo", "bar"))
+    assertNull(redis["foo"])
+  }
+
+  @Test fun expireInOneSecond() {
+    // Setup
+    redis["foo"] = "baz".encodeUtf8()
+
+    // Exercise
+    // Expire immediately
```
r3mariano, line 323: nit: 
```suggestion
    // Expire in one second
```

misk-redis/src/main/kotlin/misk/redis/Redis.kt
```diff
@@ -130,4 +131,64 @@ interface Redis {
...
+   * If a key is overwritten by RENAME, like in the case of an existing key Key_A that is
+   * overwritten by a call like RENAME Key_B Key_A, it does not matter if the original Key_A had a
+   * timeout associated or not, the new key Key_A will inherit all the characteristics of Key_B.
+   *
+   * Note that calling [expire]/[pExpire] with a non-positive timeout or [expireAt]/[pExpireAt] with
+   * a time in the past will result in the key being deleted rather than expired (accordingly, the
+   * emitted key event will be del, not expired).
+   *
+   * @return 1 if the timeout was set. 0 if the timeout was not set. e.g. key doesn't exist, or
+   * operation skipped due to the provided arguments.
```
r3mariano, line 161: I missed this in the incr/incrBy PR, but I think a boolean return type makes more sense here based on the rest of the interface.

Ideally there's more of an indication to the caller of why the operation failed, but my guess is that the underlying library doesn't provide that?

r3mariano (approved):

misk-redis/src/main/kotlin/misk/redis/Redis.kt
szabado: It doesn't seem like the underlying library provides it.

Definitely agree that a boolean return type makes more sense here, didn't see that on `del`. It doesn't make sense for the `incr`/`incrBy` code, since that returns the new value rather than a success/failure state

Issue #2128, opened 2021-09-01
Wisp Logger takeMessages() unexpectedly consumes all messages


keeferrourke:
Suppose I have two loggers that I was to check the contents of in a single test method:

```kt
// Borrowed and modified from wisp.logging.LogCollectorTest.

@Test fun `check logs happened`() {
  val logger = getLogger<LogCollectorTest>()
  val logger2 = getLogger<LoggingDummyClass>()

  logger.info("A thing happened")
  logger2.info("Another thing happened")

  assertThat(logCollector.takeMessages(LogCollectorTest::class)).containsExactly("A thing happened")
  assertThat(logCollector.takeMessages(LoggingDummyClass::class)).containsExactly("Another thing happened")
}
 ```
 
 This test fails because the first call to `takeMessages` *consumes* all the messages in the collector while filtering for the requested ones, leaving no messages left in the collector for the second call.
 
 I would expect that filtering by a KClass should mean that _only_ the log messages matching that class should be taken from the collector. Is this a bad assumption?
 
 Can we make the log collector less destructive so that we can have more powerful assertions against logs?

keeferrourke:
Hm I guess this is expected behaviour? It's very inconvenient though.

https://github.com/cashapp/misk/blob/7cf9aa310cd65134feb376fadd70897f6016ac49/wisp-logging-testing/src/test/kotlin/wisp/logging/LogCollectorTest.kt#L70

chris-ryan-square:
@keeferrourke with the above PR merged, can this issue be closed?

keeferrourke:
It could be closed, but there was some discussion about migrating to the new non-consuming behaviour as the default.

Not sure if that should be a new issue though?

chris-ryan-square:
If relevant/desired, raise a new issue in the Wisp repo.

PR #2129, opened 2021-09-01
Do not consume unmatched logs in LogCollector

keeferrourke:
Addresses #2128.

Open to other solutions! Maybe this is actually a bad idea -- my use-case may the outlier here, but I find the current behaviour very odd that `takeMessages(filter)` consumes the whole queue of collected logs and only returns the ones that matched.

### DIFF
--- a/misk-testing/src/main/kotlin/misk/logging/RealLogCollector.kt
+++ b/misk-testing/src/main/kotlin/misk/logging/RealLogCollector.kt
@@ -10,36 +10,11 @@ import javax.inject.Singleton
 import kotlin.reflect.KClass
 
 @Singleton
-internal class RealLogCollector @Inject constructor() :
-  AbstractIdleService(),
-  LogCollector,
+internal class RealLogCollector @Inject constructor(
+  private val wispQueuedLogCollector: WispQueuedLogCollector
+) : AbstractIdleService(),
+  LogCollector by wispQueuedLogCollector,
   LogCollectorService {
-
-  private val wispQueuedLogCollector = WispQueuedLogCollector()
-
-  override fun takeMessages(
-    loggerClass: KClass<*>?,
-    minLevel: Level,
-    pattern: Regex?
-  ): List<String> = takeEvents(loggerClass, minLevel, pattern).map { it.message }
-
-  override fun takeMessage(
-    loggerClass: KClass<*>?,
-    minLevel: Level,
-    pattern: Regex?
-  ): String = takeEvent(loggerClass, minLevel, pattern).message
-
-  override fun takeEvents(
-    loggerClass: KClass<*>?,
-    minLevel: Level,
-    pattern: Regex?
-  ): List<ILoggingEvent> {
-    return wispQueuedLogCollector.takeEvents(loggerClass, minLevel, pattern)
-  }
-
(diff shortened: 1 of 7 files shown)


wisp-logging-testing/src/main/kotlin/wisp/logging/WispQueuedLogCollector.kt
```diff
@@ -37,38 +37,34 @@ class WispQueuedLogCollector : LogCollector {
...
+    queue.removeAll(resultList)
+    return resultList
   }
 
   override fun takeEvent(loggerClass: KClass<*>?, minLevel: Level, pattern: Regex?): ILoggingEvent {
-    while (true) {
-      val event = takeOrNull(loggerClass, minLevel, pattern)
-      if (event != null) return event
+    require(wasStarted) { "not collecting logs: did you forget to start the service?" }
+    for (i in 1..5) {
```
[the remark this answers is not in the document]
keeferrourke: More closely matches (but not perfectly) the `poll(500)` in the previous impl.

misk-testing/src/test/kotlin/misk/logging/LogCollectorTest.kt
```diff
@@ -14,8 +15,9 @@ import kotlin.test.assertFailsWith
 
 @MiskTest(startService = true)
 class LogCollectorTest {
```
keeferrourke, line 17: I don't particularly enjoy that this test appears to be a copy of the wisp LogCollectorTest, but 🤷  I added the new test here too.

misk-testing/src/main/kotlin/misk/logging/RealLogCollector.kt
```diff
@@ -10,37 +10,12 @@ import javax.inject.Singleton
...
 
 @Singleton
-internal class RealLogCollector @Inject constructor() :
-  AbstractIdleService(),
-  LogCollector,
+internal class RealLogCollector @Inject constructor(
+  private val wispQueuedLogCollector: WispQueuedLogCollector
+) : AbstractIdleService(),
+  LogCollector by wispQueuedLogCollector,
   LogCollectorService {
```
keeferrourke, line 17: I think ideally `LogCollectorService` could be delegated to the `WispQueuedLogCollector`, but I gues wisp-logging-testing doesn't want to depend on Guava?

wisp-logging-testing/src/main/kotlin/wisp/logging/LogCollector.kt
```diff
@@ -43,7 +43,17 @@ interface LogCollector {
...
+
+  /**
+   * Takes all matching messages, optionally leaving unmatched logs in this collector.
+   */
+  fun takeMessages(
+    loggerClass: KClass<*>? = null,
+    minLevel: Level = Level.INFO,
+    pattern: Regex? = null,
+    consumeUnmatchedLogs: Boolean = false,
   ): List<String>
```
keeferrourke, line 57: I think maybe this is the interface we want to land on eventually if some folks depend on consuming all unmatched logs?

r3mariano: Yep, looks good.

r3mariano (approved):

PR #2137, opened 2021-09-09
map 409s to already_exist in grpc

halakaraki:


r3mariano (approved):

ewolak-sq:
Thank you! Good call. Even though the [official mapping|https://grpc.github.io/grpc/core/md_doc_http-grpc-status-mapping.html] doesn't do this, I think you're totally correct to map HTTP 409 to gRPC ALREADY_EXISTS.

PR #2140, opened 2021-09-10
Allow read access of endpoints

drashkov:


r3mariano (changes requested):
I think i would rather have an `@AdminReadOnlyAccess` than open this up completely. That way it's still possible to describe this access in any way we see fit (including unauthenticated).

PR #2141, opened 2021-09-14
Create a new `v2.Metric` interface and implementation which allows proper `Summary` & `Histogram` metrics

MatthewDolan:
Adds the ability to expose a prometheus histogram using misk metrics.

Creates a new `v2.Metrics` interface and implementation that replaces the `Metrics.histogram(...)` method (which as confusingly named because it actually created a prometheus `Summary` with two methods:
- `Metrics.summary(...)` which creates a prometheus `Summary`
- `Metrics.histogram(...)` which creates a prometheus `Histogram`

The reason why this pull requests creates a new interface (as opposed to updating the existing interface is that) it doesn't appear that a summary and a histogram can have the same name. At least, that's how I am reading this quote:

> Currently with https://github.com/prometheus/client_golang (I don't know about others client libs) implementation it is not possible to define an Histogram and a Summary for the same data using the same name cause _sum and _count for both Histogram and Summary would collide.

https://github.com/OpenObservability/OpenMetrics/issues/120

As of right now, it's not recommended that all current uses of `Metrics` be migrated to `v2.Metrics` because it is not backward compatible. Specifically `Metrics.histogram(...)` is not backward compatible with `v2.Metrics.histogram(...)` updating will result in broken metrics because it will cause the metric to change type.

There has been some additional offline discussion about how our existing `Metrics` should be updated so that it emits both a `Histogram` and a `Summary` but that change has been excluded from this pull request and might come in a follow up pull request.


misk-metrics-testing/src/main/kotlin/misk/metrics/FakeMetrics.kt
```diff
@@ -72,17 +91,17 @@ class FakeMetrics internal constructor() : Metrics {
...
    * Returns the 0.99th percentile for a [histogram]. In small samples this is the second largest
    * element.
    */
-  fun histogramP99(name: String, vararg labels: Pair<String, String>): Double? =
-    histogramQuantile(name, "0.99", *labels)
+  fun summaryP99(name: String, vararg labels: Pair<String, String>): Double? =
+    summaryQuantile(name, "0.99", *labels)
 
-  fun histogramQuantile(
+  fun summaryQuantile(
```
r3mariano, line 104: You'll have to keep both for now so downstream services don't break

cpakman (changes requested):
I think we don't need to force anyone to change - histogram and summary both append slightly different names to metrics:

histogram:
```
<name>_bucket
<name>_sum
<name>_count
```

summary:
```
<name>  (has quantiles)
<name>_sum
<name>_count
```

assuming prometheus library doesn't complain, we actually have no issues here if we just double write - _sum and _count should be identical between the two, and for signalfx it should be easy enough to write a drop rule to ignore <name>_bucket metrics. This way no code changes are necessary for consumers to get the advantages of either metric (though mostly likely if people have access to histogram, that'll be the better choice)

MatthewDolan:
Thanks for the feedback, @cpakman.

> histogram and summary both append slightly different names to metrics

I think what you are suggesting is similar to the issue I linked to and quoted in the description above (https://github.com/OpenObservability/OpenMetrics/issues/120).

(I know it's a little confusing that this issue isn't on the prometheus client directly, but instead on open metrics, but it references the prometheus client library which is how I found it. I could potentially submit a follow up issue to the prometheus client library.)

> assuming prometheus library doesn't complain

This is the sticking point. If I am reading the library code and the issue correctly, prometheus library will complain because of the way that it keeps metrics registered in a single map so it's not really feasible to register two metrics into a single name.

Additionally, there is a protocol problem even if we could patch or replace the client to accept both types under the same name.

The prometheus protocol explicitly defines the type for a given metric name in the scrape response. For example, section of the scrape response from one of our test services.

```
# HELP client_http_request_latency_ms count and duration in ms of outgoing client requests
# TYPE client_http_request_latency_ms summary
client_http_request_latency_ms{...,quantile="0.5",} 76.0
client_http_request_latency_ms{...,quantile="0.75",} 76.0
client_http_request_latency_ms{...,quantile="0.95",} 77.0
client_http_request_latency_ms{...,quantile="0.99",} 79.0
client_http_request_latency_ms{...,quantile="0.999",} 88.0
client_http_request_latency_ms_count{...,code="200",} 12872.0
client_http_request_latency_ms_sum{...,code="200",} 964256.0
```

Note that even though the metrics wouldn't collide like you suggest, it's not clear what `#TYPE` should be associated with this new hybrid metric because it will have two types. Maybe with the coordination of a downstream scraper, we could get that to work, but it's definitely not trivial.

r3mariano (approved):
lgtm! i think moving this to v2 is a good idea

misk-metrics-testing/src/main/kotlin/misk/metrics/FakeMetricsModule.kt
```diff
@@ -1,10 +1,15 @@
 package misk.metrics
 
+import io.prometheus.client.CollectorRegistry
 import misk.inject.KAbstractModule
+import javax.inject.Inject
+import javax.inject.Provider
+import misk.metrics.v2.FakeMetricsModule as FakeMetricsModuleV2
```
maddiehayes, line 7: some unused imports @MatthewDolan

maddiehayes (approved):
looks good to me, backwards compatible and neat. exposes defaults to users

PR #2154, opened 2021-09-27
Add wisp-task

chris-ryan-square:
TODO: add metrics - follow up PR.


r3mariano (approved):

wisp-task/README.md
```diff
@@ -0,0 +1,82 @@
...
+```kotlin
+val taskConfig = TaskConfig("taskName")
+val repeatedTaskConfig = RepeatedTaskConfig(
+  timeBetweenRunsMs = 10000L  // 10 sec delay between task runs or retries
+)
+
+val manager = RepeatedTaskManager()
+
+val newTask = manager.createTask(
+  name = taskConfig.name,
```
r3mariano, line 31: why take in a name if taskConfig also has a name, and taskConfig is required?

r3mariano: put a different way, is taskConfig really required if all it's using is the name? is there a test or sample code to show?

PR #2156, opened 2021-09-29
Add a first attempt at metrics for repeated tasks

chris-ryan-square:


wisp-task/src/main/kotlin/wisp/task/RepeatedTask.kt
```diff
@@ -94,19 +93,28 @@ class RepeatedTask(
...
-        val startTime = clock.instant()
-        val timedResult = measureTimeMillis {
-          status = task(name, taskConfig)
+        var timedResult = 0L
+        try {
+          timedResult = measureTimeMillis {
+            status = task(name, taskConfig)
+          }
+        } catch (e: Exception) {
+          repeatedTaskMetrics.retryCount.increment()
```
r3mariano, line 102: Is this the same semantics as before? I'd think that retryCount increments at the start of a retry.

wisp-task/src/main/kotlin/wisp/task/RepeatedTaskMetrics.kt
```diff
@@ -1,8 +1,12 @@
 package wisp.task
 
+import io.micrometer.core.instrument.Counter
 import io.micrometer.core.instrument.DistributionSummary
 import io.micrometer.core.instrument.MeterRegistry
 
+/**
+ *
+ */
```
r3mariano, line 9:
```suggestion
```

wisp-task/src/main/kotlin/wisp/task/RepeatedTaskMetrics.kt
```diff
@@ -14,4 +18,15 @@ class RepeatedTaskMetrics(meterRegistry: MeterRegistry) {
...
+    .description("count of successful repeated tasks")
+    .register(meterRegistry)
+
+  internal var retryCount: Counter = Counter.builder("repeated_task_retry_count")
+    .description("count of repeated tasks retries")
+    .register(meterRegistry)
+
+  internal var noWorkCount: Counter = Counter.builder("repeated_task_no_work_count")
+    .description("count of repeated tasks invocations with no work to do")
+    .register(meterRegistry)
```
r3mariano, line 31: I completely missed that we're starting to use Micrometer for Misk. If so, the naming is off and should be e.g. `repeated.task.success`, with the naming convention for Prometheus kicking in: https://micrometer.io/docs/concepts#_naming_meters

Unless of course these are existing metric names, in which case `repeated_task_duration` doesn't follow Prometheus convention, and we should probably either drop Micrometer or add a naming convention class.

chris-ryan-square: I'll switch to '.' convention for micrometer to not assume what's behind it.

It's a new implementation for the same idea behind the misk version, so there's some extensions to what was already there (just the duration).

wisp-task/src/main/kotlin/wisp/task/RepeatedTask.kt
chris-ryan-square: There was no retry metric count before.

r3mariano: It should be a failed count then.

r3mariano (approved):

PR #2162, opened 2021-10-08
Ignore @Transient fields when constructing queries

ewolak-sq:
The HibernateDatabaseQueryWebActionModule introspects entity classes to select all of their fields. It shouldn't be selecting fields annotated with `@Transient`.


misk-hibernate/src/main/kotlin/misk/hibernate/actions/HibernateDatabaseQueryWebActionModule.kt
```diff
@@ -63,9 +65,12 @@ internal class HibernateDatabaseQueryWebActionModule : KAbstractModule() {
...
       val invalidSelectPaths = setOf("rootId")
       val dbEntityDeclaredMemberProperties =
-        dbEntity.declaredMemberProperties.map { it.name }.filter {
-          !invalidSelectPaths.contains(it)
-        }
+        dbEntity.declaredMemberProperties.filter { it ->
+          !invalidSelectPaths.contains(it.name) &&
+            // Because Transient is a Java annotation, we have to check it on the underlying Java
+            // field. Searching for it on the KProperty will fail.
+            it.javaField?.getAnnotation(Transient::class.java) == null
```
adrw, line 72: good catch, thanks

adrw (approved):

PR #2167, opened 2021-10-14
feat: AccessInterceptor: allow >1 access annotations on actions.

Ubehebe:
Consider an action that should be accessible by multiple groups. A natural way to express this is:

```kt
@Group1Access
@Group2Access
@Get("/some-action")
fun someAction() {
  // ...
}
```

Currently, Misk does not support this. It is hard-coded to throw an IllegalStateException whenever an action is annotated with more than one access annotation. As far as I can tell, this has been the case forever. This limitation forces Misk users to resort to unfortunate workarounds, such as defining a `@Group1OrGroup2Access` annotation that manually aggregates the underlying services/capabilities.

This PR changes AccessInterceptor to allow multiple access annotations. This is fairly easy because the class already represents the allowed services/capabilities as sets. All that is needed is to aggregate the services/capabilities from all the access annotations instead of just one.

(As a special case, this PR makes it an error to specify `@Unauthenticated` alongside other access annotations, since this is a logical contradiction.)

### DIFF
--- a/misk-service/src/main/kotlin/misk/CoordinatedService.kt
+++ b/misk-service/src/main/kotlin/misk/CoordinatedService.kt
@@ -6,7 +6,7 @@ import com.google.common.util.concurrent.Service
 import com.google.common.util.concurrent.Service.Listener
 import com.google.common.util.concurrent.Service.State
 import javax.inject.Provider
-import misk.CoordinatedService.Companion.checkNew
+import java.util.concurrent.atomic.AtomicBoolean
 
 @Suppress("UnstableApiUsage") // Guava's Service is @Beta.
 internal class CoordinatedService(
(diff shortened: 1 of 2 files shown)


misk/src/main/kotlin/misk/security/authz/AccessInterceptor.kt
```diff
@@ -86,9 +67,25 @@ class AccessInterceptor private constructor(
           |
           |
           """.trimMargin()
-          )
+      }
+
+      // This action is explicitly marked as unauthenticated.
+      if (action.hasAnnotation<Unauthenticated>()) {
+        check(actionEntries.size == 1) {
+          val otherAnnotations = actionEntries.filterNot { it.annotation == Unauthenticated::class }
```
Ubehebe, line 75: There is no coverage for this case. What I'd like is a way to assert in a test that the injector failed to create with the given error message. Reviewers, is there a way to do this?

Ubehebe: Done, thanks! I had to create the injector with Stage.PRODUCTION to get the error to surface. I _think_ this is expected behavior.

PR #2175, opened 2021-10-18
Fix publishing

swankjesse:
We weren't applying some standard boilerplate in one of our modules
and that broke publishing for everything.

Fix by centralizing the standard boilerplate.


r3mariano (approved):
nice!

PR #2180, opened 2021-10-19
Put unack'd jobs on the deadletter queue and remove FakeTransactional…

tso:
…JobQueue

### DIFF
--- a/misk-prometheus/api/misk-prometheus.api
+++ b/misk-prometheus/api/misk-prometheus.api
@@ -16,6 +16,9 @@ public final class misk/metrics/backends/prometheus/PrometheusConfig : wisp/conf
 	public fun toString ()Ljava/lang/String;
 }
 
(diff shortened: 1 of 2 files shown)


misk-jobqueue-testing/src/main/kotlin/misk/jobqueue/FakeJobQueue.kt
```diff
@@ -183,14 +185,15 @@ class FakeJobQueue @Inject constructor(
       }
 
       result += job
-      if (!job.deadLettered && assertAcknowledged) {
-        check(job.acknowledged) { "Expected $job to be acknowledged after handling" }
+      if (!job.deadLettered && assertAcknowledged && !job.acknowledged) {
```
scotte-square, line 188: Should job not be added to result in this case?

r3mariano (approved):

PR #2194, opened 2021-11-03
Make StringConverter and converterFor visbility public.

behrozsaadat:
In order to allow misk application to validate query parameters
these functions are necessary.


deanpapastrat (approved):
LGTM.

PR #2198, opened 2021-11-04
Multi-method WebActions

quad:
This seemingly arbitrary restriction forces single method service classes.

Unholy.

I suspect this restriction dates back to [the original implementation of WebActionModule](https://github.com/cashapp/misk/blame/c7a79459c11d41dbfdbbe2fcab74fd07eef84b9f/misk/src/main/kotlin/misk/web/WebActionModule.kt#L49-L66) wherein routing checks were done inline. Those checks are now [done at the WebActionsServlet](https://github.com/cashapp/misk/blob/65a49a457e2dd0220c9f9b8fbd083953f3f94fbe/misk/src/main/kotlin/misk/web/jetty/WebActionsServlet.kt#L58-L68) … so I think this change is safe.

Perfunctory tests attached, of course.

### DIFF
--- a/misk/src/main/kotlin/misk/web/actions/WebActionFactory.kt
+++ b/misk/src/main/kotlin/misk/web/actions/WebActionFactory.kt
@@ -67,21 +67,6 @@ internal class WebActionFactory @Inject constructor(
         "on ${webActionClass.simpleName}"
     }
 
-    require(actionFunctions.size == 1) {
-      val actionFunctionNames = actionFunctions.joinToString(", ") { it.name }
-      "multiple annotated methods on ${webActionClass.simpleName}: $actionFunctionNames"
-    }
-
-    // Bind providers for each supported HTTP method.
-    val actionFunction = actionFunctions.first()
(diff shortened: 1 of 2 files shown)


swankjesse (approved):
Love it

misk/src/main/kotlin/misk/web/actions/WebActionFactory.kt
```diff
@@ -67,21 +67,6 @@ internal class WebActionFactory @Inject constructor(
         "on ${webActionClass.simpleName}"
     }
 
-    require(actionFunctions.size == 1) {
```
swankjesse, line 70: Maybe require actionFunctions.isNotEmpty? That'd indicate a user error.

swankjesse: Oh, we already have that above. Don't mind me.

quad:
Good idea. I'll make a follow-up PR.

quad:
Ahh, I see why you deleted the comment. 😅

PR #2200, opened 2021-11-07
Include testing instructions for web actions

dsanchez-Sq:
This prevents an exception when testing the web action responses with a test module


r3mariano (commented):
Thanks. I think this would be more appropriate in a separate testing section at the bottom of the file. Happy to flesh that out if you'd like.

dsanchez-Sq:
@r3mariano yeah if you can, as I am not familiar with miskweb and just thought to add this piece as it got me unstuck.

r3mariano:
Made more docs following this.

PR #2201, opened 2021-11-08
Allow binding for application interceptors in grpc and typed clients

vektah:
We can configure network interceptors using guice multibindings, but theres currently no way to configure application interceptors. This PR creates a similar multibind to add application interceptors.

### Why?

I'm using okreplay to capture in dev and replay responses without hitting the backend in CI. This cant use a network interceptor because they will error if proceed isn't called. Or as the docs say:
image


r3mariano (approved):
lg, just needs an api dump to pass the build: https://github.com/cashapp/misk/blob/master/CONTRIBUTING.md#breaking-changes

PR #2202, opened 2021-11-08
Publish with jdk 11

shellderp:


keeferrourke (approved):

shellderp:
Kicked off to verify this works https://github.com/cashapp/misk/runs/4144826850?check_suite_focus=true

keeferrourke:
Unsure of this change. @r3mariano is there a reason we're using a non LTS java for this?

shellderp:
> Unsure of this change. @r3mariano is there a reason we're using a non LTS java for this?

I believe this was the latest java when I built this task. Misk doesn't seem to specify a java compatibility version in gradle..

PR #2209, opened 2021-11-11
Record client metrics for incomplete HTTP connections.

jdm-square:
The $Pay team has strict SLAs for payments, and aborts any payment that exceeds them. Sometimes these SLAs are exceeded when a downstream service is slow to return a response, but since we cancel the connection (which triggers an exception) our metrics do not show these slow requests. This change allows RPC Outgoing dashboards to reflect the full set of connections, not just the ones that either time out or return a complete response.

### DIFF
--- a/misk/src/main/kotlin/misk/client/ClientMetricsInterceptor.kt
+++ b/misk/src/main/kotlin/misk/client/ClientMetricsInterceptor.kt
@@ -36,6 +36,13 @@ class ClientMetricsInterceptor private constructor(
       requestDurationSummary.labels(actionName, "timeout").observe(elapsedMillis)
       requestDurationHistogram.labels(actionName, "timeout").observe(elapsedMillis)
       throw e
+    } catch (e: Exception) {
+      // Something else happened while the connection was in progress and we didn't receive
+      // a complete response. We still want to record any long-running calls, however.
+      val elapsedMillis = stopwatch.stop().elapsed(TimeUnit.MILLISECONDS).toDouble()
+      requestDurationSummary.labels(actionName, "incomplete-response").observe(elapsedMillis)
+      requestDurationHistogram.labels(actionName, "incomplete-response").observe(elapsedMillis)
+      throw e
     }
   }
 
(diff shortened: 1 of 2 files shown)


misk/src/main/kotlin/misk/client/ClientMetricsInterceptor.kt
```diff
@@ -36,6 +36,13 @@ class ClientMetricsInterceptor private constructor(
       requestDurationSummary.labels(actionName, "timeout").observe(elapsedMillis)
       requestDurationHistogram.labels(actionName, "timeout").observe(elapsedMillis)
       throw e
+    } catch (e: Exception) {
+      // Something else happened while the connection was in progress and we didn't receive
+      // a complete response. We still want to record any long-running calls, however.
+      val elapsedMillis = stopwatch.stop().elapsed(TimeUnit.MILLISECONDS).toDouble()
+      requestDurationSummary.labels(actionName, "0").observe(elapsedMillis)
```
elmerland, line 43: What is the `"0"` label meant to represent?

jdm-square: The second label is the response code that can be used to filter time series in signalfx. Since there's no complete response, I chose a code of 0 which is not a valid status code.

elmerland: Given the use of `timeout` in the second label above would it be an option to use a word like `incomplete-response` or something like that?

elmerland (approved):

elmerland (approved):

swankjesse (approved):

misk/src/main/kotlin/misk/client/ClientMetricsInterceptor.kt
jdm-square: Reading through https://github.com/squareup/cash-signalfx/blob/master/modules/dashboards/rpc/outgoing/main.tf I don't see any reason why that should be a problem.

PR #2210, opened 2021-11-11
Allow the minimum thread pool size to be configured.

mplauman:
Applications that experience large, spiky bursts of activity can benefit from
having a large pool of threads waiting on standby, rather than needing to spin
them up on demand.

Current default remain the same value (8).


shellderp:
take a read https://github.com/cashapp/misk/blob/master/CONTRIBUTING.md
and sign the license agreement

mplauman:
Fixed API, and signed the license agreement.

There's quite a few changes in the API because I added the new field to the middle of the class. Will that cause issues? Otherwise I can add it to the bottom of the class's field list, which should reduce the number of changes in `misk.api` by quite a bit. Not sure how critical ordering is in there.

r3mariano (approved):
Given config is usually loaded dynamically and that there's a default value on it, this should be OK.

shellderp:
I didn't have an issue adding something in the middle of the API for DataSourceConfig. Should be fine

PR #2231, opened 2021-12-01
Dont use select * from in jooq

Sethuraman:
We had a query that started randomly failing in a deployed environment. On further investigation we found that using ctx.select(<table name>.asterisk()).from(<tablename>) is untrustworthy. Read in the comment of AvoidUsingSelectStartListener.renderEnd for more details.


misk-jooq/src/main/kotlin/misk/jooq/listeners/AvoidUsingSelectStartListener.kt
```diff
@@ -0,0 +1,56 @@
+package misk.jooq.listeners
+
+import org.jooq.ExecuteContext
+import org.jooq.impl.DefaultExecuteListener
+
+class AvoidUsingSelectStartListener : DefaultExecuteListener() {
```
hugomd, line 6:
```suggestion
class AvoidUsingSelectStarListener : DefaultExecuteListener() {
```

chris-ryan-square (approved):

PR #2236, opened 2021-12-10
Introduce misk-bom

r3mariano:
Doing this allows us to constrain transitive dependencies better.

This will be needed before upgrading Jetty.

Also moves applying kotlin and java libs to subprojects.


misk-bom/build.gradle.kts
```diff
@@ -0,0 +1,21 @@
...
+dependencies {
+  constraints {
+    // Jetty 10+ uses SLF4J 2.0-alpha, which is unstable. It works well with 1.7 substituted in.
+    // See: https://github.com/eclipse/jetty.project/issues/5943
+    api(Dependencies.slf4jApi)
+  }
+}
+
+
+publishing {
```
r3mariano, line 15: Will need to check if publishing works, but can only do that outside the PR.

chris-ryan-square (approved):

Issue #2105, opened 2021-08-05
Graceful shutdown period


ewolak-sq:
When deploying a new version of a service behind a load balancer, it's useful to "gracefully" terminate the instances of the old version while directing all new traffic to the instances of the new version. In Kubernetes-land, this is usually done by signaling to the soon-to-be-terminated instances to stop responding successfully to health checks. When the load balancer notices they're failing health checks, it will stop sending new requests to them and direct all traffic to the new instances that _are_ passing health checks. Then, after some reasonable grace period, such as 30 seconds, the old instances can shut down entirely.

From what I understand, Misk's [current shutdown behavior](https://github.com/cashapp/misk/blob/5b6e3fa2f61e466307e09f9fa3995c1fbf75544c/misk/src/main/kotlin/misk/MiskApplication.kt#L91) is to just race to shutdown (while preserving `CoordinatedService` ordering), with no provision for a grace period of 1) intentionally failing health checks 2) refusing new requests or 3) both to give the upstream load balancer an opportunity to re-route traffic.

This manifests as spurious errors in services that depend on Misk services, because sometimes they make calls during the brief shutdown period and those calls fail in weird ways due to the rushed shutdown.

ewolak-sq:
@keeferrourke @swankjesse Do either of you have opinions on this?

ewolak-sq:
For Kubernetes+Istio specifically, we may not need to do this directly inside of Misk, because Kubernetes will immediately remove the Pod from the Endpoint when terminating, and Istio will pick that up. For any sort of _external_ load balancer, though (such as F5 or AWS ALB), it would still be beneficial to gracefully signal shutdown to the LB before dying.

r3mariano:
I was looking into this, and noticed that Misk [already configures Jetty to do a graceful shutdown](https://github.com/cashapp/misk/blob/9593078083385e235dda816af539bd48bae616c3/misk/src/main/kotlin/misk/web/jetty/JettyService.kt#L248) ([SO](https://stackoverflow.com/questions/25011249/embedded-jetty-and-graceful-shutdown)). Then, because `JettyService` is in a terminating state, the readiness check also fails, even if it's still serving.

jdm-square:
I just noticed a poor shutdown experience in merbro. It starts like this (each block is newest to oldest messages due to LQ's ordering, sorry):
```
[...]

PR #2238, opened 2021-12-10
Support a configurable delay before shutting down services

jdm-square:
This is a stopgap solution for #2105. The correct fix involves supporting explicit dependencies when installing JettyService, as well as modifying the behaviour of ReadinessCheckAction and LivenessCheckAction, and I still intend to work on that. However, this is an issue that is affecting the production readiness of my team's services today, and this change will buy us breathing room.

When misk is used in conjunction with k8s, there's a small window during shutdown in which misk continues to accept new connections before k8s takes the pod out of the rotation. This change makes it possible to configure misk to wait long enough to finish servicing those requests before starting to shut down any services that may be otherwise required. By defaulting to a delay of 0ms, the default behaviour for misk applications is unchanged.


misk/src/main/kotlin/misk/MiskApplication.kt
```diff
@@ -88,6 +90,10 @@ class MiskApplication(private val modules: List<Module>, commands: List<MiskComm
     Runtime.getRuntime().addShutdownHook(object : Thread() {
       override fun run() {
         log.info { "received a shutdown hook! performing an orderly shutdown" }
+        if (injector.getExistingBinding(Key.get(WebConfig::class.java)) != null) {
+          val config = injector.getInstance<WebConfig>()
+          sleep(config.shutdown_sleep_ms.toLong())
```
jdm-square, line 95: This is my attempt to make the delay configurable. It's not clear to me what other avenues exist for making it possible for consumers to customize behaviour at this layer of the framework.

misk/src/main/kotlin/misk/MiskApplication.kt
```diff
@@ -88,6 +90,10 @@ class MiskApplication(private val modules: List<Module>, commands: List<MiskComm
     Runtime.getRuntime().addShutdownHook(object : Thread() {
       override fun run() {
         log.info { "received a shutdown hook! performing an orderly shutdown" }
+        if (injector.getExistingBinding(Key.get(WebConfig::class.java)) != null) {
```
r3mariano, line 93: I think this would make a lot more sense at the start of `JettyService#shutDown()`. Saves having to check with the injector too, and makes the functionality testable (at least, in theory—testing a `sleep` call probably won't be nice).

r3mariano (changes requested):

misk/src/main/kotlin/misk/MiskApplication.kt
jdm-square: Unfortunately that does not actually help with the problem - the issue is with other services being shut down before the jetty service is finished servicing requests.

jdm-square: I could move the sleep to the hibernate service instead, I suppose, at the risk of uncovering other services that may also be affected in the same way. Sleeping here avoids the risk of playing whack-a-mole for something that is intended to be a short-term workaround.

r3mariano (approved):

PR #2249, opened 2021-12-21
Give Limiter a nanosecond clock like it expects

ewolak-sq:
The `Supplier<Long> clock` that AbstractLimiter.Builder.clock() expects is supposed to return a
_nanosecond_ value, not a millisecond value. All of the internal machinery of concurrency-limits
assumes the clock values are in nanoseconds, so we should give it what it expects.

`VegasLimiter` divides the current `rtt` by the minimum RTT it's seen recently (`rtt_noload`),
so as long as they're reasonably large integers (say, >10), then the rounding error introduced
by the lack of timing precision here should be a non-issue. However, for _very fast_ actions
(such as `cash-oauth`'s `AuthenticateClientAction`), the resolution is impaired to the point
of causing problems. If `min_rtt` is 2.54ms, that gets rounded off to 2. Then, if a request
happens to hit a 1ms GC pause, it takes `3.54ms` and gets rounded off to 3. Assuming an
`estimatedLimit` of 20 (the default value), `VegasLimiter` then estimates `queueSize` at 7.
The Vegas algorithm tries to keep the queue size between `alpha` (default 3) and `beta`
(default 6), but with ms resolution we only have a few discrete values for `queueSize`:
[0, 7, 10, 12, ...]. So the algorithm sees `queueSize` fluttering between 0 and 7, instead of
smoothly varying between 0 and 7, and doesn't handle that well.

This is the first (tiny) fruit of [my investigation into ConcurrencyLimitsInterceptor](https://docs.google.com/document/d/1oKsBcNupc7805vE3K5S-g5wALjohbqjDPelYxqWzQ8I/edit) and its impact
on Cash App Pay services.

### DIFF
--- a/misk/src/main/kotlin/misk/web/interceptors/ConcurrencyLimitsInterceptor.kt
+++ b/misk/src/main/kotlin/misk/web/interceptors/ConcurrencyLimitsInterceptor.kt
@@ -19,6 +19,8 @@ import org.slf4j.event.Level
 import wisp.logging.getLogger
 import wisp.logging.log
 import java.net.HttpURLConnection
+import java.time.Duration
+import java.time.Instant
 
 /**
  * Detects degraded behavior and sheds requests accordingly. Internally this uses adaptive limiting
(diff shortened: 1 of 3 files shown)


jdm-square (approved):

r3mariano:
Needs a conflict resolved, but lgtm. Ping me to merge, there's a stuck build part in PRs that I can't make sense of right now.

PR #2253, opened 2022-01-05
Use jobqueue-consumers-fetch-batch-size to determine the fetch batch size from SQS

Sethuraman:
I think there is an issue in the way SQS consumer config is done in misk.

I was having a look at SQS consumers code here - https://github.com/cashapp/misk/blob/4b3560cc5be086c38fa1ece5b3ea180c628cecfd/misk-aws/src/main/kotlin/misk/jobqueue/sqs/SqsJobConsumer.kt

The way I think the code works currently is, we configure an LD flag - jobqueue-consumers and set the number of parallel consumers. This could get split into multiple pods or just start multiple threads in the same pod.  It all depends on the timing. I am going to call these threads “Get from SQS threads”.

Then each of these threads will go get messages from SQS. By default we get 10 messages at a time. These 10 messages are then processed in parallel by starting up separate threads that handle them. I am going call these “Process SQS message threads”.

We have a scenario, where we would like to process one message at a time from SQS. It’s a very low volume queue. So we set the jobqueue-consumers LD flag to be 1 for our queue. This ensured we have only one “get from sqs thread” running. But we have 10 “process sqs message” threads running at the same time.
There is a flag  on the class https://github.com/cashapp/misk/blob/4b3560cc5be086c38fa1ece5b3ea180c628cecfd/misk-aws/src/main/kotlin/misk/jobqueue/sqs/AwsSqsJobQueueConfig.kt#L20 called message_batch_size, which is by default 10. This can be changed in the config to 1 and then we will have a single message processed at a time. But the trouble is this is configured for all queues in a service.

2 ways to fix this:
1. Move the batch size to retrieve from SQS into an LD flag and default it to 10, but configurable at a queue level. <- I prefer this. Which is this PR.
2. Change the AwsSqsJobQueueConfig to have a map of batch sizes to queue name and pass that in via the code.

### DIFF
--- a/misk-aws/src/main/kotlin/misk/jobqueue/sqs/AwsSqsJobQueueConfig.kt
+++ b/misk-aws/src/main/kotlin/misk/jobqueue/sqs/AwsSqsJobQueueConfig.kt
@@ -17,6 +17,9 @@ class AwsSqsJobQueueConfig(
   /**
    * Max number of messages to pull from SQS with each request.
    */
+  @Deprecated("Since this flag is set for all queues, this has been replaced with a Launch " +
+    "Darkly flag called jobqueue-consumers-fetch-batch-size. The value of this flag is set as 10 by default " +
+    "and can be customised per queue.")
   val message_batch_size: Int = 10,
 
   /**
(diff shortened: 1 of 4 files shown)


Sethuraman:
- [ ] Create LD flag before merge

vektah (approved):
makes sense to me

swankjesse (approved):

PR #2271, opened 2022-02-01
Add ability to set max age for prometheus summaries

MatthewDolan:
R: @r3mariano @chris-ryan-square 

There was a support request to set this option for a specific metrics (http latency). Before the option can be set, it must first be exposed through the wrapping API.

### DIFF
--- a/misk-metrics-testing/src/main/kotlin/misk/metrics/FakeMetrics.kt
+++ b/misk-metrics-testing/src/main/kotlin/misk/metrics/FakeMetrics.kt
@@ -32,7 +32,8 @@ class FakeMetrics @Inject internal constructor(
     name: String,
     help: String,
     labelNames: List<String>,
-    quantiles: Map<Double, Double>
+    quantiles: Map<Double, Double>,
+    maxAgeSeconds: Long?
   ): Histogram {
     val summary = Summary.build(name, help)
       .labelNames(*labelNames.toTypedArray())
--- a/misk-metrics-testing/src/main/kotlin/misk/metrics/FakeMetrics.kt
+++ b/misk-metrics-testing/src/main/kotlin/misk/metrics/FakeMetrics.kt
@@ -32,7 +32,7 @@ class FakeMetrics @Inject internal constructor(
     help: String,
     labelNames: List<String>,
     quantiles: Map<Double, Double>,
-    maybeMaxAgeSeconds: Long?
+    maxAgeSeconds: Long?
   ): Histogram {
     val summary = Summary.build(name, help)
       .labelNames(*labelNames.toTypedArray())
@@ -42,8 +42,8 @@ class FakeMetrics @Inject internal constructor(
         }
       }
       .apply {
-        if (maybeMaxAgeSeconds != null) {
-          maxAgeSeconds(maybeMaxAgeSeconds)
+        if (maxAgeSeconds != null) {
+          this.maxAgeSeconds(maxAgeSeconds)
         }
       }
       .register(registry)
--- a/misk-metrics-testing/src/main/kotlin/misk/metrics/FakeMetrics.kt
+++ b/misk-metrics-testing/src/main/kotlin/misk/metrics/FakeMetrics.kt
@@ -5,7 +5,6 @@ import io.prometheus.client.CollectorRegistry
 import io.prometheus.client.Counter
 import io.prometheus.client.Gauge
 import io.prometheus.client.Summary
-import misk.metrics.Metrics
 import javax.inject.Inject
 import javax.inject.Singleton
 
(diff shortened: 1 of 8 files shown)


misk-prometheus/src/main/kotlin/misk/metrics/backends/prometheus/PrometheusMetrics.kt
```diff
@@ -32,8 +32,9 @@ internal class PrometheusMetrics @Inject internal constructor(
     name: String,
     help: String,
     labelNames: List<String>,
-    quantiles: Map<Double, Double>
-  ): Histogram = PrometheusHistogram(metricsV2.summary(name, help, labelNames, quantiles))
+    quantiles: Map<Double, Double>,
+    maxAgeSeconds: Long?
```
dau-square, line 36:
```suggestion
    maxAgeSeconds: Long? = null,
```

misk-prometheus/src/main/kotlin/misk/metrics/backends/prometheus/v2/PrometheusMetrics.kt
```diff
@@ -49,7 +49,8 @@ internal class PrometheusMetrics @Inject internal constructor(
     name: String,
     help: String,
     labelNames: List<String>,
-    quantiles: Map<Double, Double>
+    quantiles: Map<Double, Double>,
+    maxAgeSeconds: Long?
```
dau-square, line 53:
```suggestion
    maxAgeSeconds: Long? = null,
```

dau-square (approved):

misk-prometheus/src/main/kotlin/misk/metrics/backends/prometheus/PrometheusMetrics.kt
dau-square: nit: add default value to avoid making a breaking change with the function signature

misk-metrics/src/main/kotlin/misk/metrics/Metrics.kt
```diff
@@ -62,7 +62,8 @@ interface Metrics {
     name: String,
     help: String = "",
     labelNames: List<String>,
-    quantiles: Map<Double, Double> = defaultQuantiles
+    quantiles: Map<Double, Double> = defaultQuantiles,
+    maxAgeSeconds: Long?
```
dau-square, line 66:
```suggestion
    maxAgeSeconds: Long? = null,
```

misk-metrics/src/main/kotlin/misk/metrics/v2/Metrics.kt
```diff
@@ -71,7 +71,8 @@ interface Metrics {
     name: String,
     help: String = "",
     labelNames: List<String>,
-    buckets: List<Double> = defaultBuckets
+    buckets: List<Double> = defaultBuckets,
+    maxAgeSeconds: Long?
```
dau-square:
```suggestion
    maxAgeSeconds: Long? = null,
```

r3mariano (approved):

MatthewDolan:
I did a final review through the code and made a few changes (for legibility and accuracy). Sorry about the flux here.

I am going to re-request review even though there are currently approvals because I think it's been enough change to warrant one last set of eyes on the code before merging.

misk-metrics-testing/src/main/kotlin/misk/metrics/FakeMetrics.kt
```diff
@@ -32,7 +31,8 @@ class FakeMetrics @Inject internal constructor(
     name: String,
     help: String,
     labelNames: List<String>,
-    quantiles: Map<Double, Double>
+    quantiles: Map<Double, Double>,
+    maybeMaxAgeSeconds: Long?
```
r3mariano, line 35: nit: I'd keep this (and the other instances of it) `maxAgeSeconds: Long?`, maybe with a default.

r3mariano (approved):

misk-metrics-testing/src/main/kotlin/misk/metrics/FakeMetrics.kt
dau-square: +1 for default

dau-square (approved):

misk-metrics-testing/src/main/kotlin/misk/metrics/FakeMetrics.kt
MatthewDolan: Thanks for the feedback.

> I'd keep this (and the other instances of it) maxAgeSeconds: Long?

I was trying to avoid the code below being confusing.

```
  if (maxAgeSeconds != null) {
    maxAgeSeconds(maxAgeSeconds)
  }
```

Where `maxAgeSeconds` represents both a variable and a function on the summary builder. I guess the alternative is to use `this` to make it clear which one is a function.

> adding a default value

I am a little confused about the right use of defaults in Kotlin. There is a default on the interface:

https://github.com/cashapp/misk/pull/2271/files#diff-4a56fc32fc0a0bfbf879f1050519633c6d3d4c97afcebcddc3b4f2e74087d7e9L65-R66

This class implements that interface. Do I also need to add the default here or does it inherit the default?

If I try adding a default to this function, I get an error message:

> An overriding function is not allowed to specify default values for its parameters

dau-square: You're correct, the default should be defined in the interface (and gets inherited)

PR #2279, opened 2022-02-10
Improves handling timeout/reset connection when reading request

frojasg:
* Improves exception reporting when we failed to read a request stream
* Creates RequestBodyException to wrap jetty exception when reading a
  request


misk/src/test/kotlin/misk/web/ssl/Http2ConnectivityTest.kt
```diff
@@ -241,6 +279,30 @@ class Http2ConnectivityTest {
     }
   }
 
+  class LockInterceptor constructor(val queue: ArrayBlockingQueue<Int>) : NetworkInterceptor {
```
acostama, line 282: don't remember exactly but do we still need this?

acostama (approved):

swankjesse (approved):

misk/src/test/kotlin/misk/web/ssl/Http2ConnectivityTest.kt
frojasg: yeah, this so the test can know when they can assert the side effect on the server side after the timeout.

misk/src/test/kotlin/misk/web/ssl/Http2ConnectivityTest.kt
```diff
@@ -145,23 +154,34 @@ class Http2ConnectivityTest {
...
+        "Http2ConnectivityTest.DisconnectWithLargeResponseAction",
+        "unknown",
+        "500"
+      ).get().count.toInt()
+    ).isEqualTo(1)
   }
 
   /** Confirm we don't page oncall and we record a 499 in the metrics. **/
   @Test
   fun clientTimeoutWritingTheRequest() {
```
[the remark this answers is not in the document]
frojasg: added

PR #2289, opened 2022-02-16
Add misk-policy and misk-policy-testing READMEs

AlexSzlavik:


natvyas (approved):
Might be valuable to link toolbox as an example..

AlexSzlavik:
misk is an open  source project, probably shouldn't  make reference to internal tooling.

misk-policy/README.md
```diff
@@ -0,0 +1,69 @@
+# misk-policy
+
+The misk-policy module provides a type safe interface to policy engine side cars.
+The only supported policy engine is [Open Policy Agent (OPA)](https://www.openpolicyagent.org/docs/latest/) today.
+
+## Assumptions
+
+Setup of OPA is not included.  
+The policy modules assumes that the policy engine is accessible via HTTP(s).
```
amanymous-net, line 9: Since HTTP is the transport layer. This would be more accurately, `is accessible via a TCP socket or abstract unix domain socket`

PR #2293, opened 2022-02-21
Add strongly typed feature flag support

jakew-square:
this-is-gonna-be-good

# Introduction

**Work in progress**

This PR adds strongly typed feature flags to `wisp-feature`, `wisp-feature-testing, `wisp-launchdarkly`, `misk-feature`, `misk-feature-testing` and `misk-launchdarkly`.

The idea is to define feature flags and their attributes using data classes, without breaking the existing approach. 

Goals:

- Define your feature flags using strong types
- Provide strong feature flags in parallel with legacy feature flags, to make it easer to incrementally adopt the new approach
- Easy, fluent test support 

# Example

## Step 1: Define some flags

```kotlin
data class DisableAvatar(
  val customerToken: String,
  val clientIdentifier: String,
  val avatarType: AvatarType
) : BooleanFeatureFlag { // String, Double, Enum, Json, etc... are also available
  override val feature = Feature("disable-avatar")
  override val key = customerToken
  override val attributes = Attributes()
    .with("clientIdentifier", clientIdentifier) 
    .with("avatarType", avatarType.toString()) 
}

data class AppJsonObject(val id: Int, val message: String)
data class AppJsonFlag(
  val customerToken: String
) : JsonFeatureFlag<AppJsonObject> {
  override val returnType = AppJsonObject::class
  override val feature = Feature("app-json-flag")
  override val key = customerToken
  override val attributes = Attributes()
}
```

## Step 2: Use the flag

```kotlin
class SomeService(val featureFlags: FeatureFlags) {
  fun something() {
    val disableAvatar = featureFlags.get(
      DisableAvatar(
        customerToken = "...",
        clientIdentifier = "...",
        routeId = AvatarType.SQUARE
      )
    ) // Returns Boolean / Double / Data Class depending on the flag type

    val appJsonObject = featureFlags.get(
      JsonFeatureFlag(customerToken = "...")
    ) // Returns AppJsonObject
  }
}
```

## Step 3: Test the flag

```kotlin
@Test
fun `test the thing`() {
  val featureFlags = FakeFeatureFlags()
    .override<DisableAvatar>(true) // Make `DisableAvatar` true normally
    .override<DisableAvatar>(false) { it.avatarType == AvatarType.SQUARE } // Make `DisableAvatar` false if the avatarType is square
    .override<JsonFeatureFlag, AppJsonObject>(AppJsonObject(1, "hello"))

  // inject `featureFlags` into the system-under-test
}
```

# TODO

- [x] Many more tests (particularly for `wisp-feature-testing`)
- [X] Attempt to use this in a real dependency and make sure partial migrations work and are smooth
- [X] Figure out why `misk-feature` duplicates various interfaces of `wisp-feature`, and then figure out what to do about it. (Spoiler: copy-and-paste the interface definitions 😞 )
- [x] Update README examples (including renaming `with` -> `override`)
- [x] [Remove the strong `track` methods](https://github.com/cashapp/misk/pull/2293#discussion_r813200144)


acostama:
mind giving some examples of how these will be set in launch darkly? is it simply specifying the attributes to match the strongly typed object?

jakew-square:
@acostama: Sure thing!

The idea is launch darkly configuration remains the same as today, right now Misk supports the `string` and `number` types from LaunchDarkly which means each "strong" type ultimately needs to be mapped to a `string` or a `number`. We don't currently support `array` or `boolean` but we should be able to add it if needed.

The "strong type" to "launch darkly attribute" conversion happens as part of defining a feature flag:

```kotlin
data class MyFeatureFlag(
  // Strongly typed fields go here
  val customerToken: CustomerToken,
  val customerRegion: Region
) : BooleanFeatureFlag  { // inherited flag type must match the flag configuration in launch darkly
  val feature = Feature("my-feature-flag") // Must match a feature flag name in launch darkly
  val key = customerToken  // this field will be sent to launch darkly as the feature key
  val attributes = Attributes() // these fields will be sent to launch darkly as string/number custom attributes
    .with("customerRegion", customerRegion.toString()) // here we convert our typesafe region to a string or number
}
```

The key constraints are:

- `feature` needs to match a feature name in launch darkly
- The flag type (`BooleanFeatureFlag`/`StringFeatureFlag`/etc...) needs to match the flag type in launch darkly
- The `attributes` all need to be `string` or `number` and will be sent as custom attributes to launch darkly.

This means we aren't perfectly typesafe since we can still make mistakes when defining `MyFeatureFlag`, but we are typesafe when using and testing with `MyFeatureFlag` which I think is a bit better then the status quo 🙂

misk-feature-testing/src/main/kotlin/misk/feature/testing/FakeFeatureFlags.kt
```diff
@@ -63,6 +82,42 @@ class FakeFeatureFlags @Inject constructor(val moshi: Provider<Moshi>) : Abstrac
     return delegate.getJson(feature, key, clazz, attributes)
   }
 
+  override fun track(
```
escardin, line 85: I would generally prefer that we don't use the tracking implementations. It generally results in less flexible solutions and unexpected complexity. I like that you expanded them too in the name of completeness, but I'd rather retire them.

misk-feature-testing/src/test/java/misk/feature/testing/FakeFeatureFlagsJavaTest.java
```diff
@@ -64,4 +64,41 @@ public void configure() {
     assertThat(subject.getDouble(FEATURE, "joker5", attributes)).isEqualTo(1.0);
   }
 
+  @Test
```
escardin, line 67: Can you put tests for all the types?

escardin (commented):
Looks fine to me. Maybe some more tests needed for thoroughness.

misk-feature-testing/src/main/kotlin/misk/feature/testing/FakeFeatureFlags.kt
jakew-square: Sure thing, I'll delete the "strong" variants before merging the PR 🙂

misk-feature-testing/src/test/java/misk/feature/testing/FakeFeatureFlagsJavaTest.java
jakew-square: Absolutely! I've just added a bunch of tests for `wisp-feature-flag` which cover all of the strong types, but I'll add the remaining types to the Java tests as well.

misk-feature-testing/src/main/kotlin/misk/feature/testing/FakeFeatureFlags.kt
jakew-square: Done in https://github.com/cashapp/misk/pull/2293/commits/4106012b23811a6d3daae178cee6208eaef72af8 :+1:

misk-feature-testing/src/test/java/misk/feature/testing/FakeFeatureFlagsJavaTest.java
jakew-square: This is also done

chris-ryan-square (approved):

PR #2304, opened 2022-03-02
repoint container target to host.docker.internal if the code is running in a docker environment

ctaslimsq:
repoint container target from 127.0.0.1 to host.docker.internal if the code is running in a docker environment. 

When running this code directly on the host machine, 127.0.0.1 is a proper address to reach the docker container as the container port is mapped to the host machine. But when this code is running in a docker container, It needs to target `host.docker.internal` to ensure that the ip target address is the host machine address instead of the local docker container's address


misk-aws/src/test/kotlin/misk/jobqueue/sqs/DockerSqs.kt
```diff
@@ -58,6 +59,7 @@ internal object DockerSqs : ExternalDependency {
     }
   }
 
+
```
chris-ryan-square, line 62: Missing changes?

chris-ryan-square (approved):

PR #2306, opened 2022-03-03
Make FakeFeatureFlag overrides composable with FakeFeatureFlagsOverrideModule

acostama:
In the current setup, all test feature flag overrides have to live in the same guice module. For large services we end up duplicating feature flag overrides in many places. 

This change allows us to have one `FakeFeatureFlagsModule` and many `FakeFeatureFlagsOverrideModule`. Modules at any level can install their own overrides independently using the familiar api of `FakeFeatureFlagsModule#withOverride`

### DIFF
--- a/misk-feature-testing/src/main/kotlin/misk/feature/testing/FakeFeatureFlagsModule.kt
+++ b/misk-feature-testing/src/main/kotlin/misk/feature/testing/FakeFeatureFlagsModule.kt
@@ -59,8 +59,9 @@ class FakeFeatureFlagsModule(
    *   override(Feature("foo"), true)
    * })
    * ```
+   *
+   * For overriding in many modules see [FakeFeatureFlagsOverrideModule]
    */
-  @Deprecated(message = "replaced by FakeFeatureFlagsOverrideModule")
   fun withOverrides(lambda: FakeFeatureFlags.() -> Unit): FakeFeatureFlagsModule {
     overrides.add(FakeFeatureFlagsOverrideModule(qualifier, lambda))
     return this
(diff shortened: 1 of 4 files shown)


adrw (approved):
Nice improvement to support bigger projects

misk-feature-testing/src/main/kotlin/misk/feature/testing/FakeFeatureFlagsModule.kt
```diff
@@ -1,34 +1,48 @@
...
 import misk.inject.toKey
+import javax.inject.Inject
 import kotlin.reflect.KClass
 
 /**
  * Binds a [FakeFeatureFlags] that allows tests to override values.
+ *
+ * To define overrides (especially test defaults) use [FakeFeatureFlagsOverrideModule].
+ * In a given misk service's test setup, there is one [FakeFeatureFlagsModule] installed and many
+ * [FakeFeatureFlagsOverrideModule] installed.
```
swankjesse, line 21: Wanna make it so there’s no asymmetry? Ideally I just install `FakeFeatureFlagsModule` one or more times.

swankjesse (approved):

misk-feature-testing/src/main/kotlin/misk/feature/testing/FakeFeatureFlagsModule.kt
acostama: yeah let me try that!

acostama: I got stuck since `FakeFeatureFlagsModule` takes an annotation.. it seems to me that there is no way of knowing if we have previously bound a FakeFeatureFlags with the given annotation

acostama: I was imagining installing an object module for all the `FakeFeatureFlags` bindings but that doesn't work here

acostama:
I am going to walk back the withOverrides deprecation to keep it open to installing the module multiple times (without being confusing) but will merge as-is to unblock

PR #2319, opened 2022-03-11
Extend LeaseManager to support Leases as AutoCloseable resources

keeferrourke:
Scratching an itch with the Lease interface.

Here's an implementation independent extension function to make leases auto-closeable.

Intention is to use it like:

```kt
@Inject lateinit val LeaseManager leaseManager

leaseManager.acquireOrNull("some-lease")?.use { lease ->
  /* Do critical stuff that needs the lease */
}
```

### DIFF
--- a/misk-zookeeper/src/test/kotlin/misk/clustering/zookeeper/ZkLeaseTest.kt
+++ b/misk-zookeeper/src/test/kotlin/misk/clustering/zookeeper/ZkLeaseTest.kt
@@ -23,14 +23,19 @@ import org.mockito.Mockito.times
 import org.mockito.Mockito.verify
 import org.mockito.Mockito.verifyNoMoreInteractions
 import wisp.lease.Lease
+import wisp.lease.acquireOrNull
 import java.util.concurrent.CountDownLatch
+import java.util.concurrent.LinkedBlockingDeque
+import java.util.concurrent.LinkedBlockingQueue
 import java.util.concurrent.TimeUnit
 import java.util.concurrent.atomic.AtomicBoolean
 import javax.inject.Inject
 
+@Suppress("UsePropertyAccessSyntax")
 @MiskTest(startService = true)
 internal class ZkLeaseTest {
-  @MiskTestModule private val module = Modules.combine(MiskTestingServiceModule(), ZkLeaseTestModule())
+  @MiskTestModule private val module =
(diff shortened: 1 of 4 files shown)


misk-zookeeper/src/test/kotlin/misk/clustering/zookeeper/ZkLeaseTest.kt
```diff
@@ -289,6 +294,28 @@ internal class ZkLeaseTest {
     assertThat(acquireCalled.get()).isTrue()
   }
 
+  @Test fun usableWithAutoCloseable() {
```
keeferrourke, line 297: Dunno if this is the right place for this test.

adrw (approved):

wisp-lease/src/main/kotlin/wisp/lease/Extensions.kt
```diff
@@ -0,0 +1,24 @@
...
+ *
+ * Use like
+ *
+ * ```
+ * leaseManager.acquireOrNull("some-lease")?.use { lease ->
+ *   // Do something with the lease.
+ * }
+ * ```
+ */
+fun LeaseManager.acquireOrNull(name: String): AutoCloseableLease? {
```
aerb, line 21: One potential downside of this is it's still possible to acquire and then forget to close a lease. I've seen the `useX` convention used in the kotlin stdlib that avoids this by just accepting a lambda directly. Example that jumps to mind is the [useLines](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.io/java.io.-file/use-lines.html) method. 
Eg: 
```kotlin
File("").useLines { lines -> println(lines) }
```

This may be a bit awkward her - I'm not sure what the behavior should be if you can't acquire ...

keeferrourke: You raise a good point, which is also still a problem with the prescribed usage without this extension. 
Leases are held only for some time, so if a user forgets to close the lease, it will be let go eventually.

Something like `leaseManager.useLease { lease -> ... }` would be nicer from that perspective, however it would probably have to come with the stipulation that the lambda doesn't run if the lease can't be acquired. Unsure if that would be surprising. Perhaps `leaseManager.maybeUseLease { lease -> ... }`?  Dunno! 🤷

PR #2325, opened 2022-03-17
Provide more helpful error message when duplicate migration versions present

mhickman:
When the migration list is long and a duplicate is near the middle, it's
pretty hard to see what the duplicates are with the current implementation.


adrw (approved):
LG

PR #2363, opened 2022-04-13
add bounded next int to FakeRandom

szabado:


adrw:
What's the motivation for this? What does it solve?

szabado:
> What's the motivation for this? What does it solve?

The motivation for this is to make it possible to specify what `nextInt(bount: Int)` will return in tests, so that tests can be more precise. `java.util.Random.nextInt(bound)` calls `next(bits: Int)`, and while overriding the implementation of `next()` is possible it's relatively involved and too low level IMO.

The default java Random's `nextInt(bound: Int)` does not call `nextInt()` under the hood.

adrw:
Merged in https://github.com/cashapp/misk/commit/e254e4d85c5700c3f7be6669cef8882c14fc682b

PR #2377, opened 2022-04-22
Refactor external key access

yan:


misk-crypto/src/main/kotlin/misk/crypto/ExternalKeyResolver.kt
```diff
@@ -0,0 +1,45 @@
...
+ * [ExternalKeyResolver] implements an [KeyResolver] that fetches Tink keysets from an external
+ * source, such as an S3 bucket. If multiple sources are registered (by binding implementations
+ * of [KeyResolver]), the first one to contain the key (via [ExternalKeySource.keyExists]) is
+ * the key that is used.
+ *
+ * If a key is not found, an [ExternalKeyManagerException] exception is raised.
+ */
+class ExternalKeyResolver @Inject constructor(
+  @ExternalDataKeys override val allKeyAliases: Map<KeyAlias, KeyType>,
+
```
yoavamit, line 16: super nit: empty line

misk-crypto/src/main/kotlin/misk/crypto/ExternalKeyResolver.kt
```diff
@@ -0,0 +1,45 @@
...
+  private val externalKeySources: Set<ExternalKeySource>,
+) : KeyResolver {
+
+  private fun getRemoteKey(alias: KeyAlias): Key {
+    for (keySource in externalKeySources) {
+      logger.info("checking ${keySource.javaClass.name} for $alias")
```
yoavamit, line 22: nit: I be lieve the more "kotlinized" way to write is `keySource::class.java` instead of `keySource.javaClass`

misk-crypto/src/main/kotlin/misk/crypto/ExternalKeyResolver.kt
```diff
@@ -0,0 +1,45 @@
...
+      if (keySource.keyExists(alias)) {
+        keySource.getKey(alias)?.let { return it }
+        logger.warn("$alias is reported to exist, but not able to retrieve; continuing")
```
yoavamit, line 25: WDYT about changing the line above to `keySource.getKey(alias)!!`?
If the key isn't retrieved even though it exists, I think it's better to report that error instead of hiding it behind a more general warning log message.

yoavamit (approved):

PR #2388, opened 2022-05-03
Update key loading and add extra logging

yan:
* Bind key resolvers as an eager singleton.
 * Remove extraneous key caching from S3KeySource
 * Update handling of public keys

### DIFF
--- a/misk-crypto/src/main/kotlin/misk/crypto/CryptoModule.kt
+++ b/misk-crypto/src/main/kotlin/misk/crypto/CryptoModule.kt
@@ -24,7 +24,6 @@ import misk.crypto.pgp.PgpDecrypterProvider
 import misk.crypto.pgp.PgpEncrypter
 import misk.crypto.pgp.PgpEncrypterProvider
 import misk.inject.KAbstractModule
-import misk.inject.asSingleton
 import okio.ByteString
 import okio.ByteString.Companion.toByteString
 import org.bouncycastle.jce.provider.BouncyCastleProvider
--- a/misk-crypto/src/main/kotlin/misk/crypto/CryptoModule.kt
+++ b/misk-crypto/src/main/kotlin/misk/crypto/CryptoModule.kt
@@ -24,6 +24,7 @@ import misk.crypto.pgp.PgpDecrypterProvider
 import misk.crypto.pgp.PgpEncrypter
 import misk.crypto.pgp.PgpEncrypterProvider
 import misk.inject.KAbstractModule
+import misk.inject.asSingleton
 import okio.ByteString
 import okio.ByteString.Companion.toByteString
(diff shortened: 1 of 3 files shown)


misk-crypto/src/main/kotlin/misk/crypto/ExternalKeyResolver.kt
```diff
@@ -19,10 +19,15 @@ class ExternalKeyResolver @Inject constructor(
...
+      logger.info("getRemoteKey: checking ${keySource::class.qualifiedName} for $alias")
       if (keySource.keyExists(alias)) {
-        keySource.getKey(alias)?.let { return it }
+        keySource.getKey(alias)?.let {
+          logger.info("getRemoteKey: retrieved ${it.key_name}; returning")
+          return it
+        }
         logger.warn("$alias is reported to exist, but not able to retrieve; continuing")
+      } else {
+        logger.info("getRemoteKey: ${keySource::class.qualifiedName} doesn't have the key")
```
amanymous-net, line 30: nit: not sure if this else is explicitly necessary, since we're already reporting the positive case.

amanymous-net (approved):

misk-crypto/src/main/kotlin/misk/crypto/CryptoModule.kt
```diff
@@ -96,7 +96,7 @@ class CryptoModule(
       newMultibinder<ExternalKeySource>()
       multibind<ExternalKeySource>().to<S3KeySource>()
 
-      keyManagerBinder.addBinding().to<ExternalKeyResolver>()
+      keyManagerBinder.addBinding().to<ExternalKeyResolver>().asEagerSingleton()
```
yoavamit: Better use `.in(Singleton::class)`.
Using `.asEagerSingleton()` means the class will be instantiated immediately, which will also force all object it depends on to be instantiated as well.

misk-crypto/src/main/kotlin/misk/crypto/S3KeySource.kt
```diff
@@ -58,26 +58,31 @@ class S3KeySource @Inject constructor(
   } ?: defaultS3
 
   // N.B. The path we're using for the object is based on _our_ region, not where the bucket lives
-  private fun objectPath(alias: String, type: KeyType): String {
-    if (type === KeyType.HYBRID_ENCRYPT) {
-      return "$alias/public"
-    }
-    return "$alias/${s3.regionName.toLowerCase()}"
+  private fun objectPath(alias: String): String {
+    val basename = if (allKeyAliases[alias] == KeyType.HYBRID_ENCRYPT)
```
yoavamit, line 62: nit: missing `{`, `}`

yoavamit (approved):
LGTM, pending the singleton change :)

PR #2390, opened 2022-05-06
Propose HttpActionScopeSeedDataInterceptor

JGulbronson:
At Faire, we want to be able to have ActionScoped values in all of our scopes. For example we want to seed something like an `ActionInfo` data class, which contains the type of action (Job, HTTP request, RPC, etc), as well as some metadata about the Action (perhaps path). We could use to help tag metrics with the action type they came from.

Currently there is no way to modify the seed data that Misk passes to the ActionScope. This is a proposal to change that, and allow users to bind their own seed-data interceptors, modifying the map before passing it to `ActionScope::enter`.


misk/src/main/kotlin/misk/web/HttpActionScopeSeedDataInterceptor.kt
```diff
@@ -0,0 +1,22 @@
+package misk.web
+
+import com.google.inject.Key
+import misk.Action
+import misk.scope.ActionScope
+
+/** Allows users to seed the [ActionScope] entered when handling an HTTP request. */
+interface HttpActionScopeSeedDataInterceptor {
```
JGulbronson, line 8: Not sold on the name

misk/src/main/kotlin/misk/web/HttpActionScopeSeedDataInterceptor.kt
```diff
@@ -0,0 +1,22 @@
...
+   *
+   * Returns a new seed data map which is used to enter the [ActionScope].
+   */
+  fun intercept(
+    seedData: Map<Key<*>, Any?>,
+    pathPattern: PathPattern,
+    action: Action,
+  ): Map<Key<*>, Any?>
+
+  // TODO(jeff) - Do we want this to follow the other Interceptor patterns and use a Factory?
```
JGulbronson, line 21: Open to going the Factory route, wasn't sure if we needed it, or if we wanted to be consistent regardless.

misk/src/test/kotlin/misk/web/ActionScopedWebDispatchTest.kt
```diff
@@ -38,6 +41,20 @@ internal class ActionScopedWebDispatchTest {
     assertThat(response.body!!.string()).isEqualTo("hello Thor")
   }
 
+  @Test
+  fun actionScopeSeedDataCanBeModified() {
```
JGulbronson, line 45: I feel like this tests the functionality okay and is the right test class, but I'm not sure if it's the most readable.

misk/src/main/kotlin/misk/web/HttpActionScopeSeedDataInterceptor.kt
swankjesse: If your expected implementations need to parse or dispatch on PathPattern, then it’s worthwhile going with a factory. That way you can compute and validate eagerly. Like if an annotation is malformed or whatever.

misk/src/main/kotlin/misk/web/HttpActionScopeSeedDataInterceptor.kt
```diff
@@ -0,0 +1,22 @@
...
+
+/** Allows users to seed the [ActionScope] entered when handling an HTTP request. */
+interface HttpActionScopeSeedDataInterceptor {
+  /**
+   * Accepts the current seed data, along with the [PathPattern] and [Action] for the [BoundAction]
+   * that is being handled.
+   *
+   * Returns a new seed data map which is used to enter the [ActionScope].
+   */
+  fun intercept(
```
swankjesse, line 15: I’m just pondering alternative approaches.

We’ve got this:

```
class ActionScope {
  fun enter(seedData: Map<Key<*>, Any?>): ActionScope { ... }
  fun close() { ... }
}
```

We could add something like this?

```
interface ActionScopeInterceptor {
  fun enter(seedData: Map<Key<*>, Any?>, chain: Chain): AutoCloseable

  interface Chain {
    fun proceed(seedData: Map<Key<*>, Any?>): AutoCloseable
  }
}
```

The upside there is I could seed closeable things, or do other work when the scope is closed.

You’ll also see this doesn’t declare web stuff . . . I’ve omitted the path pattern & action. I think if you need these, we do ’em in a factory for the interceptors. Again so we can do eager validation and optimize things.

swankjesse (commented):
Lets pair on this!

misk/api/misk.api
```diff
@@ -1452,6 +1452,10 @@ public abstract interface class misk/web/FeatureBinding$Subject {
 	public abstract fun takeReturnValue ()Ljava/lang/Object;
 }
 
+public abstract interface class misk/web/HttpActionScopeSeedDataInterceptor {
```
swankjesse, line 1455: @JGulbronson you’ll need to re-dump the API

swankjesse (approved):

JGulbronson:
Works locally, but it seems the tests are refusing to run w/ Github Actions

swankjesse:
Yeah, test failure looks unrelated.
```
ClientServerTraceTest > traceHopsFromClientToServerToServer() FAILED
    java.lang.IllegalArgumentException: no spans!
        at misk.testing.ConcurrentMockTracer.take(ConcurrentMockTracer.kt:20)
        at misk.testing.ConcurrentMockTracer.take(ConcurrentMockTracer.kt:29)
        at misk.tracing.ClientServerTraceTest.traceHopsFromClientToServerToServer(ClientServerTraceTest.kt:109)
```

swankjesse (approved):

PR #2397, opened 2022-05-22
Add Log Test extension to control Log Level

frojasg:


misk-testing/src/main/kotlin/misk/testing/LogLevelExtension.kt
```diff
@@ -0,0 +1,46 @@
...
+
+import ch.qos.logback.classic.Level
+import ch.qos.logback.classic.Logger
+import org.junit.jupiter.api.extension.BeforeEachCallback
+import org.junit.jupiter.api.extension.ExtensionContext
+import org.slf4j.LoggerFactory
+import javax.inject.Inject
+
+/**
+ * Helps you set the log level for misk tests. It would default all test by default to Log
```
frojasg, line 11: TODO: improve the doc 😅

misk-testing/src/main/kotlin/misk/testing/LogLevelExtension.kt
```diff
@@ -0,0 +1,46 @@
...
+    if(context == null || context.element.isEmpty) {
+      return Level.INFO
+    }
+    return  context.element.get().annotations.firstNotNullOfOrNull { it as? LogLevel }
+      ?.let { mapLevel(it.level) }
+      ?: findParent(context)
+      ?: Level.INFO
+  }
+
+  private fun findParent(context: ExtensionContext): Level? {
```
frojasg, line 31: This method lets you annotate the test class so all the tests within that class (including nested tests) will use that log level. It might be a bit of an overkill, and I'm open to removing it.

misk-testing/src/main/kotlin/misk/testing/LogLevelExtension.kt
r3mariano: imo doc on LogLevel is more valuable since that one is external-facing.

r3mariano (approved):

PR #2405, opened 2022-06-02
Introduce a Gunzip network interceptor in-place of Jetty's.

alvinsee:
## Context
At present, it is not possible to intercept an un-inflated request body that was gzip compressed - the default Jetty service installing the GzipHandler which is preemptively inflates the request when the `Content-Encoding=gzip` header appears in the request. There are use cases where we would like to intercept the request body and by surfacing a `NetworkInterceptor` that implements the gunzip operation, developers may programmatically define the interception chain.

## This PR
- Removes the assignment of `inflateBufferSize` in `JettyService` to disable installation of `GzipHttpInputInterceptor` provisioned by `GzipHandler`
- Make `HttpCall#requestHeaders` mutable in-order to modify incoming request headers
- Binds a new misk default NetworkInterceptor that performs gunzip on an incoming request body buffer and modifies appropriate header values

## Test Plan
GzipTest.kt test suite to use GunzipRequestBodyInterceptor instead GzipHttpInputInterceptor.

### DIFF
--- a/misk/src/main/kotlin/misk/web/interceptors/GunzipRequestBodyInterceptor.kt
+++ b/misk/src/main/kotlin/misk/web/interceptors/GunzipRequestBodyInterceptor.kt
@@ -1,0 +1,43 @@
+package misk.web.interceptors
+
+import misk.Action
+import misk.web.NetworkChain
+import misk.web.NetworkInterceptor
+import misk.web.WebConfig
+import okio.GzipSource
+import okio.buffer
+import org.eclipse.jetty.server.handler.gzip.GzipHandler
+import javax.inject.Inject
+import javax.inject.Singleton
+
+/**
+ * Inflates a gzipped compressed request. The interceptor should be installed if the default
+ * [GzipHandler] for handling such requests was opted out via [WebConfig.gunzip].
+ */
+class GunzipRequestBodyInterceptor private constructor() : NetworkInterceptor {
+  override fun intercept(chain: NetworkChain) {
+    val contentEncoding = chain.httpCall.requestHeaders[CONTENT_ENCODING]
+      ?: return chain.proceed(chain.httpCall)
+    if (contentEncoding.lowercase() == GZIP || COMMA_GZIP.matches(contentEncoding)) {
+      chain.httpCall.takeRequestBody()?.let {
+        chain.httpCall.putRequestBody(GzipSource(it).buffer())
+      }
+    }
+    chain.proceed(chain.httpCall)
+  }
+
+  @Singleton
+  class Factory @Inject internal constructor() : NetworkInterceptor.Factory {
+    override fun create(action: Action) = INTERCEPTOR
+
+    private companion object {
+      private val INTERCEPTOR = GunzipRequestBodyInterceptor()
+    }
+  }
+
+  private companion object {
+    private const val CONTENT_ENCODING = "Content-Encoding"
+    private const val GZIP = "gzip"
+    private val COMMA_GZIP = Regex(".*, *gzip")
+  }
+}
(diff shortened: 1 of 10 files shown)


misk/src/main/kotlin/misk/web/interceptors/GunzipRequestBodyInterceptor.kt
```diff
@@ -0,0 +1,36 @@
...
+import javax.inject.Singleton
+
+/**
+ * Inflates a gzipped compressed request. The interceptor should be installed if the default
+ * [GzipHandler] for handling such requests was opted out via [WebConfig.gunzip].
+ */
+class GunzipRequestBodyInterceptor private constructor() : NetworkInterceptor {
+  override fun intercept(chain: NetworkChain) {
+    val contentEncoding = chain.httpCall.requestHeaders["Content-Encoding"]
+    if (contentEncoding?.lowercase() == "gzip") {
```
andrewjenkins, line 20: Should we also check for `x-gzip`? (Just judging by the Mozilla [Content-Encoding page](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Encoding#directives), if the existing implementation doesn't check for this then feel free to disregard)

andrewjenkins (approved):
🔥

misk/src/main/kotlin/misk/web/interceptors/GunzipRequestBodyInterceptor.kt
alvinsee: https://www.eclipse.org/jetty/javadoc/jetty-9/org/eclipse/jetty/server/handler/gzip/GzipHandler.html

Afaict from the specs, the interceptor should only observe for `gzip` and/or presence of it in a comma separated list - which I should update to reflect that.

AlexSzlavik:
Have you actually tried to see if this works with ARFHeaders? is the gzip timestamp a concern at all?

alvinsee:
> Have you actually tried to see if this works with ARFHeaders? is the gzip timestamp a concern at all?

No, since that is an internal implementation and I wouldnt say theres an easy way to include a local build of misk into our projects. The gzip timestamp _should_ not matter in this case because the current hypothesis is that we were gunzipping the request followed by the ARF computation. We want to reverse the order meaning we preserve and do not inflate the request when performing the computation.

jdm-square:
> I wouldnt say theres an easy way to include a local build of misk into our projects

I've followed https://cash-dev-guide.squarecloudservices.com/tools/polyrepo/dependencies/#local-development to test out local changes in misk while building other polyrepo projects.

misk/src/main/kotlin/misk/web/interceptors/GunzipRequestBodyInterceptor.kt
```diff
@@ -0,0 +1,43 @@
...
+import okio.buffer
+import org.eclipse.jetty.server.handler.gzip.GzipHandler
+import javax.inject.Inject
+import javax.inject.Singleton
+
+/**
+ * Inflates a gzipped compressed request. The interceptor should be installed if the default
+ * [GzipHandler] for handling such requests was opted out via [WebConfig.gunzip].
+ */
+class GunzipRequestBodyInterceptor private constructor() : NetworkInterceptor {
```
swankjesse, line 17: Make this internal and install it by default?

swankjesse (approved):
One thing I don’t like is how the Misk user now has a bunch of options they need to worry about that was previously handled automatically.

Could you change to remove the configuration option, and instead automatically install the interceptor always? That changes this PR to be more of a pure refactoring!

(We will need to figure out _where_ in the network interceptors pipeline to install it, so that it can happen after security but before other things. Hmmmm.... I’m open to suggestions on this.)

PR #2417, opened 2022-06-10
Expose annotations for injected network interceptors to handle content encoding

alvinsee:
## Context
Developers need a way to provision network interceptors ahead of the Misk defaults. 

## This PR
- Exposes a new annotation `@BeforeContentEncoding`
- Add a new annotation `@ForContentEncoding`
- Concatenates network interceptors with said annotations before interceptors annotated by `@MiskDefault`

### DIFF
--- a/misk/api/misk.api
+++ b/misk/api/misk.api
@@ -1546,6 +1546,9 @@ public final class misk/web/formatter/ClassNameFormatter$Companion {
 	public final fun format (Lkotlin/reflect/KClass;)Ljava/lang/String;
 }
 
+public abstract interface annotation class misk/web/interceptors/BeforeContentEncoding : java/lang/annotation/Annotation {
+}
+
 public abstract interface class misk/web/interceptors/ConcurrencyLimiterFactory {
 	public abstract fun create (Lmisk/Action;)Lcom/netflix/concurrency/limits/Limiter;
 }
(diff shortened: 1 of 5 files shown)


misk/src/main/kotlin/misk/web/interceptors/EarlyBind.kt
```diff
@@ -0,0 +1,20 @@
...
+ */
+@Qualifier
+@Retention(AnnotationRetention.RUNTIME)
+@Target(
+  AnnotationTarget.CLASS,
+  AnnotationTarget.VALUE_PARAMETER,
+  AnnotationTarget.FIELD,
+  AnnotationTarget.TYPE
+)
+annotation class EarlyBind
```
swankjesse, line 20: Hmmm . . . I was thinking ‘Early’ in the sense that the execute before the other interceptors. The word _bind_ here suggests that they’re bound early?

I’d like another word for this. Maybe something about them being at the beginning of the list? Or executed first? 

Another option – we could make it clear that these ones run before the content is transformed?

`@BeforeContentEncoding` ? That’s very direct and non-surprising.

alvinsee: Naming is hard. I'll take `@BeforeContentEncoding` and also introduce another internal annotation `@ContentEncoding` so that the namespaces make sense

swankjesse (approved):

PR #2418, opened 2022-06-10
Add support for provenance in OPA API

adt123:
See https://www.openpolicyagent.org/docs/latest/rest-api/#provenance for more details on how provenance works in OPA.

### DIFF
--- a/misk-policy/src/main/kotlin/misk/policy/opa/Response.kt
+++ b/misk-policy/src/main/kotlin/misk/policy/opa/Response.kt
@@ -6,7 +6,18 @@ package misk.policy.opa
  */
 data class Response<T>(
   val decision_id: String?,
-  val result: T?
+  val result: T?,
+  val provenance: Provenance?
 )
 
-interface OpaResponse
+data class Provenance(
+  val version: String?,
+  val build_commit: String?,
+  val build_timestamp: String?,
+  val build_hostname: String?,
+  val revision: String?
+)
+
+interface OpaResponse {
+  var provenance: Provenance?
+}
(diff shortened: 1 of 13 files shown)


adt123:
@AlexSzlavik good call, I like the idea of making this a config option in the module!

misk-policy/src/main/kotlin/misk/policy/opa/Response.kt
```diff
@@ -6,7 +6,18 @@ package misk.policy.opa
...
+data class Provenance(
+  val version: String?,
+  val build_commit: String?,
+  val build_timestamp: String?,
+  val build_hostname: String?,
+  val revision: String?
+)
+
+interface OpaResponse {
+  var provenance: Provenance?
```
adt123, line 22: @AlexSzlavik I couldn't think of a better way to include this in the response to the user. Adding this change would mean that whoever uses this package currently will have to add something like `override var provenance: Provenance? = null` in their data class definitions that inherit from `OpaResponse`. Is that okay?

AlexSzlavik: my suggestion (after pairing with @cameronhotchkies) is to turn the OpaRequest and OpaResponse into abstract classes and giving provenance a default assignment of null. This would require that implementations of the abstract class change syntax to OpaResponse() / OpaRequest(). I think that's a cleaner implementation all together. We'll need to fixup the 3 repos that use this module (assuming no one outside block uses this feature).

amanymous-net (approved):
is there a plan to push updates to other repos, following this so as to not break them?

PR #2419, opened 2022-06-10
Add guice test.fieldbinder to misks test

frojasg:
Misks integrates with Guice [BoundFieldModule](https://github.com/google/guice/wiki/BoundFields#binding-fields-using-boundfieldmodule) to declaratively bind fields defined in misk tests


swankjesse:
I think mocks are bad and so I'm disinclined to encourage their use.

frojasg:
@swankjesse The binder annotation can be used with fakes as well. I added a test case to show how that would look like. 

I would be open to move the mock part to a different TestExtension. However, the order of operation is very important to get it right, and I'm afraid it could cause some dx issues.

swankjesse:
Discussed offline. I’m totally in on `@Bind`. Eager for the next commit.

misk-testing/src/test/kotlin/misk/testing/fieldbind/BindFieldsUsingProviderTest.kt
```diff
@@ -0,0 +1,23 @@
...
+class BindFieldsUsingProviderTest {
+  @Inject lateinit var ticketDispenser: TicketDispenserService
+  @Bind val counterService: Provider<CountService> = object : Provider<CountService> {
+    override fun get(): CountService {
+      return FakeCountService(42)
+    }
+  }
+
+  @Test
+  fun `bind CounterService using mockito`() {
```
swankjesse, line 20: is this name wrong?

frojasg: yup. good catch 😅

swankjesse (approved):
I love where this landed

PR #2421, opened 2022-06-14
Do not warn on unknown LD json fields

escardin:
The warning here is of minimal value. It's normal to update definitions which might have newer fields unknown to older clients. It's not an error state. In cases where someone has fat fingered the definition, it's much more likely that they'll only notice that their definition is wrong when everything explodes anyway.

It is of value when debugging such an explosion, but we should aim for a better way of identifying the issue.

See https://github.com/squareup/cash-franklin/pull/15662


swankjesse:
What’s the harm?

swankjesse:
Discussed offline! Verdict is to keep this warning and change the impacted callsite specifically.

PR #2427, opened 2022-06-30
Support marking service dependencies for the jetty service.

jdm-square:
This change makes it possible to address a long-standing shutdown issue—if a misk application has web action handlers that interact with a database, the misk service coordinator has no idea that the actions depend on the database transacter service being available. During shutdown, this can lead to the DB connection being closed while web actions are still in progress, triggering intermittent errors and risk of data loss.

Because the Jetty service is private to Misk, and the application DB is defined in the application, making the Jetty service aware of this dependency is convoluted. This PR makes it possible for the application to pass down a list of other services that the Jetty service should depend upon, which will enforce a consistent shutdown order.

### DIFF
--- a/misk/src/main/kotlin/misk/web/MiskWebModule.kt
+++ b/misk/src/main/kotlin/misk/web/MiskWebModule.kt
@@ -1,5 +1,7 @@
 package misk.web
 
+import com.google.common.util.concurrent.Service
+import com.google.inject.Key
 import com.google.inject.Provider
 import com.google.inject.Provides
 import com.google.inject.TypeLiteral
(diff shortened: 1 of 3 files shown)


misk/src/main/kotlin/misk/web/MiskWebModule.kt
```diff
@@ -86,12 +89,15 @@ import org.eclipse.jetty.util.thread.ExecutorThreadPool
 import org.eclipse.jetty.util.thread.QueuedThreadPool
 import org.eclipse.jetty.util.thread.ThreadPool
 
-class MiskWebModule(private val config: WebConfig) : KAbstractModule() {
+class MiskWebModule(
+  private val config: WebConfig,
+  private val jettyDependsOn: List<Key<out Service>> = emptyList(),
```
swankjesse, line 94: 👍🏻

swankjesse (approved):

swankjesse:
Yep, this is long overdue. Thanks for doing this.

PR #2428, opened 2022-07-01
Add envelope encryptor

choi-calvin:
This is a module to use envelope encryption with Tink and Cloud KMS. Currently, only encryption is supported and DEKs are generated using `AES256_GCM`.

### DIFF
--- a/misk-gcp/README.md
+++ b/misk-gcp/README.md
@@ -159,3 +159,6 @@ class GoogleSpannerModuleTest {
   }
 }
 ```
+
+## Envelope Encryption (Cloud KMS)
+See [README](src/main/kotlin/misk/cloud/gcp/kms/README.md).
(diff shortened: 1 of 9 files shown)


yoavamit:
I think this is a very interesting addition to the open source `misk` framework.

A few comments/question:

1. Right now, this module is GCP centric - the config only supports using a GCP KMS. WDYT about trying to open up this abstraction to make it KMS agnostic?
2. What's the main use case for using KMS envelope encryption? Usually, KMS is used to encrypt other encryption keys - it's not meant to be used in high QPS ops.
3. Personally, I'd start with adding this binding in the Square version - `skim`, and see if you can generalize it afterwards to this project.

choi-calvin:
Closing to move to Skim

PR #2431, opened 2022-07-06
Introduce an annotation for MiskCaller in tests

r3mariano:
This allows setting up an `ActionScoped<MiskCaller>` in tests conveniently.

Can't do something more generic because annotation classes are limited: https://kotlinlang.org/docs/annotations.html#constructors


staktrace (approved):
Nice!

PR #2449, opened 2022-07-19
adding getPermitsRemaining()

snigdhabanda:
A call to getPermitsRemaining() exposes the permits left to be consumed. For example, if you're rate limiting at 2 permits / sec with a permit count of 1L and a timeout period of 0s, you'd have 1 permit remaining after the first acquisition. If you increase the timeout period to 500ms, you'd have 2 permits remaining after the first acquisition. This functionality is salient for monitoring (ie. setting an alert once getPermitsRemaining() falls below a threshold).

### DIFF
--- a/misk-core/src/main/kotlin/misk/sampling/RateLimiter.kt
+++ b/misk-core/src/main/kotlin/misk/sampling/RateLimiter.kt
@@ -6,6 +6,8 @@ import java.time.Duration
 import java.util.concurrent.TimeUnit
 import java.util.concurrent.atomic.AtomicLong
 import javax.inject.Inject
+import kotlin.math.ceil
+import kotlin.math.floor
 
 /**
  * A deterministic testable rate limiter that uses two variables:
@@ -39,7 +41,7 @@ class RateLimiter private constructor(
   private val atomicAllocatedUntil = AtomicLong(ticker.read())
 
   /** The size of our window where we can borrow bytes from the future. */
-  private var windowSizeNs = TimeUnit.SECONDS.toNanos(1L)
+  private val windowSizeNs = TimeUnit.SECONDS.toNanos(1L)
 
   /**
    * Attempt to acquire [permitCount] permits, sleeping up to [timeout] if necessary for them to
@@ -81,20 +83,30 @@ class RateLimiter private constructor(
   ): Duration? {
     while (true) {
       val allocatedUntil = atomicAllocatedUntil.get()
+
       val permitsPerSecond = this.permitsPerSecond // Sample this volatile only once.
 
       val maxRequestSize = windowSizeNs.nanosToPermits(permitsPerSecond)
+      println(maxRequestSize)
       if (permitCount > maxRequestSize) return null
 
       val now = ticker.read()
+
       val timeoutNs = unit.toNanos(timeout)
 
       // If this acquire succeeds, this is the time we're consuming permits through.
       val newAllocatedUntil = maxOf(allocatedUntil, now) +
         permitCount.permitsToNanos(permitsPerSecond)
+      val permits = permitCount.permitsToNanos(permitsPerSecond)
+      println("now $now")
+      println("allocatedUntil $allocatedUntil")
+      println("permits $permits")
+      println("newAllocatedUntil $newAllocatedUntil")
 
       // We only sleep for permits until the beginning of our window.
       val sleepNs = newAllocatedUntil - now - windowSizeNs
+      println("sleepNs $sleepNs")
+      println("timeoutNs $timeoutNs")
 
       // We'd have to sleep too long for this number of permits. Fail fast.
       if (sleepNs > timeoutNs) return null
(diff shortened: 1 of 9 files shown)


misk-core/src/test/kotlin/misk/sampling/RateLimiterTest.kt
```diff
@@ -66,13 +66,13 @@ class RateLimiterTest {
...
-    assertThat(rateLimiter.tryAcquire(1L, 499, TimeUnit.MILLISECONDS)).isTrue()
+    assertThat(rateLimiter.tryAcquire(1L, 0, TimeUnit.MILLISECONDS)).isTrue()
     assertThat(ticker.nowMs).isEqualTo(0L)
 
-    assertThat(rateLimiter.tryAcquire(1L, 499, TimeUnit.MILLISECONDS)).isTrue()
+    assertThat(rateLimiter.tryAcquire(1L, 0, TimeUnit.MILLISECONDS)).isTrue()
     assertThat(ticker.nowMs).isEqualTo(0L)
 
-    assertThat(rateLimiter.tryAcquire(1L, 499, TimeUnit.MILLISECONDS)).isFalse()
+    assertThat(rateLimiter.tryAcquire(1L, 0, TimeUnit.MILLISECONDS)).isFalse()
```
ean5533, line 75: Why these changes?

misk-core/src/main/kotlin/misk/sampling/RateLimiter.kt
```diff
@@ -107,6 +107,45 @@ class RateLimiter private constructor(
...
+      println("windowSizeNs $windowSizeNs")
+      println("sleepNs $sleepNs")
+
+      val timeoutNs = unit.toNanos(timeout)
+
+      // We'd have to sleep too long for this number of permits. Fail fast.
+      if (sleepNs > timeoutNs) return 0L
+
+      // Try to consume permits! If atomicAllocatedUntil changed, we lost a race; loop to try again.
+      if (!atomicAllocatedUntil.compareAndSet(allocatedUntil, newAllocatedUntil)) continue
```
ean5533, line 136: When I saw the method name `getPermitsRemaining` I assumed that it just returned how many permits were remaining, but it looks like this method actually acquires permits (like `tryAcquire`) and then returns how many permits were remaining after acquiring. In other words, it looks like this method is a reimplementation of `tryAcquire`, but it just returns different information.

I think we should do one of two things:

1. Rename this method to something like `tryAcquire` (though we can't use that name exactly because it would conflict), and possibly deprecate the existing method.
2. Trim down this method significantly so that all it does is get how many permits are remaining, and has no side effects at all, including acquiring any permits.

The second option feels simpler. I imagine these methods be used by the caller like:

```
if(!rateLimiter.tryAcquire(...)) {
  throw RuntimeException("too busy, go away")
}

val permitsRemaining = rateLimiter.getPermitsRemaining()
metricRegistry.recordGauge("permits-remaining-percent", permitsRemaining/maxPermits)
```

On one hand, this would be subject to some race conditions, because two threads might take out two permits at the same time, and then both threads would record that there is `X` percent remaining. On the other hand, I don't think it really matters if this number is subject to race conditions; all anyone will typically care about is the `min` value that occurs in a given time period.

tl;dr I recommend chopping this method way down so that all it does is calculate the `permitsRemaining` and returns that.

misk-core/src/test/kotlin/misk/sampling/RateLimiterTest.kt
snigdhabanda: Reverted these

misk-core/src/test/kotlin/misk/sampling/RateLimiterTest.kt
```diff
@@ -66,53 +79,69 @@ class RateLimiterTest {
...
     assertThat(ticker.nowMs).isEqualTo(0L)
 
     // Exhausted.
+    assertThat(rateLimiter.getPermitsRemaining(TimeUnit.MILLISECONDS, 100)).isEqualTo(0L)
     assertThat(rateLimiter.tryAcquire(1L, 100, TimeUnit.MILLISECONDS)).isFalse()
     assertThat(ticker.nowMs).isEqualTo(0L)
 
     rateLimiter.permitsPerSecond = 10L
 
+    assertThat(rateLimiter.getPermitsRemaining(TimeUnit.MILLISECONDS, 100)).isEqualTo(1L)
```
snigdhabanda, line 118: This assertion is currently failing because we have reached the maxNs in our current bucket. However, since our rate limiting changed, we should be allowing extension into the next bucket of time. I know we decided to not factor in permitCount, but I think this is the case where it's needed.

misk-core/src/test/kotlin/misk/sampling/RateLimiterTest.kt
```diff
@@ -66,53 +79,69 @@ class RateLimiterTest {
...
     assertThat(rateLimiter.tryAcquire(1L, 100, TimeUnit.MILLISECONDS)).isTrue()
     assertThat(ticker.nowMs).isEqualTo(100L)
+    assertThat(rateLimiter.getPermitsRemaining(TimeUnit.MILLISECONDS, 100)).isEqualTo(0L)
   }
 
   @Test
   fun `rate limit decreases`() {
     rateLimiter.permitsPerSecond = 10L
 
+    assertThat(rateLimiter.getPermitsRemaining(TimeUnit.MILLISECONDS, 100)).isEqualTo(1L)
```
snigdhabanda, line 128: Similar to above, this assertion fails. I would think here that only permit is allowed, but the behavior is that 10 permits are actually allowed. I think taking permitCount into account would once again prevent this.

misk-core/src/test/kotlin/misk/sampling/RateLimiterTest.kt
ean5533: The `permitCount` parameter from `tryAcquire` represents how many permits the caller is asking for. This should not have any impact on the calculation of how many permits are available.

(disussed in another forum; we're changing the calculation to be simpler, but still not using `permitCount`)

misk-core/src/main/kotlin/misk/sampling/RateLimiter.kt
```diff
@@ -107,6 +108,33 @@ class RateLimiter private constructor(
     }
   }
 
+  /**
+   * Returns the permits remaining, given a time unit and timeout, after tryAcquire() has been invoked.
```
ean5533, line 112: I think this comment is a bit misleading. There's no requirement for this method to be called before or after `tryAcquire` has been invoked; it can be called any time, even if `tryAcquire` is never invoked.

Perhaps something like this?

> Returns the maximum number of permits that could have been acquired by a call to [tryAcquire], assuming the caller passed the same [timeout] and [unit].

ean5533 (approved):
One nit about the doc comment, but otherwise this LGTM.

I recommend getting a review from @isafan or @mightyguava if possible since they wrote/reviewed the original implementation. Otherwise, someone like @adrw or @swankjesse would be good.

misk-core/src/main/kotlin/misk/sampling/RateLimiter.kt
```diff
@@ -81,17 +81,18 @@ class RateLimiter private constructor(
...
       if (permitCount > maxRequestSize) return null
-
       val now = ticker.read()
+
       val timeoutNs = unit.toNanos(timeout)
 
       // If this acquire succeeds, this is the time we're consuming permits through.
-      val newAllocatedUntil = maxOf(allocatedUntil, now) +
-        permitCount.permitsToNanos(permitsPerSecond)
+      val newAllocatedUntil =
```
adrw, line 94: nit: I prefer the previous line break

misk-core/src/main/kotlin/misk/sampling/RateLimiter.kt
```diff
@@ -107,6 +108,35 @@ class RateLimiter private constructor(
     }
   }
 
+  /**
+   * Returns the maximum number of permits that could have
+   * been acquired by a call to [tryAcquire], assuming the caller
+   * passed the same [timeout] and [unit].
+   *
+   * @return the permits remaining.
```
adrw, line 116: nit: I think this `@return` line is redundant and implied by the function name and could be removed, thanks for the above description though!

misk-core/src/main/kotlin/misk/sampling/RateLimiter.kt
```diff
@@ -107,6 +108,35 @@ class RateLimiter private constructor(
...
+   */
+  fun getPermitsRemaining(
+    unit: TimeUnit,
+    timeout: Long
+  ): Long {
+    val allocatedUntil = atomicAllocatedUntil.get()
+    println("getPermitsRemaining.allocatedUntil: $allocatedUntil")
```
adrw, line 123: this should be removed before merge

adrw: (I'm assuming it is there solely for debugging during development and not intended to land in master)

adrw (approved):

PR #2456, opened 2022-07-26
extend MiskConfig api

mightyguava:
The old interface only allowed override files on the filesystem. The new interface uses the existing `classpath:` and `filesystem:` syntax to allow both jar resources and files to be passed in.

The new interface also adds a `JsonNode` option to allow applications to customize using data from other sources, like system properties.


misk/api/misk.api
```diff
@@ -750,8 +750,9 @@ public final class misk/config/MiskConfig {
 	public static final field INSTANCE Lmisk/config/MiskConfig;
 	public static final fun filesInDir (Ljava/lang/String;Ljava/io/FilenameFilter;)Ljava/util/List;
 	public static synthetic fun filesInDir$default (Ljava/lang/String;Ljava/io/FilenameFilter;ILjava/lang/Object;)Ljava/util/List;
-	public final fun flattenYamlMap (Ljava/util/Map;)Lcom/fasterxml/jackson/databind/JsonNode;
```
mightyguava, line 753: This shouldn't have been public API and there don't seem to be usages.

staktrace (approved):

PR #2459, opened 2022-07-26
Resolving divide by 0 error 

snigdhabanda:
Adding in error handling to `getPermitsRemaining()` in the case that a user sets the allowed `qps` to 0.

### DIFF
--- a/misk-core/src/main/kotlin/misk/sampling/RateLimiter.kt
+++ b/misk-core/src/main/kotlin/misk/sampling/RateLimiter.kt
@@ -118,7 +118,7 @@ class RateLimiter private constructor(
     unit: TimeUnit,
     timeout: Long
   ): Long {
-    if (permitsPerSecond == 0L) return 0L
+    if (permitsPerSecond <= 0L) return 0L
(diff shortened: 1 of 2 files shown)


ean5533 (approved):
What, you don’t like having `NaN` permits remaining?

snigdhabanda:
> What, you don’t like having `NaN` permits remaining?

bwamp bwamp bwamp

ean5533 (approved):

adrw (approved):

misk-core/src/main/kotlin/misk/sampling/RateLimiter.kt
```diff
@@ -118,6 +118,7 @@ class RateLimiter private constructor(
     unit: TimeUnit,
     timeout: Long
   ): Long {
+    if (permitsPerSecond == 0L) return 0L
```
tyiu, line 121: I think we could also short-circuit on negatives to make it a bit easier to understand on those edge cases. i.e.

```kotlin
if (permitsPerSecond <= 0L) return 0L
```

Tests for that scenario would be good as well.

PR #2474, opened 2022-08-04
DontRetryException takes a cause exeception

acostama:
The retry handler will rethrow the DontRetryException and as is loses all context of what the throw exception was

### DIFF
--- a/misk-core/api/misk-core.api
+++ b/misk-core/api/misk-core.api
@@ -5,6 +5,7 @@ public abstract interface class misk/backoff/Backoff {
 
 public final class misk/backoff/DontRetryException : java/lang/Exception {
 	public fun <init> (Ljava/lang/String;)V
+	public fun <init> (Ljava/lang/String;Ljava/lang/Exception;)V
 }
 
 public final class misk/backoff/ExponentialBackoff : misk/backoff/Backoff {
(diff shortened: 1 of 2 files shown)


ean5533 (approved):

jdm-square (approved):

staktrace (approved):
This is going to be source-compatible but binary-incompatible with existing usages. So if a downstream consumer is compiled against a version of misk without this change but picks up a new version at runtime it's going to cause an exception. Please make sure the balance-mover and sc-jobqueue repos (which both create `DontRetryException`) get updated to the latest version after this is merged through.

acostama:
> This is going to be source-compatible but binary-incompatible with existing usages. So if a downstream consumer is compiled against a version of misk without this change but picks up a new version at runtime it's going to cause an exception. Please make sure the balance-mover and sc-jobqueue repos (which both create `DontRetryException`) get updated to the latest version after this is merged through.

@staktrace is it worth explicitly overloading instead?

ean5533:
We could just `class DontRetryException @JvmOverloads constructor(message: String, exception: Exception? = null)` if we want to preserve binary compatibility

acostama:
ah did an explicit overload but just saw @ean5533's comment no preference either way for me

ean5533:
I think using `@JvmOverloads` would be slightly more idiomatic (certainly less code), but I don't think it's a big deal either way.

PR #2475, opened 2022-08-05
feat(misk-policy-testing): Add version argument for OPA image.

kerrnel-cash:
The OPA image pulled for testing may not match the version in production,
leading to behavior mismatches. This adds a constructor argument to specify
the image version, which defaults to latest-debug.

Refs: PRODSQUAD-153

### DIFF
--- a/misk-policy-testing/api/misk-policy-testing.api
+++ b/misk-policy-testing/api/misk-policy-testing.api
@@ -14,9 +14,9 @@ public final class misk/policy/opa/LocalOpaService : com/google/common/util/conc
 	public static final field Companion Lmisk/policy/opa/LocalOpaService$Companion;
 	public static final field DEFAULT_POLICY_DIRECTORY Ljava/lang/String;
 	public static final field OPA_CONTAINER_NAME Ljava/lang/String;
-	public static final field OPA_DOCKER_IMAGE Ljava/lang/String;
+	public static final field OPA_DOCKER_IMAGE_BASE Ljava/lang/String;
 	public static final field OPA_EXPOSED_PORT I
-	public fun <init> (Ljava/lang/String;Z)V
+	public fun <init> (Ljava/lang/String;ZLjava/lang/String;)V
 }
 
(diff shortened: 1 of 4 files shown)


cashapp-yzaslavsky:
Nice, so  how can I use this to provide a different usage? what should i do as a toolbox developer who would like to test their rego files?

kerrnel-cash:
> 

Can you define what you mean by usage? To test your rego files you put them into services/src/policy, right?

misk-policy-testing/api/misk-policy-testing.api
```diff
@@ -30,7 +30,7 @@ public final class misk/policy/opa/LocalOpaService$Companion {
 
 public final class misk/policy/opa/OpaDevelopmentModule : misk/inject/KAbstractModule {
 	public fun <init> ()V
-	public fun <init> (Ljava/lang/String;Z)V
-	public synthetic fun <init> (Ljava/lang/String;ZILkotlin/jvm/internal/DefaultConstructorMarker;)V
+	public fun <init> (Ljava/lang/String;ZLjava/lang/String;)V
+	public synthetic fun <init> (Ljava/lang/String;ZLjava/lang/String;ILkotlin/jvm/internal/DefaultConstructorMarker;)V
```
r3mariano, line 34: This is likely to cause backwards compatibility problems downstream ([there's an overly long explanation here](https://lankydan.dev/backwards-compatibility-of-kotlins-default-arguments), but in short, it's because of how this compiles down to JVM). Suggest adding overloads with the old signatures temporarily.

r3mariano (approved):

misk-policy-testing/api/misk-policy-testing.api
kerrnel-cash: Thank you, I will have a read and update this PR.

kerrnel-cash: Done.

PR #2488, opened 2022-08-17
Proposal to have bounded + balanced sqs receivers

acostama:
There have been many issues and discussions resulting from the poor tradeoff the sqs feature flags give.

One flag leads to unbalanced pods and the other risks self DOS if the auto scaler kicks in with an sqs backlog

This change proposes a simple way to use both flags to get balanced pods while still having a max upper bound

- [x] todo add test once the idea is ✅

### DIFF
--- a/misk-aws/src/main/kotlin/misk/jobqueue/sqs/AwsSqsJobQueueConfig.kt
+++ b/misk-aws/src/main/kotlin/misk/jobqueue/sqs/AwsSqsJobQueueConfig.kt
@@ -1,5 +1,8 @@
 package misk.jobqueue.sqs
 
+import misk.jobqueue.sqs.SqsJobConsumer.Companion.CONSUMERS_PER_QUEUE
+import misk.jobqueue.sqs.SqsJobConsumer.Companion.POD_CONSUMERS_PER_QUEUE
+import misk.jobqueue.sqs.SqsJobConsumer.Companion.POD_MAX_JOBQUEUE_CONSUMERS
 import misk.tasks.RepeatedTaskQueueConfig
 import wisp.config.Config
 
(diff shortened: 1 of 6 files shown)


misk-aws/src/main/kotlin/misk/jobqueue/sqs/AwsSqsJobQueueConfig.kt
```diff
@@ -52,5 +54,61 @@ class AwsSqsJobQueueConfig(
...
+  val aws_sqs_job_receiver_policy: AwsSqsJobReceiverPolicy = AwsSqsJobReceiverPolicy.ONE_FLAG_ONLY
 ) : Config
+
+
+/**
+ * AWS SQS consumers are spun by each of a service's pods. Each pod is responsible for running some
+ * number on consumers and making sure they are within the parameters of the feature-flags used.
+ *
+ * Which flags?
+ *  * [CONSUMERS_PER_QUEUE]: This flag specifies the total number of consumers across the entire
```
celr, line 68: Maybe it will be too much work to migrate everyone and it's outside of the scope of this PR, but should these feature flags really be config values?

celr (approved):

misk-aws/src/main/kotlin/misk/jobqueue/sqs/AwsSqsJobQueueConfig.kt
```diff
@@ -52,5 +54,61 @@ class AwsSqsJobQueueConfig(
...
+ */
+enum class AwsSqsJobReceiverPolicy {
+  /**
+   * This is the original policy. Naming is hard, but this policy will compute receivers as follows.
+   * First we choose one flag. If there is a configuration in [POD_CONSUMERS_PER_QUEUE], choose that
+   * flag; otherwise choose the [CONSUMERS_PER_QUEUE] flag.
+   *
+   * If the [POD_CONSUMERS_PER_QUEUE] is chosen, ALL pods will spin up the configured number of
+   * consumers. Imagine the flag is configured for 5 consumers, then
+   *  5 pods => 25 sqs consumers
```
ankitm111, line 83: is there a reason why we cannot impose a "default" max per pod and then if a service wants to increase it, they do it as an explicit action ? I know that the way to arrive at that number though.

acostama: yeah this is a good point

I think the part I want to get right is how we make sure we don't break existing things by keeping things backwards compatible and so I didn't add any defaults to the existing one.

But good catch that we should add a default given that folks might use the BALANCED_MAX policy even if the feature flag is not set. Like you said the tricky thing is finding that number, but it would reduce the risk + burden of using this policy. (my first choice for defaults is 1 or 5, ie smallest positive number or something that feels reasonable for most deployments based on "anecdata")


We could choose to avoid this config altogether and add a default and just replace the receiver algo with the new policy. I would look to the @misk-maintainers for guidance here since changing the current behavior might be risky.

acostama: - [ ] handle the case where the per pod flag is configured with a non-positive number, this would make the BALANCED_MAX policy easier to use, especially since that flags default is <=0 which means that sqs wouldn't run at all)

misk-aws/src/main/kotlin/misk/jobqueue/sqs/AwsSqsJobQueueConfig.kt
acostama: I could see it going both ways, though the feature flag has been nice to dynamically make changes without deploy 🤷

misk-aws/src/main/kotlin/misk/jobqueue/sqs/SqsJobConsumer.kt
```diff
@@ -118,7 +125,29 @@ internal class SqsJobConsumer @Inject internal constructor(
...
+      // Read the global max from the per queue flag.
+      val maxGlobal = receiversForQueue()
+
+      var result = 0
+      for (candidate in 1..maxGlobal) {
+        // Don't exceed the per pod max.
+        if (result >= maxPerPod) break
+
+        // Use the lease to enforce global max.
+        val lease = leaseManager.requestLease("sqs-job-consumer-${queue.name.value}-$candidate")
```
mhickman, line 144: super nit: I'd move DRY up the lease name definition that's also on L122

mhickman (approved):

misk-aws/src/main/kotlin/misk/jobqueue/sqs/SqsJobConsumer.kt
```diff
@@ -118,7 +125,29 @@ internal class SqsJobConsumer @Inject internal constructor(
           } else {
             true
           }
+        }.size
+
+    }
+
+    private fun balancedMaxReceivers(): Int {
+      // Read the max from the per pod flag.
+      val maxPerPod = receiversPerPodForQueue()
```
mhickman, line 134: slight nit - I think this is a _slightly_ different meaning than it's previous interpretation.

I wouldn't be opposed to introducing a third config flag for "max per queue" that is read if and only if we're using `BALANCED_MAX` as a config value.

acostama: added

mhickman (approved):
Still LGTM! :shipit:

PR #2490, opened 2022-08-19
Expose configuration for jetty HTTP header options

sfavron:
We're dealing with a problem where our request headers have grown larger
than jetty's default will accept. For affected users, this causes this
error:

upstream connect error or disconnect/reset before headers. reset reason:
remote refused stream reset

We're taking other steps to reduce our header size, but it will take a
while. In the meantime, if we expose these jetty configuration options,
we can increase the max allowed header sizes and get things back to a
working state until we can fix the root cause.


jdm-square (approved):
This looks like a reasonable configuration knob to expose!

PR #2496, opened 2022-08-23
Do not propagate scoped data in ActionScope if propagating to the sam…

tyiu:
…e thread as the original caller and it is already in scope

This is technically a bug. If a `DirectExecutor` is used, an exception would be thrown due to `java.lang.IllegalStateException: cannot begin an ActionScope on a thread that is already running in an action scope`

### DIFF
--- a/misk-action-scopes/src/main/kotlin/misk/scope/ActionScope.kt
+++ b/misk-action-scopes/src/main/kotlin/misk/scope/ActionScope.kt
@@ -2,6 +2,7 @@ package misk.scope
 
 import com.google.inject.Key
 import com.google.inject.Provider
+import java.util.UUID
 import java.util.concurrent.Callable
 import javax.inject.Inject
 import javax.inject.Singleton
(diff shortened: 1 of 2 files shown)


misk-action-scopes/src/main/kotlin/misk/scope/ActionScope.kt
```diff
@@ -47,10 +47,18 @@ class ActionScope @Inject internal constructor(
...
     check(tls.get() != null) { "not running within an ActionScope" }
 
+    val currentThreadId = Thread.currentThread().id
     val currentScopedData = tls.get().toMap()
+
     return Callable {
-      enter(currentScopedData).use {
+      // If the original thread is the same as the thread that calls the Callable and we are already
+      // in scope, then there is no need to re-enter the scope.
+      if (currentThreadId == Thread.currentThread().id && inScope()) {
```
cconroy, line 56: technically thread IDs can be reused which would make this check incorrect.

> When a thread is terminated, this thread ID may be reused.

https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/Thread.html#getId()

cconroy: Perhaps instead create some UUID and store in a threadlocal and compare against _that_...or perhaps `inScope()` is a sufficient check (not sure, but worth asking)

tyiu: You raised a good point. I replaced thread id comparison with a scoped seed data comparison. If the seed data is equal across executions, we don't attempt to re-enter the scope. If the seed data differs, it assumes that you're attempting to enter a new action scope, so we'll let it fail (by calling `enter` in an illegal state).

misk-action-scopes/src/main/kotlin/misk/scope/ActionScope.kt
```diff
@@ -25,7 +25,7 @@ class ActionScope @Inject internal constructor(
 
   /** Starts the scope on a thread with the provided seed data */
   fun enter(seedData: Map<Key<*>, Any?>): ActionScope {
-    check(tls.get() == null) {
+    check(!inScope()) {
```
tyiu, line 28: These changes are functionally equivalent. I just replaced it so that it's easier to read and understand.

misk-action-scopes/src/main/kotlin/misk/scope/ActionScope.kt
tyiu: Never mind. I got feedback your idea of UUID is probably the best approach:
https://cash.slack.com/archives/C01M5M297T7/p1661312203488089?thread_ts=1661304731.766609&cid=C01M5M297T7

r3mariano (approved):
Nice!

misk-action-scopes/src/main/kotlin/misk/scope/ActionScope.kt
```diff
@@ -20,37 +21,58 @@ class ActionScope @Inject internal constructor(
...
+    val currentThreadUUID = threadLocalUUID.get()
 
-    val currentScopedData = tls.get().toMap()
     return Callable {
-      enter(currentScopedData).use {
-        c.call()
+      // If the original thread is the same as the thread that calls the Callable and we are already
+      // in scope, then there is no need to re-enter the scope.
+      if (inScope() && currentThreadUUID == threadLocalUUID.get()) {
+        return@Callable c.call()
```
tyiu, line 71: Forgot to clean this up while I was debugging. It should be:
```kotlin
if (inScope() && currentThreadUUID == threadLocalUUID.get()) {
  c.call()
} else {
  enter(currentScopedData).use {
    c.call()
  }
}
```

I'll add an additional commit.

misk-action-scopes/src/main/kotlin/misk/scope/ActionScope.kt
```diff
@@ -20,37 +21,58 @@ class ActionScope @Inject internal constructor(
   private val providers: @JvmSuppressWildcards Map<Key<*>, Provider<ActionScopedProvider<*>>>
 ) : AutoCloseable {
   companion object {
-    private val tls = ThreadLocal<LinkedHashMap<Key<*>, Any?>>()
+    private val threadLocalScope = ThreadLocal<LinkedHashMap<Key<*>, Any?>>()
+    private val threadLocalUUID = ThreadLocal<UUID>()
```
swankjesse, line 25: Rather than having two ThreadLocals, you could create a Key for the UUID and put it in the map?

cconroy:
Thanks @tyiu!

PR #2513, opened 2022-09-07
Update setnx to return boolean so caller knows if key was set

shanegautreau:
Set if not exist (setnx) should return a result that can be used to indicate if the key existed or not, before the set call. Redis returns a 1, when the was set, and 0, otherwise. These integers can be coerced into booleans for the caller.

### DIFF
--- a/misk-redis/src/main/kotlin/misk/redis/FakeRedis.kt
+++ b/misk-redis/src/main/kotlin/misk/redis/FakeRedis.kt
@@ -158,21 +158,21 @@ class FakeRedis : Redis {
     }
   }
 
-  override fun setnx(key: String, value: ByteString) {
-    synchronized(lock) {
+  override fun setnx(key: String, value: ByteString): Boolean {
+    return synchronized(lock) {
       keyValueStore.putIfAbsent(key, Value(
         data = value,
         expiryInstant = Instant.MAX
-      ))
+      )) == null
     }
   }
 
-  override fun setnx(key: String, expiryDuration: Duration, value: ByteString) {
-    synchronized(lock) {
+  override fun setnx(key: String, expiryDuration: Duration, value: ByteString): Boolean {
+    return synchronized(lock) {
       keyValueStore.putIfAbsent(key, Value(
         data = value,
         expiryInstant = clock.instant().plusSeconds(expiryDuration.seconds)
-      ))
+      )) == null
     }
   }
 
--- a/misk-redis/src/main/kotlin/misk/redis/RealRedis.kt
+++ b/misk-redis/src/main/kotlin/misk/redis/RealRedis.kt
@@ -101,14 +101,14 @@ class RealRedis(private val jedisPool: JedisPool) : Redis {
     }
   }
 
-  /** Set a ByteArray value if it doesn't already exist. Returns 1 if set and 0, otherwise */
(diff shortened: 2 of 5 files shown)


misk-redis/src/main/kotlin/misk/redis/RealRedis.kt
```diff
@@ -101,18 +101,19 @@ class RealRedis(private val jedisPool: JedisPool) : Redis {
...
-  /** Set a ByteArray value if it doesn't already exist with an expiration. */
-  override fun setnx(key: String, expiryDuration: Duration, value: ByteString) {
-    val setParams = SetParams.setParams().ex(expiryDuration.seconds).nx()
-    jedisPool.resource.use { jedis ->
-      jedis.set(key.toByteArray(charset), value.toByteArray(), setParams)
+  /** Set a ByteArray value if it doesn't already exist with an expiration. Returns 1 if set and 0, otherwise  */
+  override fun setnx(key: String, expiryDuration: Duration, value: ByteString): Boolean {
+    return jedisPool.resource.use { jedis ->
+      jedis.setnx(key.toByteArray(charset), value.toByteArray())
+        .also { setResult -> if (setResult == 1L) jedis.expire(key, expiryDuration.seconds) } == 1L
```
shanegautreau, line 115: jedis doesn't have a setnx with expiry so I had to make 2 commands

misk-redis/src/main/kotlin/misk/redis/FakeRedis.kt
```diff
@@ -158,21 +158,21 @@ class FakeRedis : Redis {
     }
   }
 
-  override fun setnx(key: String, value: ByteString) {
-    synchronized(lock) {
+  override fun setnx(key: String, value: ByteString): Boolean {
```
swankjesse, line 161: Should this return 0 or 1 for consistency with the Redis docs?

misk-redis/src/main/kotlin/misk/redis/RealRedis.kt
```diff
@@ -101,18 +101,19 @@ class RealRedis(private val jedisPool: JedisPool) : Redis {
     }
   }
 
-  /** Set a ByteArray value if it doesn't already exist. */
-  override fun setnx(key: String, value: ByteString) {
-    jedisPool.resource.use { jedis ->
-      jedis.setnx(key.toByteArray(charset), value.toByteArray())
+  /** Set a ByteArray value if it doesn't already exist. Returns 1 if set and 0, otherwise */
```
swankjesse, line 104: This is not a boolean

misk-redis/src/main/kotlin/misk/redis/RealRedis.kt
swankjesse: ugh, so annoying. But I like this better than what we had.

misk-redis/src/main/kotlin/misk/redis/FakeRedis.kt
shanegautreau: I originally had that but noticed the wrapper uses boolean for other redis commands that return 0 or 1

swankjesse (approved):

swankjesse:
Nice!

I recommend either changing the docs to explain what True and False do, or changing the code to return an integer.

PR #2516, opened 2022-09-08
fix(misk-security): Clean up ContentSecurityPolicy (CSP) Code.

kerrnel-cash:
This expands Csp into ContentSecurityPolicy to avoid acronyms in code, and adds a proper javadoc comment to the annotation.


yoavamit (approved):
lgtm!

PR #2519, opened 2022-09-16
Plumb through dependencies to CronTask from CronModule

iwismer:
This adds the dependencies supplied to CronModule to CronTask to ensure that the modules that CronTask depends on are started up before CronTask starts.

### DIFF
--- a/misk-cron/src/main/kotlin/misk/cron/CronModule.kt
+++ b/misk-cron/src/main/kotlin/misk/cron/CronModule.kt
@@ -22,7 +22,12 @@ class CronModule(
   override fun configure() {
     install(FakeCronModule(zoneId, threadPoolSize, dependencies))
     install(ServiceModule<RepeatedTaskQueue>(ForMiskCron::class))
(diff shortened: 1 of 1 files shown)


misk-cron/src/main/kotlin/misk/cron/CronModule.kt
```diff
@@ -22,7 +22,12 @@ class CronModule(
   override fun configure() {
     install(FakeCronModule(zoneId, threadPoolSize, dependencies))
     install(ServiceModule<RepeatedTaskQueue>(ForMiskCron::class))
-    install(ServiceModule<CronTask>())
+    install(ServiceModule<CronTask>().apply {
+      dependsOn(RepeatedTaskQueue::class.toKey())
```
keeferrourke, line 26: almost:

```suggestion
      dependsOn(keyOf<RepeatedTaskQueue>(ForMiskCron::class))
```

keeferrourke (approved):

PR #2525, opened 2022-09-20
Check cluster weights in MiskCron leases to support other LeaseManager implementations

iwismer:
With the changes from zookeeper to SkimDynamoDB leases, cluster weight checks are no longer part of LeaseManager, and needs to be done by the caller. These changes ensure that the cron jobs always run on the active cluster.

### DIFF
--- a/misk-cron/src/main/kotlin/misk/cron/CronTask.kt
+++ b/misk-cron/src/main/kotlin/misk/cron/CronTask.kt
@@ -23,7 +23,7 @@ internal class CronTask @Inject constructor() : AbstractIdleService() {
     var lastRun = clock.instant()
     taskQueue.scheduleWithBackoff(INTERVAL) {
       if (clusterWeight.get() > 0) {
-        Status.OK
+        return@scheduleWithBackoff Status.OK
       }
       val lease = leaseManager.requestLease(CRON_CLUSTER_LEASE_NAME)
 
--- a/misk-cron/src/main/kotlin/misk/cron/CronTask.kt
+++ b/misk-cron/src/main/kotlin/misk/cron/CronTask.kt
@@ -1,6 +1,7 @@
 package misk.cron
 
 import com.google.common.util.concurrent.AbstractIdleService
+import misk.clustering.weights.ClusterWeightProvider
 import misk.tasks.RepeatedTaskQueue
 import misk.tasks.Status
 import wisp.lease.LeaseManager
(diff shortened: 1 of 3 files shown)


misk-cron/src/main/kotlin/misk/cron/CronTask.kt
```diff
@@ -16,14 +17,18 @@ internal class CronTask @Inject constructor() : AbstractIdleService() {
...
   @Inject @ForMiskCron private lateinit var taskQueue: RepeatedTaskQueue
+  @Inject private lateinit var clusterWeight: ClusterWeightProvider
 
   override fun startUp() {
-    val lease = leaseManager.requestLease(CRON_CLUSTER_LEASE_NAME)
     var lastRun = clock.instant()
+    taskQueue.scheduleWithBackoff(INTERVAL) {
+      if (clusterWeight.get() > 0) {
+        Status.OK
+      }
```
keeferrourke, line 27: I think you need to explicitly return from this lambda, or put everything else below this in an `else` block?

keeferrourke (commented):
could we get a test added for this?

keeferrourke:
> With the changes from zookeeper to SkimDynamoDB leases, cluster weight checks are no longer part of LeaseManager, and needs to be done by the caller. These changes ensure that the cron jobs always run on the active cluster.

I think this raises a broader question about what the default misk LeaseManager contract should be. If a LeaseManager _isn't_ responsible for checking cluster weights, then perhaps the default zk impl should have this functionality removed?

Or if it _is_ part of the contract, then maybe our private implementations _should_ check cluster weights?

iwismer:
> could we get a test added for this?

Yup! I'm planning on writing one, but I wanted to make sure this looked ok before figuring out how to test this, since none of the current tests cover this class right now.

iwismer:
> I think this raises a broader question about what the default misk LeaseManager contract should be. If a LeaseManager _isn't_ responsible for checking cluster weights, then perhaps the default zk impl should have this functionality removed?
> 
> Or if it _is_ part of the contract, then maybe our private implementations _should_ check cluster weights?

I think this is still a little up in the air, but initially it looks like the idea is that zookeeper is being deprecated and its replacement will not include cluster weights. A little disappointing, especially since this was not communicated well.

keeferrourke:
>I think this is still a little up in the air, but initially it looks like the idea is that zookeeper is being deprecated and its replacement will not include cluster weights.

Oh! I was talking about from the OSS misk framework perspective. I just think we want to be clear what the API -> Impl contract is, since we have inconsistent cluster weight management across implementations (private to Cash App, vs public to Misk [ZK]).

misk-cron/src/main/kotlin/misk/cron/CronTask.kt
chris-ryan-square: Agreed, it will need the return or else block.

keeferrourke (approved):

PR #2527, opened 2022-09-22
Add allow_public_key_retrieval to DataSourceConfig for MySQL8

adrw:
For MySQL8 and above, when TLS is unavailable then an RSA key pair exchange is attempted to provide secure communication of authentication details instead of as plaintext in the JDBC url. Within secured subnetworks, the added `allow_public_key_retrieval` config parameter gives service owners the flexibility to disable the restrictive default RSA fallback functionality.

See the description in these docs for more information: https://mysqlconnector.net/troubleshooting/retrieval-public-key/

I also added a generic parameters section to config to provide easier advanced configuration without waiting for Misk changes, I'm open to dropping the `allow_public_key_retrieval` section in favour of the more flexible `jdbc_url_query_parameters` field if that is preferred.

### DIFF
--- a/misk-jdbc/src/main/kotlin/misk/jdbc/DataSourceConfig.kt
+++ b/misk-jdbc/src/main/kotlin/misk/jdbc/DataSourceConfig.kt
@@ -80,7 +80,9 @@ data class DataSourceConfig(
   // MySQL 8+ by default requires authentication via RSA keys if TLS is unavailable.
   // Within secured subnets, overriding to true can be acceptable.
   // See https://mysqlconnector.net/troubleshooting/retrieval-public-key/
-  val allow_public_key_retrieval: Boolean = false
+  val allow_public_key_retrieval: Boolean = false,
+  // Allow setting additional JDBC url parameters for advanced configuration
+  val jdbc_url_query_parameters: Map<String, Any> = mapOf()
 ) {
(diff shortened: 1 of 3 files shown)


misk-jdbc/src/main/kotlin/misk/jdbc/DataSourceConfig.kt
```diff
@@ -76,7 +76,13 @@ data class DataSourceConfig(
...
   // Consider using this if you want Hibernate to automagically batch inserts/updates when it can.
   val jdbc_statement_batch_size: Int? = null,
-  val use_fixed_pool_size: Boolean = false
+  val use_fixed_pool_size: Boolean = false,
+  // MySQL 8+ by default requires authentication via RSA keys if TLS is unavailable.
+  // Within secured subnets, overriding to true can be acceptable.
+  // See https://mysqlconnector.net/troubleshooting/retrieval-public-key/
+  val allow_public_key_retrieval: Boolean = false,
+  // Allow setting additional JDBC url parameters for advanced configuration
+  val jdbc_url_query_parameters: Map<String, Any> = mapOf()
```
shellderp, line 85: yea I like this, makes it easy to change without having to send a misk PR, but you can always add in params after using them a while

shellderp (approved):

PR #2528, opened 2022-09-22
Include stack traces in exceptional gRPC responses

jdm-square:
Right now, when service A uses gRPC to service B and service B throws an exception, there are two main outcomes:
* service B throws a WebActionException—service A receives a GrpcException with the mapped status code and service B's exception's public message
* service B throws some other exception (e.g. NullPointerException)—service A receives a GrpcException with the code UNKNOWN and the message "internal server error"

Compared to non-gRPC responses, this is very limited information. Services that communicate over HTTP instead often receive a response body that includes a stack trace, which makes it much easier to triage new Sentry reports without having to find the corresponding error report in the callee service.

This PR extends the message provided to the gRPC response when misk intercepts an exception—we now explicitly including the stack trace from the caught exception. Additionally, gRPC messages triggered by the second case now include the exception message, rather than a catch-all "internal server error" string.

### DIFF
--- a/misk-grpc-tests/src/test/kotlin/misk/grpc/MiskClientMiskServerTest.kt
+++ b/misk-grpc-tests/src/test/kotlin/misk/grpc/MiskClientMiskServerTest.kt
@@ -130,7 +130,7 @@ class MiskClientMiskServerTest {
       val e = assertFailsWith<GrpcException> {
         routeGuide.GetFeature().execute(point)
       }
-      assertThat(e.grpcMessage).isEqualTo("unexpected latitude error!")
+      assertThat(e.grpcMessage).startsWith("unexpected latitude error!")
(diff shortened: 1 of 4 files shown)


jdm-square:
One thing that I was worried about:
`14:52:01.717 [jetty-thread-33] WARN org.eclipse.jetty.http2.hpack.HpackEncoder - Header list size too large 17025 > 8192`
Comes from `crossStatusError` in `ProtocClientMiskServerTest`.

swankjesse:
Discussed offline. My recommendation is to use an interceptor closer to the application layer, and have that interceptor wrap the application-layer exception in a WebActionException. That interceptor can promote diagnostic information like the stacktrace into the `message` field _if_ the caller is trusted, which it learns through some side-channel. Perhaps by injecting an `ActionScoped<MiskCaller>`.

PR #2529, opened 2022-09-23
Fix release gh action

adrw:


.github/workflows/Release.yml
```diff
@@ -19,7 +19,8 @@ jobs:
       - name: Configure JDK
         uses: actions/setup-java@v3
         with:
-          java-version: 11
+          distribution: 'adopt'
```
adrw, line 22: Release step was failing because this new required parameter wasn't set (it was added to the standard build/test gh action but wasn't here)

shellderp (approved):

PR #2531, opened 2022-09-26
Suppress duplicate exception logs for gRPC errors.

jdm-square:
Right now we get three log identical messages every time a thrown exception is turned into a gRPC response. This is needlessly confusing!


swankjesse (approved):

misk/src/main/kotlin/misk/web/exceptions/ExceptionHandlingInterceptor.kt
```diff
@@ -47,11 +47,14 @@ class ExceptionHandlingInterceptor(
     try {
       chain.proceed(chain.httpCall)
     } catch (th: Throwable) {
-      val response = toResponse(th)
       try {
         if (chain.httpCall.dispatchMechanism == DispatchMechanism.GRPC) {
+          // This response object is only used for determining the status code. toGrpcResponse
+          // will provide a more useful log instead.
+          val response = toResponse(th, suppressLog = true)
```
frojasg, line 54: the `GrpcErrorResponse` also has a status code. I wonder if we should use that instead so we avoid the `suppressLog` param.

jdm-square: It's a gRPC status code, so we'll need to convert it back to an HTTP response code. I'd rather not go through multiple layers of translation (since it was originally derived from an HTTP status code), due to the risk of losing specificity.

frojasg (approved):
LGTM

PR #2533, opened 2022-09-27
Support disabling LogRequestResponse by env

ean5533:


misk/src/test/kotlin/misk/web/WebSocketsTest.kt
```diff
@@ -73,4 +90,20 @@ class EchoWebSocket @Inject constructor() : WebAction {
       override fun toString() = "EchoListener"
     }
   }
+
+  @ConnectWebSocket("/echo-logging-disabled-by-env")
+  @LogRequestResponse(
+    bodySampling = 1.0,
+    errorBodySampling = 1.0,
+    disabledEnvironments = ["testing"]
```
r3mariano, line 98: I'd rather:
a) use `excludeDeployments`, but that's really a nit.
b) use an `Array<Deployment>` so it's typed.

ean5533: > use an Array<Deployment> so it's typed.

Annotations can only have primitive type params. (Also we'd have to add a dependency on wisp-deployments, though maybe that would be fine)

> use excludeDeployments, but that's really a nit.

Do you still feel that way even if the type remains a string? I would find it confusing, personally.

r3mariano: > As for deployments vs environments: do you still feel that way even if the type remains a string? I would find it confusing, personally.

I'd keep environments given we can't use the type.

r3mariano (approved):

PR #2537, opened 2022-10-03
Seed BoundAction initialSeedData with matched Action

JGulbronson:
This will allow users to inject `ActionScoped<Action>`, to get access to things like the current functions annotations inside of an `ActionScopedProvider`, the return type, etc.

No test changes because I think the mechanism of `initialSeedData` is already pretty well tested, but let me know if you disagree and I can look to add one.


frojasg (approved):
LGTM

PR #2538, opened 2022-10-03
Adds KInstallOnceModule

cconroy:
Allows for de-duped installation of a module from multiple sites. This is useful for core/library dependencies that might need to ensure installation in varying configurations.

### DIFF
--- a/misk-inject/api/misk-inject.api
+++ b/misk-inject/api/misk-inject.api
@@ -49,3 +49,9 @@ protected final class misk/inject/KAbstractModule$KotlinAnnotatedBindingBuilder 
 	public fun toProvider (Ljavax/inject/Provider;)Lcom/google/inject/binder/ScopedBindingBuilder;
 }
 
+public abstract class misk/inject/KInstallOnceModule : misk/inject/KAbstractModule {
+	public fun <init> ()V
+	public final fun equals (Ljava/lang/Object;)Z
+	public final fun hashCode ()I
+}
+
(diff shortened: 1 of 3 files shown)


swankjesse (approved):
Guice de-dupes most DSL bindings now automatically, though it can’t for `@Provides` functions so this is still great to have.

frojasg:
Is there any gotcha we should considering using this function? Is there a reason why this is not included in guice?

frojasg (approved):

cconroy:
> Is there any gotcha we should considering using this function? Is there a reason why this is not included in guice?

IMHO it should be included (or it should just support de-duping wholesale). TIL from Jesse's comment that guice will de-dupe some bindings, but historically it wouldn't de-dupe at all. In theory the canonical advice is "restructure your app to only bind each thing exactly once", but for library code especially that can become really unwieldy and cumbersome.

We've used this pattern widely for many years in an internal application framework without issue.

keeferrourke:
I'm curious, is using this functionally different than writing your module as an `object MyModule: KAbstractModule()`?

cconroy:
> I'm curious, is using this functionally different than writing your module as an `object MyModule: KAbstractModule()`?

@keeferrourke the difference is that a `KInstallOnceModule` can be safely installed multiple times (it will only be added once, and its `configure` method will only run once). If you install a subclass of `KAbstractModule` `N` times, it will be added `N` times, and there will be `N` calls to its configure method (one for each instance installed)

PR #2540, opened 2022-10-05
FakeClockModule install-once

cconroy:
Use `KInstallOnceModule` with `FakeClockModule` so that multiple testing components can safely install this module to ensure a fake clock is wired up.


cconroy:
cc: @snigdhabanda

snigdhabanda (approved):

PR #2547, opened 2022-10-18
Revert "Filter .sql files that don't match the expected file pattern"

yoavamit:
Reverts cashapp/misk#2546


staktrace (approved):
For context the reason this is being reverted is that it generally goes against conventions of having a single type of migration file in a folder. Some tools rely on that and having a folder where a bunch of different types of migration files are mixed can lead to issues. There's also a potential footgun where migration files that are inadvertently misnamed get silently ignored instead of erroring, which would be hard to debug. More details in the private slack thread at https://cash.slack.com/archives/C01M5M297T7/p1666104235111269

PR #2553, opened 2022-10-20
Adds a JVM runtime API

cconroy:
My original motivation here was to be able to easily inspect the command line flags passed to the JVM for tuning our production configuration. I went ahead and serialized the rest of the RuntimeMxBean while I was in there since classpath and the like can also be useful.

I'd like to eventually make some nice UI around this, but I've been unable to get miskweb to work locally. However, it's still a useful first step to have an API.

### DIFF
--- a/misk-admin/src/main/kotlin/misk/web/metadata/jvm/JvmMetadataAction.kt
+++ b/misk-admin/src/main/kotlin/misk/web/metadata/jvm/JvmMetadataAction.kt
@@ -1,0 +1,91 @@
+package misk.web.metadata.jvm
+
+import com.squareup.moshi.JsonAdapter
+import com.squareup.moshi.Moshi
+import misk.moshi.adapter
+import misk.web.Get
+import misk.web.RequestContentType
+import misk.web.ResponseContentType
+import misk.web.actions.WebAction
+import misk.web.dashboard.AdminDashboardAccess
+import misk.web.mediatype.MediaTypes
+import java.lang.management.RuntimeMXBean
+import javax.inject.Inject
+import javax.inject.Singleton
+
+/**
+ * Conveys information about the current JVM
+ */
+@Singleton
+class JvmMetadataAction : WebAction {
+
+  private val runtimeMxBean: RuntimeMXBean
+  private val moshiAdapter: JsonAdapter<JvmRuntimeResponse>
+
+  @Inject constructor(runtimeMxBean: RuntimeMXBean, moshi: Moshi) {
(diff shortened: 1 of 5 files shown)


misk-admin/src/main/kotlin/misk/web/metadata/jvm/JvmMetadataAction.kt
```diff
@@ -0,0 +1,91 @@
...
+@Singleton
+class JvmMetadataAction : WebAction {
+
+  private val runtimeMxBean: RuntimeMXBean
+  private val moshiAdapter: JsonAdapter<JvmRuntimeResponse>
+
+  @Inject constructor(runtimeMxBean: RuntimeMXBean, moshi: Moshi) {
+    this.runtimeMxBean = runtimeMxBean
+    // Indent to make the output more readable since the response can be quite large!
+    this.moshiAdapter = moshi.adapter<JvmRuntimeResponse>().indent("  ")
```
cconroy, line 28: in other frameworks I've seen affordance to configure a different serializer for things like admin consoles and the like (where it makes sense to have pretty output turned on generally). I didn't see any such affordance in misk so just doing this locally here.

jdm-square (approved):
This looks like a useful addition!

misk-admin/src/main/kotlin/misk/web/metadata/jvm/JvmMetadataAction.kt
```diff
@@ -0,0 +1,91 @@
...
+    val class_path: String?,
+    val library_path: String?,
+    val is_boot_class_path_supported: Boolean?,
+    val boot_class_path: String?,
+    val input_arguments: List<String>?,
+    val uptime_millis: Long?,
+    val start_time_millis: Long?,
+    val system_properties: Map<String, String>?,
+  ) {
+    companion object Factory {
```
kyeotic, line 61: I see companion objects left anonymous most of the time, and [the docs](https://kotlinlang.org/docs/object-declarations.html#companion-objects) say the name is optional and not used when referencing the contained static methods. Just curious about your approach:

Whats the motivation behind naming the companion? Would you always name it? 
If a future use case appears for another companion/static method that doesn't fit into a "factory", how would you approach it?

cconroy: My Java is showing through here: I'd normally just make an inner static factory class, but I'll remove the name here as it's not necessary or conventional

adrw:
Nice addition!

PR #2560, opened 2022-10-28
Remove constraint on prometheus client dependency

staktrace:
This is not the right place for such a constraint, because non-cash users of misk (if any such mythical creatures exist) may not want to be forced into using a particular prometheus client.


adrw:
There are other non-mythical companies that use Misk.

adrw (approved):

PR #2565, opened 2022-10-28
Upgrade wisp to 1.2.2

ThePumpingLemma:


jdm-square (approved):

adrw (approved):

staktrace:
Hmm I guess the misk repo itself requires the prometheus client version to be at 0.9.0. I think we should be able to force that without also forcing it on all the downstream consumers of misk (they would be reponsible for doing it themselves if desired).

staktrace:
The easiest way appears to be just exclude the prometheus transitive dep from the kubernetes client. I looked at the [kubernetes client code](https://github.com/kubernetes-client/java/tree/e63ca08c2d8d1ee93266dbc604993fef992926a3) and the prometheus dep is optional and only needed for a few specific APIs that we don't use in misk-clustering.

PR #2568, opened 2022-11-03
Retry on StaleStateException

mhickman:
StaleStateException is like StaleObjectStateException except it is thrown when Hibernate is issuing bulk change operations. In these instances it's safe to retry the transaction.


akansal-part-2 (approved):

frojasg (approved):
Should we try to add a test for this?

Dileksema26 (approved):

PR #2569, opened 2022-11-07
Deprecate histogram method and direct users to use summary instead

amitdev:
`Metrics.histogram` is really a summary metric and not histogram. Deprecating this method to avoid confusion. There is already `v2.Metrics.histogram` and `v2.Metrics.summary` which are correctly named.


MatthewDolan (approved):
Thanks for clarifying the docs!

PR #2570, opened 2022-11-08
Guice modularize dashboard tabs

adrw:
Modules per tab make it easier to build custom dashboards which only
have certain tabs installed or forgo the provided AdminDashboardModule
for a service specific dashboard module which doesn't include the
default menu links.


jdm-square (approved):
This looks reasonable!

PR #2572, opened 2022-11-08
Include JVM config in Config tab

adrw:
Add JVM config to the Config admin dashboard tab and install by default with the unsafe YAML parts disabled.

# Safe mode 
Screen Shot 2022-11-09 at 14 06 55

# Unsafe mode
Screen Shot 2022-11-09 at 14 08 36

### DIFF
--- a/misk-admin/src/main/kotlin/misk/web/dashboard/AdminDashboardModule.kt
+++ b/misk-admin/src/main/kotlin/misk/web/dashboard/AdminDashboardModule.kt
@@ -20,6 +20,7 @@ class AdminDashboardModule(private val isDevelopment: Boolean) : KAbstractModule
 
   override fun configure() {
     install(BaseDashboardModule(isDevelopment))
+    install(ConfigDashboardTabModule(isDevelopment))
     install(DatabaseDashboardTabModule(isDevelopment))
     install(WebActionsDashboardTabModule(isDevelopment))
     install(WebActionsOldDashboardTabModule(isDevelopment))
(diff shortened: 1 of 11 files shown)


adrw:
- [ ] Update the config tab to by default not render the YAML sections since they are still unsafe because of leaking Misk.Secrets, later once this is fixed it can be updated to default on render the YAML too

misk-admin/src/main/kotlin/misk/web/dashboard/AdminDashboardModule.kt
```diff
@@ -20,6 +20,7 @@ class AdminDashboardModule(private val isDevelopment: Boolean) : KAbstractModule
 
   override fun configure() {
     install(BaseDashboardModule(isDevelopment))
+    install(ConfigDashboardTabModule(isDevelopment))
```
adrw, line 23: We can now safely reinstall this since it is using the default SECURE mode and will not show any insecure YAML config, only the JVM and other future secure config.

misk-admin/src/main/kotlin/misk/web/metadata/config/ConfigMetadataAction.kt
```diff
@@ -51,20 +74,10 @@ class ConfigMetadataAction @Inject constructor(
...
-      val yamlFiles = linkedMapOf<String, String?>(
-        "Effective Config" to MiskConfig.toYaml(
-          config, ResourceLoader.SYSTEM
-        )
-      )
-      rawYamlFiles.map { yamlFiles.put(it.key, it.value) }
-      return yamlFiles
-    }
+  enum class ConfigTabMode {
+    SECURE, // Only show secure content which will not leak Misk secrets
```
cconroy, line 80: maybe `SAFE`/`UNSAFE_LEAK_MISK_SECRETS` would be clearer? hopefully we can fix the underlying secrets issue before too long so this doesn't much matter.

cconroy:
let's update the PR+commit description to explain that this is also installing the config module in the secure/safe mode by default.

Can you include a screenshot showing what this looks like?

cconroy (approved):

PR #2585, opened 2022-11-17
Set error on the parent (root) span if it exists

Hexcles:
For gRPC, the network status code is always 200, so the parent span won't know there's been an error. Help it out by explicitly setting the error tag to true based on the logical status code.

### DIFF
--- a/misk/src/main/kotlin/misk/web/interceptors/TracingInterceptor.kt
+++ b/misk/src/main/kotlin/misk/web/interceptors/TracingInterceptor.kt
@@ -71,6 +71,11 @@ internal class TracingInterceptor internal constructor(private val tracer: Trace
       Tags.HTTP_STATUS.set(span, chain.httpCall.statusCode)
       if (chain.httpCall.statusCode > 399) {
         Tags.ERROR.set(span, true)
+        // In case of gRPC errors, [HttpCall.networkStatusCode] is always 200, in which case the
+        // parent (likely the root) span won't be able to know about the error.
+        if (parentSpan != null) {
+          Tags.ERROR.set(parentSpan, true)
+        }
       }
     } catch (t: Throwable) {
       Tags.ERROR.set(span, true)
--- a/misk/src/test/kotlin/misk/web/interceptors/TracingInterceptorTest.kt
+++ b/misk/src/test/kotlin/misk/web/interceptors/TracingInterceptorTest.kt
@@ -142,6 +142,33 @@ class TracingInterceptorTest {
   }
 
   @Test
+  fun failedTraceWithParentSpan() {
+    val tracingInterceptor = tracingInterceptorFactory.create(
+      TracingTestAction::call.asAction(DispatchMechanism.GET)
+    )!!
+    val httpCall = FakeHttpCall(url = "http://foo.bar".toHttpUrl())
+    val chain = RealNetworkChain(
+      TracingTestAction::call.asAction(DispatchMechanism.GET),
+      tracingTestAction, httpCall,
+      listOf(
+        object : NetworkInterceptor {
+          override fun intercept(chain: NetworkChain) {
+            tracer.traceWithSpan("parent-span") {
+              chain.proceed(chain.httpCall)
+            }
+          }
+        },
+        tracingInterceptor, TerminalInterceptor(500)
+      )
+    )
+
+    chain.proceed(chain.httpCall)
+
+    val span = tracer.take("parent-span")
+    assertThat(span.tags().get(Tags.ERROR.key)).isEqualTo(true)
+  }
+
+  @Test
   fun nothingIfNotInstalled() {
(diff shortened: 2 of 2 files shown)


jdm-square (approved):
Nice find!

Hexcles:
@jdm-square could you merge this? Thanks!

jdm-square:
@Hexcles Since this merged, we're getting reports of phantom errors with 200 response codes in the datadog UI, where there are no logs of errors in the same period. This seems extremely likely to be related to this change; could you explain more about what problem this change was intended to address?

Hexcles:
@jdm-square just to make sure: are the "phantom error" reports from GRPC endpoints? If not, we might have a bug (although I struggle to see why -- is there another code path that could cause `statusCode` to be different from `networkStatusCode`? I only see GRPC.)

Now if we are talking about GRPC, then I'd argue these are real errors. Just a reminder that GRPC *always* uses 200 on the wire, while setting the actual status code in the trailer. These errors are non-zero GRPC status codes. If it were a normal HTTP request, it'd show up as a straightforward error with status code >= 400.

As far as logging goes, unfortunately we don't use misk built-in logging, and we pay attention to log grpc-status for GRPC endpoints, too, so there is no phantom -- the request log would clearly show it's a 200 HTTP response with a non-zero GRPC status code and set at ERROR level, but it is a bit weird in e.g. Datadog tracing UI where there's a prominent green 200, which takes a bit of developer education.

There's a usability problem here. You *could* have the request logs and span tags use the "logical" `statusCode` instead of `networkStatusCode`, but that's technically incorrect and may be confusing if you cross-reference other metrics.

mateuszmrozewski:
We started to have increased number errors in APM for http endpoints. One of our endpoints returns code 402 which is not an error.

jdm-square:
Ok, I've investigated the cases of 200 responses being reported as errors for my own service, and there are errors occuring in gRPC endpoints that line up. Some association is missing from the Datadog UI, because it claims that there are no associated errors or traces for those erroring responses, which is probably contributing to the confusion from other engineers encountering this.

It's not yet clear to me if the problem in the previous comment is related to this, given that the responses are not 200 responses.

jdm-square:
Actually, I can see how this code change would cause that behaviour change. This feels like either cardio is doing the wrong thing (how are they creating meaningful gRPC responses with non-200 status codes?) or we should only check for codes >499.

jdm-square:
Mmm, another reason why this behaviour ends up confusing people: one source of "phantom" errors that I've come across is when a gRPC action fails because the client closed the request. Misk internally catches this and reports it as an info log (https://github.com/cashapp/misk/blob/e356915b54ae122731d0d30a9e6b757b16f48002/misk/src/main/kotlin/misk/web/exceptions/EofExceptionMapper.kt#L12), so it's easy for an investigation to see a large spurt of errors reported from Datadog, but filtering the logs to warnings or errors to show nothing of interest.

jdm-square:
This is a code change that I believe would make @mateuszmrozewski's behaviour change go away. Any opinions on if it's the right change?
```diff
diff --git a/misk/src/main/kotlin/misk/web/interceptors/TracingInterceptor.kt b/misk/src/main/kotlin/misk/web/interceptors/TracingInterceptor.kt
index 41d6c154d..7ade6b20c 100644
--- a/misk/src/main/kotlin/misk/web/interceptors/TracingInterceptor.kt
+++ b/misk/src/main/kotlin/misk/web/interceptors/TracingInterceptor.kt
@@ -73,7 +73,7 @@ internal class TracingInterceptor internal constructor(private val tracer: Trace
         Tags.ERROR.set(span, true)
         // In case of gRPC errors, [HttpCall.networkStatusCode] is always 200, in which case the
         // parent (likely the root) span won't be able to know about the error.
-        if (parentSpan != null) {
+        if (chain.httpCall.networkStatusCode != chain.httpCall.statusCode && parentSpan != null) {
           Tags.ERROR.set(parentSpan, true)
         }
       }
```

Hexcles:
@jdm-square I have some explanation for the HTTP (non-GRPC) case:

Prior to this patch, misk will mark the `http.action` inner span as error for 4xx requests, but whether or not the service entry span (e.g. `servlet.request`) is marked as error is up to the instrumentation. In the case of Datadog, it only considers 5xx as error. So you'd have a failed `http.action` nested inside a successful `servlet.request`. This PR changed that.

I think your patch should work. But I wonder if it's better to only treat 5xx as errors (i.e. change `> 399` to `> 499`), which is more in line with the industry convention I believe.

jdm-square:
Agreed—it does seem surprising to mark 4xx traces as errors in the current setup.

mateuszmrozewski:
Thank you for investigating. Will you prepare a PR to apply the suggested patch?

jdm-square:
Yep!

PR #2586, opened 2022-11-18
Update add(Period) to be consistent with the other add functions

staktrace:
This way it returns the new value of the FakeClock rather than the old one, just like all the other add functions.


staktrace:
Closing in favor of https://github.com/cashapp/misk/pull/2587 which will delegate to wisp instead

PR #2587, opened 2022-11-18
Make Misk FakeClock wrap Wisp FakeClock

tyiu:
Changes Misk's `FakeClock` to delegate to Wisp's `FakeClock` with the intent to eventually deprecate Misk's in the future.

Also changes all the usages of FakeClock internally in Misk to use Wisp's directly instead.

### DIFF
--- a/misk-cron/build.gradle.kts
+++ b/misk-cron/build.gradle.kts
@@ -17,5 +17,6 @@ dependencies {
   api(Dependencies.wispLogging)
 
   testImplementation(Dependencies.assertj)
+  testImplementation(Dependencies.wispTimeTesting)
   testImplementation(project(":misk-testing"))
 }
--- a/misk-jooq/build.gradle.kts
+++ b/misk-jooq/build.gradle.kts
@@ -21,6 +21,7 @@ dependencies {
   api(Dependencies.wispLogging)
 
   testImplementation(Dependencies.assertj)
+  testImplementation(Dependencies.wispTimeTesting)
   testApi(project(":misk-testing"))
   testApi(project(":misk-jdbc-testing"))
 
--- a/misk-testing/build.gradle.kts
+++ b/misk-testing/build.gradle.kts
@@ -28,6 +28,7 @@ dependencies {
     exclude(group = "junit")
   }
   implementation(Dependencies.javaxInject)
+  implementation(Dependencies.wispTimeTesting)
   api(Dependencies.servletApi)
   api(Dependencies.wispContainersTesting)
   api(Dependencies.wispLogging)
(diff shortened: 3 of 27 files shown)


misk-cron/build.gradle.kts
```diff
@@ -9,6 +9,7 @@ dependencies {
   implementation(Dependencies.guice)
   implementation(Dependencies.kotlinReflection)
   implementation(Dependencies.moshiAdapters)
+  implementation(Dependencies.wispTimeTesting)
```
tyiu, line 12: Should be `testImplementation`

staktrace (approved):
LGTM!

misk-jooq/build.gradle.kts
```diff
@@ -21,6 +21,7 @@ dependencies {
   api(Dependencies.wispLogging)
 
   testImplementation(Dependencies.assertj)
+  testImplementation(Dependencies.wispTimeTesting)
```
adrw, line 24: Changing the implementation to api in the MiskTimeTesting module should let you get rid of all these added explicit testImplementation changes in other modules.

misk-testing/build.gradle.kts
```diff
@@ -28,6 +28,7 @@ dependencies {
     exclude(group = "junit")
   }
   implementation(Dependencies.javaxInject)
+  implementation(Dependencies.wispTimeTesting)
```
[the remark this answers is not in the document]
staktrace: As-is, the calling service will get it in the runtime classpath, but not in the compile-time classpath, which IMO is exactly what we want. If we change it to `api` then the calling service will get it in the compile-time classpath transitively, which will make migrating to the wisp version harder because it won't be obvious where the FakeClock is coming from. As-is, the migration will involve (a) change the code to use wisp and (b) swapping out the misk dep for the wisp dep. Either of those in isolation will cause compilation failure, therefore making it easy to detect a botched migration.

misk-jooq/build.gradle.kts
staktrace: I don't think minimizing the diff is the right optimization here. I think the patch as written is better because it makes it explicit that the FakeClock is coming from wisp rather than implicitly getting it transitively through a deprecated wrapper.

misk-testing/build.gradle.kts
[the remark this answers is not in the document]
staktrace: To be clear, by "botched migration" I mean "the calling service is using the wisp FakeClock, but getting the dependency transitively through the misk wrapper". This would be "botched" because eventually the misk wrapper will be removed, and that would break the calling service.

PR #2597, opened 2022-11-22
Improve RequestBodyFeatureBinding Unmarshaller creation

JGulbronson:
When using `mapNotNull` followed by `first`, we end up calling `create` on all of the `Unmarshaller` factories, only to throw them away and only take the first. This can cause issues if a custom factory conflicts with one of the built-in Misk ones, such as the `JsonUnmarshaller` which uses Moshi and can throw if it can't find a type adapter. This avoids calling `create` eagerly so if an earlier factory is able to create an `Unmarshaller`, the rest of the factories are skipped.


jdm-square:
Thanks!

PR #2601, opened 2022-11-23
add conditional `ServiceModule.dependsOn`

kryptonat:
**Overview**

It can be verbose to add service dependencies conditionally especially if these dependencies are chained. 

**Example**
```kotlin
object : KAbstractModule() {
  override fun configure() {
    bind<StringBuilder>().toInstance(log)
    install(ServiceModule<ProducerService>())

    val isACriticalDependencyForSomeReason = true
    install(ServiceModule<ConsumerService>()
      .dependsOn<ProducerService>()
      // Since `dependsOn` returns a new instance of [ServiceModule], a simple `.apply` can't be used here
      .let {
        if (isACriticalDependencyForSomeReason) {
          it.dependsOn<AnotherUpstreamService>()
        } else {
          it
        }
      }
    install(ServiceModule<AnotherUpstreamService>())
  }
}
```

**Post PR changes example**
```kotlin
object : KAbstractModule() {
  override fun configure() {
    bind<StringBuilder>().toInstance(log)
    install(ServiceModule<ProducerService>())

    val isACriticalDependencyForSomeReason = true
    install(ServiceModule<ConsumerService>()
      .dependsOn<ProducerService>()
      // More humanly readable
      .dependsOn<AnotherUpstreamService>(condition = isACriticalDependencyForSomeReason))
    install(ServiceModule<AnotherUpstreamService>())
  }
}
```

### DIFF
--- a/misk-service/src/main/kotlin/misk/ServiceModule.kt
+++ b/misk-service/src/main/kotlin/misk/ServiceModule.kt
@@ -116,6 +116,14 @@ class ServiceModule constructor(
     key, dependsOn + upstream, enhancedBy
   )
 
+  /**
+   * Returns the [ServiceModule] with added the upstream dependency if condition is met otherwise
+   * it returns the original [ServiceModule] without changes.
+   */
+  fun dependsOn(upstream: Key<out Service>, condition: Boolean) = if (condition) ServiceModule(
+    key, dependsOn + upstream, enhancedBy
+  ) else this
+
   fun enhancedBy(enhancement: Key<out Service>) =
     ServiceModule(key, dependsOn, enhancedBy + enhancement)
 
@@ -124,6 +132,11 @@ class ServiceModule constructor(
 
   inline fun <reified T : Service> dependsOn(qualifier: KClass<out Annotation>? = null) =
     dependsOn(T::class.toKey(qualifier))
+
+  inline fun <reified T : Service> dependsOn(
+    qualifier: KClass<out Annotation>? = null,
+    condition: Boolean,
+  ) = dependsOn(upstream = T::class.toKey(qualifier), condition = condition)
 
   inline fun <reified T : Service> enhancedBy(qualifier: KClass<out Annotation>? = null) =
     enhancedBy(T::class.toKey(qualifier))
--- a/misk-service/src/test/kotlin/misk/ServiceManagerModuleTest.kt
+++ b/misk-service/src/test/kotlin/misk/ServiceManagerModuleTest.kt
@@ -233,4 +233,41 @@ internal class ServiceManagerModuleTest {
       |AnotherUpstreamService.shutDown
       |""".trimMargin())
   }
+
+  @Test fun serviceNotProvidedUntilAllEligibleDependenciesCreated() {
+    val log = StringBuilder()
+    val injector = Guice.createInjector(
+      MiskTestingServiceModule(),
+      object : KAbstractModule() {
+        override fun configure() {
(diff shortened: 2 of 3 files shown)


adrw:
Alternatively, would this work without adding new APIs to Misk?

```kotlin
    val isACriticalDependencyForSomeReason = true
    val module = ServiceModule<ConsumerService>()
      .dependsOn<ProducerService>()
    if (isACriticalDependencyForSomeReason) {
      install(module.dependsOn<AnotherUpstreamService>())
    } else {
      install(module)
    }
    install(ServiceModule<AnotherUpstreamService>())
```

misk-service/src/main/kotlin/misk/ServiceModule.kt
```diff
@@ -125,6 +133,11 @@ class ServiceModule constructor(
   inline fun <reified T : Service> dependsOn(qualifier: KClass<out Annotation>? = null) =
     dependsOn(T::class.toKey(qualifier))
 
+  inline fun <reified T : Service> dependsOn(
+    qualifier: KClass<out Annotation>? = null,
+    condition: Boolean,
```
adrw, line 138: I wonder if accepting a lambda instead like `() -> Boolean` would be more flexible for inline expressions.

keeferrourke:
This change looks like and I'm certainly all for making nicer service APIs, but I think I'm more curious about the conditions that lead to wanting to install service dependencies in this manner.

Is this just a matter of dev/test vs production, and test versions of services not wanting or needing certain dependencies? Is this a pattern that we're seeing a lot of?

My only concern is that the more conditions in service configuration there are, the more difficult it might be to understand the service configuration.

kryptonat:
> Alternatively, would this work without adding new APIs to Misk?
```
    val isACriticalDependencyForSomeReason = true
    val module = ServiceModule<ConsumerService>()
      .dependsOn<ProducerService>()
    if (isACriticalDependencyForSomeReason) {
      install(module.dependsOn<AnotherUpstreamService>())
    } else {
      install(module)
    }
    install(ServiceModule<AnotherUpstreamService>())
```

@adrw It can start to look ugly / hard to read when there multiple conditional service dependencies - and they might not all have the same condition.

> This change looks nice and I'm certainly all for making nicer service APIs, but I think I'm more curious about the conditions that lead to wanting to install service dependencies in this manner.
> 
> Is this just a matter of dev/test vs production, and test versions of services not wanting or needing certain dependencies? Is this a pattern that we're seeing a lot of?
> 
> My only concern is that the more conditions in service configuration there are, the more difficult it might be to understand the service configuration.

That is a valid concern. This scenario came up due to a subset of services that are only included in production/staging environments only (e.g. dynamodb lease service) of which a lot of other service modules depend on.

keeferrourke:
> @keeferrourke That is a valid concern. This scenario came up due to a subset of services that are only included in production/staging environments only (e.g. dynamodb lease service) of which a lot of other service modules depend on.

I had an offline conversation about the particular concerns that led to this PR, and I'm inclined to say that while the spirit of this PR is good and well-intentioned, there is probably a better solution around structuring fake/real code within Misk services. What this conditional dependencies pattern does is just short-circuit proper test-mirroring and dependency installation. I'm pretty sure we don't want that!

Misk apps are meant to look the same at the interface level for all components involved, regardless of environment. Guava Services might have totally different implementations (like even just no-op as a test-double to satisfy dependencies), and I think that's... fine?

Happy to be convinced otherwise of course, but I just think this is solving the problem too deep in the stack.

PR #2611, opened 2022-12-06
Add retries to SchemaMigrator.initialize

adrw:
For cases where multiple JdbcModules are installed, they will race to
create the `schema_version` table in tests and one will fail. Wrapping
initialize with retries will fix this as one will succeed and on retry
the others will succeed.

Motivating use case is where a utility library (say feature flags) is
installed which relies on its own utility table being present in the
database and a separate JDBC module installed for the utility usage.

The test added previously failed without the retries change.


misk-jdbc/src/main/kotlin/misk/jdbc/SchemaMigrator.kt
```diff
@@ -11,7 +13,10 @@ import misk.vitess.target
 import wisp.logging.getLogger
 import java.sql.Connection
 import java.sql.SQLException
-import java.util.*
```
swankjesse, line 14: TY!

swankjesse (approved):

PR #2612, opened 2022-12-06
control migration files regular expression and excluded files

yoavamit:
Enable misk's `SchemaMigrator` better control over which files are used when checking and applying schema files.

### Background
Misk's `SchemaMigrator` is a simple service that loads at application startup, and applies any new database migrations statements found in the files specified by the `DataSourceConfig.migrations_resource(s)` config value.
Any file name in the path configured in `DataSourceConfig.migrations_resource(s)` that doesn't match the pattern `vNNN__some-string.sql` where `NNN` is a number, will cause an exception to be raised, and the application to crash.

### What's changed?
In this PR, I've added 2 new configuration values:
- `DataSourceConfig.migrations_resources_exclusion`. Specifies a list of files that should be skipped when the `SchemaMigrator` enumerates files in the path specified in `DataSourceConfig.migrations_resource(s)`.
- `DataSourceConfig.migrations_resources_regex`. Specifies the regular expression to match against. Any file name that doesn't match the given regular expression **and** is not listed in the excluded list above will cause an exception.


staktrace (approved):
This seems reasonable, thanks!

PR #2622, opened 2022-12-08
Add asContextElement to [ActionScope] to propagate data to coroutines

frojasg:
ActionScope stores the information on the threadlocal, which is problematic if an action scope is being used within a coroutine scope as the coroutine could be run in a different thread than the one intended.

The asContextElement method is added to propagate the ActionScope into the coroutine context.


arjan-nirh-block (approved):
thank you!

adrw (approved):
Nice addition!

PR #2623, opened 2022-12-09
Fix for bug when trying to use a `LIKE` query on a `BoxedStringType` column

tmccarthy:
Currently using a JPA `like` constraint on a field that's bound using a `BoxedStringType` leads to an attempt to unbox the string used as the `LIKE` pattern. This PR adds a test for this behaviour and adjusts `BoxedStringType` to fix this issue.

Encountered this [here](https://github.com/squareup/cash-limits/commit/327fb3f5565d4660223cd548842e07f962fa0727) (internal cash-app link).


jdm-square (approved):
I appreciate the test; after reading the comments in BoxedStringType this looks like a reasonable solution for the problem.

PR #2624, opened 2022-12-12
Add better handling for cases where the content type or unmarshaller …

leecsquare:
…are null when processing an http request.

### DIFF
--- a/misk/src/main/kotlin/misk/web/extractors/RequestBodyFeatureBinding.kt
+++ b/misk/src/main/kotlin/misk/web/extractors/RequestBodyFeatureBinding.kt
@@ -1,6 +1,7 @@
 package misk.web.extractors
 
 import misk.Action
+import misk.exceptions.BadRequestException
 import misk.web.FeatureBinding
 import misk.web.FeatureBinding.Claimer
 import misk.web.FeatureBinding.Subject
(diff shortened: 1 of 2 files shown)


frojasg (approved):
LGTM! Thanks for doing this change!

misk/src/main/kotlin/misk/web/extractors/RequestBodyFeatureBinding.kt
```diff
@@ -20,11 +21,21 @@ internal class RequestBodyFeatureBinding(
...
     val unmarshaller = mediaType?.let { type ->
       unmarshallerFactories.firstNotNullOfOrNull { it.create(type, parameter.type) }
     } ?: GenericUnmarshallers.into(parameter)
-      ?: throw IllegalArgumentException("no generic unmarshaller for ${parameter.type}")
+
+    if (unmarshaller == null) {
+      if (contentType == null) {
+        throw BadRequestException("Can't parse request: missing Content-Type header")
+      } else {
+        // If this is thrown it means an endpoint was registered with a contentType that misk
```
cconroy, line 34: My read of the code here is that we fall into this case if there is no registered unmarshaller for the requested content-type.

Certainly, if one defines an endpoint for a content-type with no ability to deserialize it, that would be a bug and should be a 500 (or...more ideally some CI failure..) That _may_ be the case here, but any incorrect content-type passed from the client _also_ puts us here. So, I think this probably wants to also be a `BadRequestException` and we need a more _specific_ check for the case of no unmarshaller registered for the endpoint's content type.

leecsquare: So one fun thing we discovered when writing the tests for this is it's very hard to get into this state. I wish we had full request/response logging available so we could understand how this happened with clientsync in the first place.

Why is it hard to get in to this state? If you pass a content type that the endpoint is not registered to accept you get an error much earlier in the flow. So to get here the content type has to be a type that the endpoint is configured to accept but we don't have an unmarhsaller.

cconroy: >  If you pass a content type that the endpoint is not registered to accept you get an error much earlier in the flow. 

Is that a 4xx or a 5xx error? It should be 4xx

leecsquare: It's a 404.

cconroy: Hmmm. I think that probably wants to be a `415: Unsupported Media Type` .. but that should probably be a different PR!

leecsquare: Agreed on both counts!

cconroy:
nit: good practice to follow the 50/72 rule for git commit messages (and transitively, the resulting PR title and descriptions). Github renders these better, and it keeps the git log in terminal nicer, etc.

https://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html

cconroy (approved):

PR #2634, opened 2022-12-21
Include apply migrations in SchemaMigrator retry block

adrw:
Failures on restart were still being seen from the `applyAll` statement not being included in the retry block.

### DIFF
--- a/misk-jdbc/src/main/kotlin/misk/jdbc/SchemaMigratorService.kt
+++ b/misk-jdbc/src/main/kotlin/misk/jdbc/SchemaMigratorService.kt
@@ -26,11 +26,13 @@ class SchemaMigratorService internal constructor(
       val type = connector.config().type
       if (type != DataSourceType.VITESS_MYSQL) {
         // Retry wrapped to handle multiple JDBC modules racing to create the `schema_version` table.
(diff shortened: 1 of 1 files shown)


misk-jdbc/src/main/kotlin/misk/jdbc/SchemaMigratorService.kt
```diff
@@ -26,11 +26,13 @@ class SchemaMigratorService internal constructor(
...
         // Retry wrapped to handle multiple JDBC modules racing to create the `schema_version` table.
-        val appliedMigrations = retry(
+        retry(
           10,
           ExponentialBackoff(Duration.ofMillis(100), Duration.ofSeconds(5))
-        ) { schemaMigrator.initialize() }
-        migrationState = schemaMigrator.applyAll("SchemaMigratorService", appliedMigrations)
+        ) {
+          val appliedMigrations = schemaMigrator.initialize()
+          migrationState = schemaMigrator.applyAll("SchemaMigratorService", appliedMigrations)
```
adrw, line 34: Failures on restart were still being seen from this statement not being included in the retry block.

shellderp (approved):

PR #2654, opened 2023-01-10
Clean up some hermit and test configuration

tso:


.github/workflows/Publish-Website.yml
```diff
@@ -17,13 +17,16 @@ jobs:
       - name: Checkout the repo
         uses: actions/checkout@v3
 
+      - name: Hermit
+        run: ./bin/hermit env -r >> $GITHUB_ENV
```
staktrace, line 21: Is doing this solving a particular problem? IMO it's preferable to just use `bin/gradle` since that explicitly uses the hermit version of gradle and manages everything under the hood.

tso: I had gotten the suggestion on my previous PR https://github.com/cashapp/misk/pull/2648

It does nicely mean I can just npm install and then miskweb in the web stuff

staktrace: Ok. I don't feel strongly about it, this is fine.

staktrace (approved):

PR #2660, opened 2023-01-12
Add return values to misk-redis HSET

keeferrourke:
`HSET key field value [field value ...]` returns the integer number of the fields that were added by the command invocation.

Previously, misk's Redis interface hid these return values.

This PR exposes them.

You can play with HSET on the doc website: https://redis.io/commands/hset/

### DIFF
--- a/misk-redis/src/main/kotlin/misk/redis/FakeRedis.kt
+++ b/misk-redis/src/main/kotlin/misk/redis/FakeRedis.kt
@@ -207,15 +207,13 @@ class FakeRedis : Redis {
         expiryInstant = Instant.MAX
       )
     }
-    val newFieldCount = if (hKeyValueStore[key]!!.data[field] != null) { 0L } else 1L
+    val newFieldCount = if (hKeyValueStore[key]!!.data[field] != null) 0L else 1L
     hKeyValueStore[key]!!.data[field] = value
     return newFieldCount
   }
 
   override fun hset(key: String, hash: Map<String, ByteString>): Long {
-    return hash.toList().fold(0L) { newFieldCount, (field, value) ->
-      newFieldCount + hset(key, field, value)
-    }
+    return hash.entries.sumOf { (field, value) -> hset(key, field, value) }
   }
 
   override fun incr(key: String): Long {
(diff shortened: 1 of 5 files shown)


misk-redis/src/main/kotlin/misk/redis/FakeRedis.kt
```diff
@@ -200,19 +200,21 @@ class FakeRedis : Redis {
...
+    val newFieldCount = if (hKeyValueStore[key]!!.data[field] != null) { 0L } else 1L
     hKeyValueStore[key]!!.data[field] = value
+    return newFieldCount
   }
 
-  override fun hset(key: String, hash: Map<String, ByteString>) {
-    hash.forEach {
-      hset(key, it.key, it.value)
+  override fun hset(key: String, hash: Map<String, ByteString>): Long {
+    return hash.toList().fold(0L) { newFieldCount, (field, value) ->
```
tso, line 216: probably prefer `hash.entries.sumOf { (field, value) -> hset(field, value) }` here since more readable i think

misk-redis/src/main/kotlin/misk/redis/FakeRedis.kt
```diff
@@ -200,19 +200,21 @@ class FakeRedis : Redis {
...
 
-  override fun hset(key: String, field: String, value: ByteString) {
+  override fun hset(key: String, field: String, value: ByteString): Long {
     if (!hKeyValueStore.containsKey(key)) {
       hKeyValueStore[key] = Value(
         data = ConcurrentHashMap(),
         expiryInstant = Instant.MAX
       )
     }
+    val newFieldCount = if (hKeyValueStore[key]!!.data[field] != null) { 0L } else 1L
```
tso, line 210: nit: `if (hKeyValueStore[key]!!.data.containsKey(field)) 0L else 1L` (mostly just remove the brackets around the `0L`

tso (approved):

misk-redis/src/main/kotlin/misk/redis/FakeRedis.kt
keeferrourke: Nice, I always forget about `sumOf`!

PR #2662, opened 2023-01-16
Deprecate TransactionalJobQueue because it does not guarantee a job will be created if a record is spooled

tristandl:
This implementation of TransactionalJobQueue is not fully transactional, in that a failure to enqueue the job (due to AWS outage etc) will not be detected, and no job will be enqueued.

Given it's several years old and the replication-based jobqueue is still not created, this deprecates the existing implementation. 

An alternate implementation would be to pass an [idempotenceKey](https://github.com/cashapp/misk/blob/1e8c0c70f7a75152b12c9f5f41adc41621c7bee6/misk-jobqueue/src/main/kotlin/misk/jobqueue/JobQueue.kt#L41-L41) and in the job handler check if it's in our DB. If not, run enqueue the job and then persist the `idempotenceKey`.

Alternately, an implementation of TransactionalJobQueue could be provided that uses replication logs to guarantee an SQS job is created from a DB record.

### DIFF
--- a/misk-transactional-jobqueue/src/main/kotlin/misk/jobqueue/TransactionalJobQueue.kt
+++ b/misk-transactional-jobqueue/src/main/kotlin/misk/jobqueue/TransactionalJobQueue.kt
@@ -15,6 +15,15 @@ import java.util.UUID
  * will typically use tbe second variant, writing jobs to the local shard in the context
  * of a transaction that modifies the contents of an entity group.
  */
+@Deprecated(
+  message = "This implementation is not strictly transactional since jobs are enqueued in a DB "
+    + "post commit hook. If the enqueueing operation fails, the DB record exists but no job is enqueued. "
+    + "Instead, replace with a standard JobQueue and pass an idempotency key to the enqueue() function, "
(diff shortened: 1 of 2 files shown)


misk-transactional-jobqueue/src/main/kotlin/misk/jobqueue/TransactionalJobQueue.kt
```diff
@@ -12,9 +12,18 @@ import java.util.UUID
  * then forwards the messages asynchronously to the actual queueing system. Two variants
  * of the [enqueue] operation exist - one which writes to the application's main shard, and
  * another which writes to the shard on which a given entity group exists. Applications
- * will typically use tbe second variant, writing jobs to the local shard in the context
+ * will typically use the second variant, writing jobs to the local shard in the context
  * of a transaction that modifies the contents of an entity group.
  */
+@Deprecated(
```
tristandl, line 18: i wasn't sure if we should deprecate the interface as well, but I think this creates more visibility in the downstream applications themselves, rather than their DI modules.

meganmckeough (approved):
🌟

tso (approved):

PR #2666, opened 2023-01-26
Allow to specify version for gcr.io/cloud-spanner-emulator/emulator image

lkerford:
Allow for configurations of the version of the image we pull for the spanner emulator

### DIFF
--- a/misk-gcp-testing/src/main/kotlin/misk/cloud/gcp/spanner/GoogleSpannerEmulator.kt
+++ b/misk-gcp-testing/src/main/kotlin/misk/cloud/gcp/spanner/GoogleSpannerEmulator.kt
@@ -126,11 +126,12 @@ class GoogleSpannerEmulator @Inject constructor(
    * Pulls a Docker container containing the Google Spanner emulator.
    */
   fun pullImage(imageVersion: String? = null) {
-    image = "$IMAGE_NAME:${imageVersion ?: "latest"}"
+    setImageVersion(imageVersion)
     Companion.pullImage()
   }
 
-  private fun doStart() {
+  private fun doStart(imageVersion: String? = null) {
+    setImageVersion(imageVersion)
     val spannerGrpcPort = ExposedPort.tcp(server.config.emulator.port)
     val spannerHttpPort = ExposedPort.tcp(server.config.emulator.port + 10)
     val ports = Ports()
@@ -266,6 +267,10 @@ class GoogleSpannerEmulator @Inject constructor(
         }
       )
     }
+  }
+
+  private fun setImageVersion(imageVersion: String?) {
+    image = "$IMAGE_NAME:${imageVersion ?: "latest"}"
   }
 
   private class SpannerServer(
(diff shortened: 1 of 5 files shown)


misk-gcp-testing/src/main/kotlin/misk/cloud/gcp/spanner/GoogleSpannerEmulator.kt
```diff
@@ -92,8 +92,9 @@ class GoogleSpannerEmulator @Inject constructor(
       .build()
     val docker: DockerClient =
       DockerClientImpl.getInstance(defaultDockerClientConfig, httpClient)
-    const val IMAGE = "gcr.io/cloud-spanner-emulator/emulator"
+    const val IMAGE_NAME = "gcr.io/cloud-spanner-emulator/emulator"
     const val CONTAINER_NAME = "misk-spanner-testing"
+    var image: String = "$IMAGE_NAME:latest"
```
MatthewDolan, line 97: This variable is making the hairs stand up on the back of my head because it creates some unintuitive side effects. Examples:

```
// no pull
doStart() // starts latest

pullImage() // pulls latest
doStart() // starts latest

pullImage("1.0.0") // pulls 1.0.0
doStart() // starts 1.0.0
```

This is sort of exacerbated by the fact that docker doesn't require images to be pulled before start (if the image isn't present at start, it will pull automatically).

What do you think about passing `imageVersion` into `doStart(...)` as well? That way it's more clear that the command starts the specific version.

MatthewDolan (commented):
Good fix! Glad to allow pinning and image version.

One comment about how `image` is passed around. But besides that, the change looks good to me.

misk-gcp-testing/src/main/kotlin/misk/cloud/gcp/spanner/GoogleSpannerEmulator.kt
lkerford: oh ya that is kinda odd, I'll update

misk-gcp-testing/src/main/kotlin/misk/cloud/gcp/spanner/GoogleSpannerEmulator.kt
```diff
@@ -124,11 +125,13 @@ class GoogleSpannerEmulator @Inject constructor(
   /**
    * Pulls a Docker container containing the Google Spanner emulator.
    */
-  fun pullImage() {
+  fun pullImage(imageVersion: String? = null) {
+    setImageVersion(imageVersion)
     Companion.pullImage()
```
MatthewDolan, line 130: What do you think of returning the image instead of setting it?

```suggestion
    Companion.pullImage(fullImageName(imageVersion))
```

Same with `doStart(...)`:

```
  private fun doStart(imageVersion: String? = null) {
    image = fulllImageName(imageVersion)
    ...
```

Then, below, update the function to return:

```
  private fun fullImageName(imageVersion: String?) {
    return "$IMAGE_NAME:${imageVersion ?: "latest"}"
  }
```

lkerford: I think that makes sense in the `doStart` but I think it would be confusing if in the following case
```
pullImage("1.0.0") // pulls 1.0.0
doStart() // starts latest
```

I think we would want
```
pullImage("1.0.0") // pulls 1.0.0
doStart() // starts 1.0.0
```

Ill update the `doStart` to not set the image but not the `pullImage`.  In  `doStart` I'll default to using the provided version if there is one and the set version if there is not

MatthewDolan (approved):
Looks good! Thanks for addressing my comments.

PR #2676, opened 2023-02-03
Add RPUSH, LPOP and RPOP commands to misk-redis client

keeferrourke:
This PR adds support for the RPUSH, LPOP and RPOP commands for redis lists.

```
RPUSH key element [element...]
```

```
LPOP key [count]
```

```
RPOP key [count]
```

The tests against the FakeRedis for these commands have been verified against a Redis REPL.

---

It also fixes a somewhat subtle bug with the FakeRedis implementation of LPUSH. Previously, FakeRedis would push the entire ordered list of elements to the head of an existing list. This contrasts Redis' actual behaviour for LPUSH, which is to push _each_ element one at a time to the head of the list.

Old behaviour:

```
redis.lpush("mylist", "a", "b", "c")
redis.lrange("mylist", 0, -1) // listOf("a", "b", "c")
```

New correct behaviour:

```
redis.lpush("mylist", "a", "b", "c")
redis.lrange("mylist", 0, -1) // listOf("c", "b", "a")
```

Tests that relied on this (wrong) behaviour have been minimally updated to reflect the behaviour as documented, and tested against a Redis REPL.

### DIFF
--- a/misk-redis/src/test/kotlin/misk/redis/FakeRedisTest.kt
+++ b/misk-redis/src/test/kotlin/misk/redis/FakeRedisTest.kt
@@ -841,4 +841,16 @@ class FakeRedisTest {
     clock.add(Duration.ofSeconds(1))
     assertThat(redis.rpop("droids", 1)).isEmpty()
   }
+
+  @Test fun lpushAndRpushAreOrderedCorrectly() {
+    // This test is pulled directly from the Redis documentation for LPUSH and RPUSH.
+    val elements = listOf("a", "b", "c").map { it.encodeUtf8() }
+
+    redis.lpush("l", *elements.toTypedArray())
+    assertThat(redis.lrange("l", 0, -1))
+      .containsExactly(*elements.toList().asReversed().toTypedArray())
+
+    redis.rpush("r", *elements.toTypedArray())
+    assertThat(redis.lrange("r", 0, -1)).containsExactly(*elements.toTypedArray())
+  }
 }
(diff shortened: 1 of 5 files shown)


keeferrourke:
Kind of scary that the fake's lpush was acting like rpush. Should add a test that compares fake redis with a real redis to prevent this kind of drift in the future.

misk-redis/src/test/kotlin/misk/redis/FakeRedisTest.kt
```diff
@@ -760,4 +761,96 @@ class FakeRedisTest {
...
+    assertThat(redis.lpop("droids", 1)).containsExactly("R2-D2".encodeUtf8())
+    assertThat(redis.lpop("droids", 3)).containsExactlyElementsOf(
+      listOf("K-2SO", "IG-88B", "IG-11").map { it.encodeUtf8() })
+    assertThat(redis.lpop("droids", 99)).containsExactlyElementsOf(
+      listOf("C-3PO", "BD-1", "BB-8", "4-LOM").map { it.encodeUtf8() })
+    assertThat(redis.lpop("droids", 1)).isEmpty()
+
+    // Expired key.
+    redis.lpush("droids", *droids.toTypedArray())
+    redis.pExpireAt("droids", clock.instant().toEpochMilli())
```
cashguy, line 780: If we don't expire them we can insert the same twice right?

cashguy (approved):

tso (approved):

adrw (approved):

misk-redis/src/test/kotlin/misk/redis/FakeRedisTest.kt
keeferrourke: I can add a test for it, if you want, but yes, Redis allows for that.

keeferrourke: actually I think there's sufficient coverage here for lists with duplicate elements and access before expiry.

PR #2678, opened 2023-02-03
Add configurable concurrency limits

deager:
This PR adds support for custom concurrency limit configuration. This will allow developers to configure:
* Limiting strategies (from the ones [defined](https://javadoc.io/static/com.netflix.concurrency-limits/concurrency-limits-core/0.3.6/com/netflix/concurrency/limits/limit/package-summary.html) by the original Netflix library)
* The maximum allowed concurrency limit (providing an upper bound failsafe)
* The initial limit used by the ratelimiter

Developers must opt-in to this new functionality by setting the `WebConfig#concurrency_limiter`. If this is not set, the existing behavior/defaults will be honored.

### DIFF
--- a/misk/src/main/kotlin/misk/web/MiskWebModule.kt
+++ b/misk/src/main/kotlin/misk/web/MiskWebModule.kt
@@ -164,7 +164,7 @@ class MiskWebModule(
       .to<MetricsInterceptor.Factory>()
 
     newMultibinder<ConcurrencyLimiterFactory>()
-    if (!config.concurrency_limiter_disabled && config.concurrency_limiter?.enabled != false) {
+    if (!config.concurrency_limiter_disabled && config.concurrency_limiter?.disabled != true) {
       // Shed calls when we're degraded.
       multibind<NetworkInterceptor.Factory>(MiskDefault::class)
         .to<ConcurrencyLimitsInterceptor.Factory>()
--- a/misk/src/main/kotlin/misk/web/MiskWebModule.kt
+++ b/misk/src/main/kotlin/misk/web/MiskWebModule.kt
@@ -164,12 +164,23 @@ class MiskWebModule(
       .to<MetricsInterceptor.Factory>()
 
     newMultibinder<ConcurrencyLimiterFactory>()
-    if (!config.concurrency_limiter_disabled) {
-      install(ConcurrencyLimitsModule(config))
-
+    if (!config.concurrency_limiter_disabled && config.concurrency_limiter?.enabled != false) {
       // Shed calls when we're degraded.
       multibind<NetworkInterceptor.Factory>(MiskDefault::class)
         .to<ConcurrencyLimitsInterceptor.Factory>()
+
+      // Configure custom concurrency limiting configuration. Use the defaults from the web config
+      // if not set in the limiter config.
+      val concurrencyLimiterConfig = config.concurrency_limiter?.copy(
+        // 2 is chosen somewhat arbitrarily here. Most services have one or two endpoints that
+        // receive the majority of traffic (power law, yay!), and those endpoints should _start up_
+        // without triggering the concurrency limiter at the parallelism that we configured Jetty
+        // to support.
+        initial_limit = config.concurrency_limiter.initial_limit
+          ?: (config.jetty_max_thread_pool_size / 2),
+        log_level = config.concurrency_limiter.log_level ?: config.concurrency_limiter_log_level,
+      )
+      concurrencyLimiterConfig?.let { install(ConcurrencyLimitsModule(it)) }
     }
 
     // Traces requests as they work their way through the system.
--- a/misk/src/main/kotlin/misk/web/MiskWebModule.kt
+++ b/misk/src/main/kotlin/misk/web/MiskWebModule.kt
@@ -178,7 +178,6 @@ class MiskWebModule(
         // to support.
         initial_limit = config.concurrency_limiter.initial_limit
           ?: (config.jetty_max_thread_pool_size / 2),
-        log_level = config.concurrency_limiter.log_level ?: config.concurrency_limiter_log_level,
       )
       concurrencyLimiterConfig?.let { install(ConcurrencyLimitsModule(it)) }
     }
--- a/misk/src/main/kotlin/misk/web/MiskWebModule.kt
+++ b/misk/src/main/kotlin/misk/web/MiskWebModule.kt
@@ -50,7 +50,8 @@ import misk.web.extractors.ResponseBodyFeatureBinding
 import misk.web.extractors.WebSocketFeatureBinding
 import misk.web.extractors.WebSocketListenerFeatureBinding
 import misk.web.interceptors.BeforeContentEncoding
-import misk.web.interceptors.ConcurrencyLimiterFactory
+import misk.web.concurrencylimits.ConcurrencyLimiterFactory
+import misk.web.concurrencylimits.ConcurrencyLimitsModule
 import misk.web.interceptors.ConcurrencyLimitsInterceptor
 import misk.web.interceptors.ForContentEncoding
 import misk.web.interceptors.GunzipRequestBodyInterceptor
@@ -164,6 +165,8 @@ class MiskWebModule(
 
     newMultibinder<ConcurrencyLimiterFactory>()
     if (!config.concurrency_limiter_disabled) {
+      install(ConcurrencyLimitsModule(config))
+
       // Shed calls when we're degraded.
       multibind<NetworkInterceptor.Factory>(MiskDefault::class)
         .to<ConcurrencyLimitsInterceptor.Factory>()
--- a/misk/src/main/kotlin/misk/web/WebConfig.kt
+++ b/misk/src/main/kotlin/misk/web/WebConfig.kt
@@ -230,12 +230,15 @@ data class CorsConfig(
 
 data class ConcurrencyLimiterConfig(
   /** If true, disables automatic load shedding when degraded. */
-  val enabled: Boolean = true,
+  val disabled: Boolean = false,
 
   /** The algorithm to use for determining concurrency limits. */
   val strategy: ConcurrencyLimiterStrategy = ConcurrencyLimiterStrategy.VEGAS,
 
-  /** Maximum allowed concurrency limit providing an upper bound failsafe. */
+  /**
+   * Maximum allowed concurrency limit providing an upper bound failsafe. At runtime, this will
+   * default to half of the jetty_max_thread_pool_size if not set.
+   */
   val max_concurrency: Int? = null,
 
   /** Initial limit used by the concurrency limiter. */
--- a/misk/src/main/kotlin/misk/web/WebConfig.kt
+++ b/misk/src/main/kotlin/misk/web/WebConfig.kt
@@ -2,6 +2,7 @@ package misk.web
 
 import misk.security.ssl.CertStoreConfig
 import misk.security.ssl.TrustStoreConfig
+import misk.web.concurrencylimits.ConcurrencyLimiterStrategy
 import misk.web.exceptions.ActionExceptionLogLevelConfig
 import org.slf4j.event.Level
 import wisp.config.Config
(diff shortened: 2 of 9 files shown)


misk/src/main/kotlin/misk/web/concurrencylimits/ConcurrencyLimiterFactory.kt
```diff
@@ -1,4 +1,4 @@
-package misk.web.interceptors
+package misk.web.concurrencylimits
```
deager, line 1: thought it made sense to move this, because it wasn't specific to the network interceptor, but happy to move it back!

Also happy to move all of this new code into the interceptors package!

misk/src/main/kotlin/misk/web/WebConfig.kt
```diff
@@ -90,6 +91,21 @@ data class WebConfig(
   /** If true, disables automatic load shedding when degraded. */
   val concurrency_limiter_disabled: Boolean = false,
 
+  /** The algorithm to use for determining concurrency limits. */
+  val concurrency_limiter_strategy: ConcurrencyLimiterStrategy = ConcurrencyLimiterStrategy.VEGAS,
```
deager, line 95: so the defaults here currently map to the [existing default](https://github.com/cashapp/misk/blob/1f766aea1bb093f86441a09a4e4316b5316f4616/misk/src/main/kotlin/misk/web/interceptors/ConcurrencyLimitsInterceptor.kt#L185-L197) (vegas limit, initialized at half of the thread pool size), a few questions:

1. Do we still need that code? what is the purpose of having a set of limiter factories? it looks like we only take the first one or use the default- does that ever alternate between limiters?

2. Should we put all of these `concurrency_limiter_*` configs in a separate config object?

misk/src/main/kotlin/misk/web/MiskWebModule.kt
```diff
@@ -164,6 +165,8 @@ class MiskWebModule(
 
     newMultibinder<ConcurrencyLimiterFactory>()
     if (!config.concurrency_limiter_disabled) {
+      install(ConcurrencyLimitsModule(config))
```
deager, line 168: is this the way we want to do this?

My understanding of how this works is that this will be configured through the `<service>-<environment>.yaml` config files under `skim.web`, is that right?

jdm-square: Correct.

misk/src/main/kotlin/misk/web/concurrencylimits/ConcurrencyLimiterStrategy.kt
```diff
@@ -0,0 +1,10 @@
+package misk.web.concurrencylimits
+
+enum class ConcurrencyLimiterStrategy {
```
adrw, line 3: Could some docs be added on what this strategy is and potentially where to learn more about the algos?

deager: 👍🏼👍🏼👍🏼👍🏼 this is really a first pass to validate the approach, but will follow up with tests and docs!

misk/src/main/kotlin/misk/web/MiskWebModule.kt
```diff
@@ -163,10 +164,23 @@ class MiskWebModule(
...
-    if (!config.concurrency_limiter_disabled) {
+    if (!config.concurrency_limiter_disabled && config.concurrency_limiter?.enabled != false) {
       // Shed calls when we're degraded.
       multibind<NetworkInterceptor.Factory>(MiskDefault::class)
         .to<ConcurrencyLimitsInterceptor.Factory>()
+
+      // Configure custom concurrency limiting configuration. Use the defaults from the web config
+      // if not set in the limiter config.
+      val concurrencyLimiterConfig = config.concurrency_limiter?.copy(
+        // 2 is chosen somewhat arbitrarily here. Most services have one or two endpoints that
```
deager, line 175: preserving Eric's comment from https://github.com/cashapp/misk/commit/b26ad282892578a7746749ea2728605419839129

misk/src/main/kotlin/misk/web/MiskWebModule.kt
```diff
@@ -163,10 +164,22 @@ class MiskWebModule(
       .to<MetricsInterceptor.Factory>()
 
     newMultibinder<ConcurrencyLimiterFactory>()
-    if (!config.concurrency_limiter_disabled) {
+    if (!config.concurrency_limiter_disabled && config.concurrency_limiter?.enabled != false) {
```
deager, line 167: if someone chooses to use the new configuration and wants to use that to disable limiting, they may do so.

(Happy to revert if others disagree!)

misk/src/main/kotlin/misk/web/WebConfig.kt
```diff
@@ -223,3 +227,17 @@ data class CorsConfig(
   /** A comma separated list of HTTP headers that are allowed to be exposed on the client. */
   val exposedHeaders: Array<String> = arrayOf()
 )
+
+data class ConcurrencyLimiterConfig(
+  /** If true, disables automatic load shedding when degraded. */
+  val enabled: Boolean = true,
```
achuang94, line 233: nit: maybe make this `disabled` instead, to keep it consistent with the other variable to disable the rate limiter in the WebConfig?

i think its slightly confusing to have 2 config variables controlling this, but they're the inverse of each other haha

misk/src/main/kotlin/misk/web/WebConfig.kt
```diff
@@ -223,3 +227,17 @@ data class CorsConfig(
   /** A comma separated list of HTTP headers that are allowed to be exposed on the client. */
   val exposedHeaders: Array<String> = arrayOf()
 )
+
+data class ConcurrencyLimiterConfig(
+  /** If true, disables automatic load shedding when degraded. */
```
achuang94, line 232: this comment describes this as a "disabled" variable, maybe should be: `If false, disables automatic load shedding`?

misk/src/main/kotlin/misk/web/WebConfig.kt
```diff
@@ -223,3 +227,17 @@ data class CorsConfig(
...
+  val enabled: Boolean = true,
+
+  /** The algorithm to use for determining concurrency limits. */
+  val strategy: ConcurrencyLimiterStrategy = ConcurrencyLimiterStrategy.VEGAS,
+
+  /** Maximum allowed concurrency limit providing an upper bound failsafe. */
+  val max_concurrency: Int? = null,
+
+  /** Initial limit used by the concurrency limiter. */
```
achuang94, line 241: should we document here that this defaults to half of the max thread pool size, if not specified?

misk/src/main/kotlin/misk/web/concurrencylimits/ConcurrencyLimitsModule.kt
```diff
@@ -0,0 +1,79 @@
...
+
+    ConcurrencyLimiterStrategy.SETTABLE ->
+      config.max_concurrency
+        ?.let { SettableLimit.startingAt(it) }
+        ?: throw IllegalStateException("SettableLimit algorithm requires 'maxConcurrency'")
+
+    ConcurrencyLimiterStrategy.FIXED ->
+      config.max_concurrency
+        ?.let { FixedLimit.of(it) }
+        ?: throw IllegalStateException("FixedLimitx algorithm requires 'maxConcurrency'")
```
achuang94, line 77:
```suggestion
        ?: throw IllegalStateException("FixedLimit algorithm requires 'maxConcurrency'")
```

misk/src/main/kotlin/misk/web/concurrencylimits/ConcurrencyLimiterStrategy.kt
adrw: Thanks for these thorough docs!

misk/src/main/kotlin/misk/web/WebConfig.kt
adrw, line 232:
```suggestion
  /** If false, disables automatic load shedding when degraded. */
```

misk/src/main/kotlin/misk/web/WebConfig.kt
```diff
@@ -223,3 +227,17 @@ data class CorsConfig(
   /** A comma separated list of HTTP headers that are allowed to be exposed on the client. */
   val exposedHeaders: Array<String> = arrayOf()
 )
+
+data class ConcurrencyLimiterConfig(
+  /** If true, disables automatic load shedding when degraded. */
+  val enabled: Boolean = true,
+
+  /** The algorithm to use for determining concurrency limits. */
+  val strategy: ConcurrencyLimiterStrategy = ConcurrencyLimiterStrategy.VEGAS,
```
adrw, line 236: Is this the same as the current default/only algo?

misk/src/main/kotlin/misk/web/WebConfig.kt
```diff
@@ -93,6 +94,9 @@ data class WebConfig(
   /** The level of log when concurrency shedding. */
   val concurrency_limiter_log_level: Level = Level.ERROR,
```
adrw, line 95: nit: It would be nice to move `concurrency_limiter_disabled` and `concurrency_limiter_log_level` into the new `concurrency_limiter: ConcurrencyLimiterConfig` backwards compatibly for better config organization. This would take a few PRs / rolling the changes out.

misk/src/main/kotlin/misk/web/MiskWebModule.kt
adrw: Could the log level also be included in your new config then?

adrw (approved):

misk/src/main/kotlin/misk/web/MiskWebModule.kt
deager: I wasn't quite sure how to handle backwards compatibility for that- would we want to make it nullable in the `ConcurrencyLimiterConfig` then fall back to the original? it looks like the only place it's being used is the in the interceptor factory, which i was somewhat reluctant to touch

misk/src/main/kotlin/misk/web/WebConfig.kt
deager: agreed! how would you recommend we do so? I wanted to touch as little as possible, but i'm happy to add the additional configurations. are you suggesting i'd go through services using this configuration and change them?

misk/src/main/kotlin/misk/web/WebConfig.kt
```diff
@@ -93,6 +94,19 @@ data class WebConfig(
...
 
+  /* Custom configuration for calculating concurrency limits */
+  val concurrency_limiter: ConcurrencyLimiterConfig? = ConcurrencyLimiterConfig(
+    disabled = concurrency_limiter_disabled,
```
deager, line 99: prefer to use the existing configurations (and their defaults)

misk/src/main/kotlin/misk/web/WebConfig.kt
deager: added documentation of this where the default is (above)!

misk/src/main/kotlin/misk/web/WebConfig.kt
adrw: Yes, I'd land equivalents in your new `ConcurrencyLimiterConfig` and update any callsites to look first at the new field in `ConcurrencyLimiterConfig` and then in the old one if not present. Then, update any services using the old fields to use the new ones. Then, delete the old fields from `WebConfig`.

misk/src/main/kotlin/misk/web/MiskWebModule.kt
adrw: Yes, I'd make it nullable in the new one and fallback, then update service code to use new field, then delete the old field and make it non-null with a default parameter.

misk/src/main/kotlin/misk/web/interceptors/ConcurrencyLimitsInterceptor.kt
```diff
@@ -172,7 +173,8 @@ internal class ConcurrencyLimitsInterceptor internal constructor(
         action = action,
         defaultLimiter = createLimiterForAction(action, quotaPath = null),
         clock = clock,
-        logLevel = config.concurrency_limiter_log_level
+        logLevel = config.concurrency_limiter?.log_level
```
deager, line 176: the `concurrency_limiter.log_level` will default to `config.concurrency_limiter_log_level` if `concurrency_limiter` isn't specified

misk/src/main/kotlin/misk/web/WebConfig.kt
deager: any concerns about Misk being _open source_?

misk/src/main/kotlin/misk/web/MiskWebModule.kt
deager: So I added it to the new config- didn't make it nullable, but gave it the same default in the class definition and created a default `ConcurrencyLimiterConfig` within `WebConfig` that uses `concurrency_limiter_disabled` to set `disabled`

PR #2679, opened 2023-02-06
Expands Metric.histogram documentation and its backing Prometheus type. 

frojasg:


misk-metrics/src/main/kotlin/misk/metrics/Metrics.kt
```diff
@@ -47,10 +47,16 @@ interface Metrics {
    *
    * Deprecated: if you really need a summary metric, use [misk.metrics.v2.Metrics.summary] instead.
    *
-   * For legacy reasons, this function is called `histogram(...)` but it's not backed
-   * by a histogram which is confusing.
+   * For legacy reasons this function is called histogram(...) but its not backed by a histogram
```
MatthewDolan, line 50:
```suggestion
   * For legacy reasons this function is called histogram(...) but it's not backed by a histogram
```

MatthewDolan (approved):

PR #2680, opened 2023-02-08
Return a 415 to help decrease developer debugging time

tso:
This commit changes the behavior of misk action matching to return a 415 rather than a 404 when a request matches a URL but not the content type. For example, if there's an action defined as a `POST /hello application/json` and we make a request to `POST /hello text/plain` we will now return a 415 as it's likely the caller is simply using the wrong media types.

### DIFF
--- a/misk/src/main/kotlin/misk/web/actions/NotFoundAction.kt
+++ b/misk/src/main/kotlin/misk/web/actions/NotFoundAction.kt
@@ -2,6 +2,7 @@ package misk.web.actions
 
 import misk.scope.ActionScoped
 import misk.security.authz.Unauthenticated
+import misk.web.BoundActionMatch
 import misk.web.Get
 import misk.web.HttpCall
 import misk.web.PathParam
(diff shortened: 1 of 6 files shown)


misk/src/main/kotlin/misk/web/actions/NotFoundAction.kt
```diff
@@ -20,44 +21,35 @@ import javax.inject.Provider
...
   @Post("/{path:.*}")
   @RequestContentType(MediaTypes.ALL)
   @ResponseContentType(MediaTypes.ALL)
   @Unauthenticated
   fun notFound(@PathParam path: String): Response<ResponseBody> {
+    val httpCall = clientHttpCall.get()
+    val alternativeActions = urlOnlyMatches(httpCall)
+    // If there are any actions that match on URL the client is likely using the wrong
+    // content type, so we'll return a "415 Unsupported Media Type".
+    val statusCode = if (alternativeActions.isNotEmpty()) 415 else 404
```
ean5533, line 38: Proposal: what if we also checked the HTTP verb (`httpCall.dispatchMechanism`) to see if we have a bound action for another verb and, if so, return `405 Method Not Allowed`?

ean5533 (approved):
Definite improvement! Curious to hear your opinion about checking verb as well

PR #2683, opened 2023-02-14
Upgrade tracing modules for datadog version 1.10.1

mpeyper:
The main change here is fixing things up for the new `addScopeListener` function on the `InternalTracer` and the removal of the `ScopeListener` interface. Other minor import changes are also included.

I also added source and target compatibility setting for the `JavaCompile` task. I don't know enough about it but it caused issues when trying to use local builds of `misk` in other projects to test things out. The solution came from [here](https://stackoverflow.com/questions/72117858/incompatible-because-this-component-declares-a-component-compatible-with-java-11).

**BREAKING CHANGE:** this is a potentially breaking change if consumers resolve a different version of the datadog libraries which can cause runtime errors when classes or class members have been changed. The version of the `dd-java-agent` used to instrument the code should also be updated to match the version of the datadog libraries being used.


misk-datadog/src/main/kotlin/misk/tracing/backends/datadog/MDCScopeListener.kt
```diff
@@ -1,34 +1,41 @@
...
-class MDCScopeListener : ScopeListener {
+// Generic exceptions are caught here to avoid crashing the critical path when setting MDC fails.
+@Suppress("TooGenericExceptionCaught")
+internal class MDCScopeListener {
   companion object {
     val log = getLogger<MDCScopeListener>()
     const val MDC_TRACE_ID = "trace_id"
     const val MDC_SPAN_ID = "span_id"
+    const val MDC_DD_TRACE_ID = "dd.trace_id"
+    const val MDC_DD_SPAN_ID = "dd.span_id"
```
mpeyper, line 19: Some of our internal libraries are also adding the trace and span IDs to these keys as well, so I added them here too. Happy to remove them if preferred not to add them.

chris-ryan-square (approved):

mpeyper:
Marking this as a draft until we are ready to merge this.

build.gradle.kts
```diff
@@ -50,6 +50,11 @@ subprojects {
     maven(url = "https://s3-us-west-2.amazonaws.com/dynamodb-local/release")
   }
 
+  tasks.withType<JavaCompile> {
+    sourceCompatibility = "11"
+    targetCompatibility = "11"
+  }
+
```
mpeyper, line 57: I need to remove this now

PR #2685, opened 2023-02-14
Add support for HLEN to misk-redis

keeferrourke:
Adds support for the [HLEN command](https://redis.io/commands/hlen/).

```
HLEN key
```

### DIFF
--- a/misk-redis/src/main/kotlin/misk/redis/FakeRedis.kt
+++ b/misk-redis/src/main/kotlin/misk/redis/FakeRedis.kt
@@ -133,6 +133,12 @@ class FakeRedis : Redis {
     }
   }
 
+  override fun hlen(key: String): Long {
+    synchronized(lock) {
+      return hKeyValueStore[key]?.data?.size?.toLong() ?: 0L
+    }
+  }
+
   override fun hmget(key: String, vararg fields: String): List<ByteString?> {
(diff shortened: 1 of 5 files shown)


mslogar-squareup (approved):

jdm-square (approved):

misk-redis/src/main/kotlin/misk/redis/FakeRedis.kt
```diff
@@ -133,6 +133,12 @@ class FakeRedis : Redis {
     }
   }
 
+  override fun hlen(key: String): Long {
+    synchronized(lock) {
```
tso, line 137: You should change this to just use `@Synchronized` on every function tbh, would avoid some indents

tso (approved):

adrw (approved):

misk-redis/src/main/kotlin/misk/redis/FakeRedis.kt
keeferrourke: Ha true. Honestly, just following the existing convention. Will remove the lock in a follow up.

keeferrourke: https://github.com/cashapp/misk/pull/2686

PR #2687, opened 2023-02-14
Fix warnings in RealRedis

keeferrourke:
1. Removes redundant docs (which often referred to wrong types).
2. Removes redundant non-null assertions (!!)
3. Fixes a possible NPE bug where Jedis can return null, but Kotlin couldn't infer the platform type properly.

### DIFF
--- a/misk-redis/src/main/kotlin/misk/redis/RealRedis.kt
+++ b/misk-redis/src/main/kotlin/misk/redis/RealRedis.kt
@@ -9,23 +9,23 @@ import redis.clients.jedis.args.ListDirection
 import redis.clients.jedis.params.SetParams
 import java.time.Duration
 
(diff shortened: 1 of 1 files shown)


misk-redis/src/main/kotlin/misk/redis/RealRedis.kt
```diff
@@ -226,7 +214,7 @@ class RealRedis(private val jedisPool: JedisPool) : Redis {
         destinationKey.toByteArray(charset),
         from,
         to
-      ).toByteString()
+      )?.toByteString()
```
keeferrourke, line 217: This was an actual bug.

jdm-square (approved):
Thanks for continually improving this code!

tso (approved):

adrw (approved):

PR #2689, opened 2023-02-15
Improve Redis#lpop and rpop

keeferrourke:


misk-redis/src/main/kotlin/misk/redis/RealRedis.kt
```diff
@@ -232,13 +232,25 @@ class RealRedis(private val jedisPool: JedisPool) : Redis {
 
   override fun lpop(key: String, count: Int): List<ByteString?> {
     return jedisPool.resource.use { jedis ->
-      jedis.lpop(key.toByteArray(charset), count).map { it?.toByteString() }
+      jedis.lpop(key.toByteArray(charset), count)?.map { it?.toByteString() } ?: emptyList()
```
keeferrourke, line 235: jedis can return a null list

misk-redis/src/main/kotlin/misk/redis/RealRedis.kt
```diff
@@ -232,13 +232,25 @@ class RealRedis(private val jedisPool: JedisPool) : Redis {
...
+  override fun lpop(key: String): ByteString? {
+    return jedisPool.resource.use { jedis ->
+      jedis.lpop(key.toByteArray(charset))?.toByteString()
     }
   }
 
   override fun rpop(key: String, count: Int): List<ByteString?> {
     return jedisPool.resource.use { jedis ->
-      jedis.rpop(key.toByteArray(charset), count).map { it?.toByteString() }
+      jedis.rpop(key.toByteArray(charset), count)?.map { it?.toByteString() } ?: emptyList()
```
keeferrourke, line 247: jedis can return a null list

adrw (approved):

misk-redis/src/main/kotlin/misk/redis/RealRedis.kt
cashguy: Will rpop with a count still throw the same error? Looks like it's using the same API? Do I need to switch to use the one without count?

cashguy (approved):

misk-redis/src/main/kotlin/misk/redis/RealRedis.kt
keeferrourke: Yeah since we're working off a v5.0 instance, we'll need to use the rpop without count.

I _guess_ I could change the underlying impl to use the rpop without count, but I don't want to forgo efficiency for clients of newer Redis.

PR #2694, opened 2023-02-17
Introduces `PeakGauge`

cconroy:
A `PeakGauge` is a variant of a regular gauge that resets to an initial value of 0 after a metric collection.

This is useful for accurately capturing maximum observed values over time. In contrast to the histogram maximum which tracks the maximum value in its sampling window. That sampling window typically covers multiple metric collections.

### DIFF
--- a/misk-metrics/src/main/kotlin/misk/metrics/v2/PeakGauge.kt
+++ b/misk-metrics/src/main/kotlin/misk/metrics/v2/PeakGauge.kt
@@ -1,0 +1,74 @@
+package misk.metrics.v2
+
+import com.google.common.util.concurrent.AtomicDouble
+import io.prometheus.client.SimpleCollector
+import javax.annotation.concurrent.ThreadSafe
+
+/**
+ * A peak gauge is a variant of a [io.prometheus.client.Gauge] that resets to an initial value of 0
+ * after a metric collection.
+ *
+ * This is useful for accurately capturing maximum observed values over time. In contrast to the
+ * histogram maximum which tracks the maximum value in its sampling window. That sampling window
+ * typically covers multiple metric collections.
+ */
+@ThreadSafe
+class PeakGauge  : SimpleCollector<PeakGauge.Child> {
+  constructor(builder: Builder) : super(builder)
+
+  class Builder : SimpleCollector.Builder<Builder, PeakGauge>() {
+    override fun create(): PeakGauge {
+      return PeakGauge(this)
+    }
+  }
+
+  class Child {
+    // For simplicity, using an atomic double with an initial value of 0 here.
+    //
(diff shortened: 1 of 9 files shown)


adrw (approved):

cconroy:
An alternate name for consideration `HighWatermarkGauge`...it's longer though and I don't think it is necessarily clearer...thoughts?

misk-metrics/src/main/kotlin/misk/metrics/v2/PeakGauge.kt
```diff
@@ -0,0 +1,74 @@
...
+    //
+    // If we cared to differentiate between the initial value due to no samples vs having
+    // samples that are equal or less than 0, then we would need to also track whether we
+    // have received an update since the last reset.
+    private val value = AtomicDouble()
+
+    /**
+     * Updates the stored value if the new value is greater.
+     *
+     *
```
arunparmar-square, line 36: nit: remove empty new lines

arunparmar-square (approved):

frojasg (approved):
I would go with `HighWatermarkGauge` in case at some point we need something similar for the lowest value and we can use `LowWatermarkGauge`. Otherwise, the naming will not be symmetrical

cconroy:
Another thing to consider: should we go ahead and add some `.peak` suffix or similar to the metric name to easily distinguish from normal gauges? We likely want to do this at the outset if we want that to avoid an intermediate migration step.

Unless there's a good reason _not_ to add that suffix, then may as well for clarity?

cconroy:
I'm assuming it makes sense to just add this to the v2 metrics since presumably everyone should be moving to that at some point...Though happy to add to v1 if that would be useful to someone.

frojasg:
It's fine to add it just to v2. 

I'm not sure about the metric name suffix. Adding the peak makes sense to me for clarity, but I'm not sure of the consequences. Should we ask the observability team?

cconroy:
> I'm not sure about the metric name suffix. Adding the peak makes sense to me for clarity, but I'm not sure of the consequences. Should we ask the observability team?

Observability recommended against the suffix. They7 cited this [article](https://www.robustperception.io/on-the-naming-of-things/) on metrics best practices, specifically this section:

> Similarly don’t put the type of the metric in the name, such as gauge, counter, summary or map. This is already implied by the existing naming scheme, adds no information and will cause confusion as you process and write rules based on the metrics. For example whether you calculated request rate from a Counter or Summary is irrelevant.

Now, this is a somewhat different question than that snippet is addressing since a `PeakGauge` is not a different metric _type_ but rather some new behavior on top of an existing type. But, we're going to start with the notion that folks will just have to manually make descriptive names for their usages.

PR #2695, opened 2023-02-21
Upgrade the prometheus client + wisp

chris-ryan-square:


lyonlai (approved):

misk/src/test/kotlin/misk/web/interceptors/ConcurrencyLimitsInterceptorTest.kt
```diff
@@ -216,7 +216,7 @@ class ConcurrencyLimitsInterceptorTest {
 
   private fun callSuccessCount(id: String): Double {
     return prometheusRegistry.getSampleValue(
-      "concurrency_limits_outcomes",
+      "concurrency_limits_outcomes_total",
```
adrw, line 219: If the Prometheus bump requires downstream metrics name changes, should service owners expect to need to update downstream CI alert formulas (ie in SignalFX, Datadog...) given changed metric names?

chris-ryan-square: Yes.  According to prometheus: https://github.com/prometheus/client_java/releases/tag/parent-0.10.0, this was done to done to match OpenMetrics.

buildSrc/src/main/kotlin/Dependencies.kt
```diff
@@ -96,9 +96,10 @@ object Dependencies {
   val openTracingOkHttp = "io.opentracing.contrib:opentracing-okhttp3:3.0.0"
   val openTracingUtil = "io.opentracing:opentracing-util:0.33.0"
   val postgresql = "org.postgresql:postgresql:42.2.23"
-  val prometheusClient = "io.prometheus:simpleclient:0.9.0"
-  val prometheusHotspot = "io.prometheus:simpleclient_hotspot:0.9.0"
-  val prometheusHttpserver = "io.prometheus:simpleclient_httpserver:0.9.0"
+  val prometheusClientBom = "io.prometheus:simpleclient_bom:0.15.0"
```
ThePumpingLemma, line 99: `0.16.0`?

misk-metrics-testing/src/main/kotlin/misk/metrics/FakeMetrics.kt
```diff
@@ -103,19 +104,22 @@ class FakeMetrics @Inject internal constructor(
...
-    sampleName: String = name
+    sampleName: String? = null
   ): Sample? {
     val metricFamilySamples = registry.metricFamilySamples()
       .asSequence()
       .firstOrNull { it.name == name }
       ?: return null
 
+    val familySampleName = sampleName
+      ?: if (metricFamilySamples.type == Collector.Type.COUNTER) "${name}_total" else name
```
mpeyper, line 115: Problem with the tests is that calling `metrics.counter("gets", "-", labelNames = listOf("status"))` creates a  metric family named `gets` but the sample name is `gets_total` so unless the suffix is added for counter metrics, the `FakeMetrics.get("gets"...)` cannot find the samples and `FakeMetrics.get("gets_total"...) cannot find the metric family.

This fix is not particularly pretty (the better fix may be to introduce new getters that apply the suffixes like for `histogram` values), but went this path the cause the least disruption in the library API for a smoother upgrade for consumers.

The same solution has been applied to the v2 metrics.

misk-metrics-testing/src/test/kotlin/misk/metrics/FakeMetricsTest.kt
```diff
@@ -101,30 +101,30 @@ class FakeMetricsTest {
...
+
+    histogram.record(600.0)
+    assertThat(metrics.histogramP50("call_times")).isEqualTo(500.0)
+    assertThat(metrics.histogramP99("call_times")).isEqualTo(600.0)
   }
 
   @Test
   internal fun `get all samples`() {
-    metrics.counter("counter", "-", listOf("foo")).labels("bar").inc()
+    metrics.counter("counter_total", "-", listOf("foo")).labels("bar").inc()
```
mpeyper, line 121: Technically this line didn't need to change, just the sample name below [see this comment for details](https://github.com/cashapp/misk/pull/2695/files#r1113709699)

PR #2701, opened 2023-02-23
Add bindConstant to ActionScopedProviderModule

JGulbronson:
At Faire we want the ability to have a default binding for a key, with the ability to override it when entering the scope. Generally we're using `Optional` to accomplish this, and the default value is an empty `Optional`. This makes that possible with less boilerplate, avoiding having to create a new class for every `Optional<T>` you want to bind.


adrw:
Do you mind posting some before/after example code of how this change improves the developer ergonomics?

swankjesse (approved):

JGulbronson:
Sure. Before, you'd have to do something like this:

```kotlin
class EmptyOptionalProvider : ActionScopedProvider<Optional<String>> {
  override fun get() = Optional.empty()
}

bindProvider(object: TypeLiteral<Optional<String>>() {}, EmptyOptionalProvider::class)
```

Now you can do
```kotlin
bindConstant(object: TypeLiteral<Optional<String>>() {}, Optional.empty())
```

No more having to redefine a provider for every class!

PR #2703, opened 2023-02-24
Explicitly set Java compile target to 11

staktrace:
This helps when including misk as an includeBuild in a larger gradle superproject to test changes locally. Otherwise the superproject build can failed with errors about being unable to find the right variant.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -1,3 +1,4 @@
+import org.gradle.api.tasks.compile.JavaCompile
 import org.gradle.api.tasks.testing.logging.TestExceptionFormat
 import org.jetbrains.dokka.gradle.DokkaTask
 import org.jetbrains.kotlin.gradle.plugin.KotlinPluginWrapper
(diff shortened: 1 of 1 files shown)


build.gradle.kts
```diff
@@ -51,21 +52,14 @@ subprojects {
 
   // Only apply if the project has the kotlin plugin added:
   plugins.withType<KotlinPluginWrapper> {
-    val compileKotlin by tasks.getting(KotlinCompile::class) {
+    tasks.withType<KotlinCompile> {
       kotlinOptions {
         jvmTarget = "11"
-
-        // TODO(alec): Enable again once Environment enum is deleted
```
staktrace, line 58: I removed this TODO since it appears to be done

build.gradle.kts
```diff
@@ -51,21 +52,14 @@ subprojects {
 
   // Only apply if the project has the kotlin plugin added:
   plugins.withType<KotlinPluginWrapper> {
-    val compileKotlin by tasks.getting(KotlinCompile::class) {
```
staktrace, line 54: I also changed the syntax of this KotlinCompile configuration to avoid duplication and keep things more consistent.

lkerford (approved):

PR #2708, opened 2023-03-03
Introduces pause detection

cconroy:
Adds a pause detector thread that sleeps for a defined resolution and records elapsed overage of that sleep time as pause time.

The detector can emit logs with configurable pause time thresholds for INFO/WARN/ERROR

The detector tracks pause time in a `Summary` and a `PeakGauge` so that operators can track total pause time, pause distribution, and the maximum pause time per collection interval.

### DIFF
--- a/misk/src/main/kotlin/misk/perf/PauseDetector.kt
+++ b/misk/src/main/kotlin/misk/perf/PauseDetector.kt
@@ -1,0 +1,140 @@
+package misk.perf
+
+import com.google.common.base.Ticker
+import com.google.common.util.concurrent.AbstractExecutionThreadService
+import io.prometheus.client.Summary
+import misk.concurrent.Sleeper
+import misk.metrics.v2.Metrics
+import misk.metrics.v2.PeakGauge
+import org.slf4j.event.Level
+import wisp.logging.getLogger
+import java.time.Duration
+import java.util.concurrent.TimeUnit.MILLISECONDS
+import java.util.concurrent.TimeUnit.NANOSECONDS
+import javax.inject.Inject
+import javax.inject.Singleton
+
+/**
+ * Detects and records pauses experienced by the VM. Garbage collection is a common source of
+ * pauses, but pauses can be caused by many things: CPU throttling, excessive swapping, excessive
+ * active threads, etc.
+ *
+ * Heavily inspired by [jhiccup](https://www.azulsystems.com/jHiccup) which is
+ * [public domain](https://creativecommons.org/publicdomain/zero/1.0/)
+ */
+@Singleton
+internal class PauseDetector @Inject constructor(
+  private val config: PauseDetectorConfig,
+  private val ticker: Ticker,
+  private val sleeper: Sleeper,
+  val metrics: Metrics,
+) : AbstractExecutionThreadService() {
+
+  /** Log levels by pause time sorted by severity descending*/
+  private val logLevels: List<LogLevels> = listOf(
+    LogLevels(config.logErrorMillis, Level.ERROR),
+    LogLevels(config.logWarnMillis, Level.WARN),
+    LogLevels(config.logInfoMillis, Level.INFO)
+  )
+
+  /**
+   * Tracks the distribution and total amount of pause time observed.
+   *
+   * (Using a summary instead of a histogram to prefer more accurate node-local data over
+   * more accurate aggregations)
+   */
+  private val pauseSummary: Summary =
+    metrics.summary("jvm_pause_time_summary_ms", "Summary in ms of pause time", listOf())
+
+  /** Tracks peak pause time. */
+  private val pausePeak: PeakGauge =
+    metrics.peakGauge("jvm_pause_time_peak_ms", "Peak gauge of pause time", listOf())
+
+  // No synchronization is necessary for these variables: they are only ever accessed by the
+  // detector thread itself OR by a test harness thread.
+  //
+  // (They are also initialized in the constructor, but that initialization has a happens-before
+  // relationship to the [Thread.start] that starts the thread).
+  /** Used to prevent generating early noise (for example in the first few seconds of startup) */
+  private var shortestObservedDeltaTimeNsec = Long.MAX_VALUE
+  /** Tracks the start time of the last invocation to [sleep] */
+  private var startTime: Long = 0
+
+  override fun run() {
+    while (isRunning) {
+      // NB: Broken into two different functions so that we can inject time delays for testing
+      // purposes.
+      sleep()
+      check()
+    }
+  }
+
+  /**
+   * Sleeps for the configured resolution
+   *
+   * NB: [sleep] and [check] must always be invoked from the same thread.
+   */
+  internal fun sleep() {
+    startTime = ticker.read()
+    if (config.resolutionMillis != 0L) {
+      sleeper.sleep(Duration.ofMillis(config.resolutionMillis))
+    }
+  }
+
+  /**
+   * Check the elapsed time after an invocation to [sleep]. Elapsed time beyond the configured
+   * resolution and the shortest observed delta is recorded as pause time.
+   *
+   * NB: [sleep] and [check] must always be invoked from the same thread.
+   */
+  internal fun check(): PauseResults {
+    val stopTime: Long = ticker.read()
+    val deltaTimeNsec: Long = (stopTime - startTime)
+
+    // We expect a monotonic ticker that should measure at least the resolution time between
+    // invocations. Guard against a non-monotonic or otherwise "fast" ticker from polluting results.
+    if (deltaTimeNsec < MILLISECONDS.toNanos(config.resolutionMillis)) {
+      val decreaseMs = config.resolutionMillis - NANOSECONDS.toMillis(deltaTimeNsec)
+      logger.info("Observed a negative pause time of ${decreaseMs}ms. Non-monotonic ticker?")
+    } else if (deltaTimeNsec < shortestObservedDeltaTimeNsec) {
+      shortestObservedDeltaTimeNsec = deltaTimeNsec
+    }
+
+    val pauseTimeNsec = deltaTimeNsec - shortestObservedDeltaTimeNsec
+    val pauseMillis = NANOSECONDS.toMillis(pauseTimeNsec)
+    if (pauseMillis >= config.metricsUpdateFloor) {
+      val pauseMillisDouble = pauseMillis.toDouble()
+      pauseSummary.observe(pauseMillisDouble)
+      pausePeak.record(pauseMillisDouble)
+    }
+
+    val level = getLoggingLevel(pauseMillis)
+    if (level != null) {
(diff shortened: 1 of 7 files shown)


misk/src/main/kotlin/misk/perf/PauseDetectorThread.kt
```diff
@@ -0,0 +1,155 @@
...
+    name = "PauseDetector"
+
+    logLevels = listOf(
+      LogLevels(config.logErrorMillis, Level.ERROR),
+      LogLevels(config.logWarnMillis, Level.WARN),
+      LogLevels(config.logInfoMillis, Level.INFO)
+    )
+
+    pauseSummary =
+      metrics.summary("jvm_pause_time_summary_ms", "Summary in ms of pause time", listOf())
```
cconroy, line 73: Seems like this fits with our metric conventions, but please let me know if there's a better name structure to use here!

FWIW `framework.perf.pause` is the metric path used by the Square version

frojasg (approved):
This looks great! Thank for the change!

misk/src/main/kotlin/misk/perf/PauseDetectorThread.kt
```diff
@@ -0,0 +1,155 @@
...
+) : Thread() {
+  private val logger = getLogger<PauseDetectorThread>()
+
+  /** Log levels by pause time sorted by severity descending*/
+  private val logLevels: List<LogLevels>
+
+  /**
+   * Tracks the distribution and total amount of pause time observed.
+   *
+   * (Using a summary instead of a histogram to prefer more accurate node-local data over
```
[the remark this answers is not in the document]
frojasg: TIL: you can't group by pod with histogram metrics 😅

misk/src/main/kotlin/misk/perf/PauseDetectorThread.kt
```diff
@@ -0,0 +1,155 @@
...
+
+  /** Tracks peak pause time. */
+  private val pausePeak: PeakGauge
+
+  /* If false, do not schedule another iteration. */
+  @Volatile
+  private var keepRunning = true
+
+
+  // No synchronization is necessary for these variables: they are only ever accessed by the
```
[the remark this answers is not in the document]
frojasg: if you make this class `internal` you can ensure nobody outside misk can use this class, which can alleviate some of the concern.

misk/src/main/kotlin/misk/perf/PauseDetectorConfig.kt
```diff
@@ -0,0 +1,21 @@
...
+
+/**
+ * Configuration for the [PauseDetectorThread]
+ */
+data class PauseDetectorConfig(
+  /** The delay between detector runs. If 0, the detector runs in a spin loop. */
+  val resolutionMillis: Long = 1,
+
+  /** The minimum number of millis paused before logging at info. -1 for never. */
+  val logInfoMillis: Long = 1000,
```
swankjesse, line 11: 👍🏻  good default

misk/src/main/kotlin/misk/perf/PauseDetectorThread.kt
```diff
@@ -0,0 +1,155 @@
...
+/**
+ * Detects and records pauses experienced by the VM. Garbage collection is a common source of
+ * pauses, but pauses can be caused by many things: CPU throttling, excessive swapping, excessive
+ * active threads, etc.
+ *
+ * Heavily inspired by [jhiccup](https://www.azulsystems.com/jHiccup) which is
+ * [public domain](https://creativecommons.org/publicdomain/zero/1.0/)
+ */
+@Singleton
+class PauseDetectorThread @Inject constructor(
```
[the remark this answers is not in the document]
swankjesse: Rather than extending `Thread`, consider extending `AbstractExecutionThreadService` and get Misk to manage your lifecycle for free

misk/src/main/kotlin/misk/perf/PauseDetectorThread.kt
```diff
@@ -0,0 +1,155 @@
...
+   * more accurate aggregations)
+   */
+  private val pauseSummary: Summary
+
+  /** Tracks peak pause time. */
+  private val pausePeak: PeakGauge
+
+  /* If false, do not schedule another iteration. */
+  @Volatile
+  private var keepRunning = true
```
swankjesse, line 49: AbstractExecutionThreadService gives you this for freeeee

misk/src/test/kotlin/misk/perf/PauseDetectorThreadTest.kt
```diff
@@ -0,0 +1,128 @@
...
+  class TestModule : KAbstractModule() {
+    override fun configure() {
+      // Wire up the detector
+      val config = PauseDetectorConfig(
+        resolutionMillis = 1,
+        logInfoMillis = 100,
+        logWarnMillis = 1000,
+        logErrorMillis = 10000
+      )
+      // NB: We are intentionally _not_ installing the module
```
swankjesse, line 114: Nice comment

misk/src/main/kotlin/misk/perf/PauseDetectorThread.kt
```diff
@@ -0,0 +1,155 @@
...
+
+    pauseSummary =
+      metrics.summary("jvm_pause_time_summary_ms", "Summary in ms of pause time", listOf())
+    pausePeak =
+      metrics.peakGauge("jvm_pause_time_peak_ms", "Peak gauge of pause time", listOf())
+  }
+
+  override fun run() {
+    while (keepRunning) {
+      // NB: Broken into two different functions so that we can inject time delays for testing
```
swankjesse, line 80: Nice comment

swankjesse (approved):

adrw (approved):

misk/src/main/kotlin/misk/perf/PauseDetectorThread.kt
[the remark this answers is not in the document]
adrw: +1, extending `AbstractExecutionThreadService` would be more inline with the rest of similar Misk infra.

misk/src/main/kotlin/misk/perf/PauseDetectorThread.kt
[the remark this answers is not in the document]
cconroy: Done!

misk/src/main/kotlin/misk/perf/PauseDetector.kt
```diff
@@ -0,0 +1,137 @@
...
+ * [public domain](https://creativecommons.org/publicdomain/zero/1.0/)
+ */
+@Singleton
+internal class PauseDetector @Inject constructor(
+  private val config: PauseDetectorConfig,
+  private val ticker: Ticker,
+  private val sleeper: Sleeper,
+  val metrics: Metrics,
+) : AbstractExecutionThreadService() {
+  private val logger = getLogger<PauseDetector>()
```
adrw, line 32: Could this be in a companion object so it's static?

adrw: the logLevels list and other static fields too

misk/src/test/kotlin/misk/perf/PauseDetectorTest.kt
```diff
@@ -0,0 +1,128 @@
...
+import misk.testing.MiskTest
+import misk.testing.MiskTestModule
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.jupiter.api.BeforeEach
+import org.junit.jupiter.api.Test
+import wisp.logging.LogCollector
+import java.util.concurrent.TimeUnit
+import javax.inject.Inject
+
+@MiskTest(startService = true) // NB: only starting services here to get log collection to work.
```
adrw, line 21: What does NB mean? For a longer comment line maybe put above the annotation

misk/src/test/kotlin/misk/perf/PauseDetectorTest.kt
```diff
@@ -0,0 +1,128 @@
...
+  class TestModule : KAbstractModule() {
+    override fun configure() {
+      // Wire up the detector
+      val config = PauseDetectorConfig(
+        resolutionMillis = 1,
+        logInfoMillis = 100,
+        logWarnMillis = 1000,
+        logErrorMillis = 10000
+      )
+      // NB: We are intentionally _not_ installing the module
```
adrw, line 114: What does NB mean?

adrw (approved):
LGTM

misk/src/test/kotlin/misk/perf/PauseDetectorTest.kt
cconroy: https://en.wikipedia.org/wiki/Nota_bene

(This is a convention we've used on the Square side for a long time...and I see it has worked its way into Cash as well)

misk/src/test/kotlin/misk/perf/PauseDetectorTest.kt
cconroy: Nota bene (linked in another comment). Done

cconroy:
@adrw I moved the logger to a companion object, but the other fields aren't static (log levels derives from config)

misk/src/main/kotlin/misk/perf/PauseDetector.kt
```diff
@@ -0,0 +1,140 @@
...
+  //
+  // (They are also initialized in the constructor, but that initialization has a happens-before
+  // relationship to the [Thread.start] that starts the thread).
+  /** Used to prevent generating early noise (for example in the first few seconds of startup) */
+  private var shortestObservedDeltaTimeNsec = Long.MAX_VALUE
+  /** Tracks the start time of the last invocation to [sleep] */
+  private var startTime: Long = 0
+
+  override fun run() {
+    while (isRunning) {
```
cconroy, line 64: @swankjesse @adrw this doesn't do what I expected..I was under the impression that `AbstractExecutionThreadService` could manage my running lifecycle for me such that all my code needs to do is check `isRunning` on each iteration. However, this fails checks for actually stopping the service when I try to wire this up in skim (see https://github.com/squareup/cash-skim/pull/1959)

I see some other uses of this abstract class implement their own `isRunning` flag as a volatile/AtomicBoolean...is that what I should do here? I feel like I must be missing the trick since I'd expect to have mandatory interfaces to implement if we have to handle the running state ourselves...

swankjesse: How long does it sleep for? Is it just that `pauseMillisDouble` is longer than Skim’s tolerance for the transition from STOPPING to STOPPED ?

cconroy: Turns out the issue was the misk testing module installs the `FakeSleeper` by default which makes any sleep calls block indefinitely unless the clock is manually managed. (In contrast, the `FakeTicker`'s implementation of sleep just moves the clock forward).

PR to address here: https://github.com/cashapp/misk/pull/2747

PR #2717, opened 2023-03-14
Improve fake clock a bit to better use wisp's

tso:


frojasg (approved):
LGTM

adrw (approved):

misk-testing/src/main/kotlin/misk/time/FakeClock.kt
```diff
@@ -1,35 +1,3 @@
...
-   * its current time and timezone. Not all days, months or years have the same length. See the
-   * documentation for [Period].
-   */
-  fun add(p: Period) = wispFakeClock.add(p = p)
-
-  fun add(n: Long, unit: TimeUnit) = wispFakeClock.add(n = n, unit = unit)
-
-  fun setNow(instant: Instant) = wispFakeClock.setNow(instant = instant)
-}
+typealias FakeClock = wisp.time.FakeClock
```
tso, line 3: Normally I would create a wrapper then delegate but in this case `java.time.Clock` isn't an interface so it doesn't really work

Issue #2606, opened 2022-11-28
Flaky tests when `idle_timeout` is 0


zpingcai:
It seems if we are passing a `WebConfig` with `idle_timeout=0` to `WebServerTestingModule` and make an HTTP request in test, the underlying connection cannot be closed gracefully.

For example, if we change the default value of `idle_timeout` to 0 in `TESTING_WEB_CONFIG` and run `WebTestClientTest`, all test cases will fail with the error message:
```
    java.util.concurrent.TimeoutException: Timeout waiting for the services to stop. The following services have not stopped: {STOPPING=[JettyService [STOPPING]]}
        at com.google.common.util.concurrent.ServiceManager$ServiceManagerState.awaitStopped(ServiceManager.java:598)
        at com.google.common.util.concurrent.ServiceManager.awaitStopped(ServiceManager.java:374)
        at misk.testing.MiskTestExtension$StopServicesAfterEach.afterEach(MiskTestExtension.kt:105)
        at misk.testing.MiskTestExtension$Callbacks.afterEach(MiskTestExtension.kt:128)
        at misk.testing.MiskTestExtension.afterEach(MiskTestExtension.kt:68)
        at org.junit.jupiter.engine.descriptor.TestMethodTestDescriptor.lambda$invokeAfterEachCallbacks$12(TestMethodTestDescriptor.java:257)
        at org.junit.jupiter.engine.descriptor.TestMethodTestDescriptor.lambda$invokeAllAfterMethodsOrCallbacks$13(TestMethodTestDescriptor.java:273)
        at org.junit.platform.engine.support.hierarchical.ThrowableCollector.execute(ThrowableCollector.java:73)
        at org.junit.jupiter.engine.descriptor.TestMethodTestDescriptor.lambda$invokeAllAfterMethodsOrCallbacks$14(TestMethodTestDescriptor.java:273)
        at java.base/java.util.ArrayList.forEach(ArrayList.java:1511)
        at org.junit.jupiter.engine.descriptor.TestMethodTestDescriptor.invokeAllAfterMethodsOrCallbacks(TestMethodTestDescriptor.java:272)
        at org.junit.jupiter.engine.descriptor.TestMethodTestDescriptor.invokeAfterEachCallbacks(TestMethodTestDescriptor.java:256)
        at org.junit.jupiter.engine.descriptor.TestMethodTestDescriptor.execute(TestMethodTestDescriptor.java:141)
[...]

PR #2718, opened 2023-03-14
Modify WebConfig to support shutdown idle timeout

zpingcai:
This PR addresses the issue https://github.com/cashapp/misk/issues/2606.

Background: Starting from jetty 10, setting `idleTimeout=0` will make HTTP connections unable to close gracefully (issue https://github.com/eclipse/jetty.project/issues/9019). We use this setting in most of our tests, and the new behaviour is blocking us from upgrading to the latest misk.

It's possible to avoid this issue by setting a positive value for `idleTimeout`. However, if the value is too small, we may observe flaky e2e tests as connections are being closed eagerly; if the value is too large, it slows down each of our tests by taking 1-2 extra seconds to close connections. In fact, tests in this repo also became slower after upgrading to jetty 10.

Since the root cause is not `idleTimeout=0` but `shutdownIdleTimeout=0` (check my comment at https://github.com/eclipse/jetty.project/issues/9019#issuecomment-1340522765), we want the flexibility to override `shutdownIdleTimeout` without affecting the existing setup for `idleTimeout`.


frojasg (approved):
LGTM, could you run  `./bin/gradle apiDump` to make the test happy?

PR #2722, opened 2023-03-15
Change FakeClock back to a real class

tso:
This solves backwards incompatibilities in terms of missing classes.

### DIFF
--- a/misk-testing/api/misk-testing.api
+++ b/misk-testing/api/misk-testing.api
@@ -235,6 +235,10 @@ public final class misk/testing/TemporaryFolderModule$DeleteTempFolder : org/jun
 	public fun afterEach (Lorg/junit/jupiter/api/extension/ExtensionContext;)V
 }
 
+public final class misk/time/FakeClock : wisp/time/FakeClock {
+	public fun <init> ()V
+}
+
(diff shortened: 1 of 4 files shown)


misk-testing/api/misk-testing.api
```diff
@@ -235,6 +235,10 @@ public final class misk/testing/TemporaryFolderModule$DeleteTempFolder : org/jun
 	public fun afterEach (Lorg/junit/jupiter/api/extension/ExtensionContext;)V
 }
 
+public final class misk/time/FakeClock : wisp/time/FakeClock {
```
tso, line 238: This is what this PR's goal is

misk-testing/src/main/kotlin/misk/time/FakeClock.kt
```diff
@@ -1,3 +1,7 @@
 package misk.time
 
-typealias FakeClock = wisp.time.FakeClock
+import javax.inject.Inject
+import javax.inject.Singleton
+
+@Singleton
+class FakeClock @Inject constructor() : wisp.time.FakeClock()
```
tso, line 7: I actually prefer this style for wrapping wisp code as we can simply inject this and bind normally

cmassoudi (approved):

frojasg (approved):

PR #2723, opened 2023-03-16
Makes RequestBodyLoggingInterceptor public

ean5533:
And, by necessity, makes a couple child classes public as well.

The rationale for this change is so that services can opt in to request logging through a mechanism other than Misk's `@LogRequestResponse`.

Example use case: a library that sets up web actions, where the user of the library may or may not want request logging. This change will allow the library to take in request logging parameters and set them at runtime rather than relying on values set in the annotation when the library was compiled.

See also an alternative approach: https://github.com/cashapp/misk/pull/2726

### DIFF
--- a/misk/src/main/kotlin/misk/web/interceptors/RequestLoggingInterceptor.kt
+++ b/misk/src/main/kotlin/misk/web/interceptors/RequestLoggingInterceptor.kt
@@ -24,7 +24,7 @@ private val logger = getLogger<RequestLoggingInterceptor>()
  * Logs request and response information for an action.
  * Timing information doesn't count time writing the response to the remote client.
  */
-class RequestLoggingInterceptor internal constructor(
+class RequestLoggingInterceptor(
   private val action: Action,
   private val caller: ActionScoped<MiskCaller?>,
   private val ticker: Ticker,
--- a/misk/api/misk.api
+++ b/misk/api/misk.api
@@ -1736,12 +1736,37 @@ public final class misk/web/interceptors/RequestBodyLoggingInterceptor$Factory :
 }
 
 public final class misk/web/interceptors/RequestLoggingInterceptor : misk/web/NetworkInterceptor {
+	public fun <init> (Lmisk/Action;Lmisk/scope/ActionScoped;Lcom/google/common/base/Ticker;Lmisk/random/ThreadLocalRandom;Lmisk/web/interceptors/LogRateLimiter;JJDDLmisk/web/interceptors/RequestResponseCapture;)V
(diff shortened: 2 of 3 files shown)


misk/src/main/kotlin/misk/web/interceptors/RequestLoggingInterceptor.kt
```diff
@@ -24,7 +24,7 @@ private val logger = getLogger<RequestLoggingInterceptor>()
  * Logs request and response information for an action.
  * Timing information doesn't count time writing the response to the remote client.
  */
-class RequestLoggingInterceptor internal constructor(
+class RequestLoggingInterceptor(
```
swankjesse, line 27: There’s a ton of parameters here, and I pity the poor code that has to call this constructor. I also dislike the premise of future maintainers introducing new parameters here and breaking you.

Would it work equivalently well to create an overload of `Factory.create()` that takes the parameters you need, but otherwise gets Guice to do the work for the rest? Then your code injects a Factory and calls create() rather than calling this constructor directly?

swankjesse (approved):

misk/src/main/kotlin/misk/web/interceptors/RequestLoggingInterceptor.kt
frojasg: Thinking about the use case presented here. I think it speak on how the current way to configure (compile time) is not enough. Maybe we should a mechanism to override that using the config files?

frojasg (approved):

misk/src/main/kotlin/misk/web/interceptors/RequestLoggingInterceptor.kt
ean5533: The `create` method has to conform to `NetworkInterceptor.Factory.create`, since that's what is going to get called [in WebActionFactory](https://github.com/cashapp/misk/blob/c2cd9a13b0f4dbe0b6e7c35c9883487f65aff183/misk/src/main/kotlin/misk/web/actions/WebActionFactory.kt#L204).

That said, I could just change this PR to instead add [the factory that I'm adding to Plasma](https://github.com/squareup/cash-plasma/blob/a855a5e562541e9e4cbdb30682f53e1304b6b7f8/client/src/main/kotlin/com/squareup/plasma/logging/LogPlasmaRequestResponse.kt#L31). It seems generally reusable; it's just a factory that uses a single (injected) global config instead of relying on config from its annotation.

ean5533: https://github.com/cashapp/misk/pull/2726

PR #2726, opened 2023-03-16
Makes RequestLoggingInterceptor configurable via (multi)binding

ean5533:
This PR adds a new constructor param to `RequestLoggingInterceptor.Factory` for overriding request logging on a per-endpoint basis by binding one or more `RequestLoggingConfig` objects. These configs, if present, will override whatever was configured in the `@LogRequestResponse` annotation.

The rationale for this change is similar to https://github.com/cashapp/misk/pull/2723: so that services can modify request logging params through a mechanism other than the `@LogRequestResponse` annotation params.

Example use case: a library that sets up web actions, where the user of the library may or may not want request logging. This change will allow the library to take in request logging parameters and set them at runtime rather than relying on values set in the annotation when the library was compiled.


misk/src/main/kotlin/misk/web/interceptors/RequestLoggingInterceptor.kt
```diff
@@ -144,3 +157,30 @@ class RequestLoggingInterceptor internal constructor(
...
+
+/**
+ * A set of per-action logging config overrides.
+ */
+data class RequestLoggingConfig(val actions: Map<String, Item>) {
+
+  /**
+   * This class should have all the same config options as [LogRequestResponse]. See that class for details.
+   */
+  data class Item(
```
frojasg, line 169: maybe `ActionLoggingConfig`

frojasg (approved):
LGTM

misk/src/main/kotlin/misk/web/interceptors/RequestLoggingInterceptor.kt
```diff
@@ -144,3 +148,45 @@ class RequestLoggingInterceptor internal constructor(
...
+ */
+data class ActionLoggingConfig(
+  val ratePerSecond: Long = 10,
+  val errorRatePerSecond: Long = 0,
+  val bodySampling: Double = 0.0,
+  val errorBodySampling: Double = 0.0,
+  val excludedEnvironments: List<String> = listOf(),
+) {
+  companion object {
+    fun fromAnnotation(logRequestResponse: LogRequestResponse): ActionLoggingConfig = ActionLoggingConfig(
```
adrw, line 169: `fun LogRequestResponse.toConfig() = ActionLogginConfig(...)` could be a bit more idiomatic Kotlin

misk/src/main/kotlin/misk/web/interceptors/RequestLoggingInterceptor.kt
```diff
@@ -144,3 +148,45 @@ class RequestLoggingInterceptor internal constructor(
...
+      ratePerSecond = logRequestResponse.ratePerSecond,
+      errorRatePerSecond = logRequestResponse.errorRatePerSecond,
+      bodySampling = logRequestResponse.bodySampling,
+      errorBodySampling = logRequestResponse.errorBodySampling,
+      excludedEnvironments = logRequestResponse.excludedEnvironments.toList(),
+    )
+
+    fun fromConfigMapOrAnnotation(
```
adrw, line 177: `fun List<RequestLoggingConfig>.findConfig(action: Action) = ActionLogginConfig(...)` could be a bit more idiomatic Kotlin since I think you can probably get the annotation from the action within the extension function (vs within the interceptor factory callsite).

adrw (commented):
Nice contribution!

PR #2727, opened 2023-03-16
Make TokenGenerator2 for backwards compatibility reasons

tso:


jdm-square (approved):

ean5533 (approved):
If you had described a PR like this to me a month ago, without context, and asked me if there was any situation under which I would approve it, I would have said no.

And yet, here I am, approving this PR.

PR #2729, opened 2023-03-17
Add client metrics to misk-redis

keeferrourke:
This PR adds client-side metrics to misk-redis.

1. Connection metrics:
   - `redis_client_max_connections` (gauge) (tracks configuration, serves a baseline to track pool usage)
   - `redis_client_max_idle_connections` (gauge) (tracks configuration, serves a baseline to track pool usage)
   - `redis_client_active_connections` (gauge) (tracks connections in use, helps track pool usage)
   - `redis_client_idle_connections` (gauge) (tracks connections not in use, helps track pool usage)
   - `redis_client_pool_destroyed_connections_total` (counter) (destroyed connections outside of host restarts are indicative of anomalies, but generally Jedis will do the right thing to replace destroyed connections)
  
   These metrics are exposed by plugging into the relevant lifecycle methods of the JedisPool.
   
2. Performance/latency metrics:
   - `redis_client_operation_time_seconds` (histogram) (tagged by command, e.g. command=hset)
      
   This works by dynamic proxy, so every Redis method call gets timed and recorded to a histogram

---

Notes: 

- Jedis connection pool metrics _can_ be exposed via JMX, as they use Apache common pool 2; this PR opts to add custom metrics for consistent prometheus style naming conventions, and to be robust in case the backing redis implementation ever changes.
- Misk-redis is a bit of a leaky abstraction, as it exposes Jedis types (Pipeline and Transaction); metrics are _not_ supported on pipeline or transaction calls right now. They will likely be added in a follow-up, but we'll need to add new APIs for better contained (and testable) transactions and pipelines.

### DIFF
--- a/misk-redis/src/main/kotlin/misk/redis/JedisPoolWithMetrics.kt
+++ b/misk-redis/src/main/kotlin/misk/redis/JedisPoolWithMetrics.kt
@@ -1,0 +1,60 @@
+package misk.redis
+
+import org.apache.commons.pool2.PooledObject
+import redis.clients.jedis.Jedis
+import redis.clients.jedis.JedisFactory
+import redis.clients.jedis.JedisPool
+import redis.clients.jedis.JedisPoolConfig
+import redis.clients.jedis.Protocol
+
+internal class JedisPoolWithMetrics(
+  private val metrics: RedisClientMetrics,
+  poolConfig: JedisPoolConfig,
+  replicationGroupConfig: RedisReplicationGroupConfig,
+  ssl: Boolean = true,
+  requiresPassword: Boolean = true,
+) : JedisPool(
+  poolConfig,
+  JedisFactoryWithMetrics(metrics, replicationGroupConfig, ssl, requiresPassword)
+) {
+  init {
+    metrics.maxTotalConnectionsGauge.set(this.maxTotal.toDouble())
+    metrics.maxIdleConnectionsGauge.set(this.maxIdle.toDouble())
+  }
+
+  override fun getResource(): Jedis {
+    metrics.activeConnectionsGauge.set(this.numActive.toDouble())
+    metrics.idleConnectionsGauge.set(this.numIdle.toDouble())
+    return super.getResource()
+  }
+
+  private class JedisFactoryWithMetrics(
+    private val metrics: RedisClientMetrics,
+    replicationGroupConfig: RedisReplicationGroupConfig,
+    ssl: Boolean = true,
+    requiresPassword: Boolean = true,
+  ) : JedisFactory(
+    /* host = */ replicationGroupConfig.writer_endpoint.hostname,
+    /* port = */ replicationGroupConfig.writer_endpoint.port,
+    /* connectionTimeout = */ replicationGroupConfig.timeout_ms,
+    /* soTimeout = */ replicationGroupConfig.timeout_ms,
+    /* password = */ replicationGroupConfig.redis_auth_password
+    .ifEmpty {
+      check(!requiresPassword) {
+        "This Redis client is configured to require an auth password, but none was provided!"
+      }
+      null
+    },
+    /* database = */ Protocol.DEFAULT_DATABASE,
+    /* clientName = */ null,
+    /* ssl = */ ssl,
+    /* sslSocketFactory = */ null,
+    /* sslParameters = */ null,
+    /* hostnameVerifier = */ null
+  ) {
+    override fun destroyObject(pooledJedis: PooledObject<Jedis>) {
+      metrics.destroyedConnectionsCounter.inc()
+      super.destroyObject(pooledJedis)
+    }
+  }
+}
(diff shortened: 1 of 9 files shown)


misk-redis/src/main/kotlin/misk/redis/JedisPoolWithMetrics.kt
```diff
@@ -0,0 +1,78 @@
...
+    metrics.idleConnectionsGauge.set(this.numIdle.toDouble())
+  }
+
+  private class JedisFactoryWithMetrics(
+    private val metrics: RedisClientMetrics,
+    replicationGroupConfig: RedisReplicationGroupConfig,
+    ssl: Boolean = true,
+    requiresPassword: Boolean = true,
+  ) : JedisFactory(
+    /* host = */ replicationGroupConfig.writer_endpoint.hostname,
```
adrw, line 55: Why are the named parameters commented out? Why not use named parameters, especially with so many constructor parameters?

adrw: Maybe the Java Jedis library doesn't have them generated hmmm

misk-redis/src/main/kotlin/misk/redis/RedisModule.kt
```diff
@@ -4,12 +4,31 @@ import com.google.inject.Provides
...
 import misk.ServiceModule
 import misk.inject.KAbstractModule
+import misk.metrics.v2.Metrics
 import redis.clients.jedis.JedisPool
 import redis.clients.jedis.JedisPoolConfig
 import wisp.deployment.Deployment
 
 /**
- * Configures a JedisPool to connect to a Redis instance. The use of a JedisPool ensures thread safety.
+ * Configures a [Redis] client with metrics, this also installs a [ServiceModule] for [RedisService].
```
adrw, line 13: ty for docs

misk-redis/src/main/kotlin/misk/redis/RedisModule.kt
```diff
@@ -20,27 +39,29 @@ class RedisModule(
   override fun configure() {
     bind<RedisConfig>().toInstance(redisConfig)
     install(ServiceModule<RedisService>())
+    requireBinding<Metrics>()
```
adrw, line 42: nit: I've tended to see requireBindings at the top of the module

adrw (approved):

misk-redis/src/main/kotlin/misk/redis/JedisPoolWithMetrics.kt
keeferrourke: Java doesn't have named parameters :( This is the best we can do!

misk-redis/src/main/kotlin/misk/redis/RealRedis.kt
```diff
@@ -6,88 +6,84 @@ import redis.clients.jedis.JedisPool
...
+  private val jedisPool: JedisPool,
+  private val clientMetrics: RedisClientMetrics,
+) : Redis {
   override fun del(key: String): Boolean {
-    jedisPool.resource.use { jedis ->
-      return (jedis.del(key) == 1L)
-    }
+    val keyBytes = key.toByteArray(charset)
+    return jedis { del(keyBytes) == 1L }
   }
```
kevink-sq, line 28: why remove `jedisPool.resource.use` wrappers?

misk-redis/src/main/kotlin/misk/redis/JedisPoolWithMetrics.kt
```diff
@@ -0,0 +1,78 @@
...
+  override fun returnResource(resource: Jedis?) {
+    super.returnResource(resource).also {
+      setActiveIdleConnectionMetrics()
+    }
+  }
+
+  private fun setActiveIdleConnectionMetrics() {
+    metrics.activeConnectionsGauge.set(this.numActive.toDouble())
+    metrics.idleConnectionsGauge.set(this.numIdle.toDouble())
+  }
```
kevink-sq, line 47: Why also set connection guages in overridden functions `getResource`, `returnBrokenResource `, and `returnResource` instead of just during init? Can there be memory leak from re`set`ing guages?

misk-redis/src/main/kotlin/misk/redis/RealRedis.kt
```diff
@@ -208,124 +187,137 @@ class RealRedis(private val jedisPool: JedisPool) : Redis {
...
+      val timedProxy = JedisBinaryCommands::class.cast(
+        Proxy.newProxyInstance(
+          ClassLoader.getSystemClassLoader(),
+          arrayOf(JedisBinaryCommands::class.java),
+          invocationHandler
+        )
+      )
+      timedProxy.op()
+    }
+  }
```
kevink-sq, line 309: I think you are using `JedisBinaryCommands.() -> T` to make the syntax look real nice, e.g. 

```
jedis { incr(keyBytes) }
```

just a question, but where is the `op` param in L296 being used/passed?

misk-redis/src/main/kotlin/misk/redis/JedisPoolWithMetrics.kt
keeferrourke: If I only set them during init, then they won't be updated. The get/return resource methods are what actually modify the backing numActive and numIdle values, so updating the gauge here makes sense.

misk-redis/src/main/kotlin/misk/redis/RealRedis.kt
keeferrourke: You actually got it right in the context of this comment.

`op` is used on L307. I could've called it like `op(timedProxy)`, but since `op` is written as an extension on the receiver, I called it in the same style.

PR #2731, opened 2023-03-18
Add misk-redis-testing module

keeferrourke:
# Add misk-redis-testing module

This PR adds a new module `misk-redis-testing` which exposes a DockerRedis (based on the existing DockerSqs) that can be used in tests. In doing so, we now allow FakeRedis and RealRedis to share a suite of common tests, which ensures that the in-memory Fake acts much like a real Redis instance.

This also fills a gap with missing tests for the RealRedis client, which will give us some confidence that if we extend the RealRedis client (e.g. to include client-side metrics (#2729)) that we don't break functionality that allows it to connect to Redis.

---

💡 Tip for reviewers: I recommend reviewing this commit-by-commit, as I needed to move around and update some code in doing this. Notably, the `FakeRedis::hmget` and `::hgetAll` implementations were wrong, and mismatched RealRedis behaviour.


misk-redis-testing/src/test/kotlin/misk/redis/AbstractRedisTest.kt
```diff
@@ -356,9 +284,7 @@ class FakeRedisTest {
     redis[key] = "Not a number".encodeUtf8()
 
     // Exercise
-    assertThrows<IllegalArgumentException> {
-      redis.incrBy(key, 3)
-    }
+    assertThrows<RuntimeException> { redis.incrBy(key, 3) }
```
keeferrourke, line 287: The RealRedis client actually throws a JedisDataException here. Both are RuntimeExceptions, hence this change.

I'm unsure if misk RealRedis should remap Jedis specific exceptions to something more Kotlin-y, or if the Fake should throw something other than an IllegalArgumentException, but I will not change that behaviour in this PR.

misk-redis-testing/src/test/kotlin/misk/redis/AbstractRedisTest.kt
```diff
@@ -380,14 +306,13 @@ class FakeRedisTest {
...
     redis.hset("foo", "bar", "1".encodeUtf8())
-    redis.expire("foo", 1)
 
     // Assert that the expiry hasn't taken affect yet
     redis.hincrBy("foo", "bar", 2)
     assertEquals("3", redis.hget("foo", "bar")?.utf8())
 
     // Exercise
-    clock.add(Duration.ofSeconds(1))
+    redis.expire("foo", -1)
```
keeferrourke, line 315: I did this in as many places as possible where we were just trying to validate behaviour after a key has been expired/deleted.

This is because when we test RealRedis, we can't manipulate the clock.

This is _technically_ a behaviour change, because Redis docs indicate that when you call EXPIRE (and friends) with negative arguments actually the key is _deleted_ instead of expired, but it's not meaningfully different for our usages.

misk-redis-testing/src/test/kotlin/misk/redis/AbstractRedisTest.kt
```diff
@@ -436,14 +361,16 @@ class FakeRedisTest {
 
     // Test hrandfield key [count] with values.
     assertThat(redis.hrandFieldWithValues("star wars characters", 1))
-      .containsExactlyEntriesOf(mapOf("Luke Skywalker" to "Mark Hamill".encodeUtf8()))
+      .containsAnyOf(*map.entries.toTypedArray())
```
keeferrourke, line 364: Can't control RealRedis' random

misk-redis/src/main/kotlin/misk/redis/RedisModule.kt
```diff
@@ -6,34 +6,39 @@ import misk.ServiceModule
...
-        replicationGroup.redis_auth_password,
-        true
+      jedisPoolConfig,
+      replicationGroup.writer_endpoint.hostname,
+      replicationGroup.writer_endpoint.port,
+      replicationGroup.timeout_ms,
+      replicationGroup.redis_auth_password.ifEmpty {
+        check(deployment.isFake) { "Redis auth password cannot be empty in a real environment!" }
+        null
+      },
```
keeferrourke, line 40: This is an important change. I could not for the life of me get the docker Redis image to accept an auth password at start-up without building a custom image and copying in a Redis config.

I don't think we're in the business of maintaining a docker image with a password just for the purpose of testing, so we'll allow an empty auth password in our configs **only** in fake environments.

misk-redis/src/test/kotlin/misk/redis/RedisAuthPasswordEnvTest.kt
```diff
@@ -0,0 +1,53 @@
...
+class RedisAuthPasswordEnvTest {
+  @Suppress("unused")
+  @MiskExternalDependency
+  private val dockerRedis = DockerRedis // Start Redis so that we can properly provision a client.
+
+  @Test fun `injection succeeds with password-less config in fake environments`() {
+    assertThat(DockerRedis.config.values.first().redis_auth_password).isEmpty()
+    val injector = createInjector(fakeEnv, realRedisModule)
+    val redis = injector.getInstance(keyOf<RedisConsumer>()).redis
+    assertThat(redis).isInstanceOf(RealRedis::class.java)
```
frojasg, line 30: should we add some operation to validate we can write/read to docker redis?

misk-redis-testing/src/main/kotlin/misk/redis/testing/FakeRedis.kt
```diff
@@ -0,0 +1,395 @@
...
+import java.time.Clock
+import java.time.Duration
+import java.time.Instant
+import java.util.concurrent.ConcurrentHashMap
+import javax.inject.Inject
+import kotlin.math.max
+import kotlin.math.min
+import kotlin.random.Random
+
+/** Mimics a Redis instance for testing. */
```
adrw, line 19: Could you clarify in docs that this is in-memory?

misk-redis-testing/src/test/kotlin/misk/redis/AbstractRedisTest.kt
```diff
@@ -1,37 +1,23 @@
...
-import kotlin.random.Random
 import kotlin.test.assertEquals
 import kotlin.test.assertFalse
-import kotlin.test.assertNotNull
 import kotlin.test.assertNull
 import kotlin.test.assertTrue
 
-@MiskTest
-class FakeRedisTest {
+abstract class AbstractRedisTest {
```
adrw, line 16: Nice, abstract test classes are a great pattern to have tests written against interfaces tested against multiple implementations

misk-redis-testing/src/test/kotlin/misk/redis/AbstractRedisTest.kt
```diff
@@ -1,37 +1,23 @@
...
   @Suppress("unused")
-  @MiskTestModule
-  val module: Module = Modules.combine(
-    FakeClockModule(),
-    RedisTestModule(Random(1977)), // Hardcoded random seed for hrandfield* test determinism.
-  )
+  abstract val module: Module
```
adrw, line 18: Is this field used or necessary given the abstract class doesn't have `@MiskTest` or `@MiskTestModule` annotations? I imagine that setting the module on the implementation classes is sufficiently without the abstract placeholder.

adrw (approved):

misk-redis-testing/src/test/kotlin/misk/redis/AbstractRedisTest.kt
frojasg: yeah, I would change that in a different PR. 

My thoughts there:
* FakeRedis should imitate RealRedis as close as possible. 
* Should we remap Jedis specific exceptions? I think it depends on how much you wanna hide the usage of jedis in this library. I wonder how consistent we're in other part of misk where we use other libraries. In my mind the tradeoff are:
  * [+]  The abstractions encapsulates redis concern better because you don't force misk users to know how both use misk-redis and jedis. 
  * [-] it can become a shallow abstractions where we're just mapping things from one side to another. 

I wish we could define some misk principles on how to wrap other libraries.

keeferrourke: > I wish we could define some misk principles on how to wrap other libraries.

That would be amazing. Is misk meant to be a thin kotlinized wrapper around things? Is it meant to provide its own set of opinioned abstractions? Presently... It does both, a lot.

misk-redis-testing/src/main/kotlin/misk/redis/testing/FakeRedis.kt
```diff
@@ -0,0 +1,395 @@
...
+class FakeRedis : Redis {
+  @Inject lateinit var clock: Clock
```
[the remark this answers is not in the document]
keeferrourke: Sure thing! This was just copied over, but I agree that is the pattern that is more widely used.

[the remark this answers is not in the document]
keeferrourke: Using constructor injection now.

misk-redis/src/test/kotlin/misk/redis/RedisAuthPasswordEnvTest.kt
keeferrourke: Done!

PR #2740, opened 2023-03-28
Expose the JVM uptime

cconroy:
Adds a computed gauge to capture the current JVM uptime whenever metrics are scraped.

Uptime is useful for a few things:
- Allows for easy correlation of other metrics with process startup (e.g. latencies might be slower early in a fresh VM without warm caches, pools, or full JIT)
- Allows for the correlation of elapsed time against the resulting time-series. This can be a useful operational tool to help reason about artifacts from time and space aggregation in a metrics pipeline (e.g. we know that 1000ms _should_ be the observed rate of time elapsed per second).

### DIFF
--- a/misk/src/main/kotlin/misk/monitoring/MonitoringModule.kt
+++ b/misk/src/main/kotlin/misk/monitoring/MonitoringModule.kt
@@ -1,0 +1,10 @@
+package misk.monitoring
+
+import misk.inject.KInstallOnceModule
+
+class MonitoringModule : KInstallOnceModule() {
+  override fun configure() {
+    // Use an eager binding to force instantiation of this otherwise unused class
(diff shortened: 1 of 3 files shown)


misk/src/main/kotlin/misk/monitoring/MonitoringModule.kt
```diff
@@ -0,0 +1,10 @@
+package misk.monitoring
+
+import misk.inject.KInstallOnceModule
+
+class MonitoringModule : KInstallOnceModule() {
+  override fun configure() {
+    // Use an eager binding to force instantiation of this otherwise unused class
```
adrw, line 7: ty for docs

adrw (approved):

PR #2741, opened 2023-03-29
Refactor provisioning of RuntimeMxBean

cconroy:
Refactor the provider of this to a more common location so that other parts of the framework can rely on an injected `RuntimeMxBean`.

(Immediate use case is plugging the new `MonitoringModule` in to our misk apps by default)


cconroy:
Bit of a follow-up for https://github.com/cashapp/misk/pull/2740

jdm-square (approved):

PR #2744, opened 2023-04-02
Replace eventrouter with redis pubsub

tso:
This commit removes the eventrouter package and adds the publish and subscribe APIs to misk-redis so that it can be used instead in the exemplar chat example.


misk-redis/src/main/kotlin/misk/redis/Redis.kt
```diff
@@ -439,6 +440,16 @@ interface Redis {
    * Closes the client, so it may not be used further.
    */
   fun close()
+
+  /**
+   * Subscribe to a redis channel via pubsub. This is blocking!
+   */
+  fun subscribe(jedisPubSub: JedisPubSub, channel: String)
```
tso, line 447: Possibly I shouldn't be bleeding this interface to the world, but I couldn't figure out a nice way to avoid this, lemme know if anyone has a better way lmao

samples/exemplarchat/src/main/kotlin/com/squareup/chat/ChatModule.kt
```diff
@@ -11,10 +11,10 @@ import misk.web.resources.StaticResourceEntry
 
 class ChatModule : KAbstractModule() {
   override fun configure() {
-    // TODO(adrw) finish testing this with the new StaticResourceAction
     install(WebActionModule.createWithPrefix<StaticResourceAction>("/room/"))
     multibind<StaticResourceEntry>().toInstance(
-        StaticResourceEntry("/room/", "classpath:/web/index.html"))
+      StaticResourceEntry("/room/", "classpath:/web/index.html")
```
tso, line 16: for some reason this doesn't work via my local intellij @adrw any ideas?

misk-redis-testing/src/main/kotlin/misk/redis/testing/FakeRedis.kt
```diff
@@ -402,4 +403,12 @@ class FakeRedis @Inject constructor(
   override fun close() {
     // no op
   }
+
+  override fun subscribe(jedisPubSub: JedisPubSub, channel: String) {
+    throw NotImplementedError("Fake client not implemented for this operation")
```
amcrorie, line 408: nit: I wonder if faking this would be as simple as mutableMapOf<String, MutableList<JedisPubSub>> and listeners[channel].forEach { it.onMessage(message) }

misk-eventrouter/src/main/kotlin/misk/eventrouter/api.kt
amcrorie, line 1: Just a heads up that [clientsync](https://cs.github.com/?scope=org%3Asquareup&scopeName=squareup&q=miskeventrouter) seems to be referencing miskclientrouter and not using it... not sure if it will break that and definitely think we should still do this - just FYI.

misk-redis/src/main/kotlin/misk/redis/Redis.kt
amcrorie: hot take: As someone who used these Redis APIs a lot, it was always annoying when our attempt at encapsulation got in the way of just exposing the full capabilities of the Redis API.

All to say, leaking it doesn't bother me. :)

amcrorie (approved):
Looks great! Thanks for doing that.

adrw (approved):
Really nice cleanup, does the exemplarchat still work without the surrounding React app?

tso:
> Really nice cleanup, does the exemplarchat still work without the surrounding React app?

Yeah I replaced with vue to avoid a build system, works when you go to `localhost:8080/room/blah`

misk-redis-testing/src/main/kotlin/misk/redis/testing/FakeRedis.kt
tso: the problem is for specifically the `unsubscribe` method

PR #2747, opened 2023-04-05
Improve test support for the pause detector

cconroy:
My attempt to install the detector module universally in skim revealed a couple of issues with other integration tests.

Misk's default test sleeper is the `FakeSleeper` which doesn't let a sleep terminate unless some other thread manually moves the ticker the appropriate amount. This is a strange default for misk to have, and we should probably consider changing that, but that's beyond the scope of what I want to do here (unclear how many folks might be relying on this default today). This results in test failures utilizing the `ServiceManager` as the detector service will never shut down as it is blocked on a sleep call that will never terminate.

In practice, the only time we care to have a fake ticker or sleeper in the pause detector is in `PauseDetectorTest`. Any integration test that happens to install the pause detector (e.g. via some standard common module) should not have to worry about managing the clock on its behalf.

To address these issues, this change introduces as `@ForPauseDetector` annotation so that we have more direct control over the ticker/sleeper used for the detector.

### DIFF
--- a/misk/src/main/kotlin/misk/perf/PauseDetectorModule.kt
+++ b/misk/src/main/kotlin/misk/perf/PauseDetectorModule.kt
@@ -1,17 +1,26 @@
 package misk.perf
 
+import com.google.common.base.Ticker
 import misk.ServiceModule
+import misk.concurrent.Sleeper
 import misk.inject.KAbstractModule
 
 /**
  * Install this module to run the [PauseDetector] in the background.
  */
-class PauseDetectorModule constructor(
+class PauseDetectorModule(
   val pauseDetectorConfig: PauseDetectorConfig = PauseDetectorConfig(),
 ) : KAbstractModule() {
-
   override fun configure() {
     install(ServiceModule<PauseDetector>())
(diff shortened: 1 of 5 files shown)


misk/src/main/kotlin/misk/perf/PauseDetectorModule.kt
```diff
@@ -1,17 +1,26 @@
...
   override fun configure() {
     install(ServiceModule<PauseDetector>())
     bind<PauseDetectorConfig>().toInstance(pauseDetectorConfig)
+
+    // Outside of pause detector's own testing harness, we don't want to run with a fake sleeper.
+    // A fake sleeper might never return from `sleep` which will cause service teardown errors.
+    // Another kind of fake sleeper might return immediately from sleep, but that can waste
+    // CPU cycles in CI if the pause detector is pulled into such a test.
+    bind<Sleeper>().annotatedWith<ForPauseDetector>().toInstance(Sleeper.DEFAULT)
+    // For good meausre, bind a real ticker with the real sleeper:
```
swankjesse, line 23:
```suggestion
    // For good measure, bind a real ticker with the real sleeper:
```

swankjesse (approved):

PR #2742, opened 2023-03-30
Respect Accept header when choosing return content type

JGulbronson:
Not sure I'm thrilled with this approach, I tried to keep it as backwards compatible as possible, outside of changing the `value` param of `@ResponseContentType` to be a `vararg`. Tried to follow the same implementation path as how the protobuf endpoints set up JSON variants.

Putting it up to make sure we agree on the high-level implementation first (basically just look at the tests), and then I can rework the internals if there are better suggestions.

PR #2748, opened 2023-04-06
Revert "Respect Accept header when choosing return content type"

jdm-square:
Reverts cashapp/misk#2742. It was a breaking change to a public API and is causing havoc with services that have dependencies on inconsistent versions of misk.


frojasg:
Should we close this now?

jdm-square:
Yes.

PR #2753, opened 2023-04-17
Update deprecation message.

jdm-square:
There is no Config interface in misk-config, so the deprecation message is confusing.

### DIFF
--- a/misk-core/src/main/kotlin/misk/config/Config.kt
+++ b/misk-core/src/main/kotlin/misk/config/Config.kt
@@ -8,5 +8,5 @@ package misk.config
 //    "wisp.config.Config"
 //  )
 // )
-@Deprecated("Use from misk-config instead")
+@Deprecated("Use from wisp-config instead")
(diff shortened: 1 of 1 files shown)


zhen-lu (approved):

misk-core/src/main/kotlin/misk/config/Config.kt
```diff
@@ -8,5 +8,5 @@ package misk.config
 //    "wisp.config.Config"
 //  )
 // )
-@Deprecated("Use from misk-config instead")
+@Deprecated("Use from wisp-config instead")
```
frojasg, line 11: why don't we use the deprecation message that's commented?

frojasg (approved):

misk-core/src/main/kotlin/misk/config/Config.kt
jdm-square: I don't know. It was added a long time ago.

PR #2761, opened 2023-04-27
Add queueProcessingLag metric for tracking the time between when an S…

dpurcell-sq:
…QS job is published and when it is received.


dpurcell-sq:
Tied to https://github.com/squareup/cash-sqs/pull/54

hamdanjaveed (approved):

PR #2767, opened 2023-05-03
Add dependency analysis plugin

tso:
As a consequence I've upgraded a bunch of dependencies to make our transitive dependency story better.


ean5533:
> I've also deleted the ZK package as no one uses it inside of cash.

Are we concerned about anyone using it outside of Cash? Should this change perhaps be a separate PR?

tso:
> > I've also deleted the ZK package as no one uses it inside of cash.
> 
> Are we concerned about anyone using it outside of Cash? Should this change perhaps be a separate PR?

This should be a separate PR yeah. I'll revert and then just ignore the ZK package from dep analysis

ean5533 (approved):
Wow, we had _really_ drifted a lot on our transitive dependencies. Thank you for doing the work to clean us up! Glad to see this being enforced more carefully going forward.

PR #2768, opened 2023-05-10
Update wire from 4.5.2 --> 4.6.0 and kotlin 1.7.0 --> 1.8.20

tso:


build.gradle.kts
```diff
@@ -32,6 +32,8 @@ dependencyAnalysis {
     all {
       onAny {
         severity("fail")
+        // Weird issue due to wire.
+        exclude("() -> java.io.File?")
```
ean5533, line 36: I don't even comprehend this. How could the plugin possibly thing that a lambda is a dependency?

ean5533 (approved):

build.gradle.kts
tso: might be due to https://github.com/autonomousapps/dependency-analysis-android-gradle-plugin/issues/884

chris-ryan-square (approved):

PR #2783, opened 2023-05-17
Add a function to snapshot the action scope

staktrace:
This will allow using the action scope on threads that are running okhttp async executions. Those threads are not tied to the original action thread in a coroutine-friendly way so we can't use the existing coroutine APIs to propagate the action scope. Instead we need some out-of-band mechanism, and this new function provides that ability.


staktrace:
Context for this PR (sorry, private slack channel) is at https://cash.slack.com/archives/C01M5M297T7/p1684293301465499?thread_ts=1677887172.336679&cid=C01M5M297T7

PR #2785, opened 2023-05-18
Allow POST actions to be considered for handling grpc requests

amcrorie:
Some web actions can handle any grpc since they are generic (for example, [WebProxyAction](https://github.com/cashapp/misk/blob/master/misk/src/main/kotlin/misk/web/proxy/WebProxyAction.kt)). However, they are currently not considered for handling content type grpc since the dispatch mechanism is conveyed as grpc (which doesn't match POST).

As a fix for this, we should, as a last resort (before communicating a 404), consider these POST actions if no matching GRPC action was found. 

Thinking about backwards compatibility concerns, for all affected cases today, they currently convey a 404 for these scenarios, so the risk is low of breaking clients. That said, there is some small risk of a webaction in a service today claiming to be able to support GPRC (or all content types) that could start getting hit with traffic.

### DIFF
--- a/misk/src/main/kotlin/misk/web/jetty/WebActionsServlet.kt
+++ b/misk/src/main/kotlin/misk/web/jetty/WebActionsServlet.kt
@@ -1,6 +1,7 @@
 package misk.web.jetty
 
 import misk.web.BoundAction
+import misk.web.BoundActionMatch
 import misk.web.DispatchMechanism
 import misk.web.ServletHttpCall
 import misk.web.SocketAddress
@@ -8,11 +9,13 @@ import misk.web.WebConfig
 import misk.web.actions.WebAction
 import misk.web.actions.WebActionEntry
 import misk.web.actions.WebActionFactory
+import misk.web.mediatype.MediaRange
 import misk.web.mediatype.MediaTypes
 import misk.web.metadata.webaction.WebActionMetadata
 import okhttp3.Headers
 import okhttp3.HttpUrl
 import okhttp3.HttpUrl.Companion.toHttpUrl
+import okhttp3.MediaType
 import okhttp3.MediaType.Companion.toMediaTypeOrNull
 import okio.BufferedSink
 import okio.buffer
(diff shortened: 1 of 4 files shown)


misk/src/main/kotlin/misk/web/jetty/WebActionsServlet.kt
```diff
@@ -142,10 +145,12 @@ internal class WebActionsServlet @Inject constructor(
       val requestContentType = httpCall.contentType()
       val requestAccepts = httpCall.accepts()
 
-      val candidateActions = boundActions.mapNotNull {
-        it.match(httpCall.dispatchMechanism, requestContentType, requestAccepts, httpCall.url)
+      var bestAction = findBestAction(httpCall.dispatchMechanism, httpCall.url, requestContentType, requestAccepts)
+
+      // If no actions were found, allow POST actions to be considered for handling grpc requests
```
frojasg, line 150: I think is worth mentioning this makes sense when You have gRPC calls and you want to handle it manually without using the misk abstraction on top of it.

frojasg (approved):
LGTM

swankjesse (commented):
I don’t get it.

In what situation a server advertise POST but accept gRPC? gRPC uses a different on-the-wire envelopes of its messages, strict conventions on paths, always 200 response codes, etc.

swankjesse:
(WebProxyAction was written to support static endpoints like the React dev server. If you’d like it to forward gRPC, it’s not going to work! This is because gRPC needs trailers, and WebProxyAction doesn’t know about ’em.)

amcrorie:
> I don’t get it.
> 
> In what situation a server advertise POST but accept gRPC? gRPC uses a different on-the-wire envelopes of its messages, strict conventions on paths, always 200 response codes, etc.

I have a proxy action that is attempting to forward all traffic to another endpoint. It's been working just fine for non-grpc traffic, but it isn't working for grpc, even though the logic is generic.

There are 3 possible options to make this generic logic work:
- what's in this PR to allow POST candidates to handle grpc traffic, if they claim to accept the grpc content type
- create a separate grpc proxy endpoint using @WireRpc - however, this would require using a separate feature binding than GrpcFeatureBinding that doesn't take the request/response body, handle encoding, and try to adapt it (concept below)
- do this proxy logic in an interceptor instead of an action (and check if path == "/mypath" in the interceptor)

```
      if (action.parameters.isEmpty() && wireAnnotation.requestAdapter.isEmpty()) {
        claimer.claimRequestBody()
        claimer.claimResponseBody()
        claimer.claimReturnValue()
        val responseBodyMarshaller = responseBodyMarshallerFactory.create(action)
        return ResponseBodyFeatureBinding(responseBodyMarshaller)
      }

```

amcrorie:
> (WebProxyAction was written to support static endpoints like the React dev server. If you’d like it to forward gRPC, it’s not going to work! This is because gRPC needs trailers, and WebProxyAction doesn’t know about ’em.)

Yes; I was working on a separate PR to support trailers with the Response object and RequestBridgeInterceptor.

amcrorie:
Based on feedback, decided to go with an alternate approach [here](https://github.com/cashapp/misk/pull/2832)

PR #2794, opened 2023-05-24
Update maven publish gradle plugin

tso:


build.gradle.kts
```diff
@@ -173,9 +173,8 @@ subprojects {
     apply(from = file("$rootDir/hooks.gradle"))
   }
 
-  if (!path.startsWith(":samples")) {
+  if (!path.startsWith(":samples") && !path.startsWith(":misk-bom")) {
     apply(plugin = "com.vanniktech.maven.publish")
-    apply(from = "$rootDir/gradle-mvn-publish.gradle")
```
tso, line 178: Given I'm removing this is there a point to this file??

staktrace: Why are you removing this? Won't it break the open-source publication to sonatype? (I'm just guessing, I don't know how publishing of misk works)

staktrace (approved):
Not 100% sure but I guess we can try it

build.gradle.kts
tso: Was breaking stuff with it include + wisp doesn't have it

PR #2795, opened 2023-05-25
Update misk publishing (wow so many gradle.properties are gone)

tso:


.github/workflows/Release.yml
```diff
@@ -26,9 +26,15 @@ jobs:
           sed --in-place -e "s/-SNAPSHOT/-$(date +%Y%m%d.%H%M)-$(git rev-parse --short HEAD)/" gradle.properties
           cat gradle.properties
 
+      - name: Configure JDK
```
tso, line 29: copied from tempest, do we want this?

build.gradle.kts
```diff
@@ -188,3 +189,35 @@ subprojects {
     }
   }
 }
+
+allprojects {
```
tso, line 193: should this be combined with the above `allprojects` block?

swankjesse: I believe that having lots of independent `allProjects` blocks is best for keeping logic self-contained

swankjesse (approved):

.github/workflows/Release.yml
swankjesse: This should track whatever the Build.yml does

PR #2797, opened 2023-05-25
Webactions: download file if Content-Disposition is attachment

ean5533:
When calling a web action via the `/web-actions` dashboard, the results get embedded inline.

This PR makes it so that, if [the `Content-Disposition` header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Disposition) in the response is set to `attachment; filename="<some file name>"`, then the browser will also download the response body as a file. This is the same behavior that most browsers would follow if they navigated to a page that had such a header in the response.

Note that this behavior will not be triggered if the header is just set to `attachment` with no filename; the correct behavior in that case would be to show a file download dialog, and I didn't want to get into that rat's nest.

https://github.com/cashapp/misk/assets/1847776/a7569f11-f6b9-4a0c-af01-f9b08c596765


samples/exemplar/src/main/kotlin/com/squareup/exemplar/actions/DownloadAFileWebAction.kt
```diff
@@ -0,0 +1,30 @@
...
+import misk.web.Response
+import misk.web.ResponseContentType
+import misk.web.actions.WebAction
+import misk.web.mediatype.MediaTypes
+import okhttp3.Headers.Companion.toHeaders
+import javax.inject.Inject
+import javax.inject.Singleton
+
+@Singleton
+class DownloadAFileWebAction @Inject constructor() : WebAction {
```
ean5533, line 17: Note: we don't have to keep this action in the exemplar if we don't want, it's just how I was manually testing this change

adrw: I like the idea of continuing to add to the exemplar so it has fuller examples with the mutual benefit of end-to-end testing and local development for the dashboard.

adrw (approved):
Would you be down to attach a demo video to the PR description to show it working?

misk-admin/web/tabs/web-actions/src/rewrite/WebActionSendRequest.tsx
```diff
@@ -54,6 +56,7 @@ export default function WebActionSendRequest({ webActionMetadata }: Props) {
     axiosRequest
       .then(response => {
         setResponse(JSON.stringify(response.data, null, 2))
+        maybeDownloadFile(response)
```
philfalco2, line 59: Thoughts on calling setResponse only if maybeDownloadFile returns false? And then modifying the response to 'Downloading file...' or something?

It should be fairly (hopefully) straight-forward to download the file async. Here's an example of as async axios-based method: https://github.com/squareup/ecom-phoenix-feather/blob/2f24cd31c09f7a65cc9bac3649330c12fef7cc60/resources/assets/js/pages/EditStaff.vue#L81

Which is simply called: https://github.com/squareup/ecom-phoenix-feather/blob/2f24cd31c09f7a65cc9bac3649330c12fef7cc60/resources/assets/js/pages/EditStaff.vue#L132

philfalco2 (approved):
Love to see it! Great work @ean5533. Thank you so much!!

philfalco2 (approved):

PR #2800, opened 2023-05-26
Misk Admin Dashboard v2

adrw:
This is re-implementation of the Misk Admin Dashboard using kotlinx.html, Hotwire, and Tailwind CSS instead of the existing stale and hard to maintain Misk-Web, React, Blueprint implementation.

Specifically, this PR will include:
- navbar UI
- tab loading
- example of non-Misk-Web admin tabs (ie. Hotwire)
- updating exemplar sample code

It will land alongside the existing implementation to de-risk rollout and be available at `/v2/_admin/`. After more testing with existing services, Misk-Web tabs...etc, the original implementation in the `admin-dashboard` Misk-Web tab will be deleted and this new implementation will be updated to live at `/_admin/`.

PR #2807, opened 2023-05-31
Add MiskWebTabIndexAction for per tab loading in iframe

adrw:
In preparation of https://github.com/cashapp/misk/pull/2800, this action will load a tab independently. This enables a new dashboard implementation which loads tabs within an IFrame instead of using React.

For example, the following URL will now load a tab independently within an IFrame without any navbar wrapping UI: `http://0.0.0.0:8080/api/dashboard/tab/misk-web/database`.

Screenshot 2023-05-31 at 15 32 17

This PR also introduces usage of the `kotlinx.html` DSL for building HTML responses. The new `misk-hotwire` subproject will grow to include other Kotlin HTML abstractions to support building HTML responses using Hotwire.

To support the new dashboard implementation, all Misk Web tabs will need to update their routes file similarly to the changes in this PR.


samples/exemplar/build.gradle.kts
```diff
@@ -1,20 +1,26 @@
 plugins {
   kotlin("jvm")
+  application
   id("com.squareup.wire")
 }
 
+val applicationMainClass = "com.squareup.exemplar.ExemplarServiceKt"
```
adrw, line 7: the Application plugin makes it easier to run the sample services in IntelliJ or with Gradle

misk-admin/src/main/kotlin/misk/web/dashboard/MiskWebTabIndexAction.kt
```diff
@@ -0,0 +1,91 @@
...
+class MiskWebTabIndexAction @Inject constructor(
+  @AdminDashboard private val dashboardTabs: List<DashboardTab>,
+) : WebAction {
+  @Get("$PATH/{slug}")
+  @ResponseContentType(MediaTypes.TEXT_HTML)
+  @AdminDashboardAccess
+  fun get(@PathParam slug: String?): String {
+    val dashboardTab = dashboardTabs.firstOrNull { slug == it.slug }
+    return buildHtml {
+      html {
```
adrw, line 34: This returns a Kotlin-ify'd version of the auto-generated `/_tab/<slug>/index.html` that Misk-Web's Webpack currently pops out. Putting this in Kotlin makes it easier to apply universal styles across all tabs being loaded.

misk-admin/web/tabs/config/src/routes/index.tsx
```diff
@@ -6,6 +6,10 @@ const routes = (
   <div>
     <Switch>
       <Route path="/_admin/config/" component={TabContainer} />
+      <Route
```
adrw, line 9: Thankfully this ends up being the only Misk-Web code change required of existing tabs to work in the new dashboard implementation.

misk-hotwire/src/main/kotlin/misk/hotwire/BuildHtml.kt
```diff
@@ -0,0 +1,15 @@
+package misk.hotwire
+
+import kotlinx.html.TagConsumer
+import kotlinx.html.stream.appendHTML
+
+fun buildHtml(renderer: TagConsumer<*>.() -> Unit) = StringBuilder().apply {
+  appendHTML().renderer()
+}.toString()
+
+// TODO replace with ResponseBody
```
keeferrourke, line 10: TODO?

keeferrourke (approved):

misk-admin/src/main/kotlin/misk/web/dashboard/MiskWebTabIndexAction.kt
```diff
@@ -0,0 +1,91 @@
...
+ * Kotlin backed tab loader, equivalent to /_tab/slug/index.html
+ */
+@Singleton
+class MiskWebTabIndexAction @Inject constructor(
+  @AdminDashboard private val dashboardTabs: List<DashboardTab>,
+) : WebAction {
+  @Get("$PATH/{slug}")
+  @ResponseContentType(MediaTypes.TEXT_HTML)
+  @AdminDashboardAccess
+  fun get(@PathParam slug: String?): String {
```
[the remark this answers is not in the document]
adrw: Where possible I lean away from returning Response given the extra boilerplate vs relying on Misk to properly wrap the response type. I'll change it to throw a `NotFoundException` though to get the proper 404 though rather than wrapping the error in this page.

misk-hotwire/src/main/kotlin/misk/hotwire/BuildHtml.kt
adrw: Addressed!

adrw: I'm going to pair with Jesse on finalizing this since it was a point he had raised in initial prototypes.

misk-hotwire/src/main/kotlin/misk/hotwire/BuildHtml.kt
```diff
@@ -0,0 +1,15 @@
...
+import kotlinx.html.TagConsumer
+import kotlinx.html.stream.appendHTML
+
+fun buildHtml(renderer: TagConsumer<*>.() -> Unit) = StringBuilder().apply {
+  appendHTML().renderer()
+}.toString()
+
+// TODO consider streaming response
+//fun buildHtmlResponseBody(renderer: TagConsumer<*>.() -> Unit): ResponseBody = object : ResponseBody {
+//  override fun writeTo(sink: BufferedSink) {
```
adrw, line 12: @swankjesse would this be the approach to make a streaming HTML response?

misk-admin/src/main/kotlin/misk/web/dashboard/MiskWebTabIndexAction.kt
```diff
@@ -0,0 +1,92 @@
...
+    val dashboardTab = dashboardTabs.firstOrNull { slug == it.slug }
```
keeferrourke, line 32: nit: would be better to throw the 404 here, rather than mid way through building the HTML.

```suggestion
    // TODO remove this hack when new Web Actions tab lands and old ones removed, v1 and v2 are in the same web-actions tab
    val normalizedSlug = if (slug == "web-actions-v1") "web-actions" else slug ?: "null"
    val dashboardTab = dashboardTabs.firstOrNull { slug == it.slug } ?: throw NotFoundException("No tab found for slug: $normalizedSlug")
```

then dashboardTab will be smartcast as not null below.

adrw: Fixed here https://github.com/cashapp/misk/pull/2808

PR #2809, opened 2023-06-02
Deprecate DashboardNavbarItem in favor of DashboardModule

adrw:
Rename new DashboardTabModule to DashboardModule


misk-admin/api/misk-admin.api
```diff
@@ -277,15 +282,6 @@ public final class misk/web/dashboard/DashboardTab : misk/web/dashboard/WebTab {
 	public fun toString ()Ljava/lang/String;
 }
 
-public final class misk/web/dashboard/DashboardTabModule : misk/inject/KAbstractModule {
```
adrw, line 280: `DashboardTabModule` isn't used anywhere yet, it was merged yesterday.

misk-admin/api/misk-admin.api
```diff
@@ -218,7 +218,12 @@ public final class misk/web/dashboard/DashboardHomeUrl : misk/web/dashboard/Vali
 }
 
 public final class misk/web/dashboard/DashboardModule : misk/inject/KAbstractModule {
```
adrw, line 220: The existing `DashboardModule` wasn't used anywhere outside of Misk so this is backwards compatible.

misk-admin/src/main/kotlin/misk/web/dashboard/BaseDashboardModule.kt
```diff
@@ -14,8 +18,18 @@ class BaseDashboardModule(
   private val isDevelopment: Boolean
 ): KAbstractModule() {
   override fun configure() {
-    // Install base dashboard support
-    install(DashboardModule())
+    // Setup multibindings for dashboard related components
+    newMultibinder<DashboardTab>()
```
adrw, line 22: I inlined the old `DashboardModule` since it was only installed here and I wanted to use that name for other purposes going forward.

keeferrourke (approved):

PR #2826, opened 2023-06-16
Add opt-in flag for auto MDC injection for datadog tracing

mpeyper:
We have had issues with log correlation when using coroutines and this change is the result of an internal investigation into the issue.

The idea here is that by removing our manual MDC handling we will stop interfering with the [automatic MDC injection](https://docs.datadoghq.com/tracing/other_telemetry/connect_logs_and_traces/java/?tab=slf4jandlogback#automatic-injection) that the DataDog instrumentation is already doing for us, which during the investigation appeared to work as well or better than what we are doing currently in all tested scenarios.

We are first introducing it as an opt-in flag (this PR) but will shortly make this the default and finally remove the flag as we roll through different testing stages.

### DIFF
--- a/misk-datadog/src/main/kotlin/misk/tracing/backends/datadog/DatadogTracingBackendModule.kt
+++ b/misk-datadog/src/main/kotlin/misk/tracing/backends/datadog/DatadogTracingBackendModule.kt
@@ -8,16 +8,28 @@ import misk.inject.KAbstractModule
 /**
  * Binds the datadog tracer to opentracing's [Tracer]
  */
-class DatadogTracingBackendModule : KAbstractModule() {
+
+class DatadogTracingBackendModule() : KAbstractModule() {
+
+  private var useAutoMdcInjection: Boolean = false
+
+  @Deprecated("useAutoMdcInjection is an experimental option that will be removed in a future version")
+  constructor(useAutoMdcInjection: Boolean): this() {
+    this.useAutoMdcInjection = useAutoMdcInjection
+  }
+
   override fun configure() {
     // A DDTracer is installed by the dd-java-agent in TracerInstaller, which runs before the app's main() method.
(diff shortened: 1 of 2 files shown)


misk-datadog/src/main/kotlin/misk/tracing/backends/datadog/DatadogTracingBackendModule.kt
```diff
@@ -8,16 +8,28 @@ import misk.inject.KAbstractModule
...
-class DatadogTracingBackendModule : KAbstractModule() {
+
+class DatadogTracingBackendModule() : KAbstractModule() {
+
+  private var useAutoMdcInjection: Boolean = false
+
+  @Deprecated("useAutoMdcInjection is an experimental option that will be removed in a future version")
+  constructor(useAutoMdcInjection: Boolean): this() {
+    this.useAutoMdcInjection = useAutoMdcInjection
+  }
```
mpeyper, line 19: I have added this as a seperate constructor instead of a defaulted value in the primary constructor to reduce the changes to the API as we have seen examples recently where default arguments can actually result in runtime errors when version conflicts are resolved in consumers. That said, with the default argument, the API changes look like:

```diff
public final class misk/tracing/backends/datadog/DatadogTracingBackendModule : misk/inject/KAbstractModule {
	public fun <init> ()V
+	public fun <init> (Z)V
+	public synthetic fun <init> (ZILkotlin/jvm/internal/DefaultConstructorMarker;)V
}
```

I think this would still be backwards compatible for existing consumer, however I'm not confident about it and given this will hopefully be a short lived option, I'm ok with it being a bit messier in the implementation for the added piece of mind.  However, if others strongly prefer the default value I can change it over before merging.

staktrace (approved):

PR #2827, opened 2023-06-20
Lazily supply jooq stuff

tso:


misk-jooq/src/main/kotlin/misk/jooq/JooqModule.kt
```diff
@@ -89,7 +88,7 @@ class JooqModule(
       lateinit var clock: Clock
       override fun get(): JooqTransacter {
         return JooqTransacter(
-          dslContext = dslContext(dataSourceServiceProvider.get(), clock, datasourceConfig)
+          dslContext = lazy { dslContext(dataSourceServiceProvider.get(), clock, datasourceConfig) }
```
tso, line 91: core change -- just means that the `dslContext` function won't be evaluated (and thus call `dataSourceService.get()`) until first usage which I think is completely reasonable (SchemaMigrationService will likely bear the first-time-use cost rather than any action/task)

misk-jooq/src/main/kotlin/misk/jooq/JooqModule.kt
```diff
@@ -109,33 +108,36 @@ class JooqModule(
             .withOutput(datasourceConfig.database)
         )
       )
-    return DSL.using(dataSourceService.get(), datasourceConfig.type.toSqlDialect(), settings).apply {
```
tso, line 112: formatting

misk-jooq/api/misk-jooq.api
```diff
@@ -47,13 +47,12 @@ public final class misk/jooq/JooqSession$HookType : java/lang/Enum {
 
 public final class misk/jooq/JooqTransacter {
 	public static final field Companion Lmisk/jooq/JooqTransacter$Companion;
-	public fun <init> (Lorg/jooq/DSLContext;)V
+	public fun <init> (Lkotlin/Lazy;)V
```
tso, line 50: I think this is fine because this constructor is only used within misk?

adrw (approved):

PR #2832, opened 2023-06-22
Provide grpc annotation for actions

amcrorie:
To support grpc use cases where clients don't want Wire data conversion and implement generic logic such as proxying, introducing a new Grpc type that does not attempt to convert data from proto but instead allows for the action to handle the payload on its own. Also extending the misk Response type to support trailers since they are required by grpc.


misk-actions/src/main/kotlin/misk/web/Response.kt
```diff
@@ -8,7 +8,8 @@ import okio.BufferedSink
 data class Response<out T>(
   val body: T,
   val headers: Headers = headersOf(),
-  val statusCode: Int = 200
+  val statusCode: Int = 200,
+  val trailers: Headers? = null
```
swankjesse, line 12: I think this needs to be deferred, like so:

```
val trailers = () -> Headers? = null
```

misk/src/main/kotlin/misk/web/BoundAction.kt
```diff
@@ -254,7 +254,14 @@ private class RequestBridgeInterceptor(
     if (returnValue is Response<*>) {
       httpCall.statusCode = returnValue.statusCode
       httpCall.addResponseHeaders(returnValue.headers)
+      val trailers = returnValue.trailers
```
swankjesse, line 257: Defer resolving this until the response body has been streamed?

swankjesse (approved):

misk/src/main/kotlin/misk/web/BoundAction.kt
amcrorie: I thought that the trailers are actually accumulated in [JettyUpstreamResponse](https://github.com/cashapp/misk/blob/master/misk/src/main/kotlin/misk/web/jetty/JettyServletUpstreamResponse.kt#L40) and conveyed in a lambda already such that it takes care of delaying the write until after the body is streamed?

frojasg (approved):

PR #2838, opened 2023-06-27
[misk-cron] fix: CronManager: log exceptions at ERROR level.

Ubehebe:
At head, when a CronEntry throws an exception, CronManager logs a WARNING log and swallows the exception. This makes the exception invisible to alerting systems that rely on either ERROR logs or exceptions being thrown. This means that services using Misk cron tasks could be broken without knowing about it.

This PR upgrades the log to an ERROR. This brings the failure mode closer to the "vanilla jvm" case, where exception-throwing code is called from the main thread and produces a stack trace.


Ubehebe:
[Here](https://cash.slack.com/archives/C01M5M297T7/p1687299095542259) is the context (non-public Slack).

adrw (approved):

Ubehebe:
@adrw mind merging?

PR #2841, opened 2023-06-29
Add RequestLoggingTransformer mechanism

ean5533:
This will allow services to inject custom transformers that will get applied right before request/response bodies are logged. The main motivating use case is wanting to trim out noisy data from response bodies.


misk/src/test/kotlin/misk/web/interceptors/RequestLoggingInterceptorTest.kt
```diff
@@ -354,3 +398,31 @@ internal class ConfigOverrideAction @Inject constructor() : WebAction {
   ) // these values overridden by RequestLoggingConfig binding above
   fun call(@PathParam message: String): String = "echo: $message"
 }
+
+/**
+ * A [RequestLoggingTransformer] that doesn't like seeing "Evan" in String response bodies
+ */
+internal class EvanHatingTransformer @Inject constructor() : RequestLoggingTransformer {
```
keeferrourke, line 405: I disagree with the premise of this transformer 😂 😭

ean5533: I don't condone name-based prejudice but I think it's import for Transformers to be able to express their views.

misk/src/main/kotlin/misk/web/interceptors/RequestLoggingInterceptor.kt
```diff
@@ -131,7 +134,11 @@ class RequestLoggingInterceptor internal constructor(
     val sampling = if (isError) errorBodySampling else bodySampling
     val randomDouble = random.current().nextDouble()
     if (randomDouble < sampling) {
-      val requestResponseBody = bodyCapture.get()
+      val requestResponseBody =
+        requestLoggingTransformers.foldRight(bodyCapture.get()) { transformer, body ->
+          transformer.tryTransform(body)
+        }
```
keeferrourke, line 140: It'd be nice to see a test for behaviour with multiple transformers applied to the same request. I have a feeling there might be a gotcha with ordering these transformations?

kyeotic:
> Because of this, my current stance is that there is no need to add the complexity of a configureProviders-style binding API -- just stick with vanilla multibind.

I agree

snigdhabanda (approved):

adrw (approved):

misk/src/test/kotlin/misk/web/interceptors/RequestLoggingInterceptorTest.kt
adrw: For future maintainers, I'd lean towards renaming this to something less personal.

misk/src/main/kotlin/misk/web/interceptors/RequestLoggingInterceptor.kt
```diff
@@ -45,6 +46,7 @@ class RequestLoggingInterceptor internal constructor(
     private val logRateLimiter: LogRateLimiter,
     private val deployment: Deployment,
     private val configs: Set<RequestLoggingConfig>,
+    private val requestLoggingTransformers: List<RequestLoggingTransformer>,
```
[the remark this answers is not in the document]
ean5533: It works by virtue of: https://github.com/cashapp/misk/blob/4b00d84294aabea58a3bd2a1f24ab1fa1547b45e/misk-inject/src/main/kotlin/misk/inject/KAbstractModule.kt#L67-L88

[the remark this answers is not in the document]
ean5533: We could bring it in as a `Set` just fine; honestly the only reason I brought it in as a `List` was just to make it easier to write the unit test that proved that all transformers run even if one of them throws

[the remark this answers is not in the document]
cconroy: Eh, I don't feel strongly..local consistency is maybe worth going for here but it's NBD

PR #2844, opened 2023-07-04
Add Bad Gateway Exception

blaklites:
This change adds a new status code called BAD_GATEWAY which is a 502 status code meant to be returned when the server acting as a gateway received an invalid response from the upstream server.


r3mariano (approved):

misk-actions/src/main/kotlin/misk/exceptions/Exceptions.kt
```diff
@@ -103,6 +104,9 @@ open class ClientClosedRequestException(message: String = "", cause: Throwable?
  * Base exception for when a server is acting as a gateway and cannot get a response in time.
  * The message is not exposed to the caller.
  */
```
r3mariano, line 106: Probably want this comment for gateway timeout, and a different comment for this exception.

PR #2845, opened 2023-07-04
feat!: log request and response headers

r3mariano:
Previously it would only be logged if the request included Headers as one of its args. It should just log headers, using the same redaction logic we had previously added.

Binary-incompatible change in `RequestResponseBody`.

### DIFF
--- a/misk/src/test/kotlin/misk/web/WebSocketsTest.kt
+++ b/misk/src/test/kotlin/misk/web/WebSocketsTest.kt
@@ -46,7 +46,7 @@ internal class WebSocketsTest {
     // Confirm interceptors were invoked.
     assertThat(logCollector.takeMessage(RequestLoggingInterceptor::class)).matches(
       "EchoWebSocket principal=unknown time=0.000 ns code=200 " +
-        "request=\\[JettyWebSocket\\[.* to /echo]] response=EchoListener"
+        "request=\\[JettyWebSocket\\[.* to /echo]] .* response=EchoListener .*"
     )
   }
 
--- a/misk/src/test/kotlin/misk/web/interceptors/RequestLoggingInterceptorTest.kt
+++ b/misk/src/test/kotlin/misk/web/interceptors/RequestLoggingInterceptorTest.kt
@@ -61,8 +61,9 @@ internal class RequestLoggingInterceptorTest {
       ).isSuccessful
     ).isTrue()
     assertThat(logCollector.takeMessages(RequestLoggingInterceptor::class)).containsExactly(
-      "RateLimitingIncludesBodyRequestLoggingAction principal=caller time=100.0 ms " +
-        "code=200 request=[hello] response=echo: hello"
+      "RateLimitingIncludesBodyRequestLoggingAction principal=caller time=100.0 ms code=200 " +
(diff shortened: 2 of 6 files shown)


misk/src/test/kotlin/misk/web/WebSocketsTest.kt
```diff
@@ -46,7 +46,7 @@ internal class WebSocketsTest {
     // Confirm interceptors were invoked.
     assertThat(logCollector.takeMessage(RequestLoggingInterceptor::class)).matches(
       "EchoWebSocket principal=unknown time=0.000 ns code=200 " +
-        "request=\\[JettyWebSocket\\[.* to /echo]] response=EchoListener"
+        "request=\\[JettyWebSocket\\[.* to /echo]] .* response=EchoListener .*"
```
adrw, line 49: .* feels like it will make searches more difficult? Why output anything?

misk/src/test/kotlin/misk/web/interceptors/RequestLoggingInterceptorTest.kt
```diff
@@ -143,7 +147,9 @@ internal class RequestLoggingInterceptorTest {
       .isEqualTo(500)
     val messages = logCollector.takeMessages(RequestLoggingInterceptor::class)
     assertThat(messages).containsExactly(
-      "ExceptionThrowingRequestLoggingAction principal=caller time=100.0 ms failed request=[fail]"
+      "ExceptionThrowingRequestLoggingAction principal=caller time=100.0 ms failed " +
+        "request=[fail] " +
+        "requestHeaders={accept-encoding=[gzip], connection=[keep-alive]}"
```
adrw, line 152: What's the motivation for adding header dumping to logs? Not sure what the upside is in most cases? Also would lean towards it being opt-in given the noise.

adrw (approved):

misk/src/test/kotlin/misk/web/interceptors/RequestLoggingInterceptorTest.kt
ean5533: Maybe worth adding a boolean config flag to `ActionLoggingConfig`?

misk/src/test/kotlin/misk/web/WebSocketsTest.kt
adrw: What about requestHeaders=[] and likewise for response? The .* I think will be confusing for folks since it looks like a Regex and could be confused with being part of the logged request/response preceding it.

r3mariano: Ah that is a matcher in the assertion. The logs already do something like `requestHeaders={accept-encoding=[gzip], connection=[keep-alive]}`

misk/src/test/kotlin/misk/web/interceptors/RequestLoggingInterceptorTest.kt
r3mariano: This does add to a log entry, but it does make the action log longer. I don’t think it makes request logging worse to parse, and the list of headers is heavily filtered already. If it gets bad cost-wise in our logging systems I'd be happy to revert or gate.

PR #2846, opened 2023-07-05
Bump guice to v6.0.0 and guava to v32.1.1

ThePumpingLemma:
Additionally, use BOMs for both guice and guava


misk-aws-dynamodb-testing/build.gradle.kts
```diff
@@ -10,11 +10,13 @@ plugins {
 
 dependencies {
   api(Dependencies.awsDynamodb)
+  api(Dependencies.guava)
   api(Dependencies.guice)
   api(Dependencies.javaxInject)
   api(Dependencies.tempestTestingInternal)
   api(project(":misk-aws-dynamodb"))
   api(project(":misk-inject"))
+  implementation(Dependencies.errorproneAnnotations)
```
ThePumpingLemma, line 19: I think we can actually omit these if we want, see: https://github.com/google/guice/pull/1739

chris-ryan-square (approved):

misk-aws-dynamodb-testing/build.gradle.kts
adrw: If we don't need it, let's remove it. 

Could be worth running the dependency analyzer plugin again, last time it caught stuff like this.

adrw (approved):
LG

misk-aws-dynamodb-testing/build.gradle.kts
ThePumpingLemma: I pushed a change to omit it, let's see if the tests pass. 🤞

keeferrourke (approved):

frojasg (approved):
LGTM. Thanks for doing this update

PR #2847, opened 2023-07-05
Include more enhancedBy ready service for jdbc things

adrw:


jdm-square (approved):

adrw:
I'm going to close this for now as I'm not confident this will be the right fix.

adrw:
After further review, it seems like these are beneficial.

adrw:
Going to close since it doesn't seem to be necessary with other fixes that landed.

PR #2848, opened 2023-07-05
Remove deprecation warning.

jdm-square:
When consumers try to follow the suggested replacement (eg. https://github.com/squareup/cash-suez/pull/185), they receive [compile errors](https://buildkite.com/cash/suez/builds/791#0188f8ba-0b88-4e15-8f04-c17e1baefc15). We shouldn't mark interfaces as deprecated until it's possible to act on it.

### DIFF
--- a/misk-core/src/main/kotlin/misk/config/Config.kt
+++ b/misk-core/src/main/kotlin/misk/config/Config.kt
@@ -1,12 +1,3 @@
 package misk.config
 
-// TODO(chrisryan): soft deprecating...
-// @Deprecated(
-//  message = "Use wisp.config.Config directly",
-//  replaceWith = ReplaceWith(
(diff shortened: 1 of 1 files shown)


adrw (approved):
LG but I don't think services should have that strict a setting given it makes it impossible to use deprecation annotations at all without immediate action given the compile time failure.

jdm-square:
The issue is not with services failing to compile due to the deprecation warning. The issue is when services attempt to act on those warnings but it is impossible to do so.

ean5533:
Strong agree, I've tried to act on this deprecation warning before and been confused and annoyed by it

adrw:
Definitely agree that any Deprecation annotations should be actionable and short-lived.

PR #2851, opened 2023-07-06
Revert "Set SchemaMigratorService as enhances ReadyService (#2849)"

adrw:
This reverts commit b874ffab4e0834557f4262d113418bff9f4b28ae.


adrw:
In testing, it looks like my understanding was incorrect so I'm reverting.

For service dependencies it doesn't appear like the `ReadyService` is sufficient for building the graph given it will appear to be running yet all the "enhancement" services (DB, redis...etc) are just starting up. I'll land any of those changes in a follow up PR.

jdm-square (approved):
Thanks for the doc update!

PR #2858, opened 2023-07-12
Detekt rule for public functions/constructors with default arguments and no @JvmOverloads

mmollaverdi:
The build will fail for this, because I haven't added `@JvmOverloads` to all the missing places yet and haven't re-generated api files. Wanted to get input on this rule first.

This is what a violation reported by detekt looks like locally.
Screenshot 2023-07-12 at 7 04 42 pm


detekt-rules/src/main/kotlin/detektive/javacompat/AnnotatePublicApisWithJvmOverloads.kt
```diff
@@ -0,0 +1,91 @@
...
+    checkElement(function, ElementType.FUNCTION)
+    super.visitNamedFunction(function)
+  }
+
+  override fun visitPrimaryConstructor(constructor: KtPrimaryConstructor) {
+    checkElement(constructor, ElementType.CONSTRUCTOR)
+    super.visitPrimaryConstructor(constructor)
+  }
+
+  private fun checkElement(element: KtFunction, elementType: ElementType) {
```
mmollaverdi, line 45: This is where the main change in this PR is. A custom detekt rule implementation, which find public functions and constructors with one or more default arguments which are missing `@JvmOverloads`

mmollaverdi: I've included this rule within misk just for convenience.  It should really be moved into a separate repository given this is a generic rule. That'd also simplify some of build configs which are making exceptions for this sub-project.

misk-clustering/src/main/kotlin/misk/clustering/DefaultCluster.kt
```diff
@@ -74,6 +74,7 @@ class DefaultCluster(
   }
 
   /** Triggers a change to the cluster in response members becoming ready or not ready */
+  @JvmOverloads
```
mmollaverdi, line 77: I've added `@JvmOverloads` in a few missing places, but not everywhere yet.
Undecided whether I should do that everywhere, or add a baseline file for detekt to ignore the existing violations (leaning towards the former).

misk-action-scopes/src/main/kotlin/misk/scope/ActionScopedProviderModule.kt
```diff
@@ -120,6 +122,7 @@ abstract class ActionScopedProviderModule : KAbstractModule() {
   }
 
   /** Binds an annotation qualified [ActionScoped] along with its provider */
+  @JvmOverloads
```
adrw, line 125: Do we even want this on function calls? I've usually only seen it on class constructors.

mmollaverdi: It's definitely applicable to both
https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.jvm/-jvm-overloads/#jvmoverloads

And the MethodNotFound errors can happen on both when this is present, although I'm not sure if the prior incident have been mostly on constructors or methods.

Misk also already has it on both constructors and functions in a few places
https://github.com/search?q=repo%3Acashapp%2Fmisk%20JvmOverloads&type=code

frojasg (approved):
* Could you add a README to the new module to provide some context on why and what is in there? 
* LGTM, thanks for adding this.

PR #2862, opened 2023-07-13
Keep prometheus http server running until work has ceased

tgregory-block:
This should keep both the prometheus http server and the jetty threadpool/connection metrics up until all outstanding requests have been processed


adrw (approved):
LG

tgregory-block:
@adrw would you be so kind as to merge this? I don't have misk write access

PR #2864, opened 2023-07-17
Merge Wisp to Misk

rainecp:
This merges [wisp repo](https://github.com/cashapp/wisp) as a new module in misk


r3mariano (approved):

wisp/build.gradle.kts
```diff
@@ -1,178 +1,12 @@
...
-    }
-  }
-
-  repositories {
-    mavenCentral()
-    maven(url = "https://s3-us-west-2.amazonaws.com/dynamodb-local/release")
-  }
 
-  if (!path.startsWith(":wisp-bom")) {
+  if (!path.startsWith(":wisp:wisp-bom")) {
```
adrw, line 6: Is this file necessary? I've usually seen the subdir not have a build file.

adrw (approved):
LG

wisp/build.gradle.kts
rainecp: For now, yes, because unlike Misk which adds the plugins in each submodule's `build.gradle`, wisp defines them in top level build.gradle. This minimizes the changes for this merge and makes it easier to debug and test the migration changes for wisp modules as well as look at previous changes on the same file. But in the future, we can cleanup Misk and just apply common plugins in the main build.file.

PR #2868, opened 2023-07-19
Adding users.lookupByEmail endpoint to Slack API

murada-cash:
Added a new endpoint and related data structure which lets users query slack user information based on the email. 
Example use case: for tagging users in messages we need slack user id, we can get the user id by using this endpoint and squareup email


adrw (approved):
LG

PR #2874, opened 2023-07-19
Clear MDC before and after handling requests

ean5533:
### Why?

Right now there is nothing preventing MDC tags from leaking between requests. To deal with this, most folks use some sort of helper method like `withTags` that will clean up tags when a function scope ends. This is somewhat annoying (forces additional nested function scopes), but more importantly, it means there is no way for a request handler to set MDC tags that will be included in Misk's implicit logging, particularly its `RequestLoggingInterceptor`.

By clearing MDC tags automatically between requests, users can now just `MDC.put(...)` without needing to `MDC.clear()`/`MDC.remove()` themselves, and those tags will be included in Misk's logging.

### Why in `BoundAction.scopeAndHandle`?

Other options considered were:

* In an interceptor -- however, interceptors can throw, and relying on critical ordering of interceptors feels brittle.
* In `WebActionsServlet.handleCall` -- however, this is too narrow, as it would not handle actions initiated by `JettyWebSocket`
* In `ActionScope.enter` -- however, this is too broad, as action scopes are used for more things than just web requests


misk/src/main/kotlin/misk/web/BoundAction.kt
```diff
@@ -115,8 +116,15 @@ internal class BoundAction<A : WebAction>(
       seedDataTransformers.fold(initialSeedData) { seedData, interceptor ->
         interceptor.transform(seedData)
       }
-    scope.enter(seedData).use {
-      handle(httpCall, pathMatcher)
+    
+    MDC.clear() // MDC should already be empty, but clear it again just in case
```
frojasg, line 120: I wonder if it's worth adding a test to validate the clearing is happening.

frojasg (approved):
LGTM

misk/src/main/kotlin/misk/web/BoundAction.kt
ean5533: Done!

PR #2876, opened 2023-07-19
Report gRPC exceptions as errors.

jdm-square:
Since GrpcException inherits from IOException, any gRPC exception caught by the handler is automatically reported as a warning. This is confusing and hides unexpected errors; these changes ensure that gRPC exceptions are reported as errors, while ensuring that any details of the underlying exception are not leaked outside of gRPC call chains.

### DIFF
--- a/misk/src/main/kotlin/misk/web/exceptions/GrpcExceptionMapper.kt
+++ b/misk/src/main/kotlin/misk/web/exceptions/GrpcExceptionMapper.kt
@@ -1,0 +1,11 @@
+package misk.web.exceptions
+
+import com.squareup.wire.GrpcException
+import org.slf4j.event.Level
+import javax.inject.Inject
+
+internal class GrpcExceptionMapper @Inject internal constructor() : ExceptionMapper<GrpcException> {
+  override fun toResponse(th: GrpcException) = IOExceptionMapper.INTERNAL_SERVER_ERROR_RESPONSE
+
+  override fun loggingLevel(th: GrpcException): Level = Level.ERROR
+}
--- a/misk/src/test/kotlin/misk/web/exceptions/ExceptionMapperTest.kt
+++ b/misk/src/test/kotlin/misk/web/exceptions/ExceptionMapperTest.kt
@@ -1,6 +1,8 @@
 package misk.web.exceptions
 
 import com.squareup.moshi.Moshi
+import com.squareup.wire.GrpcException
+import com.squareup.wire.GrpcStatus
 import misk.MiskTestingServiceModule
 import misk.exceptions.WebActionException
 import misk.inject.KAbstractModule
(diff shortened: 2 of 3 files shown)


misk/src/main/kotlin/misk/web/exceptions/GrpcExceptionMapper.kt
```diff
@@ -0,0 +1,44 @@
...
+    val status = try {
+      th.grpcStatusDetails?.let { Status.ADAPTER.decode(it) }
+    } catch (e: IOException) {
+      logger.error(e) {
+        "Received invalid grpc-status-details-bin header: ${th.grpcStatusDetails}"
+      }
+      null
+    }
+
+    return GrpcErrorResponse(th.grpcStatus, th.grpcMessage, status?.details ?: listOf())
```
jpulgarin, line 32: I don't think we want to do this as the default in Misk. This may leak internal details to other services. I'd probably just remove the `toGrpcResponse` method here and have it use the default one that's in `ExceptionHandlingInterceptor`.

misk/src/main/kotlin/misk/web/exceptions/GrpcExceptionMapper.kt
```diff
@@ -0,0 +1,44 @@
...
+import org.slf4j.event.Level
+import wisp.logging.getLogger
+import java.io.IOException
+import java.net.HttpURLConnection
+import javax.inject.Inject
+
+internal class GrpcExceptionMapper @Inject internal constructor() : ExceptionMapper<GrpcException> {
+  private val logger = getLogger<GrpcExceptionMapper>()
+
+  override fun toResponse(th: GrpcException) = INTERNAL_SERVER_ERROR_RESPONSE
```
jpulgarin, line 20: Would it be possible to have this class inherit everything from `IOExceptionMapper` and only override the `loggingLevel`?

jpulgarin (changes requested):

misk/src/main/kotlin/misk/web/exceptions/GrpcExceptionMapper.kt
jdm-square: I suspect the answer is no, since ExceptionMapper has a type parameter that is different.

misk/src/main/kotlin/misk/web/exceptions/GrpcExceptionMapper.kt
jdm-square: You're right.

misk/src/main/kotlin/misk/web/exceptions/GrpcExceptionMapper.kt
jpulgarin: I think you're right—if I'm not mistaken it would be possible in Java.

jpulgarin (approved):
LGTM. This will help with observability a lot, thanks!

frojasg (approved):

frojasg (approved):

misk/src/test/kotlin/misk/web/exceptions/ExceptionMapperTest.kt
```diff
@@ -61,6 +63,13 @@ internal class ExceptionMapperTest {
     assertThat(response.body?.string()).isEqualTo("internal server error")
   }
 
+  @Test
+  fun doesNotPropagateGrpcError() {
```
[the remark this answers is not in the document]
jdm-square: Good call.

PR #2877, opened 2023-07-20
Remove @JvmOverloads from Injectable constructors

mmollaverdi:
Otherwise multiple constructors annotated with `@Inject` are generated which leads to this error

> Injectable classes must have either one (and only one) constructor annotated with @Inject or a zero-argument constructor that is not private.

Also added instructions for suppressing violations of `AnnotatePublicApisWithJvmOverloads`.

Will look into not reporting this violation when a javax or guice Inject annotation is present later today.


frojasg (approved):
LGTM

PR #2878, opened 2023-07-20
Don't enforce the `AnnotatePublicApisWithJvmOverloads` rule on APIs annotated with `@Inject`

mmollaverdi:
Follow-up from https://github.com/cashapp/misk/pull/2877

### DIFF
--- a/detektive/build.gradle.kts
+++ b/detektive/build.gradle.kts
@@ -5,11 +5,13 @@ plugins {
 
 dependencies {
   compileOnly(Dependencies.detektApi)
+  compileOnly(Dependencies.detektPsiUtils)
   compileOnly(Dependencies.kotlinCompilerEmbeddable)
 
   testImplementation(Dependencies.assertj)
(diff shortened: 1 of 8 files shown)


detektive/build.gradle.kts
```diff
@@ -5,11 +5,13 @@ plugins {
 
 dependencies {
   compileOnly(Dependencies.detektApi)
+  compileOnly(Dependencies.detektPsiUtils)
   compileOnly(Dependencies.kotlinCompilerEmbeddable)
 
   testImplementation(Dependencies.assertj)
   testImplementation(Dependencies.detektTest)
   testImplementation(Dependencies.detektTestUtils)
+  testImplementation(Dependencies.guice)
```
mmollaverdi, line 14: Only needed for my test case which includes a guice annotation (detekt fails to recognise the annotation otherwise).

I've exclude this from the dependency analysis errors as a result (in the root build file)

jdm-square (approved):

PR #2883, opened 2023-07-21
feat: add memory cache to readiness check

kyeotic:
# Description

This adds a new `ReadinessCheckService` responsible for refreshing the health status and serving the cached result. `WebConfig` has been updated to include two new values to control the caching behavior for refresh interval and max age.

This updates the `apiDump`, so could constitute a breaking change.
This work is part of the Lifecycle Groups work for Misk 1.0


misk/src/main/kotlin/misk/web/actions/ReadinessCheckAction.kt
```diff
@@ -37,20 +47,42 @@ class ReadinessCheckAction @Inject internal constructor(
...
+
+    // First run needs to block and get statuses
+    if (lastUpdate == null) {
+      refreshStatuses()
+    } else {
+      // This needs to be independent of the max age check so that services can recover
+      // If refreshStatuses() is never run then a failed/stale check can never recover
+      if (clock.instant().isAfter(lastUpdate!!.plusMillis(config.readiness_refresh_interval_ms.toLong()))) {
+        // Launch in the background so that we can still return the cached state
+        launch(Job()) {
```
tgregory-block, line 61: Should we get a thread from misk's health executor thread pool instead?

misk/src/main/kotlin/misk/web/actions/ReadinessCheckAction.kt
```diff
@@ -37,20 +47,42 @@ class ReadinessCheckAction @Inject internal constructor(
...
 
     for (healthCheck in failedHealthChecks) {
       logger.info("Failed health check: ${healthCheck.messages}")
     }
-    return Response("", statusCode = 503)
+    return@runBlocking Response("", statusCode = 503)
+  }
+
+  private fun refreshStatuses() {
+    statuses = healthChecks.map { it.status() }
```
tgregory-block, line 85: Should we add a synchronization mechanism around this state? We should normally never have concurrent access, but if a service has slow health checks or a frequent readiness probe it could happen

misk/src/main/kotlin/misk/web/actions/ReadinessCheckAction.kt
```diff
@@ -7,25 +7,35 @@ import misk.web.AvailableWhenDegraded
...
 
+  var lastUpdate: Instant? = null
+  var statuses: List<HealthStatus> = listOf()
+
   @Get("/_readiness")
   @ResponseContentType(MediaTypes.APPLICATION_JSON)
   @Unauthenticated
   @AvailableWhenDegraded
-  fun readinessCheck(): Response<String> {
+  fun readinessCheck(): Response<String> = runBlocking {
```
adrw, line 38: Why are we using coroutines? Misk/Jetty is blocking.

misk/src/main/kotlin/misk/web/actions/ReadinessCheckAction.kt
```diff
@@ -37,20 +47,42 @@ class ReadinessCheckAction @Inject internal constructor(
...
+      // This needs to be independent of the max age check so that services can recover
+      // If refreshStatuses() is never run then a failed/stale check can never recover
+      if (clock.instant().isAfter(lastUpdate!!.plusMillis(config.readiness_refresh_interval_ms.toLong()))) {
+        // Launch in the background so that we can still return the cached state
+        launch(Job()) {
+          refreshStatuses()
+        }
+      }
+
+      if (clock.instant().isAfter(lastUpdate!!.plusMillis(config.readiness_max_age_ms.toLong()))) {
```
adrw, line 66: Would using a Guava cache here end up being simpler than rolling our own cache invalidation logic? https://github.com/google/guava/wiki/CachesExplained

misk/src/main/kotlin/misk/web/WebConfig.kt
```diff
@@ -124,6 +124,14 @@ data class WebConfig @JvmOverloads constructor(
    *  Use this value only when necessary.
    */
   val override_shutdown_idle_timeout: Long? = null,
+
+  /** How often readiness will re-run its status check.
+   *
+   * Ensure that [readiness_refresh_interval_ms] + "readiness latency" is less than [readiness_max_age_ms] or readiness will fail."  */
+  val readiness_refresh_interval_ms: Int = 1000,
```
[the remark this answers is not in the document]
kyeotic: Thats a good idea, I will add that validation

misk/src/main/kotlin/misk/web/actions/ReadinessCheckAction.kt
kyeotic: I want the actual refresh to happen in the background, not on the request thread. Is there a better way to do this?

misk/src/main/kotlin/misk/web/actions/ReadinessCheckAction.kt
kyeotic: I was not aware of these, the cache described under [refresh](https://github.com/google/guava/wiki/CachesExplained#refresh) sounds exactly like what I want. Thanks for the tip!

misk/src/main/kotlin/misk/web/actions/ReadinessCheckAction.kt
tgregory-block: Related to the executor pool question above, I think ideally we'd inject the thread pool misk [uses here](https://github.com/cashapp/misk/blob/78a95ce4e3f0a90fbdbe6d05486adcc96ff13145/misk/src/main/kotlin/misk/web/jetty/JettyService.kt#L70
https://github.com/tgregory-block/misk/blob/78a95ce4e3f0a90fbdbe6d05486adcc96ff13145/misk/src/main/kotlin/misk/web/jetty/JettyService.kt#L70) and then use its `submit` method. Would probably need to install the pool as an executor service module to make it injectable

misk/src/main/kotlin/misk/web/actions/ReadinessCheckAction.kt
tgregory-block: [Added a comment below](https://github.com/cashapp/misk/pull/2883#discussion_r1272733408) about installing the thread pool as an executor service so that it's injectable, then using that instead of the coroutine job

misk/src/main/kotlin/misk/web/actions/ReadinessCheckAction.kt
kyeotic: Hmm, @adrw I'm not sure I can get the correct semantics out of this without leaving in all of the logic that's already here. I want 4 things:

1. The initial load to fire after services have all started. This requires not running health checks on construction.
2. The cached value to be served.
3. The cache to refresh after REFRESH_TIME
4. An ERROR to happen if the cache is older than MAX_AGE

I don't see a way to get that last behavior. The `LoadingCache` will never return `null`, it will always run the `load` function. The cache does not return the age of the item. The load function cannot access the age of the item. So in order to detect Health Checks that are taking too long to run we would have to add our own stale-time tracking.

misk/src/main/kotlin/misk/web/actions/ReadinessCheckAction.kt
kyeotic: So first, I'm still learning how threads and coroutines work in Kotlin, so I'm happy to be corrected on any of this.

Reading the [coroutine docs](https://kotlinlang.org/docs/coroutine-context-and-dispatchers.html#dispatchers-and-threads) it sounds like the default dispatcher for a job is the parent's context. I think that means that by leaving the context blank, e.g. `launch(Job())` that the `healthExecutor` will be used, since its the context of the parent.

Does that sound right to you?

tgregory-block: I'm also not super familiar with kotlin coroutines (more familiar with Swift's oddly enough, but Kotlin coroutines are also designed around structured concurrency so it looks fairly similar), but my reading of the docs is that:
* `launch` with no specified `Dispatcher` will run in the context of its current coroutine scope
* Here, that scope comes from the `runBlocking` that's added in the PR
* `runBlocking` uses "an event loop on the current thread" by default

So indeed, it sounds like it will run on the thread the action already uses from `healthExecutor`. I'm not sure if there's any reason to be concerned about using `launch` in combination with a thread pool executor (e.g. any weirdness between the coroutine's "event loop" and the thread pool's own synchronous queue for work management?) @adrw are you able to shed any light on this?

kyeotic: I have a couple reservations about extracting the `healthExecutor`.

First, I'm not sure that using a thread pool with 4-8 threads is what we want. If concurrent requests to readiness are made I think we want at most 1 thread refreshing the status. I think we want a dedicated thread (or single thread pool) for that.

Second, the complexity impact is awkward.

* `healthExecutor` is only created when there is a `health_port` mapped. From a DI perspective this means creating an optional binding and moving that logic out of `JettyService` into another module, then handling that nullability in both `JettyService` and `ReadinessCheckAction`.
  * In `ReadinessCheckAction` its not actually nullable, but will still be getting a nullable value.
* It exposes that executor for use in other places, which is almost certainly unwanted.

I would prefer to create a private executor inside of `ReadinessCheckAction`. @tgregory-block thoughts?

tgregory-block: Ah that's a good idea. It keeps one thread for running the checks themselves and leaves `healthExecutor` to handle the web requests, and also gets rid of the concurrency issue.

misk/src/main/kotlin/misk/web/actions/ReadinessCheckAction.kt
```diff
@@ -37,20 +62,57 @@ class ReadinessCheckAction @Inject internal constructor(
...
+    val (lastUpdate, statuses) = cachedStatus
+      ?: refreshStatuses()
+
+    // This needs to be independent of the max age check so that services can recover
+    // If refreshStatuses() is never run then a failed/stale check can never recover
+    if (
+      clock.instant().isAfter(lastUpdate.plusMillis(config.readiness_refresh_interval_ms.toLong()))
+      // don't launch concurrent refreshes
+      && refreshExecutor.activeCount == 0
+    ) {
```
tgregory-block, line 79: It might be easier to reason about the refresh logic if we packaged up the check status refresh into a recurring task and let a `RepeatedTaskQueue` manage the threading, and expose the most recent check result and timestamp on some singleton

misk/src/main/kotlin/misk/web/actions/ReadinessCheckAction.kt
```diff
@@ -1,31 +1,56 @@
...
 @Singleton
 class ReadinessCheckAction @Inject internal constructor(
   private val serviceManagerProvider: Provider<ServiceManager>,
-  @JvmSuppressWildcards private val healthChecks: List<HealthCheck>
+  @JvmSuppressWildcards private val healthChecks: List<HealthCheck>,
+  val config: WebConfig,
+  val clock: Clock
 ) : WebAction {
 
+  @Volatile
```
tgregory-block, line 35: Do we need `Volatile`, since the thread pool is bound to a single thread?

kyeotic: The read comes from a different thread than the write, since its not on the refreshPool

misk/src/main/kotlin/misk/web/actions/ReadinessCheckAction.kt
```diff
@@ -40,9 +69,35 @@ class ReadinessCheckAction @Inject internal constructor(
...
     }
 
-    val failedHealthChecks = healthChecks
-      .map { it.status() }
-      .filter { !it.isHealthy }
+    // Only null on first run
+    val (lastUpdate, statuses) = cachedStatus
+      // Block to get the initial status and enqueue background refresh
+      ?: refreshStatuses().also {
+        taskQueue.scheduleWithBackoff(Duration.ofMillis(config.readiness_refresh_interval_ms.toLong())) {
```
frojasg, line 76: Should we move this to separate `AbstractIdleService`? I worry that here we could schedule multiple job  (once every time this webaction is called.

tgregory-block: Agreed this should be moved into its own class

kyeotic: This schedule action is only called when `cachedStatus == null`, which is only true on the first run... though I guess if two requests came in at the same time that could still happen.

kyeotic: @tgregory-block Do you think an abstract service replaces the need for the `RepeatedTaskQueue`? I think it does, since it covers the run-as-single-thread benefit, but I'm curious what you think?

tgregory-block: Something in that class would still periodically need to run the refresh, so personally I'd stick the task queue in the abstract service. The abstract svc does get its own threads in `startup()` and `shutdown()` but they're not accessible after those methods return

misk/src/main/kotlin/misk/web/actions/ReadinessCheckAction.kt
```diff
@@ -37,20 +62,57 @@ class ReadinessCheckAction @Inject internal constructor(
...
+      refreshExecutor.execute {
+        refreshStatuses()
+      }
     }
 
-    val failedHealthChecks = healthChecks
-      .map { it.status() }
-      .filter { !it.isHealthy }
+    if (clock.instant().isAfter(lastUpdate.plusMillis(config.readiness_max_age_ms.toLong()))) {
+      logger.info("Failed health check: last status check is older than max age")
```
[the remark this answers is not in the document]
kyeotic: The log level for a failed health check is `info`, I'm not sure this is more serious than that.

misk/src/main/kotlin/misk/web/MiskWebModule.kt
```diff
@@ -119,6 +123,12 @@ class MiskWebModule @JvmOverloads constructor(
       ServiceModule<JettyConnectionMetricsCollector>()
         .enhancedBy<ReadyService>()
     )
+    install(
+      ServiceModule<ReadinessCheckService>()
+        .enhancedBy<ReadyService>()
```
tgregory-block, line 128: Was going to say we should probably use `dependsOn` and shut this down as soon as we get sigterm but [apparently](https://github.com/kubernetes/kubernetes/issues/107473) k8s probes after shutdown are a little flaky. Maybe add that issue in a comment?

misk/src/main/kotlin/misk/web/MiskWebModule.kt
```diff
@@ -246,6 +256,8 @@ class MiskWebModule @JvmOverloads constructor(
     // Bind build-in actions.
     install(WebActionModule.create<StatusAction>())
     install(WebActionModule.create<ReadinessCheckAction>())
+    install(ServiceModule<RepeatedTaskQueue>(ReadinessRefreshQueue::class))
```
tgregory-block, line 259: Can we drop this? It looks like it's getting installed twice, here and L131

misk/src/main/kotlin/misk/web/actions/ReadinessCheckService.kt
```diff
@@ -0,0 +1,89 @@
...
+  private val config: WebConfig,
+  private val clock: Clock,
+  private val serviceManagerProvider: Provider<ServiceManager>,
+  @JvmSuppressWildcards private val healthChecks: List<HealthCheck>,
+  @ReadinessRefreshQueue private val taskQueue: RepeatedTaskQueue,
+): AbstractIdleService() {
+  var status: CachedStatus? = null
+    private set
+
+  private val isServiceActive = config.health_port >= 0
```
tgregory-block, line 32: Should we just always run this? Setting `health_port != -1` isn't required yet, so there could be a service that doesn't have it. In that case, will always be null and readiness probes will always fail

kyeotic: I don't think so, but I'm far from certain. The `JettyService` does not create a health connector unless `webConfig.health_port >= 0`, which I think means the Readiness action is not accessible. If thats the case then running a health check every 2 seconds that is inaccessible is pure waste. We could be more explicit about this, but then existing configurations that rely on this behavior will not work.

tgregory-block: The readiness action is always accessible (without it the pod will never start). Setting `health_port != -1` only causes misk to create a dedicated thread pool for handling requests on that port. When `health_port == -1` the health probes still come in, and are handled by `ReadinessCheckAction`, they just use the same thread pool as all other requests

kyeotic: Ok, I've removed this bit

misk/src/main/kotlin/misk/web/actions/ReadinessCheckAction.kt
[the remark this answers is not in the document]
tgregory-block: Oh, so it is

misk/src/main/kotlin/misk/web/actions/ReadinessCheckService.kt
```diff
@@ -0,0 +1,83 @@
...
+  private val clock: Clock,
+  private val serviceManagerProvider: Provider<ServiceManager>,
+  @JvmSuppressWildcards private val healthChecks: List<HealthCheck>,
+  @ReadinessRefreshQueue private val taskQueue: RepeatedTaskQueue,
+): AbstractIdleService() {
+  var status: CachedStatus? = null
+    private set
+
+  init {
+    require(config.readiness_refresh_interval_ms < config.readiness_max_age_ms || config.health_port <= 0) {
```
tgregory-block, line 33: nit - should probably drop the `|| config.health_port <= 0` since that requirement applies either way

tgregory-block (approved):
LGTM :shipit: 🙏

misk/src/main/kotlin/misk/web/actions/ReadinessCheckService.kt
```diff
@@ -0,0 +1,83 @@
...
+      "readiness_refresh_interval_ms must be less than readiness_max_age_ms"
+    }
+  }
+
+  override fun startUp() {
+    logger.info { "starting readiness service" }
+
+    val delay = Duration.ofMillis(config.readiness_refresh_interval_ms.toLong())
+    taskQueue.schedule(delay) {
```
cconroy, line 42: I think this `delay` is an initial delay here. But, I expect we probably want to run these checks without delay for the first run (use `0` here and `delay` in the result computation below).

misk/src/main/kotlin/misk/web/MiskWebModule.kt
kyeotic: This `enhancedBy` is a commonality for all services in this module. Are you worried about its usage here specifically, or just the general lack of dependency?

misk/src/main/kotlin/misk/web/actions/ReadinessCheckService.kt
```diff
@@ -0,0 +1,83 @@
...
+    for (service in servicesNotRunning) {
+      logger.info("Service not running: $service")
+    }
+
+    // Don't do healthchecks if services haven't all started. The app isn't in a good state yet,
+    // and a health check could end up triggering random errors that we don't want to flood the
+    // logs with.
+    if (!servicesNotRunning.isEmpty()) return
+
+    val statuses = healthChecks.map { it.status() }
```
cconroy, line 68: this will be doing all the individual health checks serially here. that's probably _mostly_ fine, but it is going to mean slower refresh than in parallel.

in the limit, this can mean one slow check can delay the updating of health status for everyone else, but we are bounded by the max age (since we'll go unhealthy regardless if it's slower than that to refresh). 

Updating to parallel execution here would help though still requires each check round to wait for all checks to complete before scheduling another one.

Service Container attempts to address this by having a thread per check to drive its refresh which works to a point (it does have a static limit on that thread pool size which if exceeded then puts you at some risk of queue delay though in practice I don't think anyone has more than 50 health checks)

kyeotic: This check has always been serial, and I think thats intentional. When this check was handled by the Action directly it was part of a very limited executor (max 8 threads). Im hesitant to increase the thread consumption of this by a significant or variable amount.

This is happening in the background.

misk/src/main/kotlin/misk/web/actions/ReadinessCheckAction.kt
cconroy: Even if we're updating status only from one thread, we're going to be reading it from another. So, we need to consider the memory visibility of our cached status. `@Volatile` seems adequate for the current structure but we could consider alternate concurrency primitives if desired.

misk/src/main/kotlin/misk/web/actions/ReadinessCheckService.kt
```diff
@@ -0,0 +1,83 @@
...
+import wisp.logging.getLogger
+import java.time.Clock
+import java.time.Duration
+import java.time.Instant
+import javax.inject.Inject
+import javax.inject.Provider
+import javax.inject.Singleton
+
+@Singleton
+class ReadinessCheckService @Inject constructor(
```
frojasg, line 22: this can be internal

misk/src/main/kotlin/misk/web/MiskWebModule.kt
```diff
@@ -312,3 +329,7 @@ class MiskWebModule @JvmOverloads constructor(
     val miskCallerType = object : TypeLiteral<MiskCaller?>() {}
   }
 }
+
+@BindingAnnotation
+@Target(AnnotationTarget.FIELD, AnnotationTarget.FUNCTION, AnnotationTarget.VALUE_PARAMETER)
+annotation class ReadinessRefreshQueue
```
frojasg, line 335: this can be internal

misk/src/main/kotlin/misk/web/actions/ReadinessCheckService.kt
```diff
@@ -0,0 +1,83 @@
...
+    logger.info { "starting readiness service" }
+
+    taskQueue.schedule(Duration.ofMillis(0)) {
+      refreshStatuses()
+      Result(Status.OK, Duration.ofMillis(config.readiness_refresh_interval_ms.toLong()))
+    }
+  }
+
+  override fun shutDown() {
+    logger.info { "stopping readiness service" }
```
frojasg, line 49: how long is the shutdown time? I worry we can shutdown and stop doing the `refreshStatuses` and cause a cache expiration case in the `ReadinessCheckAction`

misk/src/main/kotlin/misk/web/actions/ReadinessCheckService.kt
```diff
@@ -0,0 +1,83 @@
...
+    if (!servicesNotRunning.isEmpty()) return
+
+    val statuses = healthChecks.map { it.status() }
+
+    // Get time AFTER health checks have completed
+    val lastUpdate = clock.instant()
+    status = CachedStatus(lastUpdate, statuses)
+  }
+
+  data class CachedStatus(
```
frojasg, line 75: this can be internal

misk/src/main/kotlin/misk/web/actions/ReadinessCheckAction.kt
```diff
@@ -1,48 +1,42 @@
...
-    }
+    val (lastUpdate, statuses) = readinessService.status
+      // Null until first run has finished, which means not yet ready
+      ?: return Response("", statusCode = 503)
 
-    if (!servicesNotRunning.isEmpty()) {
-      // Don't do healthchecks if services haven't all started. The app isn't in a good state yet,
-      // and a health check could end up triggering random errors that we don't want to flood the
-      // logs with.
+    if (clock.instant().isAfter(lastUpdate.plusMillis(config.readiness_max_age_ms.toLong()))) {
```
frojasg, line 34: should we push all this logic to the `readinessService`?

misk/src/main/kotlin/misk/web/actions/ReadinessCheckService.kt
frojasg: FWIW: every healthChecks are running on separate threads, and the [status](https://guava.dev/releases/19.0/api/docs/com/google/common/util/concurrent/Service.html#state()) should return immediately ... which makes me wonder what's the value of this cache 😅 ?

frojasg: oh. I got confused between the `HealthCheck` status and the [Service](https://guava.dev/releases/19.0/api/docs/com/google/common/util/concurrent/Service.html#state()) status

frojasg: I worry this loop can take longer than `readiness_max_age_ms` and take the application down. We should add some tracing to get a sense how much each `HealthCheck` is taking, and do something if we're taking longer than `readiness_max_age_ms`

misk/src/main/kotlin/misk/web/actions/ReadinessCheckService.kt
kyeotic: I dont understand the concern. If you are shutting down the pod to stop doing work like health checks, right?

misk/src/main/kotlin/misk/web/actions/ReadinessCheckAction.kt
kyeotic: I left this in the action because its a check against slowness or failures from the service. To put this in the service it would need to be on the getter... but then what would it do? Throw an error? Return null (which would be indistinguishable from "has not yet run")?

I think it makes more sense here.

misk/src/main/kotlin/misk/web/actions/ReadinessCheckService.kt
kyeotic: Should it be? It could be useful for apps to override the service with their own implementation, and I don't see any reason for the framework to prohibit this.

misk/src/main/kotlin/misk/web/MiskWebModule.kt
kyeotic: Should it be? It could be useful for apps to override the service with their own implementation, and I don't see any reason for the framework to prohibit this.

misk/src/main/kotlin/misk/web/actions/ReadinessCheckService.kt
kyeotic: This is a nested class, so it follow its parents visibility

misk/src/main/kotlin/misk/web/actions/ReadinessCheckService.kt
kyeotic: We discussed this in slack. A spot check of all health checks in datadog shows only one instance of > 10s response in the last 2 days, and 2 instances of > 5 seconds. Bumping this to 10 seconds as that case should be an error.

misk/src/main/kotlin/misk/web/actions/ReadinessCheckService.kt
```diff
@@ -0,0 +1,83 @@
...
+
+@Singleton
+class ReadinessCheckService @Inject constructor(
+  private val config: WebConfig,
+  private val clock: Clock,
+  private val serviceManagerProvider: Provider<ServiceManager>,
+  @JvmSuppressWildcards private val healthChecks: List<HealthCheck>,
+  @ReadinessRefreshQueue private val taskQueue: RepeatedTaskQueue,
+): AbstractIdleService() {
+  var status: CachedStatus? = null
```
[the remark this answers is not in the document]
kyeotic: updated

misk/src/main/kotlin/misk/web/actions/ReadinessCheckService.kt
kyeotic: This is a nested class, if it doesn't follow its parents visibility it could make the parent unusable

misk/src/main/kotlin/misk/web/MiskWebModule.kt
frojasg: The values on the `ReadinessRefreshQueue` is exposed in the `readiness_refresh_interval_ms` config. I'm not sure what would be value of exposing this.

kyeotic: We discussed this out of band, Ive marked it as internal

misk/src/main/kotlin/misk/web/actions/ReadinessCheckService.kt
kyeotic: We discussed this out of band, Ive marked it as internal

frojasg (approved):

PR #2885, opened 2023-07-24
Add getset support to misk-redis

salamagd:
Adds support for getting the current value and setting a new value on a single key in an atomic command. Introduces the `getset` and `getsetnx` commands, which can both be used with or without an `expiryDuration`, with the latter only setting the new value if there is a current, non-expired value at the specified key. Behind the scenes, this uses the Redis `SET` command with the `GET` option. `getsetnx` (ie. the use of `SET` with both `GET` and `NX` options) is only supported with Redis version 7.0 and later.

This change also bumps the Docker Redis `alpine` image version used for testing from `6.2` to `7.0`, the current stable version.

### DIFF
--- a/misk-redis-testing/src/main/kotlin/misk/redis/testing/DockerRedis.kt
+++ b/misk-redis-testing/src/main/kotlin/misk/redis/testing/DockerRedis.kt
@@ -31,7 +31,7 @@ object DockerRedis : ExternalDependency {
   private const val port = 6379
   private val hostname = if (isRunningInDocker) "host.docker.internal" else "localhost"
   private val logger = getLogger<DockerRedis>()
-  private const val redisVersion = "6.2"
+  private const val redisVersion = "7.0"
 
   private val client by lazy { JedisPool(hostname, port) }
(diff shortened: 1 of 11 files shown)


adrw (approved):
LG

misk-redis-testing/src/main/kotlin/misk/redis/testing/DockerRedis.kt
```diff
@@ -31,7 +31,7 @@ object DockerRedis : ExternalDependency {
   private const val port = 6379
   private val hostname = if (isRunningInDocker) "host.docker.internal" else "localhost"
   private val logger = getLogger<DockerRedis>()
-  private const val redisVersion = "6.2"
+  private const val redisVersion = "7.0"
```
keeferrourke, line 34: We don't support Redis 7 yet!

keeferrourke (changes requested):

salamagd:
Closing for now due to not getting Redis 7 support for a while.

Issue #2863, opened 2023-07-16
Migration to Guice 7+


Nava2:
Wanted to make an issue for tracking migration to Guice 7.0.0. 

@ThePumpingLemma migrated to using Guice 6.0.0 (#2846), however, there remains issues with projects wanting to use `misk-inject` with Guice 7+. The largest issue is that when a project tries to use Guice 7+, it can not inject classes that are labelled with `javax.inject.XXX` annotations. Guice 6+ happily accepts _both_ `jakarta` and `javax`.

## Proposed path forwards

It _seems_ like the path to migration would be twp steps:

1. Replace all usages of `javax.inject.XXX` with `com.google.inject.XXX`
    * This enables projects to transition to using `jakarta.inject` in their own projects, while consuming `misk-inject` @ 6.0.0
2. _Eventually_ cut over to Guice 7.0.0 at _some point_ 
3. _Optional:_ Replace `com.google.inject.XXX` with `jakarta.inject.XXX`

## Open question

1.  misk doesn't seem to use something like `detekt` to enable [forbidding imports](https://detekt.dev/docs/rules/style/#forbiddenimport), is there a way for us to enforce only using `com.google.inject.XXX`? 
  * IntelliJ can also be adjusted to ignore certain imports for "Auto import" ([Link](https://www.jetbrains.com/help/idea/creating-and-optimizing-imports.html#exclude-from-auto-import)), but there is no intellij configuration within misk by default

P.S. I'm happy to do the find-and-replace 😄

r3mariano:
👋 I'm one of the engineers working on frameworks at Cash. 

We'd have to migrate our internal downstream services before we're able to do the migration to Guice 7. Will talk to the team about where to prioritise that. Given Guice 7 doesn't really have anything that isn't in Guice 6 at the moment, I'd lean towards doing the cutover later.

Using detekt would be great to check for imports!

r3mariano:
I'm curious btw, why `com.google.inject` instead of `jakarta.inject` if they both work? I think I'd prefer to use the more generic one of the two where possible.

Nava2:
If we used Jakarta, it would force usage of 6.0+.. which seems fine... 🤦🏻 Admittedly, I'm biased as our downstream services are not on 6.0 yet and it would force our update - which we need to do.

Nava2:
I agree strongly with holding off on the cutover until later, but doing the initial migration to Jakarta seems good to get done.

PR #2890, opened 2023-07-31
feat: [Guice 6] Migrate to use `jakarta.inject.*` for injection annotations

Nava2:
Part of #2863.

This replaces usages to enable migration to [Guice 7+](https://github.com/google/guice/wiki/Guice700). This is the first (only?) step to allow projects using Guice 6.0.0 or Guice 7.0.0 to use misk-inject.

By using `jakarta.inject` instead of `javax.inject`, injections work in both 6+ and 7+ versions. The new project, `:misk-inject-guice7-test` sets up a test project with the classpath forced to Guice 7+. This acts as a guard project for basic usages. As needed, we can add more content.

The exception is using `com.google.inject.Provider` as Guice 6 does not provide bindings that work with `jakarta.inject.Provider` in all cases, making this a frustrating requirement.


detekt.yaml
```diff
@@ -21,6 +21,22 @@ complexity:
   LongParameterList:
     active: false
 
+style:
```
Nava2, line 24: Bans imports in future code

Nava2:
cc @adrw / @r3mariano

r3mariano (approved):

Nava2:
@r3mariano I'm unsure how to actually land this change. Could you advise on process or documentation?

Nava2:
@jdm-square looks like there's some dependency cleanups and general lint things. Is there an autoformat somewhere.. was hoping the detekt formatting rules were in place but seem to not be. 

Also looks like I need to regenerate the API lockfiles, easy enough.

jdm-square:
I don't see any linting issues that require formatting, only the missing api dump and dependency changes.

r3mariano:
> @r3mariano I'm unsure how to actually land this change. Could you advise on process or documentation?

@Nava2 Nothing particular for the OSS repo. We're already aware of the change and can handle the change in our internal services (i.e. upgrade everyone to 6). Releases already happen automatically, we switched to a date-based versioning scheme that publishes for every change in main.

There is an incoming change to add more (non-Guice) modules, hopefully that won't cause a conflict.

jdm-square:
```
CronTest > leaseDenied() FAILED
    org.opentest4j.AssertionFailedError: 
    expected: 1
     but was: 0
        at java.base@17.0.7/jdk.internal.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
        at java.base@17.0.7/jdk.internal.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:77)
        at java.base@17.0.7/jdk.internal.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
        at java.base@17.0.7/java.lang.reflect.Constructor.newInstanceWithCaller(Constructor.java:499)
        at app//misk.cron.CronTest.leaseDenied(CronTest.kt:78)
```
I'll run it again to see if it's an intermittent error.

Nava2:
@jdm-square / @r3mariano could one of you enable the tests to run again? :) This should be bueno now!

Nava2:
Thanks to whomever enabled the CI, I can't actually merge this :shipit:

PR #2891, opened 2023-08-01
Add a couple MDC attributes for SQS consumers

tgregory-block:
Adds the following MDC attributes:
* `misk.job_queue.queue_type` - in case we ever implement another type of JobConsumer
* `misk.job_queue.queue_name` - the name of the queue being consumed
* `misk.job_queue.job_id` - A new [structured](https://cash-dev-guide.squarecloudservices.com/observability/logs/#adding-structured-fieldsnamespaces-to-log-messages) attribute for the job ID. Retained the current one for backwards compat


misk-aws/src/main/kotlin/misk/jobqueue/sqs/SqsJobConsumer.kt
```diff
@@ -196,6 +198,9 @@ internal class SqsJobConsumer @Inject internal constructor(
               // Run the handler and record timing
               try {
                 MDC.put(SQS_JOB_ID_MDC, message.id)
+                MDC.put(SQS_JOB_ID_STRUCTURED_MDC, message.id)
+                MDC.put(SQS_QUEUE_NAME_MDC, message.queueName.value)
+                MDC.put(SQS_QUEUE_TYPE_MDC, SQS_QUEUE_TYPE)
```
adrw, line 203: What is the value in this if it's always the same queue type?

adrw (approved):

misk-aws/src/main/kotlin/misk/jobqueue/sqs/SqsJobConsumer.kt
tgregory-block: The new tags aren't specific to SQS (unlike `SQS_JOB_ID_MDC` which is prefixed with `sqs_`), so I added this in case we implement other job consumers in the future. It will be hardcoded in each JobConsumer implementation. I can remove it for the time being if you prefer,it does seem unlikely a service would be wired up to multiple queue types

adrw: Ah makes sense, thanks!

PR #2892, opened 2023-08-01
PauseDetector: Summary->Histogram

cconroy:
We are seeing a bit more CPU usage than anticipated from the detector, and profiling shows that this is due to the CKMSQuantile calculation in prometheus.

The prometheus docs
[note](https://prometheus.io/docs/practices/histograms/) that summaries are more expensive locally than histograms. While the distribution quantiles can be interesting, we can sacrifice some accuracy there for lower detector overhead.

See https://cash.slack.com/archives/C01M5M297T7/p1690797686714589?thread_ts=1677880396.602369&cid=C01M5M297T7 and hat tip to @mateuszmrozewski for finding this issue.

### DIFF
--- a/misk/src/main/kotlin/misk/perf/PauseDetector.kt
+++ b/misk/src/main/kotlin/misk/perf/PauseDetector.kt
@@ -2,7 +2,7 @@ package misk.perf
 
 import com.google.common.base.Ticker
 import com.google.common.util.concurrent.AbstractExecutionThreadService
-import io.prometheus.client.Summary
+import io.prometheus.client.Histogram
 import misk.concurrent.Sleeper
 import misk.metrics.v2.Metrics
 import misk.metrics.v2.PeakGauge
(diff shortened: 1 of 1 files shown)


misk/src/main/kotlin/misk/perf/PauseDetector.kt
```diff
@@ -40,11 +40,11 @@ internal class PauseDetector @Inject constructor(
...
    * Tracks the distribution and total amount of pause time observed.
    *
-   * (Using a summary instead of a histogram to prefer more accurate node-local data over
-   * more accurate aggregations)
+   * (We prefer histogram to summary here because the latter has higher CPU overhead, this runs
+   * ideally at close to 1k QPS, and we don't strictly require accurate quantiles.)
    */
-  private val pauseSummary: Summary =
-    metrics.summary("jvm_pause_time_summary_ms", "Summary in ms of pause time", listOf())
+  private val pauseHistogram: Histogram =
```
cconroy, line 46: No code should be depending on this metric, we don't have any official dashboards for pause detection yet, and we have no known external dashboards or monitors making use of this summary. So, I think it should be safe to skip the backwards-compatibility dance here of keeping the old one and having some deprecation period as there is effectively no active usage of this metric other than writing to it.

adrw (approved):

frojasg (approved):

PR #2893, opened 2023-08-02
Bump aws v1 SDK

tgregory-block:
Following a [similar change](https://cash.slack.com/archives/C0253PG5N/p1690921482546079) in the java monorepo, this bumps the AWS SDK to take advantage of the new JSON protocol for interacting with SQS, which AWS claims reduces e2e message processing latency by up to 23%


tgregory-block:
@jdm-square Would you mind giving this another look? The SQS emulator misk used for local e2e tests hadn't been released in several years, and didn't support the JSON protocol, so I switched it to [ElasticMQ](https://github.com/softwaremill/elasticmq) instead, which is under active development and does support the JSON protocol. 

The only changes needed as a result of the container were:
1. Bumping the API client to `1.12.520` instead of `516`, as the exception marshalling wasn't mapping correctly with the emulator on `516` 
2. Changing the assertions on `queueProcessingLag` count down to `1`. This is because this metric isn't emitted when a message's receive count is `> 1`. The tests that needed the changes were all testing retries, which leads me to believe the old emulator we were using was buggy, and not properly setting `ApproximateReceiveCount`

misk-aws/src/test/kotlin/misk/jobqueue/sqs/DockerSqs.kt
```diff
@@ -43,12 +44,14 @@ internal object DockerSqs : ExternalDependency {
       // NB(mmihic): Because the client port is embedded directly into the queue URLs, we have to use
       // the same external port as we do for the internal port
       val exposedClientPort = ExposedPort.tcp(clientPort)
-      withImage("pafortin/goaws")
+      withImage("softwaremill/elasticmq:1.4.2")
         .withName("sqs")
         .withCmd(listOf("goaws"))
```
adrw, line 49: Is this command still the same?

misk-aws/src/test/kotlin/misk/jobqueue/sqs/SqsJobQueueTest.kt
```diff
@@ -414,7 +414,7 @@ internal class SqsJobQueueTest {
       sqsMetrics.jobsDeadLettered.labels(queueName.value, queueName.value).get()
     ).isEqualTo(0.0)
     assertThat(sqsMetrics.sqsDeleteTime.count(queueName.value, queueName.value)).isEqualTo(1)
-    assertThat(sqsMetrics.queueProcessingLag.count(queueName.value, queueName.value)).isEqualTo(3)
+    assertThat(sqsMetrics.queueProcessingLag.count(queueName.value, queueName.value)).isEqualTo(1)
```
adrw, line 417: Is this expected? Why is this not 3 anymore?

adrw (approved):

misk-aws/src/test/kotlin/misk/jobqueue/sqs/SqsJobQueueTest.kt
tgregory-block: See `2` [above](https://github.com/cashapp/misk/pull/2893#issuecomment-1662872500), I think the previous emulator was buggy and not wasn't updating `ApproximateReceiveCount`

misk-aws/src/test/kotlin/misk/jobqueue/sqs/DockerSqs.kt
tgregory-block: Thanks, removed. Their entry point seems to ignore arguments which is why this wasn't causing an issue

misk-aws/src/test/kotlin/misk/jobqueue/sqs/SqsJobQueueTest.kt
adrw: Ah it's approx, thanks for clarifying.

misk-aws/src/test/kotlin/misk/jobqueue/sqs/DockerSqs.kt
```diff
@@ -22,7 +23,7 @@ import wisp.logging.getLogger
 internal object DockerSqs : ExternalDependency {
 
   private val log = getLogger<DockerSqs>()
-  private const val clientPort = 4100
+  private const val clientPort = 9324
```
adrw, line 26: Is this going to conflict with anything or does this Docker image just have a different exposed port?

adrw (approved):

misk-aws/src/test/kotlin/misk/jobqueue/sqs/DockerSqs.kt
tgregory-block: The latter, elastic mq's default port is just different from the older image

PR #2897, opened 2023-08-03
Fix DashboardModule.createMenuLink

adrw:
Links which included `/` would fail on page load.

This PR also moves further up in the lifecycle slug and URL path prefix validation for better DX (failures will happen on boot, not on page load).

### DIFF
--- a/misk-admin/api/misk-admin.api
+++ b/misk-admin/api/misk-admin.api
@@ -339,7 +339,7 @@ public final class misk/web/dashboard/DashboardTabLoaderEntry {
 	public fun toString ()Ljava/lang/String;
 }
 
-public final class misk/web/dashboard/DashboardTabProvider : javax/inject/Provider {
+public final class misk/web/dashboard/DashboardTabProvider : misk/web/dashboard/ValidWebEntry, javax/inject/Provider {
 	public field accessAnnotationEntries Ljava/util/List;
 	public fun <init> (Ljava/lang/String;Ljava/lang/String;Ljava/lang/String;Ljava/lang/String;Ljava/lang/String;Ljava/lang/String;Ljava/util/Set;Ljava/util/Set;Lkotlin/reflect/KClass;)V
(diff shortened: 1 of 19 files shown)


misk-jdbc-testing/src/main/kotlin/misk/jdbc/ExtendedQueryExecutionListener.kt
```diff
@@ -47,7 +47,7 @@ open class ExtendedQueryExecutionListener : QueryExecutionListener, MethodExecut
     if (queryInfoList == null) return
 
     for (info in queryInfoList) {
-      val query = info.query.toLowerCase()
+      val query = info.query.lowercase()
```
adrw, line 50: `.toLowerCase()` is deprecated, `.lowercase()` is the replacement API

misk-admin/api/misk-admin.api
```diff
@@ -339,7 +339,7 @@ public final class misk/web/dashboard/DashboardTabLoaderEntry {
 	public fun toString ()Ljava/lang/String;
 }
 
-public final class misk/web/dashboard/DashboardTabProvider : javax/inject/Provider {
+public final class misk/web/dashboard/DashboardTabProvider : misk/web/dashboard/ValidWebEntry, javax/inject/Provider {
```
lkerford, line 342: Would you mind explaining how this solve the issue? I'm just not familiar enough with `ValidWebEntry` . The PR looks good but I'm a bit uncertain how it resolved our weird spinner issue

adrw: `ValidWebEntry` enforces that the `DashboardTabProvider` has been constructed with a valid slug and URL path prefix. This currently happens later in the lifecycle when the  `DashboardTabProvider` creates the `DashboardTab` on `get()` but that error is opaque and was difficult to debug.

adrw: The weird spinner issue was from an endpoint failing with the above invalid slug error. The change to the `DashboardModule` fixed the `createMenuLink` to ensure that the slug is always valid.

ean5533 (approved):

PR #2904, opened 2023-08-05
Add sqldelight module with a RetriableTransacter

mhickman:
I haven't added any guice bindings or anything (yet.) Just the
implementation of the Transacter for now.


misk-sqldelight-testing/api/misk-sqldelight-testing.api
```diff
@@ -0,0 +1,47 @@
+public final class misk/sqldelight/testing/Movies {
```
tso, line 1: can you move the `misk-sqldelight-testing` package into testFixtures inside of `misk-sqldelight`? or, since i think it wouldn't be used in consumer apps inside of `misk-sqldelight/test`?

misk-sqldelight/src/main/kotlin/misk/sqldelight/RetryingTransacter.kt
```diff
@@ -0,0 +1,107 @@
...
+// NB: all options should be immutable types as copy() is shallow.
+data class TransacterOptions(
+  val maxAttempts: Int = 3,
+  val minRetryDelayMillis: Long = 100,
+  val maxRetryDelayMillis: Long = 500,
+  val retryJitterMillis: Long = 400,
+)
+
+class RetryingTransacter(
+  private val delegate: Transacter,
```
tso, line 25: for most apps this isn't going to quite work i think? usually you have to inject like a `@MyAppDb transacter` right

tso (commented):
I think this looks pretty good but would like alec's thoughts!

misk-sqldelight-testing/api/misk-sqldelight-testing.api
mhickman: the reason I had it as a separate package was I needed the gradle sqldelight plugin to run + generate the sqldelight interact but I didn't want it part of the `main` package exposed as an artifact

misk-sqldelight/src/main/kotlin/misk/sqldelight/RetryingTransacter.kt
mhickman: I didn't even add an `@Inject` annotation here, the only thing is just calling the constructor for now, I want to think more about how to hook this up with guice - there's some subtleties because, in the test here (for example) the thing the _imlemenents_ `delegate: Transacter` is generated by the gradle plugin.

misk-sqldelight-testing/build.gradle.kts
```diff
@@ -0,0 +1,26 @@
+plugins {
+  kotlin("jvm")
+  `java-library`
+  id("app.cash.sqldelight")
```
mhickman, line 4: I didn't add the `id("com.vanniktech.maven.publish.base")` plugin here - I think that means this gradle module shouldn't publish to maven but would be nice to get confirmation. My intention is for this package to be internal-only.

AlecKazakova:
just looking at API I think this is nice. Very clean. Did not review implementation

misk-sqldelight/src/test/kotlin/misk/sqldelight/RetryingTransacterTest.kt
```diff
@@ -0,0 +1,161 @@
...
+    var tries = 0
+    val oneRetryTransacter = transacter.maxAttempts(2)
+    assertThrows<OptimisticLockException> {
+      oneRetryTransacter.transaction {
+        oneRetryTransacter.transaction {
+          tries++
+          throw OptimisticLockException("fake transient exception")
+        }
+      }
+    }
```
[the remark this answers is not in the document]
mhickman: I've done a crude implementation of 2 tracking the state on a threadlocal. The actual transacter internals from sqldelight _does_ track if we're in a transaction or not - but we don't have access to that in the scope of `RetryTransacter`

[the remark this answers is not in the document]
frojasg: I think this work for now. I worry this can have unintended consequences if folks are using coroutines during their db transaction, but let's keep this for now and iterate if it's needed.

frojasg (approved):
Thanks for working on this

frojasg (approved):
LGTM

misk-sqldelight/src/test/kotlin/misk/sqldelight/SqlDelightTestModule.kt
```diff
@@ -0,0 +1,92 @@
...
+  }
+
+  @Provides
+  @Singleton
+  fun provideMoviesDatabase(
+    @SqlDelightTestdb jdbcDriver: JdbcDriver,
+  ): MoviesDatabase {
+    val moviesDatabase = MoviesDatabase.invoke(jdbcDriver)
+
+    return object : MoviesDatabase, RetryingTransacter(moviesDatabase) {
```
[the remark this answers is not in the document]
mhickman: Yep! It's definitely ugly now, I was hesitant to do more than this. I think at very least I'll add a guice module to provide a `JdbcDriver` - it's tricksy to add one that provides `MoviesDatabase` since that's part of the codegen from sqldelight.

PR #2910, opened 2023-08-11
Remove pointless enhanced by

tso:


misk/src/main/kotlin/misk/web/MiskWebModule.kt
```diff
@@ -117,18 +117,9 @@ class MiskWebModule @JvmOverloads constructor(
...
-      ServiceModule<JettyConnectionMetricsCollector>()
-        .enhancedBy<ReadyService>()
-    )
-    install(
-      ServiceModule<ReadinessCheckService>()
-        .enhancedBy<ReadyService>()
-    )
+    install(ServiceModule<JettyThreadPoolMetricsCollector>().enhancedBy<ReadyService>())
+    install(ServiceModule<JettyConnectionMetricsCollector>().enhancedBy<ReadyService>())
+    install(ServiceModule<ReadinessCheckService>())
```
tso, line 122: enhancedBy would mean that readyness service has to wait for this to start -- but this is service is fine to start before, or after readyservice.

personally, I'd prefer if we instead made this dependOn readyservice and remove the checks that all services have started in this service, but that's just me

thoughts?

tgregory-block: I added the enhanced by to the metrics collectors so we would continue to collect those metrics during shutdown, until all services that intake work have halted (same for rationale for adding it to the `PrometheusHttpService` in the same PR). That be valuable to keep?

For `ReadinessCheckService` there was [some discussion](https://github.com/cashapp/misk/pull/2883/files#r1275257368) about that; ultimately I think it would be fine either way

tso: > I added the enhanced by to the metrics collectors so we would continue to collect those metrics during shutdown, until all services that intake work have halted (same for rationale for adding it to the PrometheusHttpService in the same PR). That be valuable to keep?

yeah this makes sense as you want to stop after all the work from (jetty, sqs, task queues) has been finished

PR #2915, opened 2023-08-15
fix: guard against bad service init with provider of health checks

kyeotic:
There appears to be an issue with the DynamoDB service bindings, which produces this error:

```
IllegalArgumentException: `dynamoDb` is only usable while the service is running
```

This change attempts to defend against it by lazily requesting the health checks, which will occur after all services are ready.


breeseAtSquare (approved):

frojasg (approved):
LGTM

tgregory-block (approved):

PR #2916, opened 2023-08-15
addLast Moshi adapters where appropriate

ean5533:
Moshi adapters have an order-dependent [precedence](https://github.com/square/moshi#precedence). The adapters added by Misk are all being added via `add()` -- even the general purpose ones -- which means it is impossible to override them with user-provided adapters. For example, right now it is impossible to write any adapter for any wire class and get it to actually be used by the injected `Moshi`, because the `WireOnlyJsonAdapterFactory` will take precedence.

This PR adds an optional `addLast` parameter to `MoshiAdapterModule`. If set to true, adapters will be added to Moshi via `addLast()`, which means they will have a lower precedence than any adapter added by `add()`.

This PR also changes `WireOnlyJsonAdapterFactory` and `KotlinJsonAdapterFactory` to be added last.


misk/src/test/kotlin/misk/moshi/MoshiModuleTest.kt
```diff
@@ -42,6 +44,24 @@ internal class MoshiModuleTest {
     assertThat(jsonAdapter.fromJson(json)).isEqualTo(value)
   }
 
+  /** Ensure we our custom Wire Message features, including ignoring unknown fields */
```
kryptonat, line 47: nit: "Ensure we use(?) ..."

kryptonat (approved):

wisp/wisp-moshi/src/main/kotlin/wisp/moshi/MoshiBuild.kt
```diff
@@ -9,18 +9,29 @@ import com.squareup.moshi.kotlin.reflect.KotlinJsonAdapterFactory
...
 
-    jsonAdapters.forEach { jsonAdapter ->
-        when (jsonAdapter) {
-            is JsonAdapter.Factory -> builder.add(jsonAdapter)
-            else -> builder.add(jsonAdapter)
-        }
+  jsonAdapters.forEach { jsonAdapter ->
+    when (jsonAdapter) {
+      is JsonAdapter.Factory -> builder.add(jsonAdapter)
+      else -> builder.add(jsonAdapter)
```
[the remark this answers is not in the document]
ean5533: It's to force the correct overload of `add` to be called. Moshi.Builder has several [overloads](https://github.com/square/moshi/blob/master/moshi/src/main/java/com/squareup/moshi/Moshi.kt#L211-L217), including one that takes an `Any` param and one that takes a `JsonAdapter.Factory` param, and these overloads do different things.

misk/src/test/kotlin/misk/moshi/MoshiModuleTest.kt
```diff
@@ -42,6 +44,24 @@ internal class MoshiModuleTest {
     assertThat(jsonAdapter.fromJson(json)).isEqualTo(value)
   }
 
+  /** Ensure we get our custom Wire Message features, including ignoring unknown fields */
+  @Test
+  fun wireFeaturesAreEnabled() {
+    val jsonAdapter = moshi.adapter<HelloRequest>()
```
adrw, line 50: What are these added tests exercising? They seem to be identical function body and not sure how they relate to the Moshi addLast changes?

wisp/wisp-moshi/src/main/kotlin/wisp/moshi/MoshiBuild.kt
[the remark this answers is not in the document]
adrw: Mind adding a comment? It isn't immediately obvious so could be a risk for future maintainers.

wisp/wisp-moshi/src/main/kotlin/wisp/moshi/MoshiBuild.kt
```diff
@@ -9,18 +9,29 @@ import com.squareup.moshi.kotlin.reflect.KotlinJsonAdapterFactory
...
     }
+  }
 
-    // Install last so that user adapters take precedence.
-    builder.add(KotlinJsonAdapterFactory())
+  // addLast so that user adapters take precedence.
+  jsonLastAdapters.forEach { jsonAdapter ->
+    when (jsonAdapter) {
```
adrw, line 28: comment here too pls

adrw (approved):

adrw:
Would be nice to see a motivating use case in tests that shows how someone would use the new addLast in a Module to get expected effect.

misk/src/test/kotlin/misk/moshi/MoshiModuleTest.kt
ean5533: This test (`wireFeaturesAreEnabled`) is quite similar to `kotlinFeaturesAreEnabled`, but differs in that it's specifically using a wire class, which means it triggers the application of `WireOnlyJsonAdapterFactory`. This isn't strictly related to my PR, it's just a test coverage gap that I'm plugging.

`customAdapterIsPreferredOverWireAdapter` is quite similar to `customAdapterIsPreferred`, but again is using a wire class. This test would fail without my changes (because `WireOnlyJsonAdapterFactory` was taking precedence).

adrw: Could you update the test name or comments to emphasize what you're specifically exercising?

adrw: Otherwise it seems unrelated to what it's confirming functionality for

ean5533: Can you suggest what the comments should change to? I'm not sure how else to explain it. Do the [comments on the pre-existing tests](https://github.com/cashapp/misk/blob/1548ad66fa278f67ce466dc4dfee480d63ddb410/misk/src/test/kotlin/misk/moshi/MoshiModuleTest.kt#L26) make sense to you?

ean5533: I'm going to go ahead and merge this so that it doesn't get stale while I'm on PTO, but I'll open a followup PR to make any comment or test method name changes

PR #2917, opened 2023-08-15
Prefer Histogram over Summaries

cconroy:
Summaries can be an order of magnitude more expensive in terms of CPU than Histograms. We should not be encouraging migration to Summaries for v1 histogram usage.


cconroy:
see https://cash.slack.com/archives/C915WS04Q/p1691080551547879

frojasg (approved):

ean5533 (approved):

adrw (approved):

PR #2919, opened 2023-08-17
Add connection_idle_timeout config to DataSourceConfig

tklovett:
h/t to @mpeyper, from whom I've copied all of his work in https://github.com/squareup/cash-cloud-kotlin/pull/491.

Current config creates [logs](https://square.datadoghq.com/logs?query=%22idleTimeout%20is%20close%20to%20or%20more%20than%20maxLifetime%2C%20disabling%20it.%22%20&cols=host%2Cservice&index=%2A&messageDisplay=inline&stream_sort=desc&viz=stream&from_ts=1692279430554&to_ts=1692365830554&live=true) similar to this one:
```
nacho_chip_factory - idleTimeout is close to or more than maxLifetime, disabling it.
```

Adding config option that defaults to `connection_max_lifetime - 10s` to prevent the warning from appearing and enable configuration if desired. The idea here is to keep it as close the the current logic (i.e. timeout when the connection is about to be killed), minus the warning log.

Note that there isn't a test around this logic as there isn't existing tests for `DataSourceService` and I didn't feel it was complex enough to warrant creating a test suite for this minor change. I could move the logic into `DataSourceConfig` if we prefer but I wasn't sure how the yaml -> class conversion works and if it wouldn't pick up the default logic if it was overridden in the yaml file so went with running it when the value is used instead. Another option is to not have default logic and rather just default it to a fixed value (e.g. `50s`) and force consumers to configure both if they are changing the `connection_max_lifetime` value. Happy to adjust the approach if there are any concerns.

Related to [support thread](https://cash.slack.com/archives/C915WS04Q/p1688478025994719)
Fixes [CCPOKR-8559](https://ccp-cashapp.atlassian.net/browse/CCPOKR-8559)


mpeyper (approved):

misk-jdbc/src/main/kotlin/misk/jdbc/DataSourceService.kt
```diff
@@ -72,6 +73,10 @@ class DataSourceService @JvmOverloads constructor(
     hikariConfig.poolName = qualifier.simpleName
     hikariConfig.connectionTimeout = config.connection_timeout.toMillis()
     hikariConfig.validationTimeout = config.validation_timeout.toMillis()
+    hikariConfig.idleTimeout = config.connection_idle_timeout.let {
+      val idleTimeout = it ?: config.connection_max_lifetime.minus(DEFAULT_IDLE_TIMEOUT_OFFSET)
```
adrw, line 77: How would it ever be null in this case within a let block? Or is it shadowing another it variable?

adrw (approved):

misk-jdbc/src/main/kotlin/misk/jdbc/DataSourceService.kt
tklovett: Its a `.let` rather than a `?.let`, so it really is an `it: Duration?`. The equivalent would be:
```
    hikariConfig.idleTimeout =
      (config.connection_idle_timeout ?: config.connection_max_lifetime.minus(
        DEFAULT_CONNECTION_IDLE_TIMEOUT_OFFSET
      )).toMillis()
```

adrw: Yes, I much prefer that final iteration. Thanks!

ean5533: Ah right, I shouldn't code with morning brain 😓

mpeyper:
(Not sure if I’m replying to the right comment chain from my email, but
this is intended for the thread about the `.let` vs `?.let` vs `?:` call)

I must not had had enough coffee that day as I much prefer `:? ` version
too. I remember struggling to make it half way readable but also iterated a
bunch of times on having it default in the constructor, default in a getter
then finally defaulting it here so probably couldn’t see the forest for the
trees at this point.

Can you roll what ever the final version is back into the
`cash-cloud-kotlin` version as well so they stay consistent please?

PR #2920, opened 2023-08-18
Deprecate no-argument @Authenticated behavior

sq-msamuel:
The default behaviour of `@Authenticated` or `@Authenticated(services=[], credentials=[])` is for it to allow all accesses, so long as there is a MiskCaller.

This is arguably surprising on its own accord, and is brittle - for example if you add an additional access annotation, it no longer has this effect.

This PR makes the following changes:
- Log a startup warning (from the factory) if an endpoint uses the empty-Authenticated behavior
- Add an AllowAnyService decorator, for endpoints that publish an API to their whole service mesh, and a corresponding ExcludeServiceFromWildcards binding so that environments can exclude access proxies (or anything else) from AllowAnyService
- Move MiskCaller.isAllowed() into AccessInterceptor.isAuthorized(), as this isn't really a property of a caller, it is a feature of the access interceptor


misk/src/main/kotlin/misk/security/authz/AccessAnnotation.kt
```diff
@@ -50,3 +51,22 @@ inline fun <reified T : Annotation> AccessAnnotationEntry(
...
+ * an AccessAnnotationEntry.
+ *
+ * Usage:
+ *  multibind<String, ExcludeServiceFromWildcards>().toInstance("web-proxy")
+ *  multibind<String, ExcludeServiceFromWildcards>().toInstance("access-proxy")
+ */
+@Target(AnnotationTarget.PROPERTY, AnnotationTarget.VALUE_PARAMETER)
+@Retention(AnnotationRetention.RUNTIME)
+@BindingAnnotation
+annotation class ExcludeServiceFromWildcards
```
frojasg, line 72: I wonder if it would be better to add this to `AccessAnnotationEntry` so it can be configured by endpoint

misk/src/main/kotlin/misk/security/authz/AccessInterceptor.kt
```diff
@@ -8,18 +9,19 @@ import misk.exceptions.UnauthenticatedException
 import misk.exceptions.UnauthorizedException
 import misk.scope.ActionScoped
 import wisp.logging.getLogger
-import jakarta.inject.Inject
 import kotlin.reflect.KClass
 
 class AccessInterceptor private constructor(
```
tso, line 14: it's weird to me that a private constructor ends up in the public api files, maybe a bug with our api dump?

misk/src/main/kotlin/misk/security/authz/AccessControlModule.kt
```diff
@@ -10,5 +10,6 @@ class AccessControlModule : ActionScopedProviderModule() {
   override fun configureProviders() {
     multibind<ApplicationInterceptor.Factory>().to<AccessInterceptor.Factory>()
     newMultibinder<AccessAnnotationEntry>()
+    newMultibinder<String, ExcludeServiceFromWildcards>()
```
tso, line 13: I wonder if this is the cleanest way

Alternatives would be... an `excludes` in the `AllowAll` annotation?

misk/src/main/kotlin/misk/security/authz/AccessInterceptor.kt
```diff
@@ -69,41 +88,51 @@ class AccessInterceptor private constructor(
...
+        allowAnyService,
+        excludeServiceFromWildcards.toSet()
       )
     }
 
     private fun Authenticated.toAccessAnnotationEntry() = AccessAnnotationEntry(
       Authenticated::class, services.toList(), capabilities.toList()
     )
 
-    private fun Unauthenticated.toAccessAnnotationEntry() = AccessAnnotationEntry(
```
tso, line 99: nice cleanup, this didn't make sense

misk-actions/src/main/kotlin/misk/MiskCaller.kt
```diff
@@ -29,6 +29,7 @@ data class MiskCaller @JvmOverloads constructor(
   }
 
   /** Determine based on allowed capabilities/services if the caller is permitted */
+  @Deprecated("This has been inlined into AccessInterceptor.isAuthorized()")
```
tso, line 32: why was this deprecated? what are services supposed to change to using?

sq-msamuel: My belief is that only AccessInterceptor or re-implementations of AccessInterceptor would want to use it.

misk/src/main/kotlin/misk/security/authz/AccessControlModule.kt
sq-msamuel: I wanted other modules (or the installing module) to easily add/set entries.

misk/src/main/kotlin/misk/security/authz/AccessInterceptor.kt
```diff
@@ -1,5 +1,6 @@
 package misk.security.authz
 
+import jakarta.inject.Inject
```
frojasg, line 3: wrong annotation

sq-msamuel: This is just a rearrangement of the imports (my IDE did it automatically)

misk/src/main/kotlin/misk/security/authz/AccessInterceptor.kt
```diff
@@ -27,9 +28,23 @@ class AccessInterceptor private constructor(
     return chain.proceed(chain.args)
   }
 
+  /** Check whether the caller is allowed to access this endpoint */
+  private fun isAuthorized(caller: MiskCaller): Boolean {
```
sq-msamuel, line 32: I moved this from lower in the file to make it clearer that it's not part of Factory

frojasg (approved):
LGTM

PR #2934, opened 2023-08-29
Add a way to add content to dashboard index page

adrw:
For instructions to fix missing access permissions or other content, this adds a way to inject arbitrary content into the dashboard home page.

# Authenticated
Screenshot 2023-08-31 at 10 00 51

# Unauthenticated
Screenshot 2023-08-31 at 10 43 29

Note: additional homepage content isn't shown at all if there are no authenticated tabs.


tso (approved):
looks legit

PR #2939, opened 2023-08-31
change ld client initialization success failure metric to gauge

adrianliu0815:


adrianliu0815:
@chris-ryan-square @staktrace @r3mariano could you pls help stamp this one? 
I occasionally needed to make some small changes in this repo, but not entirely sure whats the right process to get misk PR reviewed.

jdm-square (approved):

jdm-square:
Your best bet is contacting #misk-discuss for reviews.

staktrace (approved):
I verified using github codesearch and local grep that nobody appears to be using `FAILED_COUNTER_NAME` and `SUCCESS_COUNTER_NAME` fields so this change should be safe from an API point of view. I don't know if there are existing dashboards using these metrics and might fail but presumably if this was broken before then there shouldn't be a problem with the change.

adrianliu0815:
> presumably

yes nobody used it as it never worked

PR #2940, opened 2023-08-31
Add support for custom healthcheck execution for misk-aws-dynamodb

sangho-block:
### About
Support custom health checks for each dynamodb mapper/table for `misk-aws-dynamodb` clients. 
Custom health check provides clients to use different strategy for dynamodb health check over the [default one](https://github.com/cashapp/misk/blob/074409c49ba01c19d04bddc7a0b6f396a9377e8f/misk-aws-dynamodb/src/main/kotlin/misk/dynamodb/DynamoDbHealthCheck.kt#L18), which uses `AmazonDynamoDb.describeTable` control pane API.

### Why
The default health check's use of `describeTable` can be an issue when a large fleet of Misk applications regularly run periodic health checks against DynamoDB tables, as there is an account-level TPS threshold for control pane API requests per [AWS documentation](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/ServiceQuotas.html).

> DescribeTable/ListTables
> _You can submit up to 2,500 requests per second of read-only (DescribeTable, ListTables) control plane API requests in any combination._

This can lead to applications occasionally failing health checks due to unpredictable and elevated latency and shared account level threshold for control plane API calls.

Also, concurrent control plane API requests to the same table can be throttled, which can affect if a single service has many replicas.

This change allows clients to use alternative HealthCheck mechanisms per each mapper table.

### DIFF
--- a/misk-aws-dynamodb/src/main/kotlin/misk/dynamodb/RealDynamoDbModule.kt
+++ b/misk-aws-dynamodb/src/main/kotlin/misk/dynamodb/RealDynamoDbModule.kt
@@ -9,14 +9,14 @@ import com.amazonaws.services.dynamodbv2.AmazonDynamoDBStreamsClientBuilder
 import com.amazonaws.services.dynamodbv2.datamodeling.DynamoDBTable
 import com.google.common.util.concurrent.AbstractService
 import com.google.inject.Provides
+import jakarta.inject.Inject
+import jakarta.inject.Singleton
 import misk.ReadyService
 import misk.ServiceModule
 import misk.cloud.aws.AwsRegion
 import misk.exceptions.dynamodb.DynamoDbExceptionMapperModule
 import misk.healthchecks.HealthCheck
 import misk.inject.KAbstractModule
-import jakarta.inject.Inject
-import jakarta.inject.Singleton
 import kotlin.reflect.KClass
 import kotlin.reflect.full.findAnnotation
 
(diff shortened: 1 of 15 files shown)


misk-aws-dynamodb/src/main/kotlin/misk/dynamodb/RealDynamoDbModule.kt
```diff
@@ -25,15 +25,17 @@ import kotlin.reflect.full.findAnnotation
  * used to create a DynamoDbMapper for querying of a DynamoDb table.
  *
  * @param requiredTableTypes a list of mapper classes annotated [DynamoDBTable].
+ * @param customTableHealthChecks map of mapper class to [HealthCheck] for custom Healthcheck execution.
  */
 open class RealDynamoDbModule @JvmOverloads constructor(
   private val clientConfig: ClientConfiguration = ClientConfiguration(),
   vararg requiredTableTypes: KClass<*>,
+  private val customTableHealthChecks: Map<KClass<*>, HealthCheck> = mapOf(),
```
frojasg, line 33: The problem with this approach is you would need to be able to instanciate the health check when installing the module. I wonder if using a `mapBinder` so the HealthCheck can be injected things constructed by guice.

sangho-block:
Closing per https://github.com/cashapp/misk/pull/2948

PR #2946, opened 2023-09-15
Always Use Tailwind Play CDN

adrw:
Long term, a Gradle plugin or something to build production CSS packages per service would be best. For now, we'll use the play CDN so that there is ease of development between environments.

### DIFF
--- a/misk-tailwind/src/main/kotlin/misk/tailwind/TailwindHtmlLayout.kt
+++ b/misk-tailwind/src/main/kotlin/misk/tailwind/TailwindHtmlLayout.kt
@@ -26,12 +26,14 @@ fun TagConsumer<*>.TailwindHtmlLayout(appRoot: String, title: String, playCdn: B
         type = "image/x-icon"
         href = "/static/favicon.ico"
       }
-      if (playCdn) {
+      // TODO add Gradle plugin to comb through service JAR to build minified Tailwind CSS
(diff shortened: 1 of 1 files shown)


misk-tailwind/src/main/kotlin/misk/tailwind/TailwindHtmlLayout.kt
```diff
@@ -26,12 +26,14 @@ fun TagConsumer<*>.TailwindHtmlLayout(appRoot: String, title: String, playCdn: B
         type = "image/x-icon"
         href = "/static/favicon.ico"
       }
-      if (playCdn) {
+      // TODO add Gradle plugin to comb through service JAR to build minified Tailwind CSS
+      // Until then, use play CDN so all CSS is present for UI from Misk or internal libaries/services
+//      if (playCdn) {
```
ean5533, line 31: Nit: it's kind of deceptive that the `playCdn` parameter now does nothing, but I recognize that it'd be a pain to delete it and then re-add it just for an intermediary state, so 👍

ean5533 (approved):

misk-tailwind/src/main/kotlin/misk/tailwind/TailwindHtmlLayout.kt
adrw: Agreed, thanks for understanding. Aiming for it to only be a temporary change.

PR #2947, opened 2023-09-15
Adding optional callback parameter to misk retry function

MarcelloLins:
### Context
Misk's `retry` wrapper is used by [tens of services](https://github.com/search?q=org%3Asquareup+import+misk.backoff.retry&type=code) at Cash. This helper is mostly used to wrap RPC/HTTP calls as follows:

```
val response = retry(upTo = 3, withBackoff = FlatBackoff()) {
          myDependencyClient.aRPCCall(params)
    }
```
If the code wrapped by `retry` throws any exception that isn't of type `DontRetryException`, the `retry` block will swallow the exception, apply the backoff policy, and retry the same block up to the maximum defined number of retries.

### Problem
This exception swallowing makes it harder for clients to log or emit metrics of retries, which decreases overall service observability across cash. One possible solution to this problem is to wrap your own call in a `try / catch`, log / emit metrics, and rethrow the exception so that the `retry` wrapper can catch it again, swallow it, and reapply the backoff policy.

```
val response = retry(upTo = 3, withBackoff = FlatBackoff()) {
          try {
          myDependencyClient.aRPCCall(params)
         } catch (e: Exception) {
            logger.log("something went wrong: ${e.message}")
            metrics.emit("a metric")
            
            throw exception
         }
    }
```

Despite being a feasible solution, I couldn't find evidence that any service is doing this currently.

### Solution
This PR introduces an optional parameter of type `(retryCount: Int, exception: Exception) -> Unit)` (meaning a function that returns no value, and receives two parameters: retryCount and exception). This callback function is called (if supplied) when a retry happens within the `retry` wrapper.

The block above would be written like this:
```
val observabilityCallback: (retries: Int, e: Exception) -> Unit = { retries, e ->
        logger.log("something went wrong: ${e.message}")
       metrics.emit("a metric")
    }
val response = retry(upTo = 3, withBackoff = FlatBackoff(), onRetry = observabilityCallback) {
          myDependencyClient.aRPCCall(params)
    }
```

This should hopefully nudge other engineers to add better observability into their services when using this wrapper, while keeping the library backwards compatible and flexible.

### DIFF
--- a/misk-core/src/main/kotlin/misk/backoff/Retries.kt
+++ b/misk-core/src/main/kotlin/misk/backoff/Retries.kt
@@ -2,8 +2,9 @@ package misk.backoff
 
 /**
  * Retries the provided function up to a certain number of times, applying the given backoff
- * between each retry. The retry function is provided with current retry count, in case this is
- * relevant
+ * between each retry. If provided, the onRetry callback is called when a retry happens, allowing
+ * clients to perform a task (log, emit metrics) every time a retry occurs.
+ * The retry function is provided with current retry count, in case this is relevant.
  */
 @JvmOverloads
 fun <A> retry(
--- a/misk-core/src/main/kotlin/misk/backoff/Retries.kt
+++ b/misk-core/src/main/kotlin/misk/backoff/Retries.kt
@@ -5,7 +5,13 @@ package misk.backoff
  * between each retry. The retry function is provided with current retry count, in case this is
  * relevant
  */
-fun <A> retry(upTo: Int, withBackoff: Backoff, f: (retryCount: Int) -> A): A {
+@JvmOverloads
+fun <A> retry(
+  upTo: Int,
+  withBackoff: Backoff,
+  onRetry: ((retryCount: Int, exception: Exception) -> Unit)? = null,
+  f: (retryCount: Int) -> A,
+  ): A {
   require(upTo > 0) { "must support at least one call" }
 
   withBackoff.reset()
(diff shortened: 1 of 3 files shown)


misk-core/src/main/kotlin/misk/backoff/Retries.kt
```diff
@@ -2,10 +2,17 @@ package misk.backoff
...
+ * clients to perform a task (log, emit metrics) every time a retry occurs.
+ * The retry function is provided with current retry count, in case this is relevant.
  */
-fun <A> retry(upTo: Int, withBackoff: Backoff, f: (retryCount: Int) -> A): A {
+@JvmOverloads
+fun <A> retry(
+  upTo: Int,
+  withBackoff: Backoff,
+  onRetry: ((retryCount: Int, exception: Exception) -> Unit)? = null,
+  f: (retryCount: Int) -> A,
```
adrw, line 14:
```suggestion
  block: (retryCount: Int) -> A,
```

Let's rename this to block to be more idiomatic, f isn't used as a named parameter anywhere so this won't break anything.

https://github.com/search?q=org%3Asquareup+%22%2C+f+%3D%22+language%3AKotlin+&type=code

adrw (approved):

misk-core/src/main/kotlin/misk/backoff/Retries.kt
MarcelloLins: I agree !

PR #2948, opened 2023-09-15
Remove dynamodb healthcheck as it serves little purpose

tso:


misk-aws-dynamodb/src/main/kotlin/misk/dynamodb/RealDynamoDbService.kt
```diff
@@ -0,0 +1,21 @@
+package misk.dynamodb
+
+import com.amazonaws.services.dynamodbv2.AmazonDynamoDB
+import com.google.common.util.concurrent.AbstractIdleService
+import jakarta.inject.Inject
+import jakarta.inject.Singleton
+
+@Singleton
+class RealDynamoDbService @Inject constructor(
```
tso, line 9: I wonder if all the test services should do this as well?

jemiahw (approved):
Approving the idea / intent here. I agree this healthcheck is not effective, and the calls to DescribeTable cause some knock on problems elsewhere. In partial outages where some percentage of calls to DynamoDB are failing, you're better off leaving the service running partially degraded rather than taking it offline completely. And for services that do more than just talk to DynamoDB (maybe they're using DAX or have calls that don't include DynamoDB), you don't necessarily want a DynamoDB outage to take the pods offline entirely when they could still potentially serve the other calls.

Better is to have alerting on service error rate and ideally on DynamoDB client error rate, rather than a healthcheck that can cascade problems.

Thanks for the PR.

adrw (approved):
Nice cleanup

PR #2949, opened 2023-09-15
Create legacyHistogram method in both metrics v1 and v2 to enable migration to v2 and v1 removal

escardin:
Setting up for renaming all existing uses of v1 histogram to legacy histogram, and then changing all the metrics v1 imports to a v2 import (as there won't be any conflicts)

Also marked all of v1 as deprecated


misk-prometheus/src/main/kotlin/misk/metrics/backends/prometheus/v2/PrometheusMetrics.kt
```diff
@@ -75,6 +77,26 @@ internal class PrometheusMetrics @Inject internal constructor(
...
+    maxAgeSeconds: Long?
+  ): misk.metrics.Histogram = PrometheusHistogram(
+    summary(
+      name,
+      help,
+      labelNames,
+      quantiles,
+      maxAgeSeconds
+    )
+  )
```
MatthewDolan, line 98: If this function is simply invoking `summary(...)`, what's the advantage of creating this method as opposed to using openrewrite to rewrite functions to use `summary(...)` directly?

escardin: It's not a direct invocation. The summary is wrapped in a PrometheusHistogram(), and that needs to be unwound. My overall goal with this PR is to turn migrating from V1 to V2 renaming the usages of histogram to legacyHistogram, and then changing all usages of the V1 Metrics class to the V2 Metrics class (i.e. just an import change).

The overall idea is to simplify the migration of V1 & V2 to micrometer by only having to actually migrate V2 with the more complex recipes.

There is a task to write a recipe that migrates this to a summary (or histogram), but it's one of the later tasks.

MatthewDolan: Ah, I completely missed that extra part. Thanks for clarifying.

MatthewDolan (approved):

PR #2952, opened 2023-09-18
Fix StringConverter to work for types with generics

luanpotter:
Currently the only way custom Path Parameter types are supported is via a static `valueOf` method being provided by the class. This works out-of-the-box for enums and works fine for any type without generics. However, the reflection used to find the `valueOf` method:

```
val javaClass = type.javaType as Class<Any>?
```

Makes an incorrect assumption to cast the java type as `Class<Any>`. If the type has generic parameters, it will be a `ParametrizedTypeImpl`, not a `Class`. Since we are only getting the class here to find a static method, and do not care about generic parameter types either way, we can just add a call to `rawType` to get the actual `Class` object desired.

This PR changes that (should be completely backwards compatible) and adds a test case for the type with generics.

A more advanced option would be to add support for custom converters to be registered via a centralized Registry of sorts. That is a significantly bigger shift from the current implementation though, but if that is preferred I can help with a PR for that instead.

[The implementation of `rawType` provided by moshi](https://github.com/square/moshi/blob/24cf12cce60c531c14f9edeb89f86f04e7240a10/moshi/src/main/java/com/squareup/moshi/Types.kt#L160) gives more clarity around why `Type` and `Class` are different and how to convert one to the other.


misk/src/main/kotlin/misk/web/extractors/StringConverter.kt
```diff
@@ -49,14 +50,11 @@ private fun numberConversionWrapper(wrappedFunc: StringConverter): StringConvert
...
   try {
-    @Suppress("UNCHECKED_CAST")
-    val javaClass = type.javaType as Class<Any>?
-    val valueOfMethod = javaClass?.getMethod("valueOf", String::class.java)
-    if (valueOfMethod != null) {
-      return { param -> valueOfMethod(null, param) }
-    }
-  } catch (e: ClassCastException) {
-  } catch (e: NoSuchMethodException) {
+    val javaClass = type.javaType.rawType
```
adrw, line 53: Thanks for the explanation in PR description on this change

adrw (approved):

frojasg (approved):
LGTM. Thanks for adding this change

PR #2954, opened 2023-09-19
[CCPOKR-8875] Extract misk-api module

katukota:
This PR is the first part for creating a compatibility layer for all kotlin frameworks (Armeria and Misk).
Epic: https://ccp-cashapp.atlassian.net/browse/CCPOKR-7619


misk-action-scopes/src/main/kotlin/misk/scope/ActionScope.kt
```diff
@@ -57,12 +57,14 @@ class ActionScope @Inject internal constructor(
   }
 
   /** Starts the scope on a thread with the provided seed data */
-  fun enter(seedData: Map<Key<*>, Any?>): ActionScope {
+  override fun enter(seedData: Map<out Any, Any?>): Scope {
+    val seedMap: Map<Key<*>, Any?> = seedData as Map<Key<*>, Any?>
```
damar-block, line 61: We might need to revisit this when we write the Armeria adapter that extends Scope, since not sure the assumption that Key class will be used.

misk-action-scopes/src/main/kotlin/misk/scope/ActionScope.kt
```diff
@@ -83,7 +85,15 @@ class ActionScope @Inject internal constructor(
...
   /** Returns true if currently in the scope */
-  fun inScope(): Boolean = threadLocalScope.get() != null
+  override fun inScope(): Boolean = threadLocalScope.get() != null
+
+  override fun <T> get(type: Class<T>): T? {
+    if (!inScope()) {
+      return null
+    }
+
+    return get(Key.get(type))
```
damar-block, line 95: As it is today, this class has dependency on Guice since it has Guice providers as property which are used when a Key is given.  I wonder if we want to remove this dependency and use a more default Supplier interface.

gradle.properties
```diff
@@ -5,3 +5,4 @@ org.gradle.vfs.watch=true
 org.gradle.configuration-cache=true
 org.gradle.configureondemand=true
 org.gradle.parallel=true
+VERSION_NAME=2023.09.18.135915-444f476-SNAPSHOT
```
damar-block, line 8: Probably need to remove - this is for testing

misk-action-scopes/src/test/kotlin/misk/scope/ActionScopedTest.kt
```diff
@@ -15,6 +15,7 @@ import org.junit.jupiter.api.BeforeEach
 import org.junit.jupiter.api.Test
 import java.util.Optional
 import jakarta.inject.Inject
+import javax.swing.Action
```
damar-block, line 18: Remove?

misk-action-scopes/src/test/kotlin/misk/scope/ActionScopedTest.kt
```diff
@@ -169,7 +170,7 @@ internal class ActionScopedTest {
 
     )
     scope.enter(seedData).use { actionScope ->
-      runBlocking(actionScope.asContextElement()) {
+      runBlocking((actionScope as ActionScope).asContextElement()) {
```
damar-block, line 173: I guess we can only determine if `asContextElement` needs to be part of the interface (and then this cast can be removed) when we implement the armeria side. I think it's ok to go with that version for now.

misk-api/src/main/kotlin/misk/scope/Scope.kt
```diff
@@ -0,0 +1,9 @@
+package misk.scope
+
+interface Scope : AutoCloseable {
```
damar-block, line 3: Need to add some comments as this will be interface which should be properly documented

gradle.properties
katukota: Done

misk-action-scopes/src/test/kotlin/misk/scope/ActionScopedTest.kt
katukota: Done

misk-action-scopes/api/misk-action-scopes.api
```diff
@@ -14,13 +14,14 @@ public abstract class misk/concurrent/WrappingListeningExecutorService : com/goo
 	protected abstract fun wrap (Ljava/util/concurrent/Callable;)Ljava/util/concurrent/Callable;
 }
 
-public final class misk/scope/ActionScope : java/lang/AutoCloseable {
+public final class misk/scope/ActionScope : misk/scope/Scope {
 	public static final field Companion Lmisk/scope/ActionScope$Companion;
 	public final fun asContextElement ()Lkotlin/coroutines/CoroutineContext$Element;
 	public fun close ()V
-	public final fun enter (Ljava/util/Map;)Lmisk/scope/ActionScope;
+	public fun enter (Ljava/util/Map;)Lmisk/scope/Scope;
```
kyeotic, line 21: Why is this being renamed? `ActionScope` is in use in a ton of places; not just code, but documentation. Its also a _good name_, it tells you that the scope of the thing is the `Action`. The new name does not tell you _what_ the scope is, and implies that it could be _any scope_.

misk-api/src/main/kotlin/misk/web/HttpCall.kt
```diff
@@ -35,6 +35,7 @@ interface HttpCall {
   var requestHeaders: Headers
 
   /** Cookies derived from request's "Cookie" header, if any */
+  //TODO: Remove cookies
```
adrw, line 38: Could more reasoning for why be added here?

misk-api/src/main/kotlin/misk/client/ClientApplicationInterceptorFactory.kt
```diff
@@ -1,5 +1,6 @@
 package misk.client
 
+import misk.client.ClientAction
```
adrw, line 3: Is this import necessary?

misk-api/src/main/kotlin/misk/Action.kt
```diff
@@ -38,10 +37,6 @@ data class Action(
       }
     }
 
-  internal inline fun <reified T : Annotation> parameterAnnotatedOrNull(): KParameter? {
```
adrw, line 41: Is this unused everywhere including internally?

misk-action-scopes/src/main/kotlin/misk/scope/ActionScopedProviderModule.kt
```diff
@@ -13,6 +13,7 @@ import misk.inject.toKey
 import misk.inject.typeLiteral
 import java.lang.reflect.Type
 import jakarta.inject.Inject
+import misk.ApplicationInterceptor
```
adrw, line 16: Is this import necessary?

misk-action-scopes/src/main/kotlin/misk/scope/ActionScope.kt
```diff
@@ -57,12 +57,14 @@ class ActionScope @Inject internal constructor(
   }
 
   /** Starts the scope on a thread with the provided seed data */
-  fun enter(seedData: Map<Key<*>, Any?>): ActionScope {
+  override fun enter(seedData: Map<out Any, Any?>): Scope {
```
adrw, line 60: I do worry about the compatibility layer approach if the end result is erasing lots of types with `Any`. That seems to result in net worse APIs and usage for the slim benefit of not including an extra library in Armeria projects. 

How would a future Misk user know that this `enter(seedData)` parameter isn't actually `Map<Any, Any>` but immediately is cast to `Key` until their code surprisingly blows up in tests or runtime on cast failure?

adrw (commented):
1. I'd like more clarity on the compat layer refactor plans, is switching most APIs to `Any` a good long term strategy for maintainable code? Kotlin is used as a typed language for the benefits of the type system and it seems like this approach will gradually strip those benefits from Misk users.
2. I think this is a breaking change and probably needs to be staged to first land the new changes while leaving the old in place, roll out internally with a migration job, and then delete the old fixtures.

misk-action-scopes/api/misk-action-scopes.api
adrw: `Scope` is the new interface which will have a corresponding Armeria implementation to the `ActionScope` Misk implementation. I agree though, on first glance a confusing change.

adrw:
Could a PR description be added (a good general practice and even more so for OSS repos) so that sweeping changes like this can be better understood when users/maintainers look back on commit history?

misk-api/src/main/kotlin/misk/scope/Scope.kt
```diff
@@ -1,9 +1,11 @@
 package misk.scope
 
+import kotlin.reflect.KType
+
 interface Scope : AutoCloseable {
-  fun <T> get(type: Class<T>): T?
+  fun <T> get(key: KType): T
```
adrw, line 6: Have tests been done that this will support annotations or other Guice like qualifiers on types?

misk-action-scopes/src/main/kotlin/misk/scope/ActionScope.kt
katukota: Agreed. For now, we are keeping the Guice module in misk-api.

misk-action-scopes/src/main/kotlin/misk/scope/ActionScope.kt
katukota: For now, we are keeping the Guice module in misk-api.

misk-api/src/main/kotlin/misk/web/actions/WebSocketListener.kt
```diff
@@ -1,5 +1,6 @@
 package misk.web.actions
 
+import misk.web.actions.WebSocket
```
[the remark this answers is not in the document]
katukota: Removed.

misk-api/src/main/kotlin/misk/client/ClientApplicationInterceptorFactory.kt
katukota: Removed.

misk-api/src/main/kotlin/misk/web/HttpCall.kt
katukota: @rainecp changed this to use okhttp3 cookie instead.

misk-api/src/main/kotlin/misk/Action.kt
katukota: There were 2 usages of this function. I replaced them inline.

misk-action-scopes/src/main/kotlin/misk/scope/ActionScopedProviderModule.kt
katukota: Removed.

misk/src/main/kotlin/misk/client/ClientInvocationHandler.kt
```diff
@@ -29,6 +29,7 @@ import kotlin.reflect.KFunction
 import kotlin.reflect.full.findAnnotation
 import kotlin.reflect.full.functions
 import kotlin.reflect.full.memberFunctions
+import misk.client.ClientAction
```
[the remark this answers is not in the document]
katukota: Removed.

misk-api/src/main/kotlin/misk/scope/Scope.kt
katukota: Yes, I tested with different Kotlin types to see if we can replace Guice but none of them actually worked as expected. For now, reverted to using Guice in misk-api.

misk-action-scopes/src/test/kotlin/misk/scope/ActionScopedTest.kt
katukota: Yes, we have decided to not add "asContextElement" in the generic interface for now. I will add it if required by the downstream dependencies.

PR #2958, opened 2023-09-21
Move Prometheus Metrics into Misk Metrics as it's the only impelmentation

escardin:
Rework Misk's Prometheus Metrics implementation to delegate to Misk Metrics, and move the implementation into there.

This is a prereq for removing this project, and I need to consolidate all the classes and usages into as few as possible to make further migration work straight forward.

I'll do another PR moving the implementation classes for the prom backend into misk metrics once this is done. I'm trying to keep breaking changes to the absolute minimum.

I did a bunch of searches for usages of the prometheus types here, and realistically only the modules are ever referenced directly, and in most cases I would say it's a mistake.


misk-metrics/api/misk-metrics.api
```diff
@@ -1,18 +1,14 @@
 public final class misk/metrics/FakeMetrics : misk/metrics/Metrics {
-	public fun counter (Ljava/lang/String;Ljava/lang/String;Ljava/util/List;)Lio/prometheus/client/Counter;
```
tso, line 2: these methods are unused?

tso (approved):
looks good, unsure if you'll need to wait a few months before removing `misk-prometheus` or not -- i'd lean towards probably

escardin:
> looks good, unsure if you'll need to wait a few months before removing `misk-prometheus` or not -- i'd lean towards probably

Most likely it'll get dropped once I have a complete migration off of misk metrics.

misk-metrics/api/misk-metrics.api
escardin: Not unused, but instead not overridden.

Used to be Metrics was an interface, now it's the implementation. Since there's no difference in the fake implementation, we don't need the overrides anymore.

PR #2959, opened 2023-09-21
Rework testFixtures FakeMetrics to use extention methods

escardin:
By doing this, we no longer actually need FakeMetrics, nor FakeMetricsModule really. Instead we can use the real Metrics type. This doesn't mean that the testFixture FakeMetrics goes away entirely, instead it will hold the extension methods that are used to make testing easier.

As these are based around the prometheus CollectorRegistry, we will need to eventually port them to a micrometer backed registry as part of the migration. This will result in either new extension methods, or rewriting the uses of these ones into micrometer equivalents, as it tends to be more robust for testing purposes.

### DIFF
--- a/misk-metrics/src/test/kotlin/misk/metrics/v2/FakeMetricsTest.kt
+++ b/misk-metrics/src/test/kotlin/misk/metrics/v2/FakeMetricsTest.kt
@@ -1,78 +1,89 @@
 package misk.metrics.v2
 
 import io.prometheus.client.Collector.MetricFamilySamples.Sample
+import io.prometheus.client.CollectorRegistry
 import jakarta.inject.Inject
+import misk.metrics.FakeMetricsModule
+import misk.metrics.get
+import misk.metrics.getAllSamples
+import misk.metrics.summaryCount
+import misk.metrics.summaryMean
+import misk.metrics.summaryP50
+import misk.metrics.summaryP99
+import misk.metrics.summarySum
 import misk.testing.MiskTest
 import misk.testing.MiskTestModule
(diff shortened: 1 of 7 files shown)


misk-metrics/src/test/kotlin/misk/metrics/v2/FakeMetricsTest.kt
```diff
@@ -1,140 +1,151 @@
 package misk.metrics.v2
 
 import io.prometheus.client.Collector.MetricFamilySamples.Sample
+import io.prometheus.client.CollectorRegistry
 import jakarta.inject.Inject
+import misk.metrics.FakeMetricsModule
+import misk.metrics.get
+import misk.metrics.getAllSamples
```
adrw, line 8: I'm trying to understand this approach. So if I'm writing a test, I need to remember each time to use and import this extension method instead of one with the same name on the actual `CollectorRegistry` class?

escardin: The idea is that this class goes away, and you inject CollectorRegistry and use it. The extension methods pop up normally like other methods on the class in Intellij's auto complete.

escardin: The collector registry doesn't have any of these methods, and I don't see a ton of value in adding our own extension just for some convenience.

adrw:
Checked it out locally and the autocomplete in IntelliJ works well. Let's land this and see how it rolls out.

PR #2961, opened 2023-09-25
Services can stop gracefully even when a dependency has failed

zpingcai:
When a service fails to start, misk starts orderly shutdown by calling `serviceManager.stopAsync()` in a jvm shutdown hook. However, all services that the failed service depend on cannot have a graceful shutdown because they are waiting eternally for it to have the state `State.TERMINATED`. This PR will fix the issue.

### DIFF
--- a/misk-service/src/main/kotlin/misk/CoordinatedService.kt
+++ b/misk-service/src/main/kotlin/misk/CoordinatedService.kt
@@ -64,8 +64,8 @@ internal class CoordinatedService(
     }
   }
 
-  private fun isTerminated(): Boolean {
-    return state() == State.TERMINATED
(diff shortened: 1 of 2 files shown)


zpingcai:
CC: @Hexcles

zpingcai:
Tests passed locally 🤔

keeferrourke (approved):
Thanks for fixing this!

zpingcai:
@keeferrourke Thanks for reviewing it! Could you rerun the tests and make a release on maven?

keeferrourke:
> It seems `CronTest` is failing which is irrelevant to this PR, and it passed locally 🤔

Seems like this might be flaky. There are a number of retries built into that test to poll for pending tasks. Sad.

PR #2962, opened 2023-09-26
More corner cases for graceful shutdown on service failure

zpingcai:
I noticed this PR https://github.com/cashapp/misk/pull/2961 missed a few corner cases:

1. All services depending on the fail-to-start service are stuck on shutdown. This is because the outer services (i.e., `CoordinatedService`) are in `STARTING` state, so Guava will call `doCancelStart()` instead of `doStop()` ([proof](https://github.com/google/guava/blob/5a0af313bb89e72419aec20aa232e73d95356cb3/guava/src/com/google/common/util/concurrent/AbstractService.java#L280C19-L280C19)). 
2. If a service fails to stop, all services it depends on may stuck on shutdown. This is because the service doesn't call `directDependsOn.forEach { it.stopIfReady() }` when it fails.

### DIFF
--- a/misk-service/src/main/kotlin/misk/CoordinatedService.kt
+++ b/misk-service/src/main/kotlin/misk/CoordinatedService.kt
@@ -32,6 +32,7 @@ internal class CoordinatedService(
 
             override fun failed(from: State, failure: Throwable) {
               outerService.notifyFailed(failure)
+              directDependsOn.forEach { it.stopIfReady() }
             }
           },
           MoreExecutors.directExecutor()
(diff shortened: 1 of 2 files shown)


misk-service/src/main/kotlin/misk/CoordinatedService.kt
```diff
@@ -32,6 +32,7 @@ internal class CoordinatedService(
 
             override fun failed(from: State, failure: Throwable) {
               outerService.notifyFailed(failure)
+              directDependsOn.forEach { it.stopIfReady() }
```
zpingcai, line 35: Starting any downstream dependencies after a start failure definitely makes no sense, but I think it makes sense to stop as many upstream dependencies as possible after a stop failure. There are little benefits to hang on a stop failure and allow the upstream to run.

keeferrourke (approved):

PR #2964, opened 2023-09-27
Move Prometheus Metrics into Misk Metrics as it's the only impelmentation #2

escardin:
This does not remove the prometheus metrics project, since we don't want to break compatibility before fully migrating off of it.

This time with no breaking changes and instead kinda gross factory methods.


misk-metrics/src/main/kotlin/misk/metrics/Histogram.kt
```diff
@@ -14,18 +15,32 @@ import java.util.concurrent.TimeUnit
  * A sample implementation can be found in PrometheusHistogram
  */
 interface Histogram {
+
+  fun getHistogram(): Summary
```
tso, line 19: there's a few internal implementations of Histogram who this will be backwards incompatible for

misk-metrics/src/main/kotlin/misk/metrics/Metrics.kt
```diff
@@ -15,6 +14,8 @@ import io.prometheus.client.Gauge
   level = DeprecationLevel.WARNING
 )
 interface Metrics {
+  fun getMetrics(): Metrics
```
tso, line 17: same here

misk-metrics/src/main/kotlin/misk/metrics/MetricsModule.kt
```diff
@@ -0,0 +1,24 @@
...
+import misk.inject.asSingleton
+
+class MetricsModule : KAbstractModule() {
+  override fun configure() {
+    bind<CollectorRegistry>().toProvider(CollectorRegistryProvider::class.java).asSingleton()
+
+  }
+
+  /**
+   * In order to make it possible to install this module multiple times, we make this binding not
```
tso, line 16: would `InstallOnceModule` (forgetting exact name) solve this as well? or no

tso (approved):
seems legit, what's the plan for the interface having a non-default implementation?

PR #2969, opened 2023-10-05
Swallow the invalid path exception and also call destroy

tso:


misk/src/main/kotlin/misk/web/jetty/JettyService.kt
```diff
@@ -326,7 +329,15 @@ class JettyService @Inject internal constructor(
...
-      server.stop()
+      try {
+        server.stop()
+      } catch (_: InvalidPathException) {
+        // Currently we get a nul character exception since an abstract socket address is
+        // distinguished from a regular unix socket by the fact that the first byte of
+        // the address is a null byte ('\0'). The address has no connection with filesystem
+        // path names.
+      }
+      server.destroy()
```
tso, line 340: the previous jetty stop hook called this, so we should too

PR #2970, opened 2023-10-05
Stop jetty last when shutdown_sleep_ms is set to greater than 0

tso:
This commit, combined with a previous removal of a jetty shutdown hook, means that shutdown_sleep_ms will sleep for N ms to allow requests to drain.

The shutdown process with shutdown_sleep_ms=30000 will be:
1. Services that are upstream of jetty shutdown
2. Jetty is told to shutdown -- We sleep for 30 seconds
3. Jetty dependencies shutdown
4. Jetty itself shutsdown (for real this time)

The 30 second sleep gives us time to drain the requests that are inflight, and keeping jetty alive means that our istio sidecar remains open allowing outbound requests to suceed until the end.

### DIFF
--- a/misk/src/main/kotlin/misk/web/jetty/JettyService.kt
+++ b/misk/src/main/kotlin/misk/web/jetty/JettyService.kt
@@ -37,6 +37,7 @@ import org.eclipse.jetty.util.ssl.SslContextFactory
 import org.eclipse.jetty.util.thread.ThreadPool
 import org.eclipse.jetty.websocket.server.config.JettyWebSocketServletContainerInitializer
 import wisp.logging.getLogger
+import java.lang.Thread.sleep
 import java.net.InetAddress
 import java.nio.file.InvalidPathException
 import java.util.EnumSet
@@ -44,8 +45,6 @@ import java.util.concurrent.SynchronousQueue
 import java.util.concurrent.ThreadPoolExecutor
 import java.util.concurrent.TimeUnit
 import javax.servlet.DispatcherType
-
-private val logger = getLogger<JettyService>()
 
 @Singleton
 class JettyService @Inject internal constructor(
@@ -324,11 +323,11 @@ class JettyService @Inject internal constructor(
     }
   }
 
-  override fun shutDown() {
-    val stopwatch = Stopwatch.createStarted()
-    logger.info("Stopping Jetty")
-
+  internal fun stop() {
     if (server.isRunning) {
+      val stopwatch = Stopwatch.createStarted()
+      logger.info("Stopping Jetty")
+
       try {
         server.stop()
       } catch (_: InvalidPathException) {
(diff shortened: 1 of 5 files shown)


misk/src/main/kotlin/misk/web/jetty/JettyService.kt
```diff
@@ -324,11 +323,11 @@ class JettyService @Inject internal constructor(
     }
   }
 
-  override fun shutDown() {
-    val stopwatch = Stopwatch.createStarted()
-    logger.info("Stopping Jetty")
-
+  internal fun stop() {
     if (server.isRunning) {
+      val stopwatch = Stopwatch.createStarted()
```
tso, line 328: since this is called twice (harmlessly) when shutdown_sleep_ms is 0 we only want to log if we're still running

misk/src/main/kotlin/misk/web/jetty/JettyService.kt
```diff
@@ -337,15 +336,34 @@ class JettyService @Inject internal constructor(
         // the address is a null byte ('\0'). The address has no connection with filesystem
         // path names.
       }
-      server.destroy()
+
+      logger.info { "Stopped Jetty in $stopwatch" }
     }
 
     if (healthExecutor != null) {
```
tso, line 343: this should be fine to be called twice, but we don't really care about errors that late into shutdown anyway -- i.e. i can always add the check later

adrw (approved):

misk/src/main/kotlin/misk/web/jetty/JettyService.kt
```diff
@@ -337,15 +336,34 @@ class JettyService @Inject internal constructor(
         // the address is a null byte ('\0'). The address has no connection with filesystem
         // path names.
       }
-      server.destroy()
```
tso, line 340: turns out we weren't supposed to call destroy hehe oop, it just causes a meaningless error log for people

misk/src/main/kotlin/misk/web/jetty/JettyService.kt
```diff
@@ -337,15 +336,34 @@ class JettyService @Inject internal constructor(
...
+    // We need jetty to shut down at the very end to keep outbound connections alive
+    // (due to sidecars). As such, we wait for `shutdown_sleep_ms` so that our
+    // in flight requests drain, but we don't shut down dependencies until after.
+    //
+    // The true jetty shutdown occurs in stop() above, called from MiskApplication.
+    //
+    // Ideally we could call jetty.awaitInflightRequests() but that's not available
+    // for us.
+    if (webConfig.shutdown_sleep_ms > 0) {
+      sleep(webConfig.shutdown_sleep_ms.toLong())
```
tgregory-block, line 359: Could we run into issues where we try to shut down "work producing" services and Jetty is in the middle of that list? e.g. we get a shutdown order like 

1. Cron
2. Jetty
3. SQS

wouldn't SQS's shutdown be delayed by `shutdown_sleep_ms` and happen after Jetty?

tso: yea, sqs shutdown could happen after the sleeping which is fine in my opinion

tso: i think the proper fix to that, if we think it's a problem, is to wait for kubernetes 1.28 or to shutdown many things in parallel when possible

tgregory-block (approved):

PR #2967, opened 2023-10-04
[DONT MERGE] Trying revapi for detecting breaking changes

mmollaverdi:
**Not to be merged - Only for a demo at this stage**
 
Running `gradle :misk-action-scopes:check` on this change (revapi adds a task to check), I get the following error, which is a good description of the breaking change.

```
Execution failed for task ':misk-action-scopes:revapi'.
> There were Java public API/ABI breaks reported by revapi:

  java.method.numberOfParametersChanged: The number of parameters of the method have changed.

  old: method <T, A extends java.lang.annotation.Annotation> void misk.scope.ActionScopedProviderModule::bindProvider(kotlin.reflect.KClass<T>, kotlin.reflect.KClass<? extends misk.scope.ActionScopedProvider<? extends T>>, java.lang.Class<A>)
  new: method <T, A extends java.lang.annotation.Annotation> void misk.scope.ActionScopedProviderModule::bindProvider(kotlin.reflect.KClass<T>, kotlin.reflect.KClass<? extends misk.scope.ActionScopedProvider<? extends T>>, java.lang.Class<A>, java.lang.String)

  SOURCE: BREAKING, BINARY: BREAKING

  From old archive: misk-action-scopes-2023.10.02.170330-4694630.jar
  From new archive: misk-action-scopes-0.0-SNAPSHOT.jar

  If this is an acceptable break that will not harm your users, you can ignore it in future runs like so for:
...
```

Good things to note
- Unlike `binary-compatibility-validator`, this doesn't complain about new methods being added (non-breaking changes) 🎉 
- It detected that the change to the other `bindProvider` method, where a new argument is added with a default value and an accompanying `@JvmOverloads` annotation, is not a breaking change.

PR #2974, opened 2023-10-11
Tag the misk repo with the version on publish

staktrace:


staktrace:
For context, this is needed for https://github.com/cashapp/misk/pull/2967 but is good practice in general as it allows folks to more easily find the code for a particular version.

samw-block (approved):
My thumb is more of a rubberstamp based on your testing here:

https://github.com/squareup/kartikaya

.github/workflows/Build.yml
```diff
@@ -97,3 +100,9 @@ jobs:
       - name: Publish the artifacts
         run: |
           gradle clean publish --stacktrace
+
+      - name: Tag Misk repo
+        uses: mathieudutour/github-tag-action@v6.1
+        with:
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+          custom_tag: ${{ steps.assign_version.outputs.version }}
```
adrw, line 108: Does this resolve to `GITHUB_OUTPUT` env variable?

adrw (approved):

.github/workflows/Build.yml
staktrace: The `GITHUB_OUTPUT` env var is a magic one that exports things out of the step so it can be used by other steps. Docs at https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-an-output-parameter

PR #2975, opened 2023-10-11
Install OffsetDateTimeAdapter by default

Patrax:
Adding another java.time class adapter by default to MoshiModule.


adrw (approved):
Good to land without the Hermit gradle/node version bumps.

PR #2979, opened 2023-10-13
Fix VirtualThread executor

jvmakine:
Turns out `{ VirtualThreads.getDefaultVirtualThreadsExecutor() }` was creating a `Runnable` that broke the jetty executor creation


tso (approved):
oops! sorry thank you

PR #2981, opened 2023-10-13
Duplicate MiskMetricsModules should be identical

escardin:
This was supposed to be the same as the prometheus one, such that I could safely delete the prometheus project at a later date.

### DIFF
--- a/misk-metrics/src/main/kotlin/misk/metrics/MetricsModule.kt
+++ b/misk-metrics/src/main/kotlin/misk/metrics/MetricsModule.kt
@@ -9,7 +9,9 @@ import misk.inject.asSingleton
 class MetricsModule : KAbstractModule() {
   override fun configure() {
     bind<CollectorRegistry>().toProvider(CollectorRegistryProvider::class.java).asSingleton()
-
+    bind<HistogramRegistry>().toProvider(HistogramRegistryProvider::class.java).asSingleton()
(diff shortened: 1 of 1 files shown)


Patrax (approved):

misk-metrics/src/main/kotlin/misk/metrics/MetricsModule.kt
```diff
@@ -9,7 +9,9 @@ import misk.inject.asSingleton
 class MetricsModule : KAbstractModule() {
   override fun configure() {
     bind<CollectorRegistry>().toProvider(CollectorRegistryProvider::class.java).asSingleton()
-
+    bind<HistogramRegistry>().toProvider(HistogramRegistryProvider::class.java).asSingleton()
```
adrw, line 12: Is there a `.toProvider<T>()` method we could use instead?

adrw (approved):

keeferrourke (approved):

misk-metrics/src/main/kotlin/misk/metrics/MetricsModule.kt
escardin: I just copied what was here previously. It was grosser than I'd have liked. How about I look into it for a follow on pr when i'm back making changes? I think ultimately this module will not be necessary, or very, very simple.

ean5533 (approved):
Confirmed that this makes the two `MetricsModule`s functionally identical

PR #2983, opened 2023-10-16
Add revapi for binary compatibility checks on CI

mmollaverdi:
Refer to this PR for the errors revapi fails with for breaking changes:
https://github.com/cashapp/misk/pull/2967

**Out of Scope: Running this task locally**
 
This won't run on local builds at the moment. I'm not quiet sure what gradle task folks run locally, I usually run `check`. The build is currently set up in a way which [makes](https://github.com/cashapp/misk/blob/a64f303d375336e23cf0de40d9759c0300f26618/build.gradle.kts#L217-L231) the CI test tasks depend on `check`, so making `check` depend on `revapi` means revapi would also run as part of each test step on CI, which is not idea.

I'll start some follow up conversation on this topic on slack.

### DIFF
--- a/.github/workflows/Linting.yml
+++ b/.github/workflows/Linting.yml
@@ -30,6 +30,22 @@ jobs:
       - name: API Check
         run: gradle apiCheck
 
+  breakingChangeDetection:
+    name: ABI Breaking Change Detection
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+
+      - name: Hermit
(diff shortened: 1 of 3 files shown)


.github/workflows/Linting.yml
```diff
@@ -30,6 +30,22 @@ jobs:
...
 
+  breakingChangeDetection:
+    name: ABI Breaking Change Detection
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
```
mmollaverdi, line 41: This is needed for fetching tags which are used for determining previous version to compare APIs against.

PR #2985, opened 2023-10-16
feat: Add controls for requeuing jobs

michiel-at-block:
In certain cases it is desirable to halt execution of a job and to add it back to the queue (e.g. if a service handling long-running jobs is shutting down, or if a long-running job failed and should be retried asap). This change extends Job to allow moving the Job back onto its queue, and provides the option to hint the delay after which the job is eligible for handling again.

In detail, the method `retryLater(Duration)` is added to `Job` to signal that the job should be retried after waiting for the provided `Duration`. For SQS specifically this is achieved by setting the [visibility timeout](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/APIReference/API_ChangeMessageVisibility.html) to the provided duration.


tso (approved):

adrw (approved):
LG, could be worth getting a stamp from someone on pubsub or the platform team which owns/manages the SQS integration.

PR #2986, opened 2023-10-17
Add runBlocking method to actionScope

mhickman:
In misk people frequently invoke `runBlocking` without passing the approriate
actionScope context to propagate the scope. This results in difficult to diagnose
bugs.

Adding this small utility that should just do the right thing, inside or outside
of a current scope.

### DIFF
--- a/misk-action-scopes/api/misk-action-scopes.api
+++ b/misk-action-scopes/api/misk-action-scopes.api
@@ -24,6 +24,8 @@ public final class misk/scope/ActionScope : java/lang/AutoCloseable {
 	public final fun propagate (Ljava/util/concurrent/Callable;)Ljava/util/concurrent/Callable;
 	public final fun propagate (Lkotlin/jvm/functions/Function0;)Lkotlin/jvm/functions/Function0;
 	public final fun propagate (Lkotlin/reflect/KFunction;)Lkotlin/reflect/KFunction;
+	public final fun runBlocking (Lkotlin/coroutines/CoroutineContext;Lkotlin/jvm/functions/Function2;)Ljava/lang/Object;
+	public final fun runBlocking (Lkotlin/jvm/functions/Function2;)Ljava/lang/Object;
 	public final fun snapshotActionScope ()Ljava/util/Map;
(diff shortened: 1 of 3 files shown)


adrw (approved):
Nice addition

misk-action-scopes/src/main/kotlin/misk/scope/coroutine/ActionScopedCoroutineBuilder.kt
```diff
@@ -0,0 +1,33 @@
...
+import kotlinx.coroutines.CoroutineScope
+import misk.scope.ActionScope
+import javax.inject.Inject
+import javax.inject.Singleton
+import kotlin.coroutines.CoroutineContext
+
+/**
+ * Wraps a [kotlinx.coroutines.runBlocking] to propagate the current action scope.
+ */
+@Singleton
```
acostama, line 12: it seems like usage of this is
```kotlin
class MyClass @Inject constructor(val actionScopedCoroutineBuilder: ActionScopedCoroutineBuilder) {
  fun doSomething() {
    actionScopedCoroutineBuilder.runBlocking {
      // do async things
    }
  }
}

```
is it worth adding this layer of indirection? seems like we get the same by creating some extension methods?
```kotlin
class MyClass @Inject constructor(val actionScope: ActionScope) {
  fun doSomething() {
    actionScope.runBlocking {
      // do async things
    }
  }
}

```

mhickman: oh I like that!

frojasg: +1 I would prefer to extend `actionScope`

mhickman: modified to just add on `actionScope`

johnkangg (approved):

PR #2987, opened 2023-10-18
Fix production guice stage initialization errors

tso:
This commit fixes DataSourceService to lazily provide
a DataSource to those who need it. It additionally fixes
a weird bug inside of DashboardPageLayout which was using
an ActionScoped<HttpUrl> as part of a field parameter.

- [ ] I acknowledge that by merging this I will fix any internal SqlDelight users who're doing using `Provider<DataSource>`


misk-admin/api/misk-admin.api
```diff
@@ -894,7 +894,6 @@ public final class misk/web/v2/DashboardPageLayout {
 	public static synthetic fun build$default (Lmisk/web/v2/DashboardPageLayout;Lkotlin/jvm/functions/Function4;ILjava/lang/Object;)Ljava/lang/String;
 	public final fun headBlock (Lkotlin/jvm/functions/Function1;)Lmisk/web/v2/DashboardPageLayout;
 	public final fun newBuilder ()Lmisk/web/v2/DashboardPageLayout;
-	public final fun path (Ljava/lang/String;)Lmisk/web/v2/DashboardPageLayout;
```
tso, line 897: unused based on my searches

PR #2988, opened 2023-10-19
Add wrappers for zadd and zscore methods for redis sorted set

romilbm:
Some basic commands for [Redis SortedSet ](https://redis.io/docs/data-types/sorted-sets/).
[ZADD](https://redis.io/commands/zadd/) - basic command to add members to SortedSet 
[ZSCORE](https://redis.io/commands/zscore/) - get the score of the member in the SortedSet

### DIFF
--- a/misk-redis/src/main/kotlin/misk/redis/FakeRedis.kt
+++ b/misk-redis/src/main/kotlin/misk/redis/FakeRedis.kt
@@ -11,6 +11,7 @@ import java.time.Duration
 import java.time.Instant
 import java.util.concurrent.ConcurrentHashMap
 import jakarta.inject.Inject
+import java.util.SortedMap
 import kotlin.math.max
 import kotlin.math.min
 import kotlin.random.Random
@@ -33,6 +34,9 @@ class FakeRedis : Redis {
   /** A nested hash map for hash operations. */
   private val hKeyValueStore =
     ConcurrentHashMap<String, Value<ConcurrentHashMap<String, ByteString>>>()
+
+  private val sortedSetKeyValueStore =
+    ConcurrentHashMap<String, Value<SortedMap<ByteString, Double>>>()
 
   /** A hash map for list operations. */
   private val lKeyValueStore = ConcurrentHashMap<String, Value<List<ByteString>>>()
@@ -396,6 +400,27 @@ class FakeRedis : Redis {
   }
 
   override fun publish(channel: String, message: String) {
+    throw NotImplementedError("Fake client not implemented for this operation")
+  }
+
+  override fun zadd(
+    key: String,
+    score: Double,
+    member: ByteString,
+    options: Set<String>
+  ): Long {
+    throw NotImplementedError("Fake client not implemented for this operation")
+  }
+
+  override fun zadd(
+    key: String,
+    scoreMembers: Map<ByteString, Double>,
+    options: Set<String>,
+  ): Long {
+    throw NotImplementedError("Fake client not implemented for this operation")
+  }
+
+  override fun zscore(key: String, member: String): Double? {
     throw NotImplementedError("Fake client not implemented for this operation")
   }
 }
--- a/misk-redis/src/main/kotlin/misk/redis/RealRedis.kt
+++ b/misk-redis/src/main/kotlin/misk/redis/RealRedis.kt
@@ -2,6 +2,7 @@ package misk.redis
 
 import okio.ByteString
 import okio.ByteString.Companion.toByteString
+import okio.use
 import redis.clients.jedis.JedisPool
 import redis.clients.jedis.JedisPubSub
 import redis.clients.jedis.Pipeline
@@ -9,6 +10,7 @@ import redis.clients.jedis.Transaction
 import redis.clients.jedis.args.ListDirection
 import redis.clients.jedis.commands.JedisBinaryCommands
 import redis.clients.jedis.params.SetParams
+import redis.clients.jedis.params.ZAddParams
 import java.lang.reflect.InvocationHandler
 import java.lang.reflect.Method
 import java.lang.reflect.Proxy
@@ -302,6 +304,47 @@ class RealRedis(
     jedisPool.resource.use { jedis -> jedis.publish(channel, message) }
   }
 
+  override fun zadd(
+    key: String,
+    score: Double,
+    member: ByteString,
+    options: Set<String>): Long {
+    return jedisPool.resource.use { jedis ->
+      jedis.zadd(key.toByteArray(charset), score, member.toByteArray(), getZAddParams(options))
+    }
+  }
+
+  override fun zadd(
+    key: String,
+    scoreMembers: Map<ByteString, Double>,
+    options: Set<String>): Long {
+    val params = getZAddParams(options)
+    val keyBytes = key.toByteArray(charset)
+    val scoreMembersBytes = scoreMembers.entries.associate { it.key.toByteArray() to it.value }
+
+    return jedisPool.resource.use { jedis -> jedis.zadd(keyBytes, scoreMembersBytes, params)}
+  }
+
+  override fun zscore(key: String, member: String): Double? {
+    return jedisPool.resource.use { it.zscore(key, member) }
+  }
+
+  private fun getZAddParams(options: Set<String>): ZAddParams {
+    val params = ZAddParams()
+
+    options.forEach {
+      when(ZAddOptions.valueOf(it.uppercase())) {
+        ZAddOptions.XX -> params.xx()
+        ZAddOptions.NX -> params.nx()
+        ZAddOptions.LT -> params.lt()
+        ZAddOptions.GT -> params.gt()
+        ZAddOptions.CH -> params.ch()
+      }
+    }
+
+    return params
+  }
+
   // Gets a Jedis instance from the pool, and times the requested method invocations.
   private fun <T> jedis(op: JedisBinaryCommands.() -> T): T {
     return jedisPool.resource.use { jedisResource ->
@@ -330,5 +373,13 @@ class RealRedis(
   companion object {
     /** The charset used to convert String keys to ByteArrays for Jedis commands. */
     private val charset = Charsets.UTF_8
+  }
+
+  enum class ZAddOptions(value: String) {
+    XX("XX"),
+    NX("NX"),
+    LT("LT"),
+    GT("GT"),
+    CH("CH")
   }
 }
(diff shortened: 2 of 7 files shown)


misk-redis/src/main/kotlin/misk/redis/RealRedis.kt
```diff
@@ -302,6 +304,46 @@ class RealRedis(
     jedisPool.resource.use { jedis -> jedis.publish(channel, message) }
   }
 
+  override fun zadd(
+    key: String,
+    score: Double,
+    member: ByteString,
+    options: Set<String>): Long {
+    return jedisPool.resource.use { jedis ->
+      jedis.zadd(key.toByteArray(charset), score, member.toByteArray(), getZAddParams(options))}
```
tso, line 313: nit weird format with `}` not on a new line

misk-redis/src/main/kotlin/misk/redis/RealRedis.kt
```diff
@@ -302,6 +304,46 @@ class RealRedis(
     jedisPool.resource.use { jedis -> jedis.publish(channel, message) }
   }
 
+  override fun zadd(
+    key: String,
+    score: Double,
+    member: ByteString,
+    options: Set<String>): Long {
```
tso, line 311: should this be a `Set< ZAddOptions>`?

tso (approved):

misk-redis/src/main/kotlin/misk/redis/RealRedis.kt
romilbm: I thought not exposing the callers to that might be better. It is an internal enum. We ultimately need `ZAddParams`. 
The other option was to have that argument be `ZAddParams`.
I am open to suggestions if folks think `Set< ZAddOptions>` or `ZAddParams` is better. Let me know.

misk-redis/src/main/kotlin/misk/redis/FakeRedis.kt
```diff
@@ -398,4 +402,25 @@ class FakeRedis : Redis {
...
+  override fun zadd(
+    key: String,
+    scoreMembers: Map<ByteString, Double>,
+    options: Set<String>,
+  ): Long {
+    throw NotImplementedError("Fake client not implemented for this operation")
+  }
+
+  override fun zscore(key: String, member: String): Double? {
+    throw NotImplementedError("Fake client not implemented for this operation")
```
Rizziepit, line 424: Can you implement these? You won't be able to write tests without these in most Misk codebases, because folks generally don't have Redis set up in Docker

misk-redis/src/test/kotlin/misk/redis/AbstractRedisTest.kt
```diff
@@ -373,8 +374,8 @@ abstract class AbstractRedisTest {
 
   @Test fun hrandFieldUnsupportedCount() {
     // The Redis HRANDFIELD specification dictates different behaviour when count is negative.
-    // This behaviour cannot be adhered to by our implementation until Jedis fixes this bug:
-    // https://github.com/redis/jedis/issues/3017
+    // This behaviour cannot be adhered to by our implementation until redis fixes this bug:
+    // https://github.com/redis/redis/issues/3017
```
Rizziepit, line 378: This points to the wrong PR now?

misk-redis/src/main/kotlin/misk/redis/FakeRedis.kt
Rizziepit: Hmm.... what's the difference between this FakeRedis and the textFixtures FakeRedis? Looks like there's been some changes to this codebase since I last work on it

misk-redis/src/main/kotlin/misk/redis/RealRedis.kt
```diff
@@ -302,6 +304,46 @@ class RealRedis(
...
+
+    options.forEach {
+      when(it.uppercase()) {
+        ZAddOptions.XX.name -> params.xx()
+        ZAddOptions.NX.name -> params.nx()
+        ZAddOptions.LT.name -> params.lt()
+        ZAddOptions.GT.name -> params.gt()
+        ZAddOptions.CH.name -> params.ch()
+      }
+    }
```
Rizziepit, line 342: To throw an error if they provide an invalid param:
```suggestion
    val params = ZAddParams()

    options.forEach {
      when(ZAddOptions.valueOf(it.uppercase())) {
        ZAddOptions.XX -> params.xx()
        ZAddOptions.NX -> params.nx()
        ZAddOptions.LT -> params.lt()
        ZAddOptions.GT -> params.gt()
        ZAddOptions.CH -> params.ch()
      }
    }
```

misk-redis/src/test/kotlin/misk/redis/AbstractRedisTest.kt
```diff
@@ -750,4 +751,55 @@ abstract class AbstractRedisTest {
     assertNull(redis.hget("foo", "bar"))
     assertNull(redis["foo"])
   }
+
+  @Test fun zAddTest() {
+    redis.del("foo");
```
Rizziepit, line 756: Curious if storage is not reset before each test?

misk-redis/src/test/kotlin/misk/redis/AbstractRedisTest.kt
```diff
@@ -750,4 +751,55 @@ abstract class AbstractRedisTest {
...
+    redis.zadd("foo", 1.0, aMemberBytes, setOf("lt"));
+    assertEquals(1.0, redis.zscore("foo", "a"));
+
+    // gt: only update existing elements if the new score is greater than the current score.
+    redis.zadd("foo", 0.0, bMemberBytes, setOf("gt"));
+    assertEquals(1.0, redis.zscore("foo", "b"));
+    redis.zadd("foo", 2.0, bMemberBytes, setOf("gt"));
+    assertEquals(2.0, redis.zscore("foo", "b"));
+
+    redis.del("foo");
```
Rizziepit, line 786: Ditto on the storage not being reset

misk-redis/src/testFixtures/kotlin/misk/redis/testing/FakeRedis.kt
```diff
@@ -411,4 +421,89 @@ class FakeRedis @Inject constructor(
...
+
+  override fun zscore(key: String, member: String): Double? {
+    if (sortedSetKeyValueStore[key] == null) return null
+    return sortedSetKeyValueStore[key]!!.data[member.encodeUtf8()]
+  }
 }
+
+fun main() {
+
+}
```
Rizziepit, line 509:
```suggestion
```

misk-redis/src/testFixtures/kotlin/misk/redis/testing/FakeRedis.kt
```diff
@@ -411,4 +421,89 @@ class FakeRedis @Inject constructor(
...
+    member: ByteString,
+    options: Set<String>
+  ): Long {
+    verifyOptions(options)
+    var newFieldCount = 0L
+    var elementsChanged = 0L
+    val optionsUpper = options.map { it.uppercase() }.toMutableList()
+    if (optionsUpper.isEmpty() || (optionsUpper.size == 1 && optionsUpper.contains(CH.name))) {
+      optionsUpper.add(XX.name)
+      optionsUpper.add(NX.name)
```
Rizziepit, line 437: Aren't these two options mutually exclusive?

misk-redis/src/main/kotlin/misk/redis/FakeRedis.kt
romilbm: There is class level doc which says.
```
/** Mimics a Redis instance for testing. */
@Deprecated(
  message = "Moved to misk-redis-testing. This Fake does not emulate redis correctly.",
  replaceWith = ReplaceWith("misk.redis.testing.FakeRedis")
) 
class FakeRedis : Redis {
```
In the FakeRedisTests, the `misk.redis.testing.FakeRedis` gets used. I can still add a note in the error to direct folks to the other FakeRedis.

misk-redis/src/test/kotlin/misk/redis/AbstractRedisTest.kt
romilbm: Ahh.. At some point I did Find & Replace.. I missed reverting this change. Will do that.

misk-redis/src/main/kotlin/misk/redis/FakeRedis.kt
Rizziepit: Oh great. I missed this. Feel free to ignore my comments then

misk-redis/src/test/kotlin/misk/redis/AbstractRedisTest.kt
romilbm: I don't see anything obvious in the code (like a @BeforeEach). I just tried without those deletes and it still works.

misk-redis/src/testFixtures/kotlin/misk/redis/testing/FakeRedis.kt
romilbm: No. NX, LT, GT are mutually exclusive per the doc. 
In the absence of any options it behaves as though it has NX and XX.

Rizziepit: NX = Only add new elements. Don't update already existing elements.
XX = Only update elements that already exist. Don't add new elements.

So zadd is a no-op if it has both these options. Check out the command in `redis-cli`. It's not possible to specify both:
Screenshot 2023-10-19 at 3 21 55 PM

romilbm: the default behavior (without any options) is to add new elements and update existing. So it makes sense that from the UX p.o.v. that both NX and XX are not allowed together. I'll add it to the verification logic. 
For the implementation, I think it makes sense to treat them (NX, XX) as separate operations and do them together when no options are specified or just do the specified one. 
wdyt?

Rizziepit: > For the implementation, I think it makes sense to treat them (NX, XX) as separate operations and do them together when no options are specified or just do the specified one.

If by "do them together when no options are specified" you mean do an upsert, that is correct. zadd's default is to do an upsert

romilbm: Yes!

PR #2989, opened 2023-10-19
Add micrometer backed metrics as V3 Misk Metrics

escardin:
Long run plan is to remove these and use micrometer directly, however we need to bridge the migration with an interim set of helper methods that we can more easily remove during a later stage of the migration


misk-metrics/src/main/kotlin/misk/metrics/v2/Metrics.kt
```diff
@@ -184,9 +213,124 @@ interface Metrics {
...
+    name: String,
+    description: String = "",
+    tags: Iterable<Tag> = listOf(),
+    quantiles: Map<Double, Double> = misk.metrics.v3.defaultQuantiles,
+    maxAgeSeconds: Long = TimeUnit.MINUTES.toMillis(10)
+  ) = getMetricsV3().summary(name, description, tags, quantiles, maxAgeSeconds)
+
   companion object {
-    fun factory(registry: CollectorRegistry) = object : Metrics {
+    fun factory(
```
escardin, line 315: I did it this way to ensure API compatability.

misk-metrics/src/main/kotlin/misk/metrics/MetricsModule.kt
```diff
@@ -23,14 +32,34 @@ class MetricsModule : KAbstractModule() {
...
+    override fun get(): V2Metrics {
+      return V2Metrics.factory(registry, metricsV3)
+    }
+  }
+
+  internal class MeterRegistryProvider @Inject constructor(
+    private val registry: CollectorRegistry,
+  ) : Provider<PrometheusMeterRegistry> {
+    override fun get(): PrometheusMeterRegistry {
+      return PrometheusMeterRegistry(PrometheusConfig.DEFAULT,registry, Clock.SYSTEM)
```
keeferrourke, line 55: Could we inject the Clock here? Or does using FakeClock in tests cause problems with MeterRegistry?

misk-metrics/src/main/kotlin/misk/metrics/v3/Metrics.kt
```diff
@@ -0,0 +1,257 @@
...
+/**
+ * defaultBuckets is a default set of buckets which assumes the value is in milliseconds (ms).
+ *
+ * It contains 58 buckets which range from 1ms to 1hr.
+ *
+ * Adapted from the default M3 buckets.
+ *
+ * https://github.com/m3db/m3/blob/v1.1.0/src/x/instrument/methods.go#L57-L83
+ */
+val defaultBuckets = listOf(
```
keeferrourke, line 198: Nit: Could this be called `millisBuckets` then, or some such?

misk-metrics/src/main/kotlin/misk/metrics/v3/Metrics.kt
```diff
@@ -0,0 +1,257 @@
...
+  ) = Gauge
+    .builder(name, supplier)
+    .description(description)
+    .tags(tags)
+    .register(registry)
+
+  /**
+   * histogram creates a new `Histogram` prometheus type with the supplied parameters.
+   *
+   * NOTE: `misk.metrics.v2.Metrics` is NOT backward compatible with `misk.metrics.Metrics`.
```
keeferrourke, line 74: Nit: This is in the v3 package, but talking about v2.

misk-metrics/src/test/kotlin/misk/metrics/v3/PeakGaugeTest.kt
```diff
@@ -0,0 +1,101 @@
...
+   * Verifies that [PeakGauge.collect] returns correct information and from and resets each
+   * [PeakGauge].
+   */
+  @Test fun `collection returns correct information and resets each child`() {
+    val registry = CollectorRegistry(true)
+    val metrics = Metrics(PrometheusMeterRegistry(PrometheusConfig.DEFAULT, registry, Clock.SYSTEM))
+
+    val gauge1 = Metrics.Companion.PeakGauge()
+    metrics.gauge(
+      "some_name", gauge1::getAndClear, "some help text", listOf(
```
keeferrourke, line 51: This seems a little unwieldy. Should there be a `metrics.v3.Metrics#peakGauge(name: String, description: String, tags: List<Tag>)` method?

That would help migrate from v2's peakGauge as well.

misk-metrics/src/main/kotlin/misk/metrics/MetricsModule.kt
escardin: Unfortunately it's a micrometer clock :(

misk-metrics/src/main/kotlin/misk/metrics/v3/Metrics.kt
escardin: It's default buckets because that's what exists in v2. We actually don't want buckets in micrometer, it's an anti-pattern. The bucketing I put into the micrometer versions of things is to keep things consistent until we can do a separate migration away from that, as I'm concerned that it may change things in an unexpected way.

misk-metrics/src/main/kotlin/misk/metrics/v2/Metrics.kt
```diff
@@ -184,9 +213,124 @@ interface Metrics {
...
+   * See https://prometheus.github.io/client_java/io/prometheus/client/Counter.html for more info.
+   *
+   * @param name the name of the metric which will be supplied to prometheus.
+   *  Must be unique across all metric types.
+   * @param description a human-readable description of what the metric does for display in the metrics UI.
+   * @param tags a list of key/value pairs to categorize the metric.
+   */
+  @Deprecated(
+    message = "Misk Metrics V2 is Deprecated, please use V3. This is a bridge method to aid migration to V3",
+    level = DeprecationLevel.WARNING
```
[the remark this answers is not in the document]
escardin: Unfortunately they're not directly equivalent such that a replaceWith would then compile :(

misk-metrics/src/main/kotlin/misk/metrics/v2/Metrics.kt
```diff
@@ -31,6 +44,10 @@ interface Metrics {
    * @param help human-readable help text that will be supplied to prometheus.
    * @param labelNames the names (a.k.a. keys) of all the labels that will be used for this metric.
    */
+  @Deprecated(
+    message = "Misk Metrics V2 is Deprecated, please use V3",
+    level = DeprecationLevel.WARNING
+  )
   fun counter(
```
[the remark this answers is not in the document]
escardin: The idea is that I'm going to use moderne to move them to the v2 bridge methods, and then move them to v3 later. The reason for the bridge methods is I can do a partial migration of what exists, and only when everything is on the bridge methods can I move them to v3 directly (and then off of misk-metrics)

misk-metrics/src/test/kotlin/misk/metrics/v3/PeakGaugeTest.kt
escardin: Unfortunately I don't think so (but happy for solutions). Micrometer gauges expect to have an object they can observe, and making many copies of the same thing is likely to result in errors.

PeakGauge is used only in one or two places in cash, and it's overall necessity should be lower with micrometer and some adjustments to how we do metrics (which should come with dumping bucketing if I understood the concerns correctly)

keeferrourke: What about something like this?

https://github.com/cashapp/misk/compare/lovelace/MiskMetrics/MetricsV3...keefer.20231023.keefer.metrics-v3-peak-gauge-suggestion?expand=1

misk-metrics/src/main/kotlin/misk/metrics/v3/Metrics.kt
keeferrourke: Could it not still be called `millisBuckets` though? We had a discussion about this for the v2 API about 6 months ago. @ericloe-cash was working on a metrics helper to improve bucketing in misk metrics, not sure where that ended up though.

keeferrourke:
Closing for now.

PR #2990, opened 2023-10-20
Add deployment to DashboardIndexAccessBlock

adrw:
Providing deployment within the access block allows for more context rich access notices which take into account the `DevelopmentOnly` semantics for locally running development service.


adrw:
There is only one internal use case of this so the breaking change is worth doing.

frojasg (approved):
LGTM

jvmakine (approved):

PR #2992, opened 2023-10-20
Add wrappers for zadd, zscore methods for redis sorted set

romilbm:
Some basic commands for [Redis SortedSet ](https://redis.io/docs/data-types/sorted-sets/). [ZADD](https://redis.io/commands/zadd/) - basic command to add members to SortedSet [ZSCORE](https://redis.io/commands/zscore/) - get the score of the member in the SortedSet

### DIFF
--- a/misk-redis/src/main/kotlin/misk/redis/RealRedis.kt
+++ b/misk-redis/src/main/kotlin/misk/redis/RealRedis.kt
@@ -1,5 +1,10 @@
 package misk.redis
 
+import misk.redis.Redis.ZAddOptions.CH
+import misk.redis.Redis.ZAddOptions.GT
+import misk.redis.Redis.ZAddOptions.LT
+import misk.redis.Redis.ZAddOptions.NX
+import misk.redis.Redis.ZAddOptions.XX
 import okio.ByteString
 import okio.ByteString.Companion.toByteString
 import redis.clients.jedis.JedisCluster
@@ -11,6 +16,7 @@ import redis.clients.jedis.UnifiedJedis
 import redis.clients.jedis.args.ListDirection
 import redis.clients.jedis.commands.JedisBinaryCommands
 import redis.clients.jedis.params.SetParams
+import redis.clients.jedis.params.ZAddParams
 import redis.clients.jedis.util.JedisClusterCRC16
 import java.lang.reflect.InvocationHandler
 import java.lang.reflect.InvocationTargetException
@@ -362,6 +368,59 @@ class RealRedis(
 
   override fun flushAll() {
     unifiedJedis.flushAll()
+  }
+
+  override fun zadd(
+    key: String,
+    score: Double,
+    member: String,
+    vararg options: Redis.ZAddOptions,
+  ): Long {
+    return unifiedJedis.zadd(
+      key.toByteArray(charset),
+      score,
+      member.toByteArray(charset),
+      getZAddParams(options)
+    )
+  }
+
+  private fun getZAddParams(options: Array<out Redis.ZAddOptions>): ZAddParams {
+    val params = ZAddParams()
+
+    options.forEach {
+      when (it) {
+        XX -> params.xx()
+        NX -> params.nx()
+        LT -> params.lt()
+        GT -> params.gt()
+        CH -> params.ch()
+      }
+    }
+
+    return params
+  }
+
+  override fun zadd(
+    key: String,
+    scoreMembers: Map<String, Double>,
+    vararg options: Redis.ZAddOptions,
+  ): Long {
+    val params = getZAddParams(options)
+    val keyBytes = key.toByteArray(charset)
+    val scoreMembersBytes =
+      scoreMembers.entries.associate { it.key.toByteArray(charset) to it.value }
+    return unifiedJedis.zadd(
+      keyBytes,
+      scoreMembersBytes,
+      params
+    )
+  }
+
+  override fun zscore(key: String, member: String): Double? {
+    return unifiedJedis.zscore(
+      key.toByteArray(charset),
+      member.toByteArray(charset)
+    )
   }
 
   // Gets a Jedis instance from the pool, and times the requested method invocations.
--- a/misk-redis/src/test/kotlin/misk/redis/AbstractRedisTest.kt
+++ b/misk-redis/src/test/kotlin/misk/redis/AbstractRedisTest.kt
@@ -1,12 +1,19 @@
 package misk.redis
 
+import misk.redis.Redis.ZAddOptions.CH
+import misk.redis.Redis.ZAddOptions.GT
+import misk.redis.Redis.ZAddOptions.LT
+import misk.redis.Redis.ZAddOptions.NX
+import misk.redis.Redis.ZAddOptions.XX
 import okio.ByteString.Companion.encodeUtf8
 import org.assertj.core.api.Assertions.assertThat
 import org.assertj.core.api.Assertions.assertThatThrownBy
 import org.junit.jupiter.api.Test
 import org.junit.jupiter.api.assertThrows
 import redis.clients.jedis.args.ListDirection
+import redis.clients.jedis.exceptions.JedisDataException
 import kotlin.test.assertEquals
+import kotlin.test.assertFailsWith
 import kotlin.test.assertFalse
 import kotlin.test.assertNull
 import kotlin.test.assertTrue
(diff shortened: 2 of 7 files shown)


misk-redis/src/main/kotlin/misk/redis/RealRedis.kt
```diff
@@ -302,6 +303,51 @@ class RealRedis(
     jedisPool.resource.use { jedis -> jedis.publish(channel, message) }
   }
 
+  override fun zadd(key: String, score: Double, member: ByteString, options: Set<String>): Long {
```
frojasg, line 306: should `options` be a `Set<ZAddOptions>` ? You are already exposing the `ZAddOptions` unless you meant to make that private?

misk-redis/src/main/kotlin/misk/redis/RealRedis.kt
```diff
@@ -302,6 +303,51 @@ class RealRedis(
...
+    return jedisPool.resource.use { jedis ->
+      jedis.zadd(key.toByteArray(charset), score, member.toByteArray(), getZAddParams(options))
+    }
+  }
+
+  private fun getZAddParams(options: Set<String>): ZAddParams {
+    val params = ZAddParams()
+
+    options.forEach {
+      when(ZAddOptions.valueOf(it.uppercase())) {
```
frojasg, line 316: Should we add some sort validation before this? otherwise this will throw an exception.

misk-redis/src/main/kotlin/misk/redis/RealRedis.kt
```diff
@@ -302,6 +303,51 @@ class RealRedis(
...
+        ZAddOptions.GT -> params.gt()
+        ZAddOptions.CH -> params.ch()
+      }
+    }
+
+    return params
+  }
+
+  enum class ZAddOptions(value: String) {
+    XX("XX"),
```
frojasg, line 329: Why are you adding the `value: String`?

misk-redis/src/main/kotlin/misk/redis/RealRedis.kt
romilbm: > Should we add some sort validation before this?

Won't the validation step throw an exception too? Or did you mean to filter out invalid options? I was intending this step to throw an exception if an invalid value was provided.

romilbm: Maybe if we do it like you suggested above Set<ZAddOptions> that might eliminate the need to do this conversation/validation at all.

misk-redis/src/testFixtures/kotlin/misk/redis/testing/FakeRedis.kt
```diff
@@ -42,6 +50,9 @@ class FakeRedis @Inject constructor(
   private val hKeyValueStore =
     ConcurrentHashMap<String, Value<ConcurrentHashMap<String, ByteString>>>()
 
+  private val sortedSetKeyValueStore =
+    ConcurrentHashMap<String, Value<SortedMap<ByteString, Double>>>()
```
Rizziepit, line 54: `SortedMap` sorts by key, not value. So it won't provide you the same ordering as a Redis sorted set which sorts by "value", i.e. score. You might be better off using a map for now to keep the implementation simple.

misk-redis/src/testFixtures/kotlin/misk/redis/testing/FakeRedis.kt
```diff
@@ -411,4 +422,96 @@ class FakeRedis @Inject constructor(
...
+      ) {
+      sortedSetKeyValueStore[key]!!.data[member] = score
+      if (!exists) newFieldCount++
+      elementsChanged++
+    } else if ( // overwrite existing if score greater than and add new ones.
+        (options.contains(GT) && exists && score > sortedSetKeyValueStore[key]!!.data[member]!!)
+        || (options.contains(GT) && !exists)) {
+      sortedSetKeyValueStore[key]!!.data[member] = score
+      if (!exists) newFieldCount++
+      elementsChanged++
```
Rizziepit, line 464: Is the behaviour corrrect if you do `zadd LT XX` or `zadd GT XX`? It looks like when `LT` or `GT` is set, `XX` gets ignored

misk-redis/src/testFixtures/kotlin/misk/redis/testing/FakeRedis.kt
```diff
@@ -411,4 +422,96 @@ class FakeRedis @Inject constructor(
...
+  ): Long {
+    verifyZAddOptions(options)
+    var newFieldCount = 0L
+    var elementsChanged = 0L
+
+    if (!sortedSetKeyValueStore.containsKey(key)) {
+      sortedSetKeyValueStore[key] = Value(data = sortedMapOf(), expiryInstant = Instant.MAX)
+    }
+    val exists = sortedSetKeyValueStore[key]!!.data[member] != null
+
```
Rizziepit, line 440: I'd recommend handling the no-op `XX` and `NX` cases here and returning early, to help simplify the logic below:

```suggestion
    if ((options.contains(XX) && !exists) || (options.contains(NX) && exists)) {
      return 0
    }
 
```

I think this will also eliminate the bug I mention below

misk-redis/src/test/kotlin/misk/redis/AbstractRedisTest.kt
```diff
@@ -750,4 +758,85 @@ abstract class AbstractRedisTest {
...
+    assertEquals(4.0, redis.zscore("foo", "c"))
+
+    // gt: only update existing elements if the new score is greater than the current score.
+    redis.zadd("foo", 0.0, bMemberBytes, GT)
+    assertEquals(1.0, redis.zscore("foo", "b"))
+    redis.zadd("foo", 2.0, bMemberBytes, GT)
+    assertEquals(2.0, redis.zscore("foo", "b"))
+    // GT doesn't prevent new members from being added.
+    redis.zadd("foo", 4.0, "d".encodeUtf8(), GT)
+    assertEquals(4.0, redis.zscore("foo", "d"))
```
Rizziepit, line 802: You're missing tests for flag combos like `LT XX` or `GT XX`

romilbm: added more tests and organized them better too.

misk-redis/src/testFixtures/kotlin/misk/redis/testing/FakeRedis.kt
romilbm: This should be fixed now.

misk-redis/src/testFixtures/kotlin/misk/redis/testing/FakeRedis.kt
romilbm: in `zadd LT XX` or `zadd GT XX` the contract is this.. 
```
 for existing ones, if the score is less/more than the existing scores - update
 XX will prevent adding new ones.
```
I have updated the code now. So this should be fixed.

misk-redis/src/testFixtures/kotlin/misk/redis/testing/FakeRedis.kt
romilbm: You are right, this is a mistake on my part. I changed this. Added a comment here.

misk-redis/src/testFixtures/kotlin/misk/redis/testing/FakeRedis.kt
```diff
@@ -411,4 +427,113 @@ class FakeRedis @Inject constructor(
...
+    if (options.contains(LT) && options.contains(XX) && exists
+      && score < currentScore!!) return true
+
+    // GT XX
+    // for existing ones, the score should be more than the existing scores.
+    // XX will prevent adding new ones.
+    if (options.contains(GT) && options.contains(XX) && exists
+      && score > currentScore!!) return true
+
+    return false
```
Rizziepit, line 495: You can simplify this logic a lot if you exit early for the `XX` and `NX` modifiers.
```suggestion
    if ((options.contains(XX) && !exists) || (options.contains(NX) && exists)) {
      return false
    }

    if (options.contains(LT)) {
      return !exists || score < currentScore!!
    }

    if (options.contains(GT)) {
      return !exists || score > currentScore!!
    }
   
    return true
```

Rizziepit (approved):
Suggested some changes in comments. The most important change is to make the `zadd` and `zscore` methods consistent with similar Redis methods like `hset` and `hget`, i.e. `member` should be a `String` argument everywhere, and we should consistently do the encoding ourselves and invoke Jedis methods with byte arrays instead of strings

misk-redis/src/main/kotlin/misk/redis/RealRedis.kt
```diff
@@ -302,6 +308,43 @@ class RealRedis(
     jedisPool.resource.use { jedis -> jedis.publish(channel, message) }
   }
 
+  override fun zadd(key: String, score: Double, member: ByteString, vararg options: Redis.ZAddOptions): Long {
+    return jedisPool.resource.use { jedis ->
+      jedis.zadd(key.toByteArray(charset), score, member.toByteArray(), getZAddParams(options))
+    }
+  }
```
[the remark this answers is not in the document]
romilbm: Done.

misk-redis/src/main/kotlin/misk/redis/RealRedis.kt
```diff
@@ -302,6 +308,43 @@ class RealRedis(
...
+
+  override fun zadd(
+    key: String,
+    scoreMembers: Map<ByteString, Double>,
+    vararg options: Redis.ZAddOptions
+  ): Long {
+    val params = getZAddParams(options)
+    val keyBytes = key.toByteArray(charset)
+    val scoreMembersBytes = scoreMembers.entries.associate { it.key.toByteArray() to it.value }
+    return jedisPool.resource.use { jedis -> jedis.zadd(keyBytes, scoreMembersBytes, params)}
```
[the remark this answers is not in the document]
romilbm: Done.

misk-redis/src/main/kotlin/misk/redis/RealRedis.kt
```diff
@@ -302,6 +308,43 @@ class RealRedis(
...
+  }
+
+  override fun zscore(key: String, member: String): Double? {
+    return jedisPool.resource.use { it.zscore(key, member) }
```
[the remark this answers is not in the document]
romilbm: Done.

misk-redis/src/testFixtures/kotlin/misk/redis/testing/FakeRedis.kt
romilbm: Spoke offline. I think the existing one is easier to understand if you read the comments on each option in the docs. So keeping it.

PR #2995, opened 2023-10-31
Add return type to withTags

billymillions:


wisp/wisp-logging/src/main/kotlin/wisp/logging/Logging.kt
```diff
@@ -96,7 +96,7 @@ fun KLogger.log(level: Level, th: Throwable, vararg tags: Tag, message: () -> An
     }
 }
 
-fun withTags(vararg tags: Tag, f: () -> Unit) {
+fun <T> withTags(vararg tags: Tag, f: () -> T): T {
```
tgregory-block, line 99: Should we inline this? I'm not sure how likely it is to be used on hot paths/inner loops so it may or may not be worth it

ean5533 (approved):

PR #2996, opened 2023-10-31
add ff value into tracing span tags

adrianliu0815:


wisp/wisp-launchdarkly/src/main/kotlin/wisp/launchdarkly/LaunchDarklyFeatureFlags.kt
```diff
@@ -70,6 +71,9 @@ class LaunchDarklyFeatureFlags @JvmOverloads constructor(
             buildUser(feature, key, attributes),
         )
         checkDefaultNotUsed(feature, result)
+        GlobalTracer.get()?.activeSpan()?.let { span ->
```
lkerford, line 74: Is this something we can test?

adrianliu0815: I am curious about this too, @qi-liao , do you know?

qi-liao: no idea

PR #2997, opened 2023-11-02
Upgrade jedis to 4.4 and replace JedisPool with a unified API for interacting with Redis

rainecp:
**Why?**

Currently, all Redis client requests are going only to the leader of a redis replication group. This is a scalability bottleneck as it doesn’t allow us to leverage the replica nodes within a cluster. This also has a consequence of unnecessary cost going into vertical scaling since horizontal scaling is not available by default. This is a preliminary work to support cluster mode.

**What?**

This contains changes to convert legacy `JedisPool` to `JedisPooled` use to support single Redis instance. This is backwards compatible so services should not need to change their usage of Redis.

- Replaced `JedisPool` with `UnifiedJedis`. This is an interface that provides a unified API for interacting with Redis. Currently supports `JedisCluster`, `JedisPooled`, `JedisSharding` etc 
- Jedis version updated to 4.4, 4.0 introduced UnifiedJedis but 4.4 includes support for pipelined() and multi() which is part of our `Redis` interface
- Added custom `JedisPooledWithMetrics` similar to existing `JedisPool` implementation but with updated implementation on setting the metrics
- Replaced `JedisPoolConfig` with `ConnectionPoolConfig` in `RedisModule` but keeping the old one
- Added more test cases for watch/unwatch

**Testing**
Fixed and ran all tests that use FakeRedis, DockerRedis and RedisClientMetrics, also added new ones. Also tested manually by publishing to maven local then using this version on another service that uses Redis and made sure there are no changes necessary, existing unit tests are passing and using local redis instance also works.


misk-redis/src/main/kotlin/misk/redis/JedisPoolWithMetrics.kt
```diff
@@ -1,78 +0,0 @@
-package misk.redis
-
-import org.apache.commons.pool2.PooledObject
-import redis.clients.jedis.Jedis
-import redis.clients.jedis.JedisFactory
-import redis.clients.jedis.JedisPool
-import redis.clients.jedis.JedisPoolConfig
-import redis.clients.jedis.Protocol
-
-internal class JedisPoolWithMetrics(
```
tgregory-block, line 10: I'm working on a PR to add bucket4j support to misk, as an alternative to `cash-permit`. Unfortunately, its redis backend implementation currently requires a `JedisCluster` or a `Pool<Jedis>`. I can work on a PR to the upstream that adds support for `UnifiedJedis`, but until then could we keep `JedisPoolWithMetrics`? I'll just instantiate a dedicated pool in the bucket4j redis module

rainecp: I will be adding the `JedisCluster` support next. But yes, I can revert `JedisPoolWithMetrics` for now and clean it up on a later PR. Also I just realized it was deleted here, I just renamed it.

tso (approved):
looks good to me basically, but i have no redis expertise

tgregory-block (approved):

damar-block (approved):

keeferrourke (approved):

misk-redis/src/testFixtures/kotlin/misk/redis/testing/DockerRedis.kt
```diff
@@ -88,17 +88,15 @@ object DockerRedis : ExternalDependency {
...
       close()
     }
     composer.stop()
   }
 
   override fun beforeEach() {
-    client.resource.use { jedis ->
-      jedis.flushAll()
-    }
+    jedis.flushAll()
```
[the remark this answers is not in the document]
rainecp: Yes, it is different, it implements autocloseable.

PR #3000, opened 2023-11-02
fix(wisp-feature-testing): fix bug in FakeLegacyFeatureFlags.overrideKey that caused FakeFeatureFlags.trackJson to fail with ClassCastException

torstenrudolf:
Added two tests for `trackJson` and `trackString`.

`trackString` is working, however `trackJson` causes a `ClassCastException`.

Am I doing something wrong or is this a bug in `FakeFeatureFlags`?

logs:
```
Exception in thread "pool-1-thread-1" java.lang.ClassCastException: class misk.feature.testing.FakeFeatureFlagsTest$trackJson works$$inlined$overrideJson$2 cannot be cast to class misk.feature.testing.FakeFeatureFlagsTest$JsonFeature (misk.feature.testing.FakeFeatureFlagsTest$trackJson works$$inlined$overrideJson$2 and misk.feature.testing.FakeFeatureFlagsTest$JsonFeature are in unnamed module of loader 'app')
	at misk.feature.testing.FakeFeatureFlagsTest$trackJson works$1.invoke(FakeFeatureFlagsTest.kt:354)
	at wisp.feature.testing.FakeLegacyFeatureFlags.overrideKey$lambda$6$lambda$5$lambda$4(FakeLegacyFeatureFlags.kt:359)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1136)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:635)
	at java.base/java.lang.Thread.run(Thread.java:833)

```

**update**: added a fix for it into wisp-feature-testing. The bug was that in the json case inside `overrideKey<T>` the value is already transformed from `T` to `() -> String` and then it tries to call the tracker with this instead of a `T`. So I added the functionality to provide a backwards mapper as a quick fix. But I'm of course open to other ideas on how to fix this.


wisp/wisp-feature-testing/src/main/kotlin/wisp/feature/testing/FakeLegacyFeatureFlags.kt
```diff
@@ -369,8 +380,9 @@ constructor(
         value: T,
         attributes: Attributes = defaultAttributes
     ) {
-        val jsonValue = { moshi().adapter(T::class.java).toSafeJson(value) }
-        overrideKey(feature, key, jsonValue, attributes)
+        val m = moshi().adapter(T::class.java)
```
adrw, line 383:
```suggestion
        val adapter = moshi().adapter(T::class.java)
```

Adapter is a more idiomatic variable name and leaves the code more readable.

misk-feature/src/test/kotlin/misk/feature/testing/FakeFeatureFlagsTest.kt
```diff
@@ -321,4 +322,41 @@ internal class FakeFeatureFlagsTest {
...
+    fun listener(newValue: String) {
+      require(valueReceivedByListenerReceived == null)
+      valueReceivedByListenerReceived = newValue
+    }
+
+    subject.trackString(FEATURE, Executors.newSingleThreadExecutor(),  ::listener)
+
+    // override the featureflag, this should trigger trackString
+    subject.override(FEATURE, "newValue")
+    Thread.sleep(10)
```
frojasg, line 340: This can make this test a bit flaky. Could you use some `CountDownLatch` like this [example](https://github.com/cashapp/misk/blob/45dfef7622b6d14dbd6281722c1f9cbd3279decc/misk-aws/src/test/kotlin/misk/jobqueue/sqs/SqsJobQueueServiceTest.kt#L58C18-L58C32).

torstenrudolf: this is great, thank you!

frojasg (approved):

adrw (approved):

PR #3004, opened 2023-11-08
Allow the MYSQL_HOST env var to override the hard-coded default

staktrace:
All the other database types auto-detect if they are inside a docker container and use host.docker.internal. However, the MySQL codepath does not. Doing the same thing here would be ideal, but is tricky for reasons discussed at https://cash.slack.com/archives/C01M5M297T7/p1699385740943289. Instead, allow setting a MYSQL_HOST envvar to override the hard-coded default.

### DIFF
--- a/misk-jdbc/src/main/kotlin/misk/jdbc/DataSourceConfig.kt
+++ b/misk-jdbc/src/main/kotlin/misk/jdbc/DataSourceConfig.kt
@@ -128,7 +128,7 @@ data class DataSourceConfig @JvmOverloads constructor(
       DataSourceType.MYSQL -> {
         copy(
           port = port ?: 3306,
-          host = host ?: "127.0.0.1",
+          host = host ?: System.getenv("MYSQL_HOST") ?: "127.0.0.1",
           database = database ?: ""
         )
       }


misk-jdbc/src/main/kotlin/misk/jdbc/DataSourceConfig.kt
```diff
@@ -116,7 +116,7 @@ data class DataSourceConfig @JvmOverloads constructor(
       DataSourceType.MYSQL -> {
         copy(
           port = port ?: 3306,
-          host = host ?: "127.0.0.1",
+          host = host ?: System.getenv("MYSQL_HOST") ?: "127.0.0.1",
```
adrw, line 119: I'm curious why we need this when the host in the config could be set instead? Fine to stamp temporarily but a 3rd way to set this seems like it'll make future debugging more difficult.

adrw (approved):

misk-jdbc/src/main/kotlin/misk/jdbc/DataSourceConfig.kt
staktrace: When running a service under orc, the service might be running either directly on the host or in a docker container. In both cases it will use the same config, but we need the `host` value to be different in the two cases (`127.0.0.1` vs `host.docker.internal`). For armeria services the config yaml is loaded using hoplite which does variable expansion so things like `host: ${MYSQL_HOST:-localhost}` work in the config, but that's not the case for misk configs. Allowing an envar to override it here seemed like the simplest option, but I'm open to suggestions on how to do this differently.

nssherpa: Explicitly specifying the host in the`data_source_clusters` cluster would work but as @staktrace mentioned under orc a service can run both with or without docker. So we need this host field to be set accordingly. Passing in an envar seems to be the easiest option right now, else we'll have to add logic to unfurl the host's value given an envar.

The code above also seems to see if if the service is running under docker but it doesn't really recognize that the service is running under docker it correctly. Else we could have substituted `"127.0.0.1"` with server_hostname like we have for other data types below.
```
 val isRunningInDocker = File("/proc/1/cgroup")
      .takeIf { it.exists() }?.useLines { lines ->
        lines.any { it.contains("/docker") }
      } ?: false
```

staktrace:
Marking this as ready for review as @nssherpa has done some additional investigation and this does seem to be the way forward.

nssherpa (approved):

nssherpa:
I have also confirmed this with the db team that they always set the `host` in the resource.*.yaml file so this alternate code path for retrieving the `MYSQL_HOST` envar shouldn't be used for most cash services running in actual environments https://cash.slack.com/archives/C915WS04Q/p1708542253513679.

PR #3005, opened 2023-11-13
Enhance bucket4j functionality

tgregory-block:
This PR has two commits:
* First, it adds functionality to the `RateLimiter` interface that allows services to check the number of available tokens in a bucket, reset buckets to their max capacity, and test consumption from a bucket.
* Second, it adds performance metrics to these and the already existing operations, so services can monitor how long it takes to consume/release/etc a rate limit token

### DIFF
--- a/wisp/wisp-rate-limiting/src/main/kotlin/wisp/ratelimiting/RateLimiter.kt
+++ b/wisp/wisp-rate-limiting/src/main/kotlin/wisp/ratelimiting/RateLimiter.kt
@@ -28,11 +28,35 @@ interface RateLimiter {
   ): ConsumptionData
 
   /**
+   * This tests whether [amount] tokens are available in the bucket associated with the given key.
+   * It is essentially a dry run of [consumeToken].
+   * Note that this data may be stale when it comes back, as time has elapsed and other pods
+   * could have taken tokens in the meantime.
+   */
+  fun testConsumptionAttempt(
+    key: String,
+    configuration: RateLimitConfiguration,
+    amount: Long = 1
+  ): TestConsumptionResult
+
+  /**
    * Releases [amount] tokens back to the bucket associated with the given key
    * This will raise any exception thrown by the bucket4j proxy manager implementation, e.g.
    * subclasses of [JedisException] when using the Jedis implementation.
    */
   fun releaseToken(key: String, configuration: RateLimitConfiguration, amount: Long = 1)
+
+  /**
+   * Returns how many tokens remain in the bucket.
+   * Note that this data may be stale when it comes back, as time has elapsed and other pods
+   * could have taken tokens in the meantime.
+   */
+  fun availableTokens(key: String, configuration: RateLimitConfiguration): Long
+
+  /**
+   * Resets the bucket back to its maximum capacity
+   */
+  fun resetBucket(key: String, configuration: RateLimitConfiguration)
 
   /**
    * Executes the given function if a token is available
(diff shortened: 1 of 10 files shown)


wisp/wisp-rate-limiting/src/main/kotlin/wisp/ratelimiting/RateLimiter.kt
```diff
@@ -27,13 +27,37 @@ interface RateLimiter {
     amount: Long = 1
   ): ConsumptionData
 
+  /**
+   * This tests whether [amount] tokens are available in the bucket associated with the given key.
+   * It is essentially a dry run of [consumeToken].
+   * Note that this data may be stale when it comes back, as time has elapsed and other pods
+   * could have taken tokens in the meantime.
+   */
+  fun testConsumptionAttempt(
```
swankjesse, line 36: What’s the use case for this?

wisp/wisp-rate-limiting/src/main/kotlin/wisp/ratelimiting/RateLimiter.kt
```diff
@@ -27,13 +27,37 @@ interface RateLimiter {
     amount: Long = 1
   ): ConsumptionData
```
swankjesse, line 28: Exceptions for flow control on this are bad news

swankjesse: Ahh, I’m misreading the docs. ‘This will raise’ is not raising if the consumption fails, but raising if the underlying datastore is unreachable. We don’t typically document exceptions like that.

wisp/wisp-rate-limiting/src/main/kotlin/wisp/ratelimiting/RateLimiter.kt
```diff
@@ -27,13 +27,37 @@ interface RateLimiter {
...
 
+  /**
+   * Returns how many tokens remain in the bucket.
+   * Note that this data may be stale when it comes back, as time has elapsed and other pods
+   * could have taken tokens in the meantime.
+   */
+  fun availableTokens(key: String, configuration: RateLimitConfiguration): Long
+
+  /**
+   * Resets the bucket back to its maximum capacity
```
swankjesse, line 57: Is this dangerous? What’s the use case for this?

swankjesse (approved):

wisp/wisp-rate-limiting/src/main/kotlin/wisp/ratelimiting/RateLimiter.kt
tgregory-block: Essentially the same as `availableTokens` but it provides more information, namely the timestamp when the bucket will be refilled. Callers might want to know this if there weren't enough tokens in the bucket for their test consumption

wisp/wisp-rate-limiting/src/main/kotlin/wisp/ratelimiting/RateLimiter.kt
tgregory-block: Some rate limits are used as mutexes, and we want to reset the bucket after they're used. There may also be other cases based on business logic (e.g. I can imagine a scenario where we want reset a customer's attempts for an action under some circumstance) but that's a hypothetical on my part

wisp/wisp-rate-limiting/src/main/kotlin/wisp/ratelimiting/RateLimiter.kt
tgregory-block: Noted, I can see how that would be confusing. Will remove this the next time I touch this code

PR #3009, opened 2023-11-15
Add exception throwing functionality to FakeJobQueue

paigevpark:
There should be a way to test FakeJobQueue throwing exceptions on `enqueue`, similar to https://github.com/squareup/cash-event-client/blob/1bdd07a70b301a59fcd9ad6bbf77be5f13cae1b0/testing/src/main/kotlin/com/squareup/cash/eventclient/testing/FakeCloudEventProcessor.kt#L101


misk-jobqueue/src/testFixtures/kotlin/misk/jobqueue/FakeJobQueue.kt
```diff
@@ -37,7 +40,25 @@ class FakeJobQueue @Inject constructor(
...
+    queueName?.let { failureJobQueues.getOrPut(it, ::LinkedBlockingQueue).add(e) } ?: failureJobQueue.add(e)
+  }
+
+  private fun nextFailure(queueName: QueueName?): Exception? {
+    return queueName?.let { failureJobQueues.get(it)?.poll()} ?: failureJobQueue.poll()
+  }
+
+  @Throws
+  private fun throwIfQueuedFailure(queueName: QueueName?) {
+    nextFailure(queueName)?.let {throw(it)}
```
adrw, line 60:
```suggestion
    nextFailure(queueName)?.let { throw(it) }
```

adrw (approved):

PR #3013, opened 2023-11-20
Introduce a dynamodb backed cluster implementation

tso:


misk-clustering-dynamodb/src/test/kotlin/misk/clustering/DynamoClusterTest.kt
```diff
@@ -0,0 +1,62 @@
...
+class DynamoClusterTest {
+  @MiskTestModule
+  private val module = TestModule()
+
+  @Inject private lateinit var cluster: Cluster
+  @Inject private lateinit var dynamoClusterWatcherTask: DynamoClusterWatcherTask
+  @Inject private lateinit var fakeClock: FakeClock
+
+  @Test
+  fun basic() {
```
tso, line 42: going to add the following tests, but wanted to get earlier review:
- more than 100 entries in table (tests pagination)
- more than 1 entry in table (just sanity)
- inactive cluster doesn't put in an entry
- resource mapper still works

misk-clustering-dynamodb/src/main/kotlin/misk/clustering/dynamo/DynamoClusterModule.kt
```diff
@@ -0,0 +1,33 @@
...
+import misk.ServiceModule
+import misk.clustering.Cluster
+import misk.inject.KAbstractModule
+import misk.tasks.RepeatedTaskQueue
+import misk.tasks.RepeatedTaskQueueFactory
+import java.util.UUID
+
+class DynamoClusterModule : KAbstractModule() {
+  override fun configure() {
+    val dynamoCluster = DynamoCluster(System.getenv("MY_POD_NAME") ?: UUID.randomUUID().toString())
```
jvmakine, line 15: I don't think we should use `System.getenv("MY_POD_NAME")` here, as that is no longer guaranteed to be unique? Maybe just `UUID` always?

tso: what're we doing for logs, metrics, etc? until now we've always operated with pod name as the ID, are we changing that for datadog stuff?

if i see, for example, a high cpu usage node in datadog how do i determine if it owns many leases?

tso: i'm fine with uuid here just curious in general mostly

misk-clustering-dynamodb/src/main/kotlin/misk/clustering/dynamo/DyClusterMember.kt
```diff
@@ -0,0 +1,15 @@
...
+import software.amazon.awssdk.enhanced.dynamodb.mapper.annotations.DynamoDbPartitionKey
+
+@DynamoDbBean
+class DyClusterMember {
+  @get:DynamoDbPartitionKey
+  var name: String? = null
+
+  // Stored as epoch ms
+  @get:DynamoDbAttribute("updated_at")
+  var updated_at: Long? = null
```
jvmakine, line 14: I might prefer to keep `pod_name` as an additional non-functional metadata, similar to https://github.com/squareup/cash-dynamodb-lease/blob/48fd2dbe64e28f9c3a5fd7ab280b00ce2e83b4de/lib/src/main/kotlin/com/squareup/cash/dynamodb/lease/RealDynamoDbLease.kt#L153
and keep `name` as UUID. This would allow us to run multiple cluster clients per pod as well if we wanted to do that for some reason (not saying we should)

misk-clustering-dynamodb/src/main/kotlin/misk/clustering/dynamo/DynamoClusterWatcherTask.kt
```diff
@@ -0,0 +1,94 @@
...
+  private val clusterWeightProvider: ClusterWeightProvider,
+  private val dynamoCluster: DynamoCluster,
+) : AbstractIdleService() {
+  private val enhancedClient = DynamoDbEnhancedClient.builder()
+    .dynamoDbClient(ddb)
+    .build()
+  private val table = enhancedClient.table("misk-cluster-members", TABLE_SCHEMA)
+
+  override fun startUp() {
+    taskQueue.scheduleWithBackoff(timeBetweenRuns = Duration.ofSeconds(30)) {
```
jvmakine, line 41: We probably want to add the frequency to configuration eventually?

jvmakine (commented):
Seems very clean, nice!

misk-clustering-dynamodb/src/main/kotlin/misk/clustering/dynamo/DynamoClusterModule.kt
jvmakine: Ah, I commented later that we should have the pod_name as additional nonfunctional metadata in the table? For metrics and logs

damar-block (approved):

misk-clustering-dynamodb/src/main/kotlin/misk/clustering/dynamo/DynamoClusterWatcherTask.kt
```diff
@@ -0,0 +1,101 @@
...
+  ddb: DynamoDbClient,
+  private val clock: Clock,
+  private val clusterWeightProvider: ClusterWeightProvider,
+  private val dynamoCluster: DynamoCluster,
+  private val dynamoClusterConfig: DynamoClusterConfig,
+) : AbstractIdleService() {
+  private val enhancedClient = DynamoDbEnhancedClient.builder()
+    .dynamoDbClient(ddb)
+    .build()
+  private val table = enhancedClient.table("misk-cluster-members", TABLE_SCHEMA)
```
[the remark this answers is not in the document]
tso: good idea, i added a readme with the dynamodb specification (i THINK it's correct but need to verify)

jvmakine (approved):

PR #3016, opened 2023-11-22
moves misk-rate-limiting-bucket4j-redis to `testShardHibernate`

frojasg:
* this is a quick proof of concept to see if this makes the race conditions in ci to go away. Probably we want a better strategy to handle these


frojasg:
https://github.com/cashapp/misk/pull/3017#pullrequestreview-1743558256 is a better fix 🙏🏽

PR #3017, opened 2023-11-22
Introduce a redis shard to avoid parallel test problems

tso:


frojasg (approved):
LGTM! thanks for the fix!

PR #3023, opened 2023-11-29
Integrate JedisCluster for Redis Cluster support and implement DockerRedisCluster for testing

rainecp:
**Why?**

Redis Cluster enables horizontal scaling by distributing data and workload across multiple Redis nodes. If a master node fails, one of the slave nodes is promoted, ensuring uninterrupted service. With automatic cluster topology discovery and routing, it provides a more resilient architecture.

**What?**

This adds new classes to support cluster mode:

- `RedisClusterConfig` to configure the cluster endpoint
- `RedisClusterModule` to configure and instantiate JedisCluster
- `DockerRedisCluster` which creates a docker container with Redis Cluster for testing
- `RealRedisClusterTest` includes all existing unit tests for supported commands and new ones for distributed keys

`RealRedis` was also modified to support multi-key batch operations `mget`, `mset` and `del` by grouping keys by slot first before calling the appropriate `JedisCluster` command.

**Testing**

Added new test class for cluster mode that encapsulates existing tests for all supported redis commands. Also tested manually by publishing to maven local then using this version on another service that uses Redis. Verified `DockerRedisCluster` for existing tests and connection to a local Redis cluster.

### DIFF
--- a/misk-redis/src/main/kotlin/misk/redis/RealRedis.kt
+++ b/misk-redis/src/main/kotlin/misk/redis/RealRedis.kt
@@ -2,6 +2,7 @@ package misk.redis
 
 import okio.ByteString
 import okio.ByteString.Companion.toByteString
+import redis.clients.jedis.JedisCluster
 import redis.clients.jedis.JedisPooled
 import redis.clients.jedis.JedisPubSub
 import redis.clients.jedis.Pipeline
@@ -10,7 +11,9 @@ import redis.clients.jedis.UnifiedJedis
 import redis.clients.jedis.args.ListDirection
 import redis.clients.jedis.commands.JedisBinaryCommands
 import redis.clients.jedis.params.SetParams
+import redis.clients.jedis.util.JedisClusterCRC16
 import java.lang.reflect.InvocationHandler
+import java.lang.reflect.InvocationTargetException
 import java.lang.reflect.Method
 import java.lang.reflect.Proxy
 import java.time.Duration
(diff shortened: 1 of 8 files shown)


misk-redis/src/main/kotlin/misk/redis/RedisClusterConfig.kt
```diff
@@ -0,0 +1,21 @@
+package misk.redis
+
+import misk.config.Redact
+import redis.clients.jedis.JedisCluster.DEFAULT_MAX_ATTEMPTS
+import redis.clients.jedis.Protocol
+import wisp.config.Config
+
+class RedisClusterConfig : java.util.LinkedHashMap<String, RedisClusterReplicationGroupConfig>, Config {
```
adrw, line 8: Could you add some docs for what key/value represent? Also could/should this be a data class to match the simpler modelling of config other places?

rainecp: Good point, added more docs particularly in `RedisClusterReplicationGroupConfig` which is the data class. The one here is the top level config which I matched with `RedisConfig` (and DataSourceConfig) and they're not data class due to the constructors.

chris-ryan-square (approved):

jvmakine (approved):

misk-redis/src/main/kotlin/misk/redis/Redis.kt
```diff
@@ -226,6 +228,8 @@ interface Redis {
    * another client pushes to it or until timeout is reached. A timeout of zero can be used to block
    * indefinitely.
    *
+   * For Redis Cluster, source and destination must be in the same hash slot.
```
frojasg, line 231: What will happen if it's not? will this throw?

misk-redis/src/main/kotlin/misk/redis/RealRedis.kt
```diff
@@ -30,20 +33,64 @@ class RealRedis(
...
+    when(unifiedJedis) {
+      is JedisPooled -> {
+        val byteArrays = keyValues.map { it.toByteArray() }.toTypedArray()
+        return jedis { unifiedJedis.mset(*byteArrays) }
+      }
+      is JedisCluster -> {
+        // JedisCluster does not support multi-key mset, so we need to group by slot and perform mset for each slot
+        keyValues.toList().chunked(2).groupBy { JedisClusterCRC16.getSlot(it[0].toByteArray()) }
+          .forEach { (_, slotKeys) ->
+            jedis { unifiedJedis.mset(*slotKeys.flatten().map{it.toByteArray()}.toTypedArray())}
```
frojasg, line 88: Should we say something in the docs that this can trigger severals calls to redis?

misk-redis/src/main/kotlin/misk/redis/RealRedis.kt
```diff
@@ -30,20 +33,64 @@ class RealRedis(
...
-    val byteArrays = keyValues.map { it.toByteArray() }.toTypedArray()
-    jedis { mset(*byteArrays) }
+    when(unifiedJedis) {
+      is JedisPooled -> {
+        val byteArrays = keyValues.map { it.toByteArray() }.toTypedArray()
+        return jedis { unifiedJedis.mset(*byteArrays) }
+      }
+      is JedisCluster -> {
+        // JedisCluster does not support multi-key mset, so we need to group by slot and perform mset for each slot
+        keyValues.toList().chunked(2).groupBy { JedisClusterCRC16.getSlot(it[0].toByteArray()) }
```
frojasg, line 86: Maybe something that can be added in a following PR. Adding a tracing span here so we can identify all the redis call that were trigger by one `mset`

misk-redis/src/main/kotlin/misk/redis/RealRedis.kt
```diff
@@ -340,8 +387,12 @@ class RealRedis(
     private val clientMetrics: RedisClientMetrics,
   ) : InvocationHandler {
     override fun invoke(proxy: Any, method: Method, args: Array<out Any?>?): Any? =
-      clientMetrics.timed(method.name) {
-        method.invoke(jedisCommand, *(args ?: arrayOf()))
+      try {
+        clientMetrics.timed(method.name) {
+          method.invoke(jedisCommand, *(args ?: arrayOf()))
+        }
+      } catch (e: InvocationTargetException) {
```
frojasg, line 394: Why this was added?

frojasg (approved):
Non blocking comments. LGTM

misk-redis/src/main/kotlin/misk/redis/Redis.kt
rainecp: Yes, updated the docs.

misk-redis/src/main/kotlin/misk/redis/RealRedis.kt
rainecp: Updated

misk-redis/src/main/kotlin/misk/redis/RealRedis.kt
rainecp: Good idea, hoping this is a temp fix until Jedis handles this but can do this.

misk-redis/src/main/kotlin/misk/redis/RealRedis.kt
rainecp: For precise error handling and easier debugging since it wraps the exception thrown by the invoked method.

PR #3050, opened 2023-12-06
Start redis via gradle, not via MiskExternalDependency

tgregory-block:


samples/exemplar/src/test/kotlin/com/squareup/exemplar/RedisRateLimitedActionTests.kt
```diff
@@ -19,6 +20,7 @@ class RedisRateLimitedActionTests : AbstractRateLimitedActionTests() {
     override fun configure() {
       install(ExemplarTestModule())
       install(RedisBucket4jRateLimiterModule(DockerRedis.config, JedisPoolConfig(), useSsl = false))
+      install(RedisFlushModule())
```
adrw, line 23:
```suggestion
      install(RedisTestingModule())
```

I lean toward making it explicit in the module name that this should be installed for testing only. It also maps in functionality to JdbcTestingModule which includes similarly TruncateTablesService.

misk-redis/build.gradle.kts
```diff
@@ -28,6 +28,7 @@ dependencies {
   testFixturesApi(project(":misk-inject"))
   testFixturesApi(project(":misk-redis"))
   testFixturesApi(project(":misk-testing"))
+  testFixturesImplementation(project(":misk-service"))
```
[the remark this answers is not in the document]
tgregory-block: I think so, the compiler was yelling at me about not being able to import `ServiceModule` in `RedisFlushModule` without it

samples/exemplar/src/test/kotlin/com/squareup/exemplar/RedisRateLimitedActionTests.kt
tgregory-block: There's already a `RedisTestModule` that installs the `FakeRedis` in-memory redis. Do you think the similar naming would cause confusion? Could just add doc comments

rainecp (approved):

PR #3068, opened 2023-12-21
Support `watch` in DynamoDB-backed cluster

rainecp:
This updates `DynamoClusterWatcherTask` to use `DefaultCluster` instead of `DynamoCluster` so we could continue to support other functions like `watch` and `clusterChanged` that does not rely on the source of the cluster membership.


misk-clustering-dynamodb/src/main/kotlin/misk/clustering/dynamo/DynamoCluster.kt
```diff
@@ -6,6 +6,7 @@ import misk.clustering.Cluster.Snapshot
 import misk.clustering.ClusterHashRing
 import misk.clustering.ClusterWatch
 
+@Deprecated("Use DefaultCluster")
```
damar-block, line 9: I think since module is new and (probably?) not in use, we should just delete this file if we go that route?

misk-clustering-dynamodb/src/main/kotlin/misk/clustering/dynamo/DynamoClusterWatcherTask.kt
```diff
@@ -85,8 +87,8 @@ internal class DynamoClusterWatcherTask @Inject constructor(
           members.add(Member(item.name!!, "invalid-ip"))
         }
       }
-
-      dynamoCluster.update(members)
+      cluster.clusterChanged(membersBecomingReady = members, membersBecomingNotReady = prevMembers - members)
```
damar-block, line 90: do you know how `membersBecomingNotReady` are used? I am not sure it is correct to say that a previous member that that no longer exists is not ready. should we just always set `membersBecomingNotReady` to empty set?

misk-clustering-dynamodb/src/main/kotlin/misk/clustering/dynamo/DynamoCluster.kt
rainecp: Yes I was thinking we could also just delete this

misk-clustering-dynamodb/src/main/kotlin/misk/clustering/dynamo/DynamoClusterWatcherTask.kt
rainecp: it removes it from `readyMembers` otherwise it's still considered `ready`. Yea the variable names might be confusing in terms of how we're tracking it now in dynamoDB. But we should still remove it from the cluster, if it's not in the table anymore

misk-clustering-dynamodb/src/main/kotlin/misk/clustering/dynamo/DynamoCluster.kt
```diff
@@ -1,36 +0,0 @@
-package misk.clustering.dynamo
-
-import misk.clustering.Cluster
-import misk.clustering.Cluster.Member
-import misk.clustering.Cluster.Snapshot
-import misk.clustering.ClusterHashRing
-import misk.clustering.ClusterWatch
-
-class DynamoCluster(self: String) : Cluster {
```
lkerford, line 9: Do we have an idea of the scope of what downstream repos might not be compatible with this class being removed?

katukota (approved):

misk-clustering-dynamodb/src/main/kotlin/misk/clustering/dynamo/DynamoCluster.kt
rainecp: This was a recently added module and it's not being used yet.

jvmakine (approved):

PR #3074, opened 2024-01-08
Add wrapper for redis' zrange command

romilbm:
It will be easier to review this by the commits..
The first three commits are the important ones.
- 1st commit : Change sortedSet data structure in fake redis. Representing it as `HashMap<Double, HashSet<String>>>` which more closely represents how sorted set actually works in redis.
- 2nd commit: Add Interface of wrappers for zrange command.
- 3rd commit: The actual implementation of default (by index) and by score options of zrange.
- Rest of commits: Fixed linting, added documentation, added more tests.

### DIFF
--- a/misk-redis/src/main/kotlin/misk/redis/FakeRedis.kt
+++ b/misk-redis/src/main/kotlin/misk/redis/FakeRedis.kt
@@ -11,6 +11,9 @@ import java.time.Duration
 import java.time.Instant
 import java.util.concurrent.ConcurrentHashMap
 import jakarta.inject.Inject
+import misk.redis.Redis.ZRangeLimit
+import misk.redis.Redis.ZRangeMarker
+import misk.redis.Redis.ZRangeType
 import kotlin.math.max
 import kotlin.math.min
 import kotlin.random.Random
@@ -423,6 +426,28 @@ class FakeRedis : Redis {
   }
 
   override fun zscore(key: String, member: String): Double? {
+    throw NotImplementedError("Fake client not implemented for this operation")
+  }
+
+  override fun zrange(
+    key: String,
+    type: ZRangeType,
+    start: ZRangeMarker,
+    stop: ZRangeMarker,
+    reverse: Boolean,
+    limit: ZRangeLimit?,
+  ): List<ByteString?> {
+    throw NotImplementedError("Fake client not implemented for this operation")
+  }
+
+  override fun zrangeWithScores(
+    key: String,
+    type: ZRangeType,
+    start: ZRangeMarker,
+    stop: ZRangeMarker,
+    reverse: Boolean,
+    limit: ZRangeLimit?,
+  ): List<Pair<ByteString?, Double>> {
     throw NotImplementedError("Fake client not implemented for this operation")
   }
 }
--- a/misk-redis/src/main/kotlin/misk/redis/RealRedis.kt
+++ b/misk-redis/src/main/kotlin/misk/redis/RealRedis.kt
@@ -471,10 +471,6 @@ class RealRedis(
       ZRangeType.SCORE ->
         zrangeByScore(key, start as ZRangeScoreMarker, stop as ZRangeScoreMarker, reverse,
                       withScore, limit)
-
-      ZRangeType.LEX ->
-        throw RuntimeException("Unsupported UnifiedJedis implementation of BYLEX option for " +
-                                 "zrange command")
     }
   }
 
--- a/misk-redis/src/main/kotlin/misk/redis/RealRedis.kt
+++ b/misk-redis/src/main/kotlin/misk/redis/RealRedis.kt
@@ -494,31 +494,31 @@ class RealRedis(
       Either.Left(unifiedJedis.zrangeByScore(key.toByteArray(charset),
                                      minString.toByteArray(charset),
                                      maxString.toByteArray(charset)))
-    } else if (limit == null && !reverse && withScore) {
+    } else if (limit == null && !reverse) {
       Either.Right(unifiedJedis.zrangeByScoreWithScores(key.toByteArray(charset),
                                                         minString.toByteArray(charset),
                                                         maxString.toByteArray(charset)))
-    } else if (limit == null && reverse && !withScore) {
+    } else if (limit == null && !withScore) {
       Either.Left(unifiedJedis.zrevrangeByScore(key.toByteArray(charset),
                                              maxString.toByteArray(charset),
                                              minString.toByteArray(charset)))
-    } else if (limit == null && reverse && withScore){
+    } else if (limit == null){
       Either.Right(unifiedJedis.zrevrangeByScoreWithScores(key.toByteArray(charset),
                                                         maxString.toByteArray(charset),
                                                         minString.toByteArray(charset)))
-    } else if (limit != null && !reverse && !withScore) {
+    } else if (!reverse && !withScore) {
       Either.Left(unifiedJedis.zrangeByScore(key.toByteArray(charset),
                                              minString.toByteArray(charset),
                                              maxString.toByteArray(charset),
                                              limit.offset,
                                              limit.count))
-    } else if (limit != null && !reverse && withScore) {
+    } else if (!reverse) {
       Either.Right(unifiedJedis.zrangeByScoreWithScores(key.toByteArray(charset),
                                                         minString.toByteArray(charset),
                                                         maxString.toByteArray(charset),
                                                         limit.offset,
                                                         limit.count))
-    } else if (limit != null && reverse && !withScore) {
+    } else if (!withScore) {
       Either.Left(unifiedJedis.zrevrangeByScore(key.toByteArray(charset),
                                              maxString.toByteArray(charset),
                                              minString.toByteArray(charset),
@@ -528,7 +528,7 @@ class RealRedis(
       Either.Right(unifiedJedis.zrevrangeByScoreWithScores(key.toByteArray(charset),
                                                         maxString.toByteArray(charset),
                                                         minString.toByteArray(charset),
-                                                        limit!!.offset,
+                                                        limit.offset,
                                                         limit.count))
     }
 
--- a/misk-redis/src/main/kotlin/misk/redis/RealRedis.kt
+++ b/misk-redis/src/main/kotlin/misk/redis/RealRedis.kt
@@ -5,6 +5,11 @@ import misk.redis.Redis.ZAddOptions.GT
 import misk.redis.Redis.ZAddOptions.LT
 import misk.redis.Redis.ZAddOptions.NX
 import misk.redis.Redis.ZAddOptions.XX
+import misk.redis.Redis.ZRangeIndexMarker
+import misk.redis.Redis.ZRangeLimit
+import misk.redis.Redis.ZRangeMarker
+import misk.redis.Redis.ZRangeScoreMarker
+import misk.redis.Redis.ZRangeType
 import okio.ByteString
 import okio.ByteString.Companion.toByteString
 import redis.clients.jedis.JedisCluster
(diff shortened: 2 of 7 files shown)


misk-redis/src/main/kotlin/misk/redis/RealRedis.kt
```diff
@@ -1,10 +1,16 @@
 package misk.redis
 
+import arrow.core.Either
```
frojasg, line 3: Should be adding the arrow library for this type?

misk-redis/src/main/kotlin/misk/redis/Redis.kt
```diff
@@ -517,6 +517,144 @@ interface Redis {
...
+     * It is possible to specify an open interval.
+     *
+     * Use [ZRangeScoreMarker] to specify the start and stop for this type.
+     */
+    SCORE,
+
+    /**
+     * This is currently not supported.
+     */
+    LEX
```
frojasg, line 598: I wouldn't add this value if it's not supported

misk-redis/src/main/kotlin/misk/redis/RealRedis.kt
adrw: +1, would much rather use Kotlin standard library vs. pulling in an opinionated library for a simple type.

misk-redis/src/main/kotlin/misk/redis/FakeRedis.kt
```diff
@@ -425,4 +428,26 @@ class FakeRedis : Redis {
...
+
+  override fun zrange(
+    key: String,
+    type: ZRangeType,
+    start: ZRangeMarker,
+    stop: ZRangeMarker,
+    reverse: Boolean,
+    limit: ZRangeLimit?,
+  ): List<ByteString?> {
+    throw NotImplementedError("Fake client not implemented for this operation")
```
adrw, line 440: FakeRedis in testing seems to support, should the fake client also support this method?

misk-redis/src/main/kotlin/misk/redis/RealRedis.kt
```diff
@@ -423,6 +431,129 @@ class RealRedis(
...
+    stop: ZRangeScoreMarker,
+    reverse: Boolean,
+    withScore: Boolean,
+    limit: ZRangeLimit?,
+  ): Either<List<ByteArray?>, List<Tuple>> {
+    val minString = start.toString()
+    val maxString = stop.toString()
+
+    return if (limit == null && !reverse && !withScore) {
+      Either.Left(unifiedJedis.zrangeByScore(key.toByteArray(charset),
```
adrw, line 494: This indent formatting doesn't look idiomatic to Kotlin, also would prefer Tuple/Pair than introducing Either.

misk-redis/src/test/kotlin/misk/redis/AbstractRedisTest.kt
```diff
@@ -955,4 +960,1592 @@ abstract class AbstractRedisTest {
...
+    )
+  }
+
+  @Test fun `zrange by index test - stop index exceeds the size`() {
+    val key = "bar1"
+
+    redis.zadd(
+      key,
+      mapOf(
+        "b" to 5.0,
```
adrw, line 1068: Are all of these test fixtures different? If not, can the be collected to reduce the LOC size of each test case for better readability/maintainability?

If each text fixture set is unique then it's fine as is.

misk-redis/src/main/kotlin/misk/redis/Redis.kt
```diff
@@ -517,6 +517,144 @@ interface Redis {
     member: String
   ) : Double?
 
+  /**
+   * Returns the specified range of elements in the sorted set stored at [key].
```
adrw, line 521: Thanks for the docs!

buildSrc/src/main/kotlin/Dependencies.kt
```diff
@@ -1,6 +1,7 @@
 object Dependencies {
   val apacheCommonsLang3 = "org.apache.commons:commons-lang3:3.11"
   val apacheCommonsPool2 = "org.apache.commons:commons-pool2:2.11.1"
+  val arrowCore = "io.arrow-kt:arrow-core:1.2.1"
```
adrw, line 4:
```suggestion
```

I don't think this library is necessary for this change and warrants broader discussion before introducing to Misk.

adrw (approved):
LG other than use of Arrow. Thanks for the contribution!

buildSrc/src/main/kotlin/Dependencies.kt
romilbm: will remove.

misk-redis/src/test/kotlin/misk/redis/AbstractRedisTest.kt
romilbm: it is same in many but not all places.

misk-redis/src/main/kotlin/misk/redis/FakeRedis.kt
romilbm: FakeRedis is deprecated. None of the methods added after March 2023 are implemented here. 
```
@Deprecated(
  message = "Moved to misk-redis-testing. This Fake does not emulate redis correctly.",
  replaceWith = ReplaceWith("misk.redis.testing.FakeRedis")
)
```

romilbm:
> the ZRANGE method has a complicated signature. I wonder if we should avoid it in favor these other more specific methods.

I considered that. I wanted to draw parallels with the redis commands themselves. The specific query operations have been deprecated and replaced with a single zrange command with options in Redis iteself. My guess is jedis will deprecate the specific operations in a future version too. So I felt having a single method was a better way to model this.

misk-redis/src/main/kotlin/misk/redis/RealRedis.kt
romilbm: will remove.

misk-redis/src/main/kotlin/misk/redis/Redis.kt
```diff
@@ -517,6 +517,131 @@ interface Redis {
...
+   * The [limit] argument can be used to obtain a sub-range from the matching elements.
+   * See [ZRangeLimit] for more info.
+   */
+  fun zrange(
+    key: String,
+    type: ZRangeType = ZRangeType.INDEX,
+    start: ZRangeMarker,
+    stop: ZRangeMarker,
+    reverse: Boolean = false,
+    limit: ZRangeLimit? = null,
```
Rizziepit, line 545: Can we simply re-use the existing params class in Jedis? I worry that this reimplements a bunch of Jedis logic. This API should be very thin wrapper around Jedis
```suggestion
    params: ZRangeParams,
```

misk-redis/src/main/kotlin/misk/redis/Redis.kt
```diff
@@ -517,6 +517,131 @@ interface Redis {
...
+  /**
+   * This is similar to [zrange] but returns the scores along with the members.
+   */
+  fun zrangeWithScores(
+    key: String,
+    type: ZRangeType = ZRangeType.INDEX,
+    start: ZRangeMarker,
+    stop: ZRangeMarker,
+    reverse: Boolean = false,
+    limit: ZRangeLimit? = null,
```
Rizziepit, line 557: Ditto on reusing ZRangeParams
```suggestion
    params: ZRangeParams,
```

misk-redis/src/testFixtures/kotlin/misk/redis/testing/FakeRedis.kt
```diff
@@ -539,6 +568,146 @@ class FakeRedis @Inject constructor(
...
+  override fun zrange(
+    key: String,
+    type: ZRangeType,
+    start: ZRangeMarker,
+    stop: ZRangeMarker,
+    reverse: Boolean,
+    limit: ZRangeLimit?,
+  ): List<ByteString?> {
+    return zrangeWithScores(key, type, start, stop, reverse, limit)
+      .map { (member, _) ->  member }.toList()
```
Rizziepit, line 592: My guess is that we're not re-using `ZRangeParams` because it's not possible to inspect the parameters and provide a fake `zrange` implementation. I actually think this is preferred, i.e. throw NotImplemented in all the zrange functionality. The implementation is too complex for a fake. Folks should not be testing this stuff with a fake, they should be testing again Docker.

Rizziepit (changes requested):

misk-redis/src/main/kotlin/misk/redis/FakeRedis.kt
```diff
@@ -425,4 +428,26 @@ class FakeRedis : Redis {
   override fun zscore(key: String, member: String): Double? {
     throw NotImplementedError("Fake client not implemented for this operation")
   }
+
+  override fun zrange(
```
frojasg, line 432: Could we add some javadoc explaining why we don't implement these methods. We should nudge people to use the DockerRedis for testing if they need this.

misk-redis/src/main/kotlin/misk/redis/Redis.kt
romilbm: > Can we simply re-use the existing params class in Jedis?

ZRangeParams does not support use cases like include/exclude in ranges, infinite ranges, limits. For that, jedis also uses separate methods altogether.

misk-redis/src/main/kotlin/misk/redis/FakeRedis.kt
romilbm: There is already [doc](https://github.com/cashapp/misk/blob/master/misk-redis/src/main/kotlin/misk/redis/FakeRedis.kt#L19-L22) at the top of the class.

misk-redis/src/testFixtures/kotlin/misk/redis/testing/FakeRedis.kt
Rizziepit: Spinning up Docker slows down tests. So I think it's reasonable for folks to use a fake for trivial commands. But the z-family of commands... they should probably all be tested again Docker

misk-redis/src/main/kotlin/misk/redis/Redis.kt
Rizziepit: Do we need to support all of those? Which form will we use in Shadow?

It looks like it's possible to add a limit to `ZRangeParam`. Does that not work? Want to pair on this?
Screenshot 2024-01-09 at 12 17 21 PM

frojasg: maybe You can use a sealed class like this? 
```
sealed interface ZRangeParams 
data class ZRangeByScore(...) : ZRangeParams
data class ZRange(...) : ZRangeParams
```

misk-redis/src/testFixtures/kotlin/misk/redis/testing/FakeRedis.kt
Rizziepit: To be fair, there are some other `FakeRedis` methods that should throw NotImplemented but don't, e.g. [watch and unwatch](https://github.com/cashapp/misk/blob/29e65e5836bd6c973a98e2343dbd4a060f5cdab2/misk-redis/src/testFixtures/kotlin/misk/redis/testing/FakeRedis.kt#L402-L408). We need to have a bigger discussion about fake vs. Docker, and that's beyond the scope of this PR.

misk-redis/src/test/kotlin/misk/redis/AbstractRedisTest.kt
adrw: Would it be worth DRY-ing it up? I worry about maintenance of test files like this with unnecessary duplication and verbosity. Readability is also impacted vs. pulling most of the test fixtures into common variables so it's clear what is new in each test case.

adrw (approved):

frojasg (approved):
LGTM.

romilbm:
> Would it be worth DRY-ing it up? I worry about maintenance of test files like this with unnecessary duplication and verbosity. Readability is also impacted vs. pulling most of the test fixtures into common variables so it's clear what is new in each test case.

@adrw - Addressed this

PR #3075, opened 2024-01-09
Add service name to the config so that every service does not have to set it again.

katukota:
Every service creates a dynamo db with the service name appended to the table name. So, in the config table name will need to be updated in each table individually if not added here.


misk-clustering-dynamodb/src/test/kotlin/misk/clustering/DynamoClusterTest.kt
```diff
@@ -31,10 +31,10 @@ class DynamoClusterTest {
       install(FakeClusterWeightModule())
       install(
         InProcessDynamoDbModule(
-          DynamoDbTable("misk-cluster-members", DyClusterMember::class),
+          DynamoDbTable("$TEST_SERVICE_NAME.misk-cluster-members", DyClusterMember::class),
```
adrw, line 34: Is this change backwards compatible? Doesn't this change the location in Dynamo where Misk will look for the table?

katukota: It is, but it isn't being used in any repo currently.

rainecp (approved):

misk-clustering-dynamodb/src/main/kotlin/misk/clustering/dynamo/DynamoClusterConfig.kt
```diff
@@ -3,7 +3,8 @@ package misk.clustering.dynamo
 import wisp.config.Config
 
 data class DynamoClusterConfig @JvmOverloads constructor(
-  val table_name: String = "misk-cluster-members",
+  val appName: String = System.getenv("SERVICE_NAME") ?: "<invalid-service-name>",
```
adrw, line 6: This feels a bit weird/brittle vs. injecting `@AppName` which would work with existing service/test patterns.

PR #3078, opened 2024-01-09
Migrate misk from Dependencies.kt to a version catalog

staktrace:
This brings it in line with current idiomatic gradle dependency declarations.


adrw (approved):
Amazing

PR #3088, opened 2024-01-11
Allow disabling NotFoundAction via WebConfig

JGulbronson:
At Faire we want to be able to bind our own 404 action, and forward requests to another service. As discussed on Slack, neither the existing `NotFoundAction` or the `WebProxyAction` works for this case. Adding an option to not install the default action allows us to bind our own without having to worry about overlapping paths.


adrw (approved):

adrw:
You might need `gradle apiDump`

JGulbronson:
@adrw it says the change is breaking, but should be safe with the default parameter?

adrw:
End of the logs in the broken step it has the steps to solve, a bit annoying but it has caught actual breaking changes before.

```
gradle :misk:revapiAcceptAllBreaks --justification "{why this break is ok}"
```

PR #3089, opened 2024-01-12
Add wrappers for zremrangebyrank, zcard

romilbm:


frojasg (approved):

Rizziepit (approved):
Nice. Thank goodness these are simpler than `zrange`

PR #3091, opened 2024-01-12
Plumbing for MiskConcurrencyLimiterFeature

jclyne:
Adds an interface for a MiskConcurrencyLimiterFeature into the ConcurrencyLimitsInterceptor that allows for dynamically enabling/disabling enforcement. There is also an Optional Binding in the ConcurrencyLimitsModule, which defaults to an AlwaysEnabled implementation. The intent is that this can be overwritten in Skim, with a binding on an implementation which is backed by a LaunchDarkly flag.

This is to implement ADR0002

see: https://docs.google.com/document/d/15IQxBccR0MlfoxfTT9j51NaCxSvhAHsA37wAN-IMdhs/edit#heading=h.mpnrnewt2hso


frojasg (commented):
This is a big breaking change. It will break so many tests. I think we need to think through alternatives ways on how to introduce the feature flag  that mitigate the impact of users. 

Maybe `MiskConcurrencyLimiterEnabledFeature` should be an interface and have a simpler implementation by default?

misk/build.gradle.kts
```diff
@@ -68,6 +68,8 @@ dependencies {
   implementation(project(":misk-prometheus"))
   implementation(project(":misk-proto"))
   implementation(project(":misk-service"))
+  implementation(project(":misk-feature"))
+  runtimeOnly(Dependencies.jettyAlpnServerJava)
```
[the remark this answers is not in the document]
jclyne: misk-feature pulls in the feature flags interface

jclyne:
> This is a big breaking change. It will break so many tests. I think we need to think through alternatives ways on how to introduce the feature flag that mitigate the impact of users.
> 
> Maybe `MiskConcurrencyLimiterEnabledFeature` should be an interface and have a simpler implementation by default?

When you say it breaks so many tests, you are talking about all the changes for the unit tests? I'm not sure how to simplify that as each test has a TestModule and a binding is required for each of them

misk/src/main/kotlin/misk/MiskServiceModule.kt
```diff
@@ -29,6 +32,8 @@ class MiskRealServiceModule : KAbstractModule() {
     install(TickerModule())
     install(TokenGeneratorModule())
     install(MiskCommonServiceModule())
+    bind<MiskConcurrencyLimiterEnabledFeature>()
+      .to(RealMiskConcurrencyLimiterEnabledFeature::class.java)
```
jclyne, line 36: Should this be here? or should it go into Skim? If it were in skim, we could remove the dependency on misk-feature in the main misk module

frojasg: I like the idea to push it to `Skim` to remove the dependency here

frojasg (approved):
LGTM. Would you mind updating the PR details to reflect the change that was introduce, please?

PR #3095, opened 2024-01-15
Actions: Enable basic coroutine `WebAction`

Nava2:
The intention of this change is to make it viable to write [kotlinx.coroutines](https://kotlinlang.org/docs/coroutines-guide.html#0)-based actions. This change causes `WebAction` implementations written with `suspend` ingress points to properly execute. Without this change, accessing the current context will throw a confusing and misleading `NullPointerException`.

This change does not imply that there's no other issues with using `suspend` functions as ingress points, but it does allow experimentation.

### Reviewer questions

1. Should this be blocked behind a feature flag?
   * It is entirely opt-in by writing `suspend` actions, but also implies there is more support than exists.

### DIFF
--- a/misk/src/main/kotlin/misk/web/actions/WebActions.kt
+++ b/misk/src/main/kotlin/misk/web/actions/WebActions.kt
@@ -1,11 +1,14 @@
 package misk.web.actions
 
+import kotlinx.coroutines.CoroutineName
+import kotlinx.coroutines.runBlocking
 import misk.ApplicationInterceptor
 import misk.Chain
 import misk.web.HttpCall
 import misk.web.RealChain
 import kotlin.reflect.KFunction
 import kotlin.reflect.KParameter
+import kotlin.reflect.full.callSuspendBy
 
 fun WebAction.asChain(
   function: KFunction<*>,
@@ -25,8 +28,17 @@ fun WebAction.asChain(
         }
         argsMap[param] = arg
       }
-      return function.callBy(argsMap)
-        ?: throw IllegalStateException("Null return from WebAction")
+
+      return if (function.isSuspend) {
+        val coroutineName = CoroutineName(
+          "WebAction: ${this@asChain::class.simpleName}.${function.name}"
+        )
+        runBlocking(coroutineName) {
+          function.callSuspendBy(argsMap)
+        }
+      } else {
+        function.callBy(argsMap)
+      } ?: throw IllegalStateException("Null return from WebAction")
     }
   }
   val realChainInterceptors = interceptors + callFunctionInterceptor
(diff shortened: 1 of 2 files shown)


misk/src/main/kotlin/misk/web/actions/WebActions.kt
```diff
@@ -1,11 +1,14 @@
...
+import kotlinx.coroutines.runBlocking
 import misk.ApplicationInterceptor
 import misk.Chain
 import misk.web.HttpCall
 import misk.web.RealChain
 import kotlin.reflect.KFunction
 import kotlin.reflect.KParameter
+import kotlin.reflect.full.callSuspendBy
 
 fun WebAction.asChain(
```
Nava2, line 13: Aside: This function is only used in one place and it's internal, I wonder if it should be `internal`. It feels odd to be exposed as `public`. 

**This would break ABI compatibility.**

Nava2:
Looks like tests flaked 😢

adrw:
What's the motivating use case for using coroutines or writing suspend web actions?

Nava2:
> What's the motivating use case for using coroutines or writing suspend web actions?

General concurrency and it's a separate pattern to consider. Ktor for example is entirely built on top of coroutines.

Nava2:
> What's the motivating use case for using coroutines or writing suspend web actions?

@adrw part of the justification for this is that we'd like to experiment with coroutines at Faire. Another part is that I have some other work I'd like to integrate with coroutines. :) This opens a door, but closes none!

Nava2:
> Coroutines and non-blocking is a big departure from that, and given experiments in Cash I'm not sure that departure has much upside. 

Do you have references as to why this is? We have been considering using coroutines for parallelism due to their high support within Kotlin. 

> Some teams have experimented with it but then ended up just putting their entire action body in a runBlocking because sorting out async behaviour became too difficult conceptually and practically. 

Currently, that's the work around to make coroutines work in Misk today... 😬 What was the conceptual and practical hurdles that were seen? Happy to take this conversation somewhere else, too.

frojasg:
> What was the conceptual and practical hurdles that were seen? 

Coroutines is a different concurrency model that requires cooperative multitasking (coroutines have to voluntarily suspend execution to allow other coroutines to run) and we have run into unexpected/ugly failure mode when you use it in code that's blocking (non-cooperative). 

We're open to add support to coroutines in Misk, but it will require a lot of careful design work that needs to consider all the existing pieces that would need to change (Interceptors, ActionScope, JobHandler, outgoing rpc, etc) and the transition for code that's already using blocking systems and wants to introduce coroutines. It would be dangerous adding this type of features before that design work is done.

adrw:
I'm going to close this PR for now, but happy to discuss further in the community Slack and reconsider.

PR #3099, opened 2024-01-17
Allow users to configure Jetty output buffer size

adrw:
"We need this to have finer control over streaming / long polling
behaviours." - @Hexcles


Hexcles (approved):
Thanks!

PR #3104, opened 2024-01-22
[CCPOKR-8875] - Add minimal API module for sessions integration

mmollaverdi:
A incremental step towards a thin end-to-end implementation of a compatibility layer for Cash App's kotlin frameworks.


misk-api/src/main/kotlin/misk/api/RequestContext.kt
```diff
@@ -0,0 +1,16 @@
+package misk.api
+
+import misk.web.DispatchMechanism
+import okhttp3.HttpUrl
+
+interface RequestContext {
+  val url: HttpUrl
+  val dispatchMechanism: DispatchMechanism
```
mmollaverdi, line 8: Afaict, The `method` field in `DispatchMechanism` is the only thing used from DispatchMechanism across different interceptors in cash app repositories. I wonder I should replace this interface value with just a request method. 

I find `DispatchMechanism` a confusing concept given it captures methods or grpc/websocket.

misk-api/src/main/kotlin/misk/api/RequestContext.kt
```diff
@@ -0,0 +1,16 @@
...
+  val requestHeadersContext: RequestHeadersContext
+}
+
+interface RequestHeadersContext {
+  operator fun get(name: String): String?
+
+  fun asIterable(): Iterable<Pair<String, String>>
```
mmollaverdi, line 15: The Okhttp Headers class implements the `Iterable<Pair<String, String>>` interface.

misk/api/misk.api
```diff
@@ -1254,19 +1254,6 @@ public final class misk/web/CorsConfig {
 	public fun toString ()Ljava/lang/String;
 }
 
-public final class misk/web/DispatchMechanism : java/lang/Enum {
```
mmollaverdi, line 1257: Moved as is to misk-api

misk/src/main/kotlin/misk/web/BoundAction.kt
```diff
@@ -110,6 +111,7 @@ internal class BoundAction<A : WebAction>(
     val initialSeedData = mapOf<Key<*>, Any?>(
       keyOf<HttpServletRequest>() to request,
       keyOf<HttpCall>() to httpCall,
+      keyOf<RequestContext>() to httpCall,
```
mmollaverdi, line 114: Puts the `httpCall` instance into action scope with the `RequestContext` key in addition to the existing `HttpCall` key. This means cash libs can get a RequestContext instance out of action scope, instead of HttpCall, when they only need the fields provided by RequestContext.

misk-api/src/main/kotlin/misk/api/RequestContext.kt
jvmakine: I would prefer to keep `DispatchMechanism` still here, as it still could be useful. I would find just having `method` to be confusing as well with grpc

jvmakine (approved):

misk-api/src/main/kotlin/misk/api/RequestContext.kt
mmollaverdi: A nicer design from my perspective would be to have: 

```
sealed interface DispatchMechanism { // with a different name, so that it's not a braking change
  val method: Method
}

object Grpc : DispatchMechanism {
   override val method: Method = POST
}

object WebSocket : DispatchMechanism {
   override val method: Method = GET
}

data class Http(override val methd: Method): DispatchMechanism

enum Metod {
  // No grpc or websocket here
  GET,
  POST,
  PATCH,
  PUT,
  DELETE
}
```

We can make this change in a non-breaking change if we want. But I'm happy to come back to it later.

Even then though, I think having direct access to `method` in `RequestContext` is nice. Request method is a first-class concern for any http engine.

misk-api/src/main/kotlin/misk/api/RequestContext.kt
```diff
@@ -0,0 +1,16 @@
...
+import misk.web.DispatchMechanism
+import okhttp3.HttpUrl
+
+interface RequestContext {
+  val url: HttpUrl
+  val dispatchMechanism: DispatchMechanism
+  val requestHeadersContext: RequestHeadersContext
+}
+
+interface RequestHeadersContext {
```
mmollaverdi, line 12: This is to avoid having to create OKHttp headers out of Armeria headers. This interface is very easy to implement for both misk and Armeria.

misk-api/api/misk-api.api
```diff
@@ -0,0 +1,24 @@
...
+	public abstract fun getRequestHeadersContext ()Lmisk/api/RequestHeadersContext;
+	public abstract fun getUrl ()Lokhttp3/HttpUrl;
+}
+
+public abstract interface class misk/api/RequestHeadersContext {
+	public abstract fun asIterable ()Ljava/lang/Iterable;
+	public abstract fun get (Ljava/lang/String;)Ljava/lang/String;
+}
+
+public final class misk/web/DispatchMechanism : java/lang/Enum {
```
mmollaverdi, line 12: Moved out of the main misk module.

misk-api/src/main/kotlin/misk/api/RequestContext.kt
```diff
@@ -0,0 +1,16 @@
+package misk.api
+
+import misk.web.DispatchMechanism
+import okhttp3.HttpUrl
+
+interface RequestContext {
```
mmollaverdi, line 6: This new interface (implemented by `HttpCall`) will be used for existing code in cash libraries which only relies on a bunch of request attributes. This lets us incrementally implement the compat layer, but also decouples us from the many things which are currently packed into `HttpCall`, but not used in any cash libs afaict, such as web socket-related methods.  

Given `HttpCall` implements this interface, I've created a new action scope entry for `RequestContext` with the existing  `HttpCall` instance, which means code relying on `ActionScoped<HttpCall>`, but only using some request fields, can safely use `ActionScoped<RequestContext>` instead.

.palantir/revapi.yml
```diff
@@ -532,6 +532,9 @@ acceptedBreaks:
       justification: "Adding a new parameter to WebConfig"
   "2024.01.18.031247-621592e":
     com.squareup.misk:misk:
+    - code: "java.method.addedToInterface"
+      new: "method misk.api.RequestHeadersContext misk.web.HttpCall::getRequestHeadersContext()"
+      justification: "There is a default implementation for this meth[Bod in HttpCall"
```
adrw, line 537:
```suggestion
      justification: "There is a default implementation for this method in HttpCall"
```

misk-api/src/main/kotlin/misk/api/RequestContext.kt
adrw: If the `misk-api` project is already going to pull in OkHttp as an `api` dependency, I wonder if it'd be easier to just have an extension function that converts Armeria headers into OkHttp headers?

Adding new types everywhere really worries me about the long term maintainability of this effort, and in some cases seems a bit unnecessary.

misk-api/src/main/kotlin/misk/api/HttpRequest.kt
```diff
@@ -0,0 +1,18 @@
...
+import misk.web.DispatchMechanism
+import okhttp3.Headers
+import okhttp3.HttpUrl
+import org.jetbrains.annotations.ApiStatus
+
+/**
+ * Represents the current incoming request. Misk makes an instance of this interface available in the action scope
+ * with the `HttpRequest` key, for use in interceptors, other action scope providers, etc..
+ */
+@ApiStatus.Experimental
```
adrw, line 12: Thanks for adding this.

misk-api/src/main/kotlin/misk/api/HttpRequest.kt
```diff
@@ -0,0 +1,18 @@
...
+interface HttpRequest {
+  val url: HttpUrl
+  val dispatchMechanism: DispatchMechanism
+  /** HTTP request headers that may be modified via interception. */
```
adrw, line 16:
```suggestion
  /** HTTP request headers that may be modified by interceptors. */
```

misk/src/main/kotlin/misk/web/HttpCall.kt
```diff
@@ -24,15 +24,9 @@ sealed class SocketAddress {
 /**
  * A live HTTP call from a client for use by a chain of network interceptors.
  */
-interface HttpCall {
+interface HttpCall: HttpRequest {
 
-  /** Immutable information about the incoming HTTP request. */
```
adrw, line 29: I think this doc covers the first 3 parameters of the interface and should remain.

adrw (approved):

PR #3106, opened 2024-01-22
chore: update dependencies, resolve dependency issues.

autonomousapps:
* Updated several plugins:
    * Dependency Analysis Gradle Plugin, which required...
    * Kotlin to 1.9.x, which required...
    * Dokka to 1.9.10, and
    * Vanniktech Maven Publish plugin 0.27.0.
* The Kotlin update also necessitated some code changes. I'm not sure I resolved those in the best manner.

I then ran `buildhealth` (which didn't work before the above changes) and resolved the issues it reported. I punted on the `test` advice to keep the PR small. (Out of curiosity, why do you run `projectHealth` instead?)

There are some ABI changes. From a quick glance, it looks like they're a new method added to enum classes by newer Kotlin.

### DIFF
--- a/misk-grpc-reflect/build.gradle.kts
+++ b/misk-grpc-reflect/build.gradle.kts
@@ -32,7 +32,6 @@ dependencies {
   implementation(libs.kotlinLogging)
   implementation(libs.kotlinReflect)
   implementation(libs.okio)
-  implementation(libs.wireGrpcClient)
   implementation(libs.wireRuntime)
   implementation(libs.wireSchema)
   implementation(project(":wisp:wisp-logging"))
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -50,43 +50,45 @@ dependencyAnalysis {
   issues {
     all {
       ignoreSourceSet("testFixtures")
+      ignoreSourceSet("test")
       onAny {
         severity("fail")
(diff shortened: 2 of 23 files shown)


misk-grpc-reflect/build.gradle.kts
```diff
@@ -32,7 +32,7 @@ dependencies {
   implementation(libs.kotlinLogging)
   implementation(libs.kotlinReflect)
   implementation(libs.okio)
-  implementation(libs.wireGrpcClient)
+//  implementation(libs.wireGrpcClient)
```
jvmakine, line 35: Should we just remove these?

jvmakine (approved):

misk-grpc-reflect/build.gradle.kts
autonomousapps: yes for sure. I should have done that before converting from draft to ready for review.

Is there any documentation on how to update the ABI baselines?

jvmakine: I think you can run `gradle apiDump` and commit the changed `.api` files

jvmakine: The contributing guide should have more info here: https://github.com/cashapp/misk/blob/master/CONTRIBUTING.md#breaking-changes

autonomousapps:
@jvmakine I don't have authority to merge PRs into this repo, fyi.

PR #3115, opened 2024-01-25
[CCPOKR-7619] Move ActionScoped and ActionScopedProvider to misk-api

mmollaverdi:
So that they can be used agnostic of framework.


misk-admin/build.gradle.kts
```diff
@@ -34,6 +34,7 @@ dependencies {
   testImplementation(libs.kotlinTest)
   testImplementation(libs.wireRuntime)
   testImplementation(project(":misk-api"))
+  testImplementation(project(":misk-action-scopes"))
```
adrw, line 37: Is this import necessary/correct?

misk-admin/build.gradle.kts
```diff
@@ -18,7 +18,7 @@ dependencies {
   api(project(":wisp:wisp-config"))
   api(project(":wisp:wisp-deployment"))
   api(project(":misk"))
-  api(project(":misk-action-scopes"))
+  api(project(":misk-api"))
```
adrw, line 21: If `misk-action-scopes` already includes `misk-api` then is this necessary/correct?

build.gradle.kts
```diff
@@ -104,6 +104,13 @@ dependencyAnalysis {
         exclude("com.bucket4j:bucket4j-core")
       }
     }
+    project(":misk-action-scopes") {
+      onIncorrectConfiguration {
```
adrw, line 108: I haven't encountered this Gradle block before, how does it work? If it excludes `misk-api` does that mean sometimes we're silently failing configuration/tests for that project?

adrw (approved):

misk-admin/build.gradle.kts
mmollaverdi: yep, some tests in misk-admin use classes from action-scope and the `testImplementation` just wasn't needed because there was an `api` dependency on actions-scopes which has been replaced with misk-api now.

misk-admin/build.gradle.kts
mmollaverdi: This is addressing an issue reported by the dependency-analysis plugin Essentially misk-admin main (non-test) code now only uses classes from misk-api, and not misk-action-scopes and dependency analysis wants us to be explicit about that dependency rather than relying on a transtivive dep.

misk-admin/src/main/kotlin/misk/web/v2/DashboardIndexAction.kt
```diff
@@ -17,7 +17,6 @@ import misk.web.dashboard.DashboardTab
 import misk.web.dashboard.ValidWebEntry.Companion.slugify
 import misk.web.mediatype.MediaTypes
 import wisp.deployment.Deployment
-import wisp.deployment.deployments
```
mmollaverdi, line 20: This for some reason was imported twice.

build.gradle.kts
mmollaverdi: There is an existing one here
https://github.com/cashapp/misk/pull/3115/files#diff-c0dfa6bc7a8685217f70a860145fbdf416d449eaff052fa28352c5cec1a98c06L86

The error I was getting from the dependency-analysis plugin before adding this block was:
```
> Existing dependencies which should be modified to be as indicated:
    implementation(project(":misk-api")) (was api)
```

Essentially saying no misk-api classes are used in action-scope's public APIs. But for backwards compatibility, I want to have this an an API dependency (so that any projects depending on misk-action-scope and using the types moved to misk-api still compile).

PR #3116, opened 2024-01-26
Batched job handling

tgregory-block:
## Description
The goal of this PR is to enable "batched" job handling, i.e. for a job handler to be given all its messages at once rather than have them processed in their own threads. 

To enable this without changing how regular `JobHandler` processing is handled, we create a hierarchy for job handlers:

```mermaid
stateDiagram-v2
    AbstractJobHander --> JobHandler
    AbstractJobHander --> BatchedJobHandler
```

and switch the implementation based on which class has been bound. 

Review note: most of the new files in the `test` bundle have just been extracted from `FakeJobQueueTest.kt`, and are otherwise the same as before


tgregory-block:
Closing in favor of #3349

PR #3117, opened 2024-01-29
Fallback to action scope key subtypes, if a requested key is missing

mmollaverdi:
For example, given `HttpCall` extends `HttpRequest`, the value for `HttpCall` key will be returned when the `HttpRequest` key is requested (only if a `HttpRequest` key is not set).

See this PR for related prior work:
https://github.com/cashapp/misk/pull/3104

### DIFF
--- a/misk-action-scopes/src/main/kotlin/misk/scope/ActionScope.kt
+++ b/misk-action-scopes/src/main/kotlin/misk/scope/ActionScope.kt
@@ -183,6 +183,20 @@ class ActionScope @Inject internal constructor(
       return cachedValue as T
     }
 
+    // Search for any subtype of this key
+    // For example, given `HttpCall` extends `HttpRequest`, the value for `HttpCall` key
+    // will be returned when `HttpRequest` is requested (only if a `HttpRequest` key is not set)
+    val keyClass = key.typeLiteral.rawType
+    val subtypesThreadState = threadState.filterKeys {
+      it.annotation == key.annotation && keyClass.isAssignableFrom(it.typeLiteral.rawType)
+    }
+    if (subtypesThreadState.isNotEmpty()) {
+      val subtypeCachedValue = subtypesThreadState.values.toList()[0]
+      threadState[key] = subtypeCachedValue // cache this value
+      @Suppress("UNCHECKED_CAST")
+      return subtypeCachedValue as T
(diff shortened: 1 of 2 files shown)


jvmakine (approved):

misk/src/main/kotlin/misk/web/BoundAction.kt
```diff
@@ -111,7 +110,6 @@ internal class BoundAction<A : WebAction>(
     val initialSeedData = mapOf<Key<*>, Any?>(
       keyOf<HttpServletRequest>() to request,
       keyOf<HttpCall>() to httpCall,
-      keyOf<HttpRequest>() to httpCall,
```
adrw, line 114: If this is set here, does it error now with the changes to ActionScope?

adrw (approved):
Clever approach, I'm not anticipating any downstream breakages on account of this but would monitor to confirm after this merges.

misk/src/main/kotlin/misk/web/BoundAction.kt
mmollaverdi: nope, I'm only falling back to keys for subclasses if the key for the requested type is not set, so if a HttpRequest key is set, that'd still be returned.

misk-action-scopes/src/main/kotlin/misk/scope/ActionScope.kt
```diff
@@ -183,6 +183,20 @@ class ActionScope @Inject internal constructor(
...
+    // Search for any subtype of this key
+    // For example, given `HttpCall` extends `HttpRequest`, the value for `HttpCall` key
+    // will be returned when `HttpRequest` is requested (only if a `HttpRequest` key is not set)
+    val keyClass = key.typeLiteral.rawType
+    val subtypesThreadState = threadState.filterKeys {
+      it.annotation == key.annotation && keyClass.isAssignableFrom(it.typeLiteral.rawType)
+    }
+    if (subtypesThreadState.isNotEmpty()) {
+      val subtypeCachedValue = subtypesThreadState.values.toList()[0]
+      threadState[key] = subtypeCachedValue // cache this value
```
frojasg, line 195: nice!

frojasg (approved):
LGTM

PR #3121, opened 2024-01-30
Add a way to disable non error request logging

iwismer:
This makes changes to the request logging interceptor in order to allow disabling logging of non-error requests completely.

Currently, the logging interceptor does not allow you to set the rate limit to 0 as 0 is used to specify no limit. I didn't want to break backwards compatility to change it from 0 to -1 for no limit and 0 for no logging. Instead I've added a new config boolean that can be used to set it to not log for non-error responses.

The reason that I've added this is that a rate limit of 1 is too high for high volume APIs when there are a large number of pods handling requests. The rate limiting is done on a per pod basis, so a rate limit of 1/second with 100 pods will result in about 8.5million logs per day. I'd like a way to just turn these off!

### DIFF
--- a/misk-actions/src/main/kotlin/misk/web/interceptors/LogRequestResponse.kt
+++ b/misk-actions/src/main/kotlin/misk/web/interceptors/LogRequestResponse.kt
@@ -14,6 +14,8 @@ package misk.web.interceptors
  * If you would like to turn off rate limiting and emit all logs, set ratePerSecond and/or
  * errorRatePerSecond to 0.
  *
+ * If you would like to turn off logging for all non-error requests, set enableNonErrorLogging to false.
+ *
  * Percentage sampling is used to sample request and response bodies, with 0.0 for none and 1.0 for all.
  * Valid values are in the range [0.0, 1.0].
  *
@@ -34,4 +36,6 @@ annotation class LogRequestResponse(
   val errorBodySampling: Double = 0.0,
   /** which deploy environments will not have request/response logging enabled **/
   val excludedEnvironments: Array<String> = [],
+  /** By default log non-error responses **/
+  val enableNonErrorLogging: Boolean = true,
 )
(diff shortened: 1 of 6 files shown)


becassel (approved):

becassel:
Looks good to me, one strong potential alternative might be to use an enum logging level config (i.e. `ALL`, and `ERROR_ONLY`), but otherwise this accomplishes what we want, would defer to the Misk maintainers about their preferences (if any).

misk-actions/src/main/kotlin/misk/web/interceptors/LogRequestResponse.kt
```diff
@@ -34,4 +36,6 @@ annotation class LogRequestResponse(
   val errorBodySampling: Double = 0.0,
   /** which deploy environments will not have request/response logging enabled **/
   val excludedEnvironments: Array<String> = [],
+  /** By default log non-error responses **/
+  val enableNonErrorLogging: Boolean = true,
```
becassel, line 40: Consider using an enum instead, i.e. `val loggingLevel: LoggingLevel = LoggingLevel.ALL`, and then have another level like `ERROR_ONLY` which has your equivalent behaviour.

iwismer: Done!

misk-actions/src/main/kotlin/misk/web/interceptors/LogRequestResponse.kt
```diff
@@ -34,4 +37,17 @@ annotation class LogRequestResponse(
   val errorBodySampling: Double = 0.0,
   /** which deploy environments will not have request/response logging enabled **/
   val excludedEnvironments: Array<String> = [],
+  /** By default log non-error responses **/
+  val requestLoggingConstraints: RequestLoggingConstraints = RequestLoggingConstraints.ALL,
 )
+
+enum class RequestLoggingConstraints {
```
mhickman, line 44: what do you think of naming this `RequestLoggingMode` ?

iwismer: Naming this was hard, I'm happy to rename it to that.

frojasg (approved):
LGTM

PR #3128, opened 2024-02-03
Read gRPC status from response trailers to fix client metrics

rainecp:
Currently, our client metrics interceptor is not capturing the correct status codes for gRPC errors. It reads the `grpc-status` from the headers but Misk is sending it in the trailers from [ExceptionHandlingInterceptor](https://github.com/cashapp/misk/blob/1f9aab29f5a601e256dd1b21b0ec9b2c233fecae/misk/src/main/kotlin/misk/web/exceptions/ExceptionHandlingInterceptor.kt#L90). If it doesn't exist, it uses the response status code which defaults to 200. This could lead to missed alerts on critical issues.

We could also change `ExceptionHandlingInterceptor` to pass it in the header. But it is more common and standard to have the `grcp-status` information in the trailers. [wire](https://github.com/square/wire/blob/8ae99171c2e5262b1892acede3724fec53c50209/wire-grpc-client/src/jvmMain/kotlin/com/squareup/wire/internal/grpc.kt#L192) also reads the `grpc-status` from the response trailers first then the header. 

Tested against exemplar services: s2s between misk -> misk and misk->armeria

### DIFF
--- a/misk/src/main/kotlin/misk/client/ClientMetricsInterceptor.kt
+++ b/misk/src/main/kotlin/misk/client/ClientMetricsInterceptor.kt
@@ -6,16 +6,17 @@ import com.squareup.wire.GrpcMethod
 import com.squareup.wire.GrpcStatus
 import io.prometheus.client.Histogram
 import io.prometheus.client.Summary
+import jakarta.inject.Inject
+import jakarta.inject.Singleton
 import misk.metrics.backends.prometheus.PrometheusConfig
 import misk.metrics.v2.Metrics
 import okhttp3.Interceptor
 import okhttp3.Response
+import okhttp3.ResponseBody
 import retrofit2.Invocation
 import java.net.SocketTimeoutException
 import java.net.URL
 import java.util.concurrent.TimeUnit
-import jakarta.inject.Inject
-import jakarta.inject.Singleton
 
 class ClientMetricsInterceptor private constructor(
   val clientName: String,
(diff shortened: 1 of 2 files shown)


misk/src/main/kotlin/misk/client/ClientMetricsInterceptor.kt
```diff
@@ -66,6 +74,27 @@ class ClientMetricsInterceptor private constructor(
...
   }
 
+  private fun emitStatusCodeMetrics(
+    actionName: String,
+    elapsedMillis: Double,
+    readTrailers: Boolean,
+    response: Response
+  ) {
+    val grpcStatusHeader = if (readTrailers) {
+      /**
```
adrw, line 84: This should be inline doc style // within a function

adrw (approved):

misk/src/main/kotlin/misk/client/ClientMetricsInterceptor.kt
```diff
@@ -110,8 +142,26 @@ class ClientMetricsInterceptor private constructor(
...
+    )
   }
+
+  private class TrailerAwareResponseBody(
+    private val responseBody: ResponseBody,
+    private val func: () -> Unit
+  ) : ResponseBody() {
+    override fun contentLength() = responseBody.contentLength()
+    override fun contentType() = responseBody.contentType()
+    override fun source() = responseBody.source()
```
staktrace, line 158: I've never used it myself, but I think you should be able to use the kotlin delegation pattern (https://kotlinlang.org/docs/delegation.html) here. Then you wouldn't need these overrides that just delegate to `responseBody` as the compiler would generate them for you

staktrace (approved):
Nice, thanks!

misk/src/main/kotlin/misk/client/ClientMetricsInterceptor.kt
rainecp: Good catch, it was moved inside since it's only describing the trailers part. Updated it.

misk/src/main/kotlin/misk/client/ClientMetricsInterceptor.kt
rainecp: That's neat! But looks like it doesn't work with abstract classes, only for interfaces. I still renamed the property to convey the idea.

frojasg (approved):

PR #3129, opened 2024-02-05
Add Jetty Threadpool Readiness Health Checks

kevink-sq:
This change adds readiness health checks to the Jetty threadpool to signal the degraded state.

edit: had to run `gradle apiDump` to generate config artifacts similar to https://github.com/cashapp/misk/pull/2921/files and https://github.com/cashapp/misk/pull/3134/files

### DIFF
--- a/misk/src/main/kotlin/misk/web/MiskWebModule.kt
+++ b/misk/src/main/kotlin/misk/web/MiskWebModule.kt
@@ -19,6 +19,7 @@ import misk.api.HttpRequest
 import misk.concurrent.ExplicitReleaseDelayQueue
 import misk.exceptions.WebActionException
 import misk.grpc.GrpcFeatureBinding
+import misk.healthchecks.HealthCheck
 import misk.inject.KAbstractModule
 import misk.inject.toKey
 import misk.queuing.TimedBlockingQueue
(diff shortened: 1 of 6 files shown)


misk/src/main/kotlin/misk/web/MiskWebModule.kt
```diff
@@ -298,6 +300,7 @@ class MiskWebModule @JvmOverloads constructor(
       }
       bind<ThreadPool>().toInstance(threadPool)
       bind<MeasuredThreadPool>().toInstance(MeasuredThreadPoolExecutor(executor))
+      multibind<HealthCheck>().to<JettyThreadPoolHealthCheck>()
```
ThePumpingLemma, line 303: Don't we want this for `QueuedThreadPool` as well?

ThePumpingLemma: Also, maybe we should make this configurable so we can explicitly opt into it and test it out?

tirsen: yeah adding health checks can be a bit risky as they can cause flapping and sitting ducks, roll it out very carefully

kevink-sq: Made the health checks configurable and fixed to include `QueuedThreadPool`

ThePumpingLemma (approved):

PR #3131, opened 2024-02-06
Extract out misk client interfaces and classes into the api module for cross-framework support

mmollaverdi:
No breaking change in this PR. The classes/interfaces which have been extracted out are in the same package in the misk-api module, and the misk module has an api dependency on misk-api, which means consumers of misk should be able to use these types like before.


misk-api/src/main/kotlin/misk/client/ClientAction.kt
```diff
@@ -17,7 +17,7 @@ data class ClientAction(
   /** The return type fo the action */
   val returnType: KType
 ) {
-  internal constructor(clientName: String, method: KFunction<*>) :
```
mmollaverdi, line 20: We need to be able to create `ClientAction`s in Armeria services for cross-framework libs.

misk-api/src/main/kotlin/misk/client/NetworkInterceptorWrapper.kt
```diff
@@ -0,0 +1,12 @@
+package misk.client
+
+import okhttp3.Interceptor
+
+class NetworkInterceptorWrapper(
```
mmollaverdi, line 5: This is extracted out of misk as is (just in a new file)

PR #3133, opened 2024-02-07
Provide a `getIfInScope` on `ActionScoped` to avoid the need for injecting both ActionScope and ActionScoped

mmollaverdi:
In a bunch of places across cash-app repos, we check `actionScope.inScope`, before using an ActionScoped value (or before calling `actionScope.get(key)`). I've added a new method to `ActionScoped`, to return null if not in scope and return the scoped value otherwise, to remove the need for this extra check we currently manually do.

### DIFF
--- a/misk-action-scopes/src/main/kotlin/misk/scope/RealActionScoped.kt
+++ b/misk-action-scopes/src/main/kotlin/misk/scope/RealActionScoped.kt
@@ -8,4 +8,6 @@ internal class RealActionScoped<T> @Inject internal constructor(
   val scope: ActionScope
 ) : ActionScoped<T> {
   override fun get(): T = scope.get(key)
+
+  override fun getIfInScope(): T? = if (scope.inScope()) get() else null
 }
--- a/misk-api/build.gradle.kts
+++ b/misk-api/build.gradle.kts
@@ -1,6 +1,7 @@
 import com.vanniktech.maven.publish.JavadocJar.Dokka
 import com.vanniktech.maven.publish.KotlinJvm
 import com.vanniktech.maven.publish.MavenPublishBaseExtension
+import org.jetbrains.kotlin.gradle.tasks.KotlinCompile
 
 plugins {
   kotlin("jvm")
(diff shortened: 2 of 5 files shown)


misk-api/src/main/kotlin/misk/scope/ActionScoped.kt
```diff
@@ -4,6 +4,8 @@ package misk.scope
 interface ActionScoped<out T> {
   fun get(): T
 
+  fun getIfInScope(): T? = get()
```
mmollaverdi, line 7: Added a default return value, so that it's not a breaking change for any other classes implementing this interface across cash app repos.

frojasg (approved):
LGTM

misk-api/build.gradle.kts
```diff
@@ -19,3 +20,9 @@ configure<MavenPublishBaseExtension> {
     KotlinJvm(javadocJar = Dokka("dokkaGfm"))
   )
 }
+
+tasks.getting(KotlinCompile::class) {
+  compilerOptions {
+    freeCompilerArgs.add("-Xjvm-default=all-compatibility")
```
mmollaverdi, line 26: This tells the compiler to make the change binary compatible
https://kotlinlang.org/docs/java-to-kotlin-interop.html#all-compatibility

misk-action-scopes/src/main/kotlin/misk/scope/RealActionScoped.kt
```diff
@@ -8,4 +8,6 @@ internal class RealActionScoped<T> @Inject internal constructor(
   val scope: ActionScope
 ) : ActionScoped<T> {
   override fun get(): T = scope.get(key)
+
+  override fun getIfInScope(): T? = if (scope.inScope()) get() else null
```
adrw, line 12: I wonder whether getOrNull would be more idiomatic with Kotlin std lib naming of methods given the get in this case throws if not within a scope I think?

mmollaverdi: `getOrNull` is the name I initially also thought of. But this method only returns null if "not in scope", but throws an exception when it delegates to `get` if the key is not found. That'd the existing behaviour we need, and `getOrNull` didn't seemed like a misleading name for that.

PR #3136, opened 2024-02-07
Revert "Fallback to action scope key subtypes, if a requested key is missing

JGulbronson:
This reverts commit 880e488fb0dd114c4630143d340a57489d3ccbc9.

This totally breaks on generics and isn't backwards compatible.


adrw (approved):

JGulbronson:
@adrw looks like a flakey test that's unrelated

PR #3140, opened 2024-02-10
Correctly apply jetty configs to http2 servers

Hexcles:
Jetty configs were never applied to `HTTP2CServerConnectionFactory` before. And the SSL variant (`HTTP2ServerConnectionFactory`) only had one of the three params applied.

Create a new helper extension function to avoid code drift between HTTP2c and SSL.

### DIFF
--- a/misk/src/main/kotlin/misk/web/jetty/JettyService.kt
+++ b/misk/src/main/kotlin/misk/web/jetty/JettyService.kt
@@ -14,6 +14,7 @@ import misk.web.mediatype.MediaTypes
 import okhttp3.HttpUrl
 import org.eclipse.jetty.alpn.server.ALPNServerConnectionFactory
 import org.eclipse.jetty.http.UriCompliance
+import org.eclipse.jetty.http2.server.AbstractHTTP2ServerConnectionFactory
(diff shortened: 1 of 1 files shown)


adrw (approved):
Good catch!

Hexcles:
Not sure why the test is failing 🤔

JGulbronson (approved):
Would be great if we could follow this up with a test of some sort, but that might be more trouble than it's worth unless we expose the http connection factories somewhere to assert on.

JGulbronson:
My guess is that the test flaked because the log wasn't flushed. The code currently has a sleep, but that's not reliable. We hit the same thing at Faire with our log appenders in tests, I'll look if there's a more resilient fix.

Current code for reference
```kotlin
    override fun takeEvents(
        loggerClass: KClass<*>?,
        minLevel: Level,
        pattern: Regex?,
        consumeUnmatchedLogs: Boolean,
    ): List<ILoggingEvent> {
        sleep(100) // Give the logger some time to flush events.
        if (!consumeUnmatchedLogs) {
            return takeEventsNonConsuming(loggerClass, minLevel, pattern)
        }
        return takeEventsConsuming(loggerClass, minLevel, pattern)
    }
```

PR #3141, opened 2024-02-12
Add an ability to set the visbiility timeout on the message

din-cashapp:
**Summary:**
Similarly to the deadletter call - we can set the visibility timeout on the specific job making it invisible to the consumer for some time X. On each subsequent retry failure the delay will be longer mimicking exp backoff with jitter.

**Intended use:**
Example of the usage of this function:
```
fun handleJob(job: Job) {
    try {
          foo(job.body)
          job.acknowledge()
    catch(e: Exception) {
          job.setVisibilityTimeout()
    }
}

```
**Relevant documentation:** 
https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-visibility-timeout.html#changing-message-visibility-timeout


misk-aws/src/main/kotlin/misk/jobqueue/sqs/SqsJob.kt
```diff
@@ -55,6 +58,31 @@ internal class SqsJob(
...
+        ChangeMessageVisibilityRequest()
+          .withQueueUrl(queue.url)
+          .withReceiptHandle(message.receiptHandle)
+          .withVisibilityTimeout(visibilityTime)
+      )
+    }
+    metrics.visibilityTime.labels(queueName.value, queueName.value).set(visibilityTime.toDouble())
+  }
+
+  private fun calculateVisibilityTimeOut(maxReceiveCount: Int): Int {
```
din-cashapp, line 74: Ideally we would want to fetch the maxReceiveCount from the queue itself but it would mean making a call to fetch queue attributes so for simplicity sake I just let the users specify this number.  

I'm open to explore fetching this attribute somewhere in the queue setup process and providing it to the ResolvedQueue.

misk-aws/src/test/kotlin/misk/jobqueue/sqs/SqsJobQueueTest.kt
```diff
@@ -185,6 +185,62 @@ internal class SqsJobQueueTest {
...
+  fun setsVisibilityTimeoutOnTheMessage(receiverPolicy: String) {
+    setupReceiverPolicy(receiverPolicy)
+    val handledJobs = CopyOnWriteArrayList<Job>()
+
+    queue.enqueue(queueName, "this is my job")
+
+    val jobsReceived = AtomicInteger()
+    val allJobsCompleted = CountDownLatch(3)
+    consumer.subscribe(queueName) {
+      val sqsJob = it as SqsJob
```
din-cashapp, line 198: Decided to not update `Job` interface as different Misk clients could be extending it and visibilityTimeout seemed like an Sqs specific job feature.  
Open to discuss whether it would make sense to update `Job` interface.

misk-aws/src/main/kotlin/misk/jobqueue/sqs/SqsJob.kt
frojasg: that value can change after the value is retrieved. I feel this approach is good enough for now. it's relatively easy for oncall check these values if they're not seeing the behavior they're expecting.

misk-aws/src/test/kotlin/misk/jobqueue/sqs/SqsJobQueueTest.kt
frojasg: Maybe we can use a different name has `Job.delayForFailure()` or something like that. I feel that would be better rather than casing the real implementation.

misk-aws/src/main/kotlin/misk/jobqueue/sqs/SqsJob.kt
```diff
@@ -55,6 +58,36 @@ internal class SqsJob(
...
+    val currentReceiveCount = attributes["ApproximateReceiveCount"]?.toInt() ?: 1
+
+    val consecutiveRetryCount = (currentReceiveCount + 1).coerceAtMost(maxReceiveCount)
+    val backoff = 2.0.pow((consecutiveRetryCount - 1).toDouble()).toLong()
+    val maxDelay = 10 * 60 * 60L // We are limited with 12hrs so let's just limit it to 10hrs
+
+    val backoffWithJitter = maxDelay.coerceAtMost((backoff / 2 + Random.nextLong(0, backoff / 2)))
```
frojasg, line 86: Could we add a test to ensure we're not going over the 10 hours?

din-cashapp: Hmm this one would be a bit tricky to make - current job setup actually makes a call to the SQS.
I also don't want to expose this function making it public. I tried playing around with mocks but that didn't work well either.

misk-aws/src/main/kotlin/misk/jobqueue/sqs/SqsJob.kt
```diff
@@ -55,6 +58,36 @@ internal class SqsJob(
...
+    }
+    metrics.visibilityTime.labels(queueName.value, queueName.value).set(visibilityTime.toDouble())
+  }
+
+  private fun calculateVisibilityTimeOut(maxReceiveCount: Int): Int {
+    val currentReceiveCount = attributes["ApproximateReceiveCount"]?.toInt() ?: 1
+
+    val consecutiveRetryCount = (currentReceiveCount + 1).coerceAtMost(maxReceiveCount)
+    val backoff = 2.0.pow((consecutiveRetryCount - 1).toDouble()).toLong()
+    val maxDelay = 10 * 60 * 60L // We are limited with 12hrs after which SQS would thrown an exception so let's just put the highest timeout to 10hrs
```
hamdanjaveed, line 84: 👍

hamdanjaveed (approved):
LGTM, only thing I can think of is whether it's worth making it so folks could control the backoff behaviour but I think this works for now

misk-aws/src/main/kotlin/misk/jobqueue/sqs/SqsJob.kt
```diff
@@ -55,6 +58,36 @@ internal class SqsJob(
...
+          .withQueueUrl(queue.url)
+          .withReceiptHandle(message.receiptHandle)
+          .withVisibilityTimeout(visibilityTime)
+      )
+    }
+    metrics.visibilityTime.labels(queueName.value, queueName.value).set(visibilityTime.toDouble())
+  }
+
+  private fun calculateVisibilityTimeOut(maxReceiveCount: Int): Int {
+    val currentReceiveCount = attributes["ApproximateReceiveCount"]?.toInt() ?: 1
```
tgregory-block, line 80: nit - put `"ApproximateReceiveCount"` in a companion obj constant or enum

misk-aws/src/main/kotlin/misk/jobqueue/sqs/SqsJob.kt
```diff
@@ -55,6 +58,36 @@ internal class SqsJob(
...
+
+    val consecutiveRetryCount = (currentReceiveCount + 1).coerceAtMost(maxReceiveCount)
+    val backoff = 2.0.pow((consecutiveRetryCount - 1).toDouble()).toLong()
```
tgregory-block, line 83: Could we do `1L shl (consecutiveRetryCount - 1)`? Would avoid precision loss/casting and is much faster

misk-aws/src/main/kotlin/misk/jobqueue/sqs/SqsJob.kt
tgregory-block: You could modify the visibility timeout in the test itself, then pull from AWS again. The receive count will still be incremented, so when it fails again you'll get the next backoff amount

misk-aws/src/main/kotlin/misk/jobqueue/sqs/SqsJob.kt
tgregory-block, line 84: nit - move `maxDelay` into a companion object constant

misk-aws/src/main/kotlin/misk/jobqueue/sqs/SqsJob.kt
```diff
@@ -55,6 +58,36 @@ internal class SqsJob(
...
+    val maxDelay = 10 * 60 * 60L // We are limited with 12hrs after which SQS would thrown an exception so let's just put the highest timeout to 10hrs
+
+    val backoffWithJitter = maxDelay.coerceAtMost((backoff / 2 + Random.nextLong(0, backoff / 2)))
```
tgregory-block, line 86: nit - could replace `/ 2` with `shr 2`

misk-aws/src/main/kotlin/misk/jobqueue/sqs/SqsJob.kt
din-cashapp: hmm not sure about it being much faster as i think JVM optimizes a lot under the hood. it does look cleaner but im personally not a big fan of bit operations as it hinders readability

misk-aws/src/main/kotlin/misk/jobqueue/sqs/SqsJob.kt
din-cashapp: hmm i tried doing that but that would involve making receiptHandle on the sqs job visible, fetching the job queue url from the resolved queues and bunch of other changes. in order to make things simpler - i just moved the function into an object and added quick unit test

misk-aws/src/main/kotlin/misk/jobqueue/sqs/SqsJob.kt
```diff
@@ -85,5 +113,14 @@ internal class SqsJob(
     /** Client-assigned identifier, useful to detect duplicate messages. */
     const val JOBQUEUE_METADATA_IDEMPOTENCE_KEY = "idempotence_key"
     const val JOBQUEUE_METADATA_ORIGINAL_TRACE_ID = "original_trace_id"
+
+    /** Estimates the current visibility timeout*/
+    fun calculateVisibilityTimeOut(currentReceiveCount: Int, maxReceiveCount: Int): Int {
```
tgregory-block, line 118: nit `@VisibileForTesting`

tgregory-block (approved):

PR #3146, opened 2024-02-13
Add API for MDC

mmollaverdi:
Part of the cash app's cross-framework project

### DIFF
--- a/misk-api/src/main/kotlin/misk/logging/Mdc.kt
+++ b/misk-api/src/main/kotlin/misk/logging/Mdc.kt
@@ -1,0 +1,8 @@
+package misk.logging
+
+interface Mdc {
+  fun put(key: String, value: String?)
+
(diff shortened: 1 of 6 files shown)


misk-api/src/main/kotlin/misk/logging/Mdc.kt
```diff
@@ -0,0 +1,8 @@
+package misk.logging
+
+interface Mdc {
```
frojasg, line 3: do we need a method to clean?

mmollaverdi: We will eventually I guess. Let me add it.

mmollaverdi: Done

adrw (approved):

PR #3152, opened 2024-02-16
Expose out DockerSqs

aparajon:
`DockerSqs` is used internally by a Cash service that copies over the code directly, and this has resulted in code drift and some test issues. Exposing this class out will grant the ability to import an up-to-date version of this construct directly, and furtermore it may be of future use to other consumers who need to simulate local SQS testing.


ThePumpingLemma (commented):
I think this would need to be moved into an exported lib (e.g. a test fixture) for it to be accessible; we typically don't export test classes into lib jars.

aparajon:
I think we're likely heading in the direction of test containers for this one.

PR #3166, opened 2024-02-29
Make the HibernateExceptionMapper log level configurable through HibernateModule

qiyu-zhao:
Goal of this PR is to make the log level of Hibernate related exceptions configurable through `HibernateModule`

### Motivation
Recently we encountered some issue which resulted a `ConstraintViolationException`. Due to the existence of `ConstraintViolationExceptionMapper`, the exception was logged as `warn` instead of `error` which didn't trigger Sentry alert, which is something we want to avoid in the future


misk-hibernate/src/main/kotlin/misk/hibernate/HibernateExceptionMappers.kt
```diff
@@ -14,29 +14,35 @@ import org.slf4j.event.Level
 import jakarta.inject.Inject
 import javax.persistence.OptimisticLockException
 
-internal class RetryTransactionExceptionMapper @Inject internal constructor() :
+internal class RetryTransactionExceptionMapper @Inject internal constructor(
+  val config: HibernateExceptionLogLevelConfig? = null,
```
qiyu-zhao, line 18: I added default value to these injected dependencies to avoid `revapi` breaking change, but they seem redundant to me. I'm wondering if it makes more sense to make as an API breaking change instead?

misk-hibernate/src/main/kotlin/misk/hibernate/HibernateModule.kt
```diff
@@ -56,6 +56,7 @@ class HibernateModule @JvmOverloads constructor(
   private val readerQualifier: KClass<out Annotation>?,
   readerConfig: DataSourceConfig?,
   val databasePool: DatabasePool = RealDatabasePool,
+  logLevelConfig: HibernateExceptionLogLevelConfig? = null,
```
adrw, line 59:
```suggestion
  logLevelConfig: HibernateExceptionLogLevelConfig = HibernateExceptionLogLevelConfig(),
```

Should this default to the default config instead of null? Then don't need the null check on L70

adrw (approved):

misk-hibernate/src/main/kotlin/misk/hibernate/HibernateExceptionLogLevelConfig.kt
```diff
@@ -0,0 +1,13 @@
...
+
+import org.slf4j.event.Level
+import wisp.config.Config
+
+/**
+ * Configures the log [Level] for a HibernateException.
+ *
+ * @property log_level
+ */
+data class HibernateExceptionLogLevelConfig @JvmOverloads constructor(
```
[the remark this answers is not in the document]
qiyu-zhao: I don't think so, I tried to find one but the existing hibernate config is for `DataSouce` which I don't think this one should belong to

misk-hibernate/src/main/kotlin/misk/hibernate/HibernateModule.kt
```diff
@@ -92,6 +100,7 @@ class HibernateModule @JvmOverloads constructor(
     bind<Query.Factory>().to<ReflectionQuery.Factory>()
     bind<QueryLimitsConfig>()
       .toInstance(QueryLimitsConfig(MAX_MAX_ROWS, ROW_COUNT_ERROR_LIMIT, ROW_COUNT_WARNING_LIMIT))
+    bind<HibernateExceptionLogLevelConfig>().toInstance(logLevelConfig)
```
tgregory-block, line 103: This will break if multiple `HibernateModule`s are installed (e.g. a service has multiple databases, or wants to have multiple connection pools for some reason)

tgregory-block: I think we'll need to either add support for `.annotatedWith()` in the exception mapper, or wire this up somewhere else where it's explicitly made global and bound only once

qiyu-zhao: Did some researches about guice, `guice` can tolerate duplicated binding as long as the two bindings happened in the same level and their targets are identical, so unless someone try to bind different log level to each hibernateModule it should be fine.

tgregory-block: That's pretty unlikely to happen imo, @adrw are you good merging this as-is?

qiyu-zhao:
I'm planning to merge in this PR next monday if no objections

PR #3167, opened 2024-03-01
Add instructions to coordinate ld-sdk upgrades.

pnsq:
Comment-only PR. Adds some instructions for future devs.


gradle/libs.versions.toml
```diff
@@ -133,6 +133,13 @@ kotlinxCoroutines = { module = "org.jetbrains.kotlinx:kotlinx-coroutines-core",
 kotlinxHtml = { module = "org.jetbrains.kotlinx:kotlinx-html-jvm", version = "0.8.1" }
 kubernetesClient = { module = "io.kubernetes:client-java", version = "18.0.1" }
 kubernetesClientApi = { module = "io.kubernetes:client-java-api", version = "18.0.1" }
+# CAUTION: When updating launchdarkly SDK to a new version, ensure that:
```
adrianliu0815, line 136: Does this apply exclusively to major version upgrades (e.g., from 5.x to 6.x), or should we also follow the guidance for minor updates?

pnsq: It's probably good to follow this for all upgrades. There's no good reason to leave them on different versions with slight inconsistencies/bugs between them.

PR #3182, opened 2024-03-12
Only allow providers to be called at-most once per action scope

JGulbronson:
This changes the behavior of action-scoped providers, so that they are invoked at most once per scope. The current behavior propagates the ThreadLocal that contains key/values that have already been computed. However, if a provider hadn't been called yet then the value cannot be propagated to a new thread. Then the provider would get called once per thread, instead of once per original scope.

This propagates the already computed values, as well as the providers to new threads as a `Lazy`. This ensures that no matter how many threads the scope is propagated to, each provider is only called once. This behavior is tested [here](https://github.com/cashapp/misk/pull/3182/files#diff-812244514535aba81342e2e9cb42c660605e24537f91acc0d8cd530701e71c95R276).

There is also a subtle but beneficial change to an undocumented behavior. Because values are cached in a `Map`, if a provider returns null it cannot be cached, and will be called repeatedly until a non-null value is provided. With this change, that behavior no longer happens because the values are wrapped in a `Lazy`. The previously null value would now be stored correctly in the `Map` as a `Lazy(null)`. This means that a provider method will only be called at-most once, even if it returns a null value. I believe this is a beneficial change, and actually solves a pain point that we had to work around at Faire, which involved wrapping results in `Optional`.

### DIFF
--- a/misk-action-scopes/src/main/kotlin/misk/scope/ActionScope.kt
+++ b/misk-action-scopes/src/main/kotlin/misk/scope/ActionScope.kt
@@ -2,12 +2,12 @@ package misk.scope
 
 import com.google.inject.Key
 import com.google.inject.Provider
-import kotlinx.coroutines.asContextElement
-import java.util.UUID
-import java.util.concurrent.Callable
 import jakarta.inject.Inject
 import jakarta.inject.Singleton
 import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.asContextElement
+import java.util.UUID
+import java.util.concurrent.Callable
 import kotlin.coroutines.CoroutineContext
 import kotlin.reflect.KFunction
 import kotlin.reflect.KParameter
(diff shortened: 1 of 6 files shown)


misk-action-scopes/src/main/kotlin/misk/scope/ActionScope.kt
```diff
@@ -161,44 +191,51 @@ class ActionScope @Inject internal constructor(
...
+  }
 
-    // NB(mmihic): We don't use computeIfAbsent because computing the value of this
-    // key might require computing the values of keys on which we depend, which would
-    // cause recursive calls to computeIfAbsent which is unsupported (and explicitly
-    // detected in JDK 9+)
-    val threadState = threadLocalScope.get()
-    val cachedValue = threadState[key]
-    if (cachedValue != null) {
+  class Instance internal constructor(
```
swankjesse, line 211: So glad to have a class for this

swankjesse (approved):

PR #3187, opened 2024-03-14
Add All Metadata Action

adrw:
This will add a new endpoint and access annotation for exposing all service metadata. An example use case would be for an internal crawler service which aggregates metadata across a fleet of microservices and is able to conduct analysis or other tasks.


misk-admin/src/main/kotlin/misk/web/metadata/all/AllMetadataAccess.kt
```diff
@@ -0,0 +1,16 @@
...
+ * // Give internal scraping services access to all metadata for the service
+ * multibind<AccessAnnotationEntry>().toInstance(
+ *   AccessAnnotationEntry<AllMetadataAccess>(
+ *   services = listOf("internal_security_scraper_service"))
+ * )
+ * ```
+ */
+@Retention(AnnotationRetention.RUNTIME)
+@Target(AnnotationTarget.FUNCTION)
+annotation class AllMetadataAccess
```
AlexSzlavik, line 16: I think we actually want to use the https://github.com/cashapp/misk/blob/ee14b0c62db94432bac204810b786862df9e0ca9/misk-actions/src/main/kotlin/misk/security/authz/Authenticated.kt#L29 in practice. That's of course a deployment decision. Is it possible to bind AccessAnnotationEntries to that existing behaviour of AllowAnyService ?

misk-admin/src/main/kotlin/misk/web/metadata/config/ConfigMetadata.kt
```diff
@@ -0,0 +1,49 @@
+package misk.web.metadata.config
+
+import com.google.inject.Provider
```
AlexSzlavik, line 3: Assuming that this is a straight move from ConfigMetadataAction

AlexSzlavik (approved):

misk-admin/src/main/kotlin/misk/web/metadata/all/AllMetadataAccess.kt
adrw: Could you explain more the @Authenticated vs a custom AccessAnnotation? Why one over the other?

adrw: I imagine that it'd be preferable to have this locked down solely to an internal crawling service exclusively rather than any service?

AlexSzlavik: I don't think it should be used directly in misk, but we should retain the option of using those semantics for the AllMetadataAcceess annotation.

misk-admin/src/main/kotlin/misk/web/metadata/config/ConfigMetadata.kt
```diff
@@ -0,0 +1,49 @@
...
+  @Inject private lateinit var deployment: Deployment
+  @Inject private lateinit var config: wisp.config.Config
+  @Inject private lateinit var jvmMetadataAction: JvmMetadataAction
+  @Inject private lateinit var mode: ConfigMetadataAction.ConfigTabMode
+
+  override fun get() = ConfigMetadata(
+    resources = generateConfigResources(appName, deployment, config)
+  )
+
+  private fun generateConfigResources(
```
swankjesse, line 27: Honest question – is there a good way to get the actually-parsed config from the application bring-up, rather than re-implementing it? I don’t like gathering the metadata again here.

swankjesse (commented):
still going through this...

swankjesse (approved):
# 🚀

PR #3192, opened 2024-03-15
Expose minMessageToCompress for GrpcClientModule

mhickman:
This was added to wire gRPC clients in https://github.com/square/wire/pull/1956.
Without exposing this, wire gRPC clients default to compressing everything with
gRPC. Some gRPC server implementations don't support gzip, so this adds an
escape hatch (and also let's folks fine tune this setting for optimal performance,
should they want.)


swankjesse (approved):

adrw (commented):
LG

PR #3195, opened 2024-03-20
Encapsulate pipelines in misk-redis client

keeferrourke:
Add a `pipelining { }` method to misk redis, and deprecate the old `pipelined()` one.

**Why?**

1. `pipelined()` is broken in cluster-mode Redis, because the backing cluster implementation in Jedis does not return a `Pipeline`, it returns a `ClusterPipeline`, which are not compatible. Attempts to use `pipelined()` in a cluster-mode configuration will always result in a class cast exception.
2. `pipelined()` exposes the Jedis implementation details of misk-redis. This results in a subpar experience, as the Jedis APIs and misk-redis APIs are not compatible.

**How do I use this?**

Pipelining Redis commands can be done to improve efficiency, reducing round-trip time while talking to the redis server.

`DeferredRedis` commands always give back a Supplier, which can be read after the `pipelining` is complete.

```kt
val results = mutableListOf<Supplier<*>>
redis.piplining { 
  // This is a DeferredRedis. You can call all supported Redis commands on it.
  // Backing Jedis implementation details are 100% hidden in this scope.
  set("{customer:1}.name", "Luke SkyWalker")
  set("{customer:1}.team", "Light")
  incrBy("{customer:1}.movies_appearances", 9)
  results += mget("{customer:2}.name", "{customer:2}.team")
  // Note: It is an error to read the result of a command before the pipeline finishes.
  // ...
}
results.forEach { 
  it.get() 
  // Do something with a result.
  // Note that get() can throw, ex. if you do an incorrect operation on the wrong key type.
}
```

### DIFF
--- a/misk-redis/src/main/kotlin/misk/redis/FakeRedis.kt
+++ b/misk-redis/src/main/kotlin/misk/redis/FakeRedis.kt
@@ -394,8 +394,13 @@ class FakeRedis : Redis {
     throw NotImplementedError("Fake client not implemented for this operation")
   }
 
+  @Deprecated("Use pipelining instead.")
   override fun pipelined(): Pipeline {
     throw NotImplementedError("Fake client not implemented for this operation")
+  }
+
+  override fun pipelining(block: DeferredRedis.() -> Unit) {
+    throw NotImplementedError("Use the fake from misk.redis.testing instead.")
   }
 
   override fun close() {
--- a/misk-redis/src/main/kotlin/misk/redis/JedisPipeline.kt
+++ b/misk-redis/src/main/kotlin/misk/redis/JedisPipeline.kt
@@ -1,0 +1,78 @@
+package misk.redis
+
+import okio.Closeable
+import redis.clients.jedis.ClusterPipeline
+import redis.clients.jedis.Pipeline
+import redis.clients.jedis.Response
+import redis.clients.jedis.args.FlushMode
+import redis.clients.jedis.args.FunctionRestorePolicy
+import redis.clients.jedis.commands.PipelineBinaryCommands
+import redis.clients.jedis.commands.PipelineCommands
+
+internal sealed class JedisPipeline : Closeable, PipelineCommands, PipelineBinaryCommands {
+  class ClusterJedisPipeline(private val pipeline: ClusterPipeline) : JedisPipeline(),
+    Closeable by pipeline, PipelineCommands by pipeline, PipelineBinaryCommands by pipeline {
+
+    // The following methods must be overridden because the class inherits multiple implementations.
+
+    override fun functionDump(): Response<ByteArray> {
+      return pipeline.functionDump()
+    }
+
+    override fun functionKill(): Response<String> {
+      return pipeline.functionKill()
+    }
+
+    override fun functionFlush(): Response<String> {
+      return pipeline.functionFlush()
+    }
+
+    override fun functionFlush(mode: FlushMode?): Response<String> {
+      return pipeline.functionFlush(mode)
+    }
+
+    override fun functionRestore(serializedValue: ByteArray?): Response<String> {
+      return pipeline.functionRestore(serializedValue)
+    }
+
+    override fun functionRestore(
+      serializedValue: ByteArray?,
+      policy: FunctionRestorePolicy?
+    ): Response<String> {
+      return pipeline.functionRestore(serializedValue, policy)
+    }
+  }
+
+  class PooledPipeline(private val pipeline: Pipeline) : JedisPipeline(),
+    Closeable by pipeline, PipelineCommands by pipeline, PipelineBinaryCommands by pipeline {
+
+    // The following methods must be overridden because the class inherits multiple implementations.
+
+    override fun functionDump(): Response<ByteArray> {
+      return pipeline.functionDump()
+    }
+
+    override fun functionKill(): Response<String> {
+      return pipeline.functionKill()
+    }
+
+    override fun functionFlush(): Response<String> {
+      return pipeline.functionFlush()
+    }
+
+    override fun functionFlush(mode: FlushMode?): Response<String> {
+      return pipeline.functionFlush(mode)
+    }
+
+    override fun functionRestore(serializedValue: ByteArray?): Response<String> {
+      return pipeline.functionRestore(serializedValue)
+    }
+
+    override fun functionRestore(
+      serializedValue: ByteArray?,
+      policy: FunctionRestorePolicy?
+    ): Response<String> {
+      return pipeline.functionRestore(serializedValue, policy)
+    }
+  }
+}
--- a/misk-redis/src/main/kotlin/misk/redis/RealPipelinedRedis.kt
+++ b/misk-redis/src/main/kotlin/misk/redis/RealPipelinedRedis.kt
@@ -2,6 +2,7 @@ package misk.redis
 
 import okio.ByteString
 import okio.ByteString.Companion.toByteString
+import redis.clients.jedis.Jedis
 import redis.clients.jedis.Response
 import redis.clients.jedis.args.ListDirection
 import redis.clients.jedis.params.SetParams
@@ -41,8 +42,20 @@ internal class RealPipelinedRedis(private val pipeline: JedisPipeline) : Deferre
 
   override fun del(vararg keys: String): Supplier<Int> {
     val keysBytes = keys.map { it.toByteArray(charset) }.toTypedArray()
-    val response = pipeline.del(*keysBytes)
-    return Supplier { response.get().toInt() }
+    val responses = when (pipeline) {
+      is JedisPipeline.PooledPipeline -> listOf(pipeline.del(*keysBytes))
+      is JedisPipeline.ClusterJedisPipeline -> {
+        keysBytes.groupBy { JedisClusterCRC16.getSlot(it) }
+          .map { (_, slottedKeys) ->
+            pipeline.del(*slottedKeys.toTypedArray())
+          }
+      }
+    }
+    return Supplier {
+      responses.fold(0) { acc, response ->
+        acc + response.get().toInt()
+      }
+    }
   }
 
   override fun mget(vararg keys: String): Supplier<List<ByteString?>> {
--- a/misk-redis/src/main/kotlin/misk/redis/RealPipelinedRedis.kt
+++ b/misk-redis/src/main/kotlin/misk/redis/RealPipelinedRedis.kt
@@ -1,0 +1,522 @@
+package misk.redis
+
+import okio.ByteString
+import okio.ByteString.Companion.toByteString
+import redis.clients.jedis.args.ListDirection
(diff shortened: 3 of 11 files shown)


misk-redis/src/main/kotlin/misk/redis/FakeRedis.kt
keeferrourke, line 1: This fake needs to be deleted in a follow-up. It should be long dead.

misk-redis/src/main/kotlin/misk/redis/JedisPipeline.kt
```diff
@@ -0,0 +1,78 @@
...
+import redis.clients.jedis.ClusterPipeline
+import redis.clients.jedis.Pipeline
+import redis.clients.jedis.Response
+import redis.clients.jedis.args.FlushMode
+import redis.clients.jedis.args.FunctionRestorePolicy
+import redis.clients.jedis.commands.PipelineBinaryCommands
+import redis.clients.jedis.commands.PipelineCommands
+
+sealed class JedisPipeline : Closeable, PipelineCommands, PipelineBinaryCommands {
+  class ClusterJedisPipeline(private val pipeline: ClusterPipeline) : JedisPipeline(),
```
keeferrourke, line 13: Jedis / UnifiedJedis is a little annoying as there is no unified Pipeline interface.

We have to support JedisClustered and JedisPooled. So... I unified it.

misk-redis/src/main/kotlin/misk/redis/RealPipelinedRedis.kt
keeferrourke, line 1: This implementation is largely copied from RealRedis. There are opportunities to share some implementation details, but I haven't gotten that far yet.

misk-redis/src/main/kotlin/misk/redis/RealRedis.kt
```diff
@@ -357,10 +357,22 @@ class RealRedis(
   // pipelined() returns the jedis to the pool, despite returning a Pipeline that holds a reference
   // to the borrowed jedis connection.
   // This is a bug, and will be fixed in a follow-up.
+  @Deprecated("Use pipelining instead.")
   override fun pipelined(): Pipeline {
     return unifiedJedis.pipelined() as Pipeline
   }
```
keeferrourke, line 363: This is actually just entirely broken for clustered redis and will result in a class cast exception if you attempt to use it as is.

misk-redis/src/main/kotlin/misk/redis/RealRedis.kt
```diff
@@ -382,38 +394,22 @@ class RealRedis(
...
     return unifiedJedis.zadd(
       key.toByteArray(charset),
       score,
       member.toByteArray(charset),
-      getZAddParams(options)
+      ZAddOptions.getZAddParams(options)
     )
   }
 
-  private fun getZAddParams(options: Array<out Redis.ZAddOptions>): ZAddParams {
```
keeferrourke, line 395: I just moved this to the ZAddOptions companion.

misk-redis/src/main/kotlin/misk/redis/Redis.kt
```diff
@@ -721,3 +753,144 @@ inline fun checkHrandFieldCount(count: Long) {
     "You must request at least 1 field."
   }
 }
+
+/**
+ * Like [Redis], but returns [Supplier]s to defer value retrieval.
+ * Does not support transactions or pubsub.
+ */
+interface DeferredRedis {
```
keeferrourke, line 761: Open to better naming. Maybe it should just be called a Pipeline...? idk.

misk-redis/src/main/kotlin/misk/redis/Redis.kt
```diff
@@ -721,3 +753,144 @@ inline fun checkHrandFieldCount(count: Long) {
...
+  fun del(key: String): Supplier<Boolean>
```
keeferrourke, line 762: I originally thought of making these suspend funs that return Deferred, but I think Suppliers are better for misk's threading model?

Would appreciate other's thoughts here.

misk-redis/src/test/kotlin/misk/redis/AbstractRedisTest.kt
```diff
@@ -765,6 +766,29 @@ abstract class AbstractRedisTest {
...
+  fun `pipelining works`() {
+    val suppliers = mutableListOf<Supplier<*>>()
+    redis.pipelining {
+      suppliers.addAll(
+        listOf(
+          set("key7", "value7".encodeUtf8()),
+          get("key7"),
+          hset("key8", "field8", "value8".encodeUtf8()),
+          hget("key8", "field8"),
+          hget("key8", "field9"),
```
keeferrourke, line 779: TODO: Exhaustively test every command in the interface.

misk-redis/src/main/kotlin/misk/redis/RealPipelinedRedis.kt
```diff
@@ -0,0 +1,518 @@
...
+import redis.clients.jedis.params.ZRangeParams
+import redis.clients.jedis.resps.Tuple
+import java.time.Duration
+import java.util.function.Supplier
+
+internal class RealPipelinedRedis(private val pipeline: JedisPipeline) : DeferredRedis {
+  override fun del(key: String): Supplier<Boolean> {
+    val keyBytes = key.toByteArray(charset)
+    val response = pipeline.del(keyBytes)
+    return Supplier { response.get() == 1L }
```
keeferrourke, line 16: Note to self: pipeline response object are _actually_ suppliers. I don't think we can get around wrapping them, however, since we often don't use the raw response types.

misk-redis/src/main/kotlin/misk/redis/RealRedis.kt
```diff
@@ -357,10 +356,22 @@ class RealRedis(
...
   }
 
+  override fun pipelining(block: DeferredRedis.() -> Unit) {
+    val pipeline: JedisPipeline = when (unifiedJedis) {
+      is JedisPooled -> JedisPipeline.PooledPipeline(unifiedJedis.pipelined() as Pipeline)
+      is JedisCluster -> JedisPipeline.ClusterJedisPipeline(unifiedJedis.pipelined() as ClusterPipeline)
+      else -> throw RuntimeException("Unsupported UnifiedJedis implementation ${unifiedJedis.javaClass}")
+    }
+    pipeline.use {
+      block(RealPipelinedRedis(pipeline))
```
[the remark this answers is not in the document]
keeferrourke: the `close()` method in both impls called `sync` :)

misk-redis/src/main/kotlin/misk/redis/Redis.kt
```diff
@@ -721,3 +760,140 @@ inline fun checkHrandFieldCount(count: Long) {
...
+
+  fun hincrBy(key: String, field: String, increment: Long): Supplier<Long>
+
+  fun hrandFieldWithValues(key: String, count: Long): Supplier<Map<String, ByteString>?>
+
+  fun hrandField(key: String, count: Long): Supplier<List<String>>
+
+  fun set(key: String, value: ByteString, expiryDuration: Duration? = null): Supplier<Unit>
+
+  fun setnx(key: String, value: ByteString, expiryDuration: Duration? = null): Supplier<Boolean>
```
keeferrourke, line 797: Note: set/setnx are intentionally and slightly different from the main Redis API. 

`set` is not an operator, because it is not ergonomic to use it as an operator within a `pipelining` block.

Expiry dates are always optional with `set`, so they are moved to be the last argument with a default of no expiry.

misk-redis/src/test/kotlin/misk/redis/AbstractRedisTest.kt
```diff
@@ -1289,6 +1311,196 @@ abstract class AbstractRedisTest {
...
+    redis.pipelining {
+      suppliers.addAll(
+        listOf(
+          mset(
+            "mkey1".encodeUtf8(),
+            "mval1".encodeUtf8(),
+            "mkey2".encodeUtf8(),
+            "mval2".encodeUtf8()
+          ),
+          mget("mkey1", "mkey2"),
```
keeferrourke, line 1358: nb: in clustered mode, the underlying impl actually makes two `mset` and two `mget` calls. This keeps the slotting behaviour opaque to the caller (matching RealRedis).

misk-redis/src/main/kotlin/misk/redis/RealPipelinedRedis.kt
```diff
@@ -0,0 +1,576 @@
...
+    return Supplier { response.get() }
+  }
+
+  override fun setnx(key: String, value: ByteString, expiryDuration: Duration?): Supplier<Boolean> {
+    val keyBytes = key.toByteArray(charset)
+    val valueBytes = value.toByteArray()
+    val params = SetParams().nx().apply {
+      if (expiryDuration != null) px(expiryDuration.toMillis())
+    }
+    val response = pipeline.set(keyBytes, valueBytes, params)
```
keeferrourke, line 173: For `pipelining`, `setnx` is actually backed by `SET` with `NX` and optional `PX` arguments. This is because `SETNX` is deprecated as of Redis 2.6 and I don't want to support this with pipelining.

We should follow up and change the [implementation](https://github.com/cashapp/misk/blob/master/misk-redis/src/main/kotlin/misk/redis/RealRedis.kt#L192) of the `setnx` method without expiry durations to _also_ be backed by `SET` with the `NX` parameters.

keeferrourke:
Rebased and updated to add `getDel` from https://github.com/cashapp/misk/pull/3196

misk-redis/src/main/kotlin/misk/redis/RealPipelinedRedis.kt
mslogar-squareup: Non-blocking, possibly for a followup PR: Would be tricky to write `RealPipelinedRedis` to use the `RealRedis` implementation... but could we rewrite `RealRedis` to call `RealPipelinedRedis` with single operations? I don't love that there's this much duplication especially since there's actual business logic involved here.

mslogar-squareup (approved):

misk-redis/src/test/kotlin/misk/redis/AbstractRedisTest.kt
```diff
@@ -1278,6 +1324,203 @@ abstract class AbstractRedisTest {
...
+          get("test key"),
+          setnx("keynx", "valuenx".encodeUtf8()),
+          setnx("keynx", "valuenx2".encodeUtf8()),
+          get("keynx"),
+          incr("incr"),
+          incrBy("incr", 3),
+          get("incr"),
+          del("test key"),
+          getDel("keynx"),
+          get("keynx"),
```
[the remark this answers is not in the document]
keeferrourke: See [2dc4cbc](https://github.com/cashapp/misk/pull/3195/commits/2dc4cbc1dc8339edc6a626067360dffd935595df)

PR #3196, opened 2024-03-20
Expose redis.GETDEL

mslogar-squareup:
See https://redis.io/commands/getdel/ for docs


mhickman (approved):

adrw (approved):

keeferrourke (commented):
I know this is basic, but I'd like to see at least a happy-path test for this in `AbstractRedisTest` so we can be sure that this works as expected in pooled and cluster variants of Jedis configuration.

keeferrourke (approved):
Great!

PR #3198, opened 2024-03-21
Simplify misk-redis implementation

keeferrourke:
...by deferring to `RealPipelinedRedis`.

Follow up for https://github.com/cashapp/misk/pull/3195.

This change deduplicates a lot of logic between `RealRedis` and `RealPipelinedRedis`, by always pipelining commands.

This incurs a little overhead since we have to create, sync, and close a pipeline every time (even for just one command), but significantly simplifies the client. I also removed the old timing proxy, so hopefully that compensates for the added overhead of the pipelines.

---

I recommend reviewing this change commit-by-commit.

---

NB: the `z*` methods from #3195 which were removed in #3199 have been added back as part of this change. This is because the underlying bug that caused wrong results on some zrange operations has been resolved by upgrading to Jedis 5.

### DIFF
--- a/.palantir/revapi.yml
+++ b/.palantir/revapi.yml
@@ -6192,6 +6192,38 @@ acceptedBreaks:
       old: "parameter void misk.redis.RealPipelinedRedis::<init>(===misk.redis.JedisPipeline===)"
       new: "parameter void misk.redis.RealPipelinedRedis::<init>(===redis.clients.jedis.AbstractPipeline===)"
       justification: "RealPipelinedRedis is internal"
+  "2024.03.26.112919-01b20b7":
+    com.squareup.misk:misk-redis:
+    - code: "java.method.addedToInterface"
+      new: "method java.util.function.Supplier<java.lang.Double> misk.redis.DeferredRedis::zscore(java.lang.String,\
+        \ java.lang.String)"
+      justification: "Additive change only"
+    - code: "java.method.addedToInterface"
+      new: "method java.util.function.Supplier<java.lang.Long> misk.redis.DeferredRedis::zadd(java.lang.String,\
+        \ double, java.lang.String, misk.redis.Redis.ZAddOptions[])"
+      justification: "Additive change only"
+    - code: "java.method.addedToInterface"
+      new: "method java.util.function.Supplier<java.lang.Long> misk.redis.DeferredRedis::zadd(java.lang.String,\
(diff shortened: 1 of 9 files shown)


.palantir/revapi.yml
```diff
@@ -6099,6 +6099,11 @@ acceptedBreaks:
       new: "method void misk.redis.Redis::pipelining(kotlin.jvm.functions.Function1<?\
         \ super misk.redis.DeferredRedis, kotlin.Unit>)"
       justification: "Adds proper support for pipelining Redis commands"
+  "2024.03.21.182031-f6f5775":
+    com.squareup.misk:misk-redis:
+    - code: "java.method.removed"
+      old: "method redis.clients.jedis.Pipeline misk.redis.Redis::pipelined()"
```
keeferrourke, line 6105: I found no usages of this method in our internal repositories.

Now that we have `pipelining {}` to replace `pipelined()`, we should probably think about cutting some releases.

Perhaps cut one before this change, so folks can pick up the deprecation, and then another one afterwards?

keeferrourke:
Ah yikes these do not behave the same way for slotted keys 😬 

The pipelined versions don't do mget in the right order, and there seems to be some difference with zrange by score.

May have to revert https://github.com/cashapp/misk/pull/3195.

keeferrourke:
> May have to revert #3195.

Or perhaps remove the affected methods from DeferredRedis for the time being, then add them back via this PR.

keeferrourke:
Ok I've narrowed down the problems to just the `zrevrangeByScore` calls, which are **always** returning empty lists for some reason when they are in a pipeline.

keeferrourke:
> Ok I've narrowed down the problems with sorted set operations to just the jedis pipeline `zrevrangeByScore` calls, which are **always** returning empty lists for some reason when they are in cluster mode.
> 
> When connecting to the redis cluster via `redis-cli -p 7000 -c`, the command works as expected.

There is a bug in Jedis 4.4.6!

https://github.com/redis/jedis/blob/3577f93f91229e6ab30719a5a11be4603848c088/src/main/java/redis/clients/jedis/MultiNodePipelineBase.java#L2866 <-- the min/max arguments are flipped here!

keeferrourke:
This PR is currently blocked on one of:

- [ ] Jedis releases `4.4.8` with the bugfix mentioned above.
- [ ] misk-redis upgrades to Jedis v5 which does not have this bug, as pipelining is done via unified `AbstractPipeline`s.

adrw (approved):

krispenney (approved):

PR #3199, opened 2024-03-22
Fix pipelined with cluster mode

keeferrourke:
While working on https://github.com/cashapp/misk/pull/3198 I noticed deficiencies in the pipelined implementations of `mget` and `zrangeByScore` when running in cluster mode.

This addresses the problem with mget, and replicates the existing `batchGetAndSet` test.

It also temporarily removes the sorted set (`z*`) methods from DeferredRedis in order to avoid accidentally giving wrong results when running in cluster mode, until the pipelined version can be fixed.

---

This is a breaking change, but since `redis.pipelining {}` is **so new** (merged today), there is no way anyone is using these methods yet.


mslogar-squareup (approved):
Nice catch!

PR #3203, opened 2024-03-25
Upgrade to Jedis 5

keeferrourke:
This unblocks some work on #3198, further simplifies our redis client implementation, and upgrades to the latest major, minor, and patch version of Jedis.

---

Note: This is a major version upgrade for Jedis. This is **not** ABI compatible with previous releases.

If we are not comfortable with this change at this time, we can wait for Jedis 4.4.8 to be released instead.


misk-redis/src/main/kotlin/misk/redis/Redis.kt
```diff
@@ -746,6 +753,30 @@ interface Redis {
...
+      /**
+       * Checks that options will result in a valid ZADD command, conforming to the following:
+       *
+       * ```
+       * zadd key [NX|XX] [GT|LT] [CH] score member [score member ...]
+       * ```
+       *
+       * See the [ZADD command documentation](https://redis.io/commands/zadd) for more information.
+       */
+      internal fun verify(options: Array<out ZAddOptions>) {
```
keeferrourke, line 766: Jedis 5 no longer throws this exception for zadd. Keeping this behavior in misk for backwards compatibility reasons. I don't really know how Jedis 5 resolves incompatible options, but I think checking before making the actual call is a good idea anyway.

misk-redis/src/test/kotlin/misk/redis/RedisClientMetricsTest.kt
```diff
@@ -32,26 +31,18 @@ class RedisClientMetricsTest {
...
-  }
-
   @Inject private lateinit var collectorRegistry: CollectorRegistry
   @Inject private lateinit var redis: Redis
 
   @Test fun `connections are counted`() {
-    // No connections are created or used yet.
-    assertThat(collectorRegistry[ACTIVE_CONNECTIONS]).isEqualTo(0.0)
+    // Creating a redis client creates a connection.
+    assertThat(collectorRegistry[ACTIVE_CONNECTIONS]).isEqualTo(1.0)
```
keeferrourke, line 39: This is a behaviour change in unified Jedis from 4->5.

adrw (approved):

tgregory-block (approved):

PR #3212, opened 2024-04-02
Partial revert of config change in #3203

keeferrourke:
Getting some reports of errors related to redis `writer_endpoint` config after this change landed, so reverting this back to `isNullOrBlank` out of precaution.


keeferrourke:
Pretty sure this is something else. Will re-open if determined necessary.

PR #3213, opened 2024-04-02
Add AllowAnyUser annotation and wildcards in access annotations

adrw:


misk/src/main/kotlin/misk/security/authz/AccessInterceptor.kt
```diff
@@ -70,10 +75,11 @@ class AccessInterceptor private constructor(
         return null
       }
 
-      val allowAnyService = action.hasAnnotation<AllowAnyService>()
+      val allowAnyService = action.hasAnnotation<AllowAnyService>() || actionEntries.any { it.allowAnyService }
```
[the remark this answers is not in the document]
AlexSzlavik: This is a tangential / follow up: 
The logic in this Interceptor is getting a bit hair imo. It's hard to look at it and reason through what we expect to happen. Wonder if there is space to refactor a bit and make the policy more obvious: 
The Policy for the default AccessInterceptor is meant to be: Combine all access policies described by AccessAnnotationEntries and Authenticated annotations and if any yield an "allow" result, the overall policy is to allow access.

misk/src/main/kotlin/misk/security/authz/AccessInterceptor.kt
```diff
@@ -114,7 +120,7 @@ class AccessInterceptor private constructor(
       val allowedServices = actionEntries.flatMap { it.services }.toSet()
       val allowedCapabilities = actionEntries.flatMap { it.capabilities }.toSet()
 
-      if (!allowAnyService && allowedServices.isEmpty() && allowedCapabilities.isEmpty()) {
+      if (!allowAnyService && allowedServices.isEmpty() && !allowAnyUser && allowedCapabilities.isEmpty()) {
```
AlexSzlavik, line 123: maybe we can group the "AllowAny" checks into a common function? Or, group allow any service + allowedServices into a single function. wdyt?

misk/src/test/kotlin/misk/web/actions/TestWebActionModule.kt
```diff
@@ -187,3 +189,13 @@ class AllowAnyServicePlusAuthenticatedAccessAction @Inject constructor() : WebAc
...
   fun get() = "${scopedCaller.get()} authorized as any service".toResponseBody()
 }
+
+class AllowAnyUserAccessAction @Inject constructor() : WebAction {
+  @Inject
+  lateinit var scopedCaller: ActionScoped<MiskCaller?>
+
+  @Get("/allow_any_user_access")
+  @ResponseContentType(MediaTypes.TEXT_PLAIN_UTF8)
+  @AllowAnyUser
```
AlexSzlavik, line 199: Change this to the new Authenticated change

misk/src/main/kotlin/misk/security/authz/AccessInterceptor.kt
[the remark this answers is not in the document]
adrw: Yeah I think the deprecated annotation is sufficient and also agree that further refactoring cleanup would be beneficial.

misk/src/test/kotlin/misk/web/actions/AuthenticationTest.kt
```diff
@@ -249,6 +249,28 @@ class AuthenticationTest {
...
+  }
+
+  @Test fun testAllowAnyUserDenyServiceOnly() {
+    val caller = MiskCaller(service = "test")
+    assertThat(
+      executeRequest(
+        path = "/allow_any_user_access",
+        user = caller.user
+      )
+    )
```
adrw, line 270: nit: join with following line

misk/src/test/kotlin/misk/web/actions/AuthenticationTest.kt
```diff
@@ -249,6 +249,28 @@ class AuthenticationTest {
...
 
+  @Test fun testAllowAnyUser() {
+    val caller = MiskCaller(user = "sandy")
+    assertThat(
+      executeRequest(
+        path = "/allow_any_user_access",
+        user = caller.user
+      )
+    )
+      .isEqualTo("$caller authorized as any user")
```
adrw, line 260: nit: join with following line

AlexSzlavik (approved):

PR #3218, opened 2024-04-05
(DNM) Revert "Simplify misk-redis implementation (#3198)"

keeferrourke:
This reverts commit d3fad45e48488f1915e1fef94095fca897aff108.

We're seeing some unexpected Jedis connection pool issues which are resolved by rolling back this dependency upgrade.

---

UPDATE: I think it's probable that this change was not compatible when Jedis is configured to have a single connection in its pool. The default Jedis pool configurations support a default value of up to 8 maximum connections.


gghang-block (approved):

keeferrourke:
- [ ] Do not merge. This PR has been prepared in case we need to revert.

mslogar-squareup (approved):

PR #3219, opened 2024-04-05
Revert "Revert "Simplify misk-redis implementation (#3198)""

keeferrourke:
Reverts cashapp/misk#3218

In other words... Brings back cashapp/misk#3198.


gghang-block (changes requested):
This refactor appears to causing an issue in our service where it suddenly becomes unresponsive after failing to acquire Redis connections. I'd like to figure out why before re-merging this change to prevent this change from affecting additional services. 

- [ ] Determine root cause of Redis connection bugs before merging

keeferrourke:
- [ ] DNM until cause of connection issues that caused the revert has been determined

keeferrourke:
I'll look for other ways to simplify this.

PR #3224, opened 2024-04-10
Add sorted set operations to misk-redis pipelines

keeferrourke:
Essentially reimplements the reverted changes #3198, but in an additive-only way which is much less disruptive.

1. Introduces a **test-sources-only** redis client implementation which always delegates to pipelining operations. This allows us to ensure that RealPipelinedRedis always has test parity with RealRedis.
   - This provides a path forward for non-disruptively attempting to reproduce and diagnose connection/client configuration issues that may have been caused by just pipelining everything by default (as was done in #3198).
2. Reintroduces sorted-set operations to pipelines. This is safe to do, since we have upgraded to Jedis 5 (see #3198 for the reasoning why this was introduced and removed, before being re-introduced due to bugs in Jedis).

### DIFF
--- a/misk-redis/src/main/kotlin/misk/redis/Redis.kt
+++ b/misk-redis/src/main/kotlin/misk/redis/Redis.kt
@@ -5,8 +5,8 @@ import redis.clients.jedis.JedisPubSub
 import redis.clients.jedis.Pipeline
 import redis.clients.jedis.Transaction
 import redis.clients.jedis.args.ListDirection
-import redis.clients.jedis.exceptions.JedisDataException
 import redis.clients.jedis.params.ZAddParams
+import redis.clients.jedis.util.JedisClusterCRC16
 import java.time.Duration
 import java.util.function.Supplier
 
@@ -800,3 +800,20 @@ inline fun checkHrandFieldCount(count: Long) {
   }
 }
 
+internal fun getSlotErrorOrNull(op: String, keys: List<ByteArray>): Throwable? {
+  val slots = keys.map { JedisClusterCRC16.getSlot(it) }.distinct()
+  return if (slots.size == 1) {
+    null
+  } else {
+    RuntimeException(
+      """
+      |When using clustered Redis, keys used by one $op command must always map to the same slot, but mapped to slots $slots.
+      |You can use {hashtags} in your key name to control how Redis hashes keys to slots.
+      |For example, keys: `{customer9001}.contacts` and `{customer9001}.payments` will  hash to the same slot.
+      |
+      |See https://redis.io/topics/cluster-spec#hash-tags for more information.
+      |
+      """.trimMargin()
+    )
+  }
+}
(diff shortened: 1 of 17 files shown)


misk-redis/src/test/kotlin/misk/redis/AbstractRedisClusterTest.kt
keeferrourke, line 1: This file is not new. It was copied from RealRedisClusterTest.

misk-redis/src/test/kotlin/misk/redis/AbstractRedisTest.kt
```diff
@@ -75,44 +78,6 @@ abstract class AbstractRedisTest {
     assertThat(redis[unknownKey]).isNull()
   }
 
-  @Test
-  fun batchGetAndSetPipelined()  {
```
keeferrourke, line 79: This test is made redundant by the new PipelinedRedis*Tests.

misk-redis/src/test/kotlin/misk/redis/RealRedisClusterTest.kt
```diff
@@ -21,8 +21,7 @@ import kotlin.test.assertEquals
 import kotlin.test.assertNull
 
 @MiskTest
-class RealRedisClusterTest : AbstractRedisTest() {
```
keeferrourke, line 24: Anything removed from this file, has simply moved into the new AbstractRedisClusterTest.

misk-redis/src/test/kotlin/misk/redis/TestAlwaysPipelinedRedis.kt
keeferrourke, line 1: This implementation exists to ensure that we get test coverage and functional parity for each method on RealRedis vs the pipelined versions.

misk-redis/src/test/kotlin/misk/redis/AbstractRedisTest.kt
```diff
@@ -29,6 +29,9 @@ import kotlin.test.assertTrue
 abstract class AbstractRedisTest {
   abstract var redis: Redis
 
+  @BeforeEach
+  fun setUp() = redis.flushAll()
+
```
mslogar-squareup, line 34: Nice!

misk-redis/src/main/kotlin/misk/redis/Redis.kt
```diff
@@ -800,3 +800,20 @@ inline fun checkHrandFieldCount(count: Long) {
   }
 }
 
+internal fun getSlotErrorOrNull(op: String, keys: List<ByteArray>): Throwable? {
+  val slots = keys.map { JedisClusterCRC16.getSlot(it) }.distinct()
```
mslogar-squareup, line 804: Do we need this logic client-side? I would imagine that Redis would throw a `CROSSSLOT` for us

mslogar-squareup (approved):
Thanks for adding this! One non-blocking comment

misk-redis/src/main/kotlin/misk/redis/Redis.kt
mslogar-squareup: Non-blocking bc I see that this check isn't entirely net-new

keeferrourke: Sadly it doesn't behave as expected in pipelines unless we do the client-side check. I would prefer a more consistent experience across the client, which is why I promoted it to the non-pipelined operations that might encounter cross slot errors as well.

PR #3228, opened 2024-04-11
Fetch the max receive count from the redrive policy directly

din-cashapp:
Instead of allowing the clients to pass-in the max retry count, fetch the retry count directly from the queue itself.

### DIFF
--- a/misk-aws/src/main/kotlin/misk/jobqueue/sqs/QueueResolver.kt
+++ b/misk-aws/src/main/kotlin/misk/jobqueue/sqs/QueueResolver.kt
@@ -1,8 +1,11 @@
 package misk.jobqueue.sqs
 
 import com.amazonaws.services.sqs.AmazonSQS
+import com.amazonaws.services.sqs.model.GetQueueAttributesRequest
 import com.amazonaws.services.sqs.model.GetQueueUrlRequest
+import com.amazonaws.services.sqs.model.QueueAttributeName
 import com.amazonaws.services.sqs.model.QueueDoesNotExistException
+import com.squareup.moshi.Moshi
 import misk.cloud.aws.AwsAccountId
 import misk.cloud.aws.AwsRegion
 import misk.jobqueue.QueueName
@@ -11,6 +14,7 @@ import java.io.File
 import java.util.concurrent.ConcurrentHashMap
 import jakarta.inject.Inject
 import jakarta.inject.Singleton
+import misk.moshi.adapter
 
 @Singleton
 internal class QueueResolver @Inject internal constructor(
@@ -21,11 +25,13 @@ internal class QueueResolver @Inject internal constructor(
   @ForSqsReceiving private val defaultForReceivingSQS: AmazonSQS,
   @ForSqsReceiving private val crossRegionForReceivingSQS: Map<AwsRegion, AmazonSQS>,
   private val externalQueues: Map<QueueName, AwsSqsQueueConfig>,
-  private val dlqProvider: DeadLetterQueueProvider
+  private val dlqProvider: DeadLetterQueueProvider,
+  private val moshi: Moshi,
 ) {
   private val forSendingMapping = ConcurrentHashMap<QueueName, ResolvedQueue>()
   private val forReceivingMapping = ConcurrentHashMap<QueueName, ResolvedQueue>()
   private val hostInternalTarget = "host.docker.internal"
+  private val redrivePolicyAdapter = moshi.adapter(RedrivePolicy::class.java)
 
   fun getForSending(q: QueueName): ResolvedQueue {
     return forSendingMapping.computeIfAbsent(q) { resolve(it, false) }
@@ -53,7 +59,6 @@ internal class QueueResolver @Inject internal constructor(
     }
 
     checkNotNull(sqs) { "could not find SQS client for ${region.name}" }
-
     val queueUrl = try {
       val queue_url = sqs.getQueueUrl(
         GetQueueUrlRequest().apply {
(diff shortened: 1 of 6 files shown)


misk-aws/src/main/kotlin/misk/jobqueue/sqs/QueueResolver.kt
```diff
@@ -67,7 +72,29 @@ internal class QueueResolver @Inject internal constructor(
       throw e
     }
 
-    return ResolvedQueue(q, sqsQueueName, queueUrl, region, accountId, sqs)
+    val queueMaxRetries = extractMaxReceiveCount(sqs, queueUrl)
+    return ResolvedQueue(q, sqsQueueName, queueUrl, region, accountId, sqs, queueMaxRetries)
+  }
+
+  data class RedrivePolicy(val maxReceiveCount: Int, val deadLetterTargetArn: String)
```
tgregory-block, line 79: Is `deadLetterTargetArn` used?

misk-aws/src/main/kotlin/misk/jobqueue/sqs/SqsJob.kt
```diff
@@ -65,10 +64,11 @@ internal class SqsJob(
    *  that duration. With every subsequent retry the duration becomes longer until it hits the max
    *  value of 10hrs.
    */
-   fun delayForFailure(maxRetryCount: Int = 10) {
+  fun delayForFailure() {
```
tgregory-block, line 67: This breaks backwards compatibility - any services actually passing a value here will break. You'll need to create a new method that doesn't take an argument

misk-aws/src/main/kotlin/misk/jobqueue/sqs/QueueResolver.kt
tgregory-block, line 79: I think `RedrivePolicy` is a misnomer, this doesn't control whether we have a DLQ or after how many delivery attempts a job will be put in the DLQ, it just controls the max backoff period for retrying a job

misk-aws/src/main/kotlin/misk/jobqueue/sqs/ResolvedQueue.kt
```diff
@@ -13,7 +13,8 @@ internal class ResolvedQueue(
   val url: String,
   val region: AwsRegion,
   val accountId: AwsAccountId,
-  val client: AmazonSQS
+  val client: AmazonSQS,
+  val maxRetries: Int,
```
tgregory-block, line 17: Similarly `maxRetries` is a bit of a misnomer, that is controlled by the `max_delivery_attempts` setting on the queue, but this is just setting a ceiling on the backoff time

misk-aws/src/main/kotlin/misk/jobqueue/sqs/SqsJob.kt
din-cashapp: hmm, how about keeping it this way but just not using the value?

misk-aws/src/main/kotlin/misk/jobqueue/sqs/QueueResolver.kt
din-cashapp: oh good catch - lemme remove

din-cashapp: hmmm that's actually the name of it in AWS: https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-dead-letter-queues.html.
but if you have a suggestion on the better name please let me know!

din-cashapp: we are effectively fetching the policy in order to determine the maxRetryCount on the particular queue 🤔

misk-aws/src/main/kotlin/misk/jobqueue/sqs/ResolvedQueue.kt
din-cashapp: maxRetries is essentially the max_delivery_attempts that we retrieve from the RedrivePolicy though or am i missing some context? 🤔

misk-aws/src/main/kotlin/misk/jobqueue/sqs/QueueResolver.kt
tgregory-block: Ah my mistake, this attribute is actually populated on the queue when we create it, I thought this was something custom for this PR.

misk-aws/src/main/kotlin/misk/jobqueue/sqs/SqsJob.kt
tgregory-block: I definitely wouldn't silently stop using the parameter, that would be very surprising to callers. I think what you'll need to do is have:

```kotlin
@Deprecated("Prefer delayForFailure(), which uses the queue's maxRetryCount")
fun delayForFailure(maxRetryCount: Int = 10) { /** old code */ }

fun delayForFailure() { /** new code */ }
```

This will change the behavior for anyone not explicitly setting a value but I don't think this function is actually used anywhere so that should be fine

tgregory-block: Apologies, I forgot that `SqsJob` is internal. Since this isn't used anywhere else I think this change is fine as-is

tgregory-block (commented):
This LGTM, you'll need to update the ABI checker though

tgregory-block (approved):

misk-aws/src/main/kotlin/misk/jobqueue/sqs/ResolvedQueue.kt
adrw, line 17: Could a default value be set here? I anticipate this will cause build failures when it rolls out due to Misk version mismatch with downstream callers constructing this object.

adrw (approved):

PR #3229, opened 2024-04-11
Add generic marshaller for Unit response

mgulbronson:
For an endpoint that returns nothing — for instance if it always returns a redirect response — it's currently impossible to return an empty `Response` object without a response media type.

In `ResponseBodyMarshallerFactory` it would look for a generic marshaller that matches the return type [here](https://github.com/cashapp/misk/blob/447c6c984bff3a87473cd2650996a20a9bb1de80/misk/src/main/kotlin/misk/web/interceptors/ResponseBodyMarshallerFactory.kt#L19). This means that a response of type `Unit` would throw.

The workaround for that is to either add a `@ResponseContentType` annotation to force it to use a non-generic marshaller — even if the endpoint never actually returns that content type — or to change the response type to something like an empty string so that it can be handled by the generic `FromString` marshaller. While both of these approaches work I think a Unit response better represents what this kind of empty response actually means.

### DIFF
--- a/misk/src/main/kotlin/misk/web/marshal/Generic.kt
+++ b/misk/src/main/kotlin/misk/web/marshal/Generic.kt
@@ -94,7 +94,7 @@ object GenericMarshallers {
 
   class ToUnit(private val contentType: MediaType?) : Marshaller<Unit> {
     override fun contentType(): MediaType? = contentType
-    override fun responseBody(o: Nothing) = object : ResponseBody {
+    override fun responseBody(o: Unit) = object : ResponseBody {
       override fun writeTo(sink: BufferedSink) {}
     }
   }
--- a/misk/src/main/kotlin/misk/web/marshal/Generic.kt
+++ b/misk/src/main/kotlin/misk/web/marshal/Generic.kt
@@ -58,7 +58,8 @@ object GenericMarshallers {
     String::class.java,
     ResponseBody::class.java,
     ByteString::class.java,
-    Nothing::class.java
+    Nothing::class.java,
+    Unit::class.java,
   )
 
   private class FromResponseBody(private val contentType: MediaType?) : Marshaller<ResponseBody> {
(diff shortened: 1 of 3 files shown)


misk/src/main/kotlin/misk/web/marshal/Generic.kt
```diff
@@ -91,13 +92,21 @@ object GenericMarshallers {
     }
   }
 
+  class ToUnit(private val contentType: MediaType?) : Marshaller<Unit> {
+    override fun contentType(): MediaType? = contentType
+    override fun responseBody(o: Nothing) = object : ResponseBody {
```
Nava2, line 97:
```suggestion
    override fun responseBody(o: Unit) = object : ResponseBody {
```

I'm fairly confident that there's no way to pass `Nothing` as a parameter.

misk/src/test/kotlin/misk/web/marshal/UnitResponseTest.kt
```diff
@@ -0,0 +1,46 @@
...
+    with(webTestClient.get("/response/as-unit-response-body")) {
+      assertThat(response.code).isEqualTo(200)
+      assertThat(response.headers.toMap()).doesNotContainKey("Content-Type")
+      assertThat(response.body!!.bytes()).isEmpty()
+    }
+  }
+
+  class ReturnAsUnitResponseBody @Inject constructor() : WebAction {
+    @Get("/response/as-unit-response-body")
+    fun call() = Response(Unit)
```
adrw, line 36: So is this code change mostly allowing to return `Unit` in the Response? Or is it to handle a missing request/response content type annotation? or Both?

swankjesse (approved):

Nava2 (approved):
LGTM with a couple more test coverage cases.

misk/src/test/kotlin/misk/web/marshal/UnitResponseTest.kt
mgulbronson: Added tests for those two cases. Not sure how useful they are considering the `Unit` response is IMO a strictly better version, but it certainly doesn't hurt to prove that my claims above are actually true :)

PR #3231, opened 2024-04-12
Switch Metadata to mapbinder and expose module to wrap install

adrw:


danieloh0714 (approved):

adrw:
I might be blocked on https://github.com/cashapp/misk/issues/2863, investigating further what the best path forward is on the import violation build failures.

PR #3232, opened 2024-04-15
Allow the caller to fail config load on unknown properties

sahilm:
This is useful when we want to run automated tests on our app configuration to find invalid configs for real environments in CI. The default behaviour of ignoring unknown properties with a warning is preserved.


frojasg (approved):
LGTM! Thanks for this change

PR #3233, opened 2024-04-15
Akelsall/20240411 introduce tagged logger

afkelsall:
Viewing service logs in a system such as Datadog and utilizing Mapped Diagnostic Context (MDC) tags is a useful and quick way to diagnose service issues when a particular domain entity has some sort of failure. Often a service will have included the domain entity token as an MDC tag to allow easy viewing of all logs relating to the processing of that entity.

When an unhandled exception occurs in that service that bubbles up to Misk (Web action or Sqs job in particular), the MDC tags have already been cleaned up by the service and therefore the final log entry highlighting the exception at the end of the log trail is missing from the initial search. It requires engineers to change their search explicitly to use a relevant `trace_id` or similar to view the final exception.

A common pattern that services using misk implement themselves is to setup a `TaggedLogger` class to wrap the setting of MDC tags for the logger. For example:

```
fun processJob(job) {
  logger
    .jobId(job.jobId)
    .businessDomainToken(job.businessDomainToken)
    .asContext(
      logger.info { "Job processing started" }
      serviceCode()
      logger.info { "Job processing finished" }
    )
}

companion object {
  private val logger = this::class.getTaggedLogger()
}
```

The way this `TaggedLogger` typically works is to build a set of MDC tags and upon entering the `asContext()` call it will first record the current MDC tags (to revert after `asContext()`) and then add the additional MDC tags to the underlying log system (thread local). Any `logger.info|warn|error {}` calls will be logged with the MDC tags and at the conclusion of the `asContext()` call they will be reverted. This also allows nested `TaggedLoggers` to be built up throughout service code as more context is loaded.

The problem is that when the service code throws an unhandled exception and the Misk framework catches it and handles appropriately, the logging from Misk does not include the MDC tags at the time the exception was thrown. As the exception bubbles up to misk, the `TaggedLogger.asContext()` calls all have a `finally` clause to clean up after themselves and remove the MDC context.

The solution here is to implement the concept of a `TaggedLogger` in Misk itself and have it save to thread local storage the MDC tags at the time the exception was thrown and Misk is then able to query this storage for any additional context when logging the exception and include those tags in addition to whatever it was going to log with (including clearing out thread local tags when appropriate).


misk-aws/src/main/kotlin/misk/jobqueue/sqs/SqsJobConsumer.kt
```diff
@@ -208,7 +211,12 @@ internal class SqsJobConsumer @Inject internal constructor(
                 )
                 Status.OK
               } catch (th: Throwable) {
-                log.error(th) { "error handling job from ${queue.queueName}" }
+                val mdcTags = TaggedLogger.getThreadLocalMdcContext()
```
ThePumpingLemma, line 214: Seems like get and reset are always called in tandem; you could also have a method like `TaggedLogger.popThreadLocalMdcContext()` that does both (not picky about naming here).

misk-aws/src/main/kotlin/misk/jobqueue/sqs/SqsJobConsumer.kt
```diff
@@ -208,7 +211,12 @@ internal class SqsJobConsumer @Inject internal constructor(
...
+                  .also { TaggedLogger.resetThreadLocalMdcContext() }
+
+                // The underlying logger takes care of adding the mdc tags and cleaning them up when calling this way
```
ThePumpingLemma, line 217: Could also push this documentation down to `wisp.logging.log` if others would find that helpful.

misk-aws/src/main/kotlin/misk/jobqueue/sqs/SqsJobConsumer.kt
afkelsall: If left to my own un-imagination I would have probably named it `getAndResetThreadLocalMdcContext...` but pop is wildly better and on point thank you.

misk-aws/src/main/kotlin/misk/jobqueue/sqs/SqsJobConsumer.kt
afkelsall: Done. Even found an underlying extension function I can import and call from the same wisp file to make this an `log.error(th` instead of `log.log(Level.ERROR, th`.

ThePumpingLemma (approved):

PR #3234, opened 2024-04-15
Make AwsSqsJobHandlerSubscriptionService public...

hamdanjaveed:
... to allow usage in the service dependency graph

Users may wish to let other Services depend on the AWS JobHandler Subscription service in cases where you want to ensure that SQS job handling has begun before your service has started.

As an alternative to making all these classes `open`, we could do something like this:

```kotlin
val sqsJobHandlerSubscriptionServiceKey: Key<out Service> = Key.get(AwsSqsJobHandlerSubscriptionService::class.java)
```

And leave the `AwsSqsJobHandlerSubscriptionService` and all the rest of the classes marked as `internal`. Not sure if this would actually work and we'd wanna test this further but from our small injector test it seemed to pass.


frojasg:
I would rather expose an interface that has less implementation details.  Happy to pair to unblock this.

PR #3235, opened 2024-04-15
Wrap DataSource to allow construction without establishing DB connection

damar-block:
During work on enabling PRODUCTION stage in injector test, there are multiple cases where objects require `DataSource` object for their construction. Currently, `DataSource` object is only available after connection to DB was established, which violates Guice best practices. This PR changes this module, that the provided `DataSource` object is a wrapper that will be initialized on service startup. It will allow consumers to require this object during construction, as long as they don't use it yet.

This PR will address cases where external code depends on `DataSource`, which we can't easily modify. One example is the construction of `RateLimiter` in [MySQLBucket4jRateLimiterModule](https://github.com/cashapp/misk/blob/f03846985cd6f158cd73cf71d81e12f67e76027e/misk-rate-limiting-bucket4j-mysql/src/main/kotlin/misk/ratelimiting/bucket4j/mysql/MySQLBucket4jRateLimiterModule.kt#L37-L37).


misk-jdbc/src/main/kotlin/misk/jdbc/DataSourceWrapper.kt
```diff
@@ -0,0 +1,76 @@
...
+
+import java.io.PrintWriter
+import java.sql.Connection
+import java.sql.ConnectionBuilder
+import java.sql.ShardingKeyBuilder
+import java.util.concurrent.atomic.AtomicReference
+import java.util.logging.Logger
+import javax.sql.DataSource
+
+internal class DataSourceWrapper(private val simpleName : String?): DataSource {
```
katukota, line 11: Is this similar to HikariDataSource that was previously being used? Can we please link the original source code here so that it is easier to understand?

misk-jdbc/src/main/kotlin/misk/jdbc/DataSourceService.kt
```diff
@@ -39,9 +39,7 @@ class DataSourceService @JvmOverloads constructor(
   /** The decorated data source */
   private var _dataSource: DataSource? = null
```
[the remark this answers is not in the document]
damar-block: No we don't, I will remove.

misk-jdbc/src/main/kotlin/misk/jdbc/DataSourceWrapper.kt
damar-block: Added a comment, LMK if that is better. If not, what original source code are you referring to?

adrw:
I'd lean towards closing this as we discussed given it doesn't seem to add much safety (the target benefit) at the cost of a lot of boilerplate for a pseudo-Provider layer outside of Guice and new risks of later verification.

PR #3236, opened 2024-04-16
Use Jedis provider instead of direct dependency

damar-block:


frojasg:
Could you add a details with the motivation of this change, please?

PR #3237, opened 2024-04-16
Add aggregation support in select queries

keeferrourke:
Previously, it's possible to perform aggregations with `dynamicUniqueResult` or `dynamicList` by operating on the `root` and `criteriaBuilder`, however grouping is not exposed without access to the `CritieriaQuery` itself. This forces anyone who wants to do relatively common projected, grouped, and aggregated queries to operate on the `session.hibernateSession,` leading to more complicated code.

This change adds top-level support for aggregations and group by in misk.hibernate.Query interfaces.

Notably, this allows one to specify an `AggregationType` on `@Select` annotated methods, and adds a `@Group` annotation (similar to the existing `@Order`).

**How do I use this?**

Let's say you want to find the number of movies that were released on each date in your movies database. You might want to run a SQL query like `select release_date, count(name) from movies group by release_date`. Here's how you do that now with misk.hibernate.Query interfaces:

```kt
interface MovieQuery: Query<DbMovie> {
  @Select 
  fun datesAndCounts(session: Session): List<DateAndCount>
  
  @Group(paths = ["release_date"])
  fun groupByDate(): MovieQuery
}

data class DateAndCount(
  @Property(path = "release_date")
  val date: LocalDate,
  @Property(path = "name", aggregation = AggregationType.COUNT) 
  val count: Long?,
): Projection

  queryFactory.newQuery<MovieQuery>()
    .groupByDate()
    .datesAndCounts()
}
```

You'll get back a list of `DateAndCount` which have correctly counted the number of movie titles released on each date in the database.

| ⚠️ Warning |
| :-- |
| Forgetting to call the `groupByDate()` method on the above query will give a wrong result! Without the grouping, mySQL will give you back a date and count of all rows. This can lead to bugs! It's basically impossible to check for this though, since you might want to (for example) find the max of all columns without any grouping, etc. Testing your queries is necessary! |

`@Group` annotations accept multiple paths, since it's often useful to group aggregated results by multiple columns.

If either the Query class or the Projection are malformed, `ReflectionQuery` will type-check and complain with exactly the issue to avoid cryptic "Could not extract ResultSet" errors from hibernate.

We also now support joins in projections, by setting `@Fetch(path, forProjection = true)`, which allows for grouped aggregations across tables. For example, see b7f348a1cd4aa6f43ddb191ba89e8c34008ac2a6.

### DIFF
--- a/misk-hibernate/src/main/kotlin/misk/hibernate/Projection.kt
+++ b/misk-hibernate/src/main/kotlin/misk/hibernate/Projection.kt
@@ -1,12 +1,15 @@
 package misk.hibernate
 
-/** Marker interface for query projections. */
 interface Projection
 
 /**
  * Annotates a parameter of a [Projection] data class to indicate which column (or path of columns)
  * to populate the parameter with.
+ *
+ * Properties may be created from an [aggregation] function, which will be applied to the column.
+ * By default, no aggregation is applied.
  */
 annotation class Property(
-  val path: String
+  val path: String,
+  val aggregation: AggregationType = AggregationType.NONE
 )
--- a/misk-hibernate/src/main/kotlin/misk/hibernate/Projection.kt
+++ b/misk-hibernate/src/main/kotlin/misk/hibernate/Projection.kt
@@ -1,5 +1,21 @@
 package misk.hibernate
 
+/**
+ * Marker interface for query projections.
+ *
+ * Projections are used to define the shape of the result set of a query, often as a subset of
+ * the properties of the entity or entities being queried.
+ *
+ * For example, if we have a `DbMovie` entity with a `name`, `release_date`, and other properties
+ * we could use a projection to only select the `name` and `release_date` properties:
+ *
+ * ```
+ * data class NameAndReleaseDate(
+ *  @Property("name") var name: String,
+ *  @Property("release_date") var releaseDate: LocalDate?
+ * ) : Projection
+ * ```
+ */
 interface Projection
 
 /**
--- a/misk-hibernate/src/main/kotlin/misk/hibernate/Query.kt
+++ b/misk-hibernate/src/main/kotlin/misk/hibernate/Query.kt
@@ -150,6 +150,19 @@ fun <T, Q : Query<T>> Q.queryHint(hint: String): Q {
 /**
  * Annotates a function on a subinterface of [Query] to indicate which column (or path of columns)
  * it constrains and using which operator.
+ *
+ * You can think of Constraints as the rules used to build the `where` clause of a SQL query.
+ *
+ * For example, you can query movies by title with a method like this:
+ * ```
+ * @Constraint(path = "name")
+ * fun matchesTitle(title: String): MovieQuery
+ * ```
+ * Or query for movies released after a certain date with a method like this:
+ * ```
+ * @Constraint(path = "release_date", operator = Operator.GT)
+ * fun releasedAfter(date: LocalDate): MovieQuery
+ * ```
  */
 annotation class Constraint(
   val path: String,
@@ -192,9 +205,11 @@ enum class Operator {
 }
 
 /**
- * Annotates a function on a subinterface of [Query] to execute a `SELECT` query. Functions with
+ * Annotates a function on a [Query] interface to execute a `SELECT` query. Functions with
  * this annotation must return a `List` to fetch multiple rows results, or a regular type to fetch
  * a unique result.
+ *
+ * [Select] annotated methods may return single column values, or [Projection]s of multiple columns.
  */
 annotation class Select(
   val path: String = "",
@@ -202,8 +217,8 @@ annotation class Select(
 )
 
 /**
- * Annotates a function on a subinterface of [Query] to indicate which columns to order the
- * the selected columns.
+ * Annotates a function on a [Query] interface to indicate by which columns to order the
+ * results. Defaults to ascending order.
  */
 annotation class Order(
   val path: String,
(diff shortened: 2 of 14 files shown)


misk-hibernate/src/test/kotlin/misk/hibernate/AggregationQueryTest.kt
```diff
@@ -0,0 +1,282 @@
...
+    install(JdbcTestingModule(PrimitivesDb::class))
+    install(HibernateModule(PrimitivesDb::class, primitivesConfig.data_source))
+    install(object : HibernateEntityModule(PrimitivesDb::class) {
+      override fun configureHibernate() {
+        addEntities(DbPrimitiveTour::class)
+      }
+    })
+
+    val moviesConfig = MiskConfig.load<MoviesConfig>("moviestestmodule", TESTING)
+    val dataSourceConfig = moviesConfig.mysql_data_source
```
keeferrourke, line 233: Note that this test uses a MySQL data source instead of a Vitess data soource because there is (fixed) bug in some older versions of vitess which can't compare VARCHAR types in aggregations. Misk's VitessCluster pulls an image from 2020 (possibly v6?). This bug was fixed in 2021. We'll need to upgrade Misk's Vitess SHA in order to run this test under Vitess, but that would definitely be a separate change.

keeferrourke: The error that is thrown on the currently used image is `vtgate: types are not comparable: VARCHAR vs VARCHAR`

misk-hibernate/src/main/kotlin/misk/hibernate/Projection.kt
```diff
@@ -1,12 +1,31 @@
...
+ * ```
+ */
 interface Projection
 
 /**
  * Annotates a parameter of a [Projection] data class to indicate which column (or path of columns)
  * to populate the parameter with.
+ *
+ * Properties may be created from an [aggregation] function, which will be applied to the column.
+ * By default, no aggregation is applied.
```
swankjesse, line 26: Love these docs

misk-hibernate/src/main/kotlin/misk/hibernate/Query.kt
```diff
@@ -150,6 +150,19 @@ fun <T, Q : Query<T>> Q.queryHint(hint: String): Q {
 /**
  * Annotates a function on a subinterface of [Query] to indicate which column (or path of columns)
  * it constrains and using which operator.
+ *
+ * You can think of Constraints as the rules used to build the `where` clause of a SQL query.
```
swankjesse, line 154: YES!

eeSeeGee (approved):

misk-hibernate/src/main/kotlin/misk/hibernate/ReflectionQueryFactory.kt
```diff
@@ -840,18 +916,80 @@ internal class ReflectionQuery<T : DbEntity<T>>(
...
+        // Primitives are special. We have to ensure we are using the boxed types.
+        val doubleType = Double::class.createType(nullable = true).typeLiteral().rawType
+        val longType = Long::class.createType(nullable = true).typeLiteral().rawType
+        val numberType = Number::class.createType(nullable = true).typeLiteral().rawType
+
+        fun typeToString(type: Class<*>): String = "${type.simpleName}?"
+
+        val errors = mutableListOf<String>()
+        when (aggregation) {
+          AggregationType.AVG -> {
```
swankjesse, line 959: These aggressive type checks are really nice

misk-hibernate/src/main/kotlin/misk/hibernate/ReflectionQueryFactory.kt
```diff
@@ -840,18 +916,80 @@ internal class ReflectionQuery<T : DbEntity<T>>(
...
         result[javaMethod] = selectMethodHandler
       }
+
+      private fun collectAggregationTypeErrors(
+        aggregation: AggregationType,
+        elementType: TypeLiteral<*>,
+        propertyName: String,
+      ) : MutableList<String> {
+        // Primitives are special. We have to ensure we are using the boxed types.
+        val doubleType = Double::class.createType(nullable = true).typeLiteral().rawType
```
swankjesse, line 951: move these to a companion?

misk-hibernate/src/test/kotlin/misk/hibernate/AggregationQueryTest.kt
```diff
@@ -0,0 +1,326 @@
...
+    )
+    install(object : HibernateEntityModule(Movies::class) {
+      override fun configureHibernate() {
+        addEntities(DbMovie::class, DbActor::class, DbCharacter::class)
+      }
+    })
+  }
+}
+
+private interface WrongTypesPrimitiveTourQuery : Query<DbPrimitiveTour> {
```
swankjesse, line 293: I was hoping to find this

swankjesse (approved):

misk-hibernate/src/main/kotlin/misk/hibernate/Projection.kt
frojasg: same! Kudos

keeferrourke: ❤️

misk-hibernate/src/main/kotlin/misk/hibernate/ReflectionQueryFactory.kt
keeferrourke: [c82a0c1](https://github.com/cashapp/misk/pull/3237/commits/c82a0c151ef7d1f95a3837d21f153c50f92ce6dd)

PR #3240, opened 2024-04-18
CCPOKR-13542 | Issue Fix: Dynamodb table "misk-cluster-members" is not deleting any records

katukota:
RCA: Currently TTL field was set to `updated_at` field which was set to `ms` granularity. It should be at `seconds` granularity. Added a new field to use as TTL.

### DIFF
--- a/misk-clustering-dynamodb/src/main/kotlin/misk/clustering/dynamo/DyClusterMember.kt
+++ b/misk-clustering-dynamodb/src/main/kotlin/misk/clustering/dynamo/DyClusterMember.kt
@@ -15,4 +15,8 @@ internal class DyClusterMember {
 
   @get:DynamoDbAttribute("pod_name")
   var pod_name: String? = null
+
+  // Stored as epoch sec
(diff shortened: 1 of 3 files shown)


damar-block (approved):
LGTM, but I do suggest adding a test this time. See [here](https://github.com/cashapp/misk/blob/81ea336182dc970a56dc67f7957deb65b77a5859/misk-clustering-dynamodb/src/test/kotlin/misk/clustering/DynamoClusterTest.kt#L83-L83) example where we can poke into the table and retrieve the item in order to inspect it.

katukota:
> LGTM, but I do suggest adding a test this time. See [here](https://github.com/cashapp/misk/blob/81ea336182dc970a56dc67f7957deb65b77a5859/misk-clustering-dynamodb/src/test/kotlin/misk/clustering/DynamoClusterTest.kt#L83-L83) example where we can poke into the table and retrieve the item in order to inspect it.

Added a test case to verify TTL.

PR #3242, opened 2024-04-18
Fix NullPointerException in ReflectionQuery.SelectMethodHandler#toValue

keeferrourke:
There's a bug in projected queries where an NPE can be thrown if there are no matching rows. This is because we incorrectly pass an `Any?` into a method that wants `Any`. Oops!

This affected both single value projections and data class projection queries.

### DIFF
--- a/misk-hibernate/src/main/kotlin/misk/hibernate/ReflectionQueryFactory.kt
+++ b/misk-hibernate/src/main/kotlin/misk/hibernate/ReflectionQueryFactory.kt
@@ -412,8 +412,7 @@ internal class ReflectionQuery<T : DbEntity<T>>(
   @Singleton
   internal class Factory @Inject internal constructor(
     private var queryLimitsConfig: QueryLimitsConfig
-  ) :
-    Query.Factory {
+  ) : Query.Factory {
     private val queryMethodHandlersCache = CacheBuilder.newBuilder()
       .build(object : CacheLoader<KClass<*>, Map<Method, QueryMethodHandler>>() {
         override fun load(key: KClass<*>) = queryMethodHandlers(key)
@@ -509,7 +508,7 @@ internal class ReflectionQuery<T : DbEntity<T>>(
         typedQuery.list()
       }
       reflectionQuery.checkRowCount(returnList, rows.size)
-      val list = rows.map { toValue(it) }
+      val list = rows.mapNotNull { toValue(it) }
       return if (returnList) list else list.firstOrNull()
     }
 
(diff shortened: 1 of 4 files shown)


frojasg (approved):
LGTM

keeferrourke:
> LGTM

Oops, there was more :) I'll get another review from you when I'm done!

misk-hibernate/src/test/kotlin/misk/hibernate/AggregationQueryTest.kt
```diff
@@ -223,6 +224,60 @@ class AggregationQueryTest {
...
+      assertThat(assertThrows<IllegalArgumentException> {
+        queryFactory.newQuery<NoDataQuery>()
+          .listWrongNullability(session)
+      }).hasMessage(
+        """
+        |There were problems with your Projection class:
+        |  Constructor parameter maybeI1 is not nullable, but query returned null for maybeI1. Mark this parameter as nullable to avoid this error.""".trimMargin()
+      )
+    }
+  }
```
keeferrourke, line 279: TODO: This test actually belongs with ReflectionQuery (it has nothing to do with aggregations!) and should operate on the movie transacter. But it works!

keeferrourke: Hmm this test might not actually be that useful. It worked locally, but then I discovered my mysql was running in a weird sql_mode=NO_ENGINE_SUBSTITUTION, instead of the default, which forbids the types of queries which can result in these problems.

I think I'll remove it, along with the checks that attempt to match up nullability in projection properties.

misk-hibernate/src/main/kotlin/misk/hibernate/ReflectionQueryFactory.kt
```diff
@@ -528,11 +527,37 @@ internal class ReflectionQuery<T : DbEntity<T>>(
...
       return when {
+        row == null -> null
         constructor == null -> row
         properties.size == 1 -> constructor.call(row)
-        else -> constructor.call(*(row as Array<*>))
+        else -> {
+          val args = row as Array<*>
+          // If we got back an array of null values, which can't be assigned to the constructor,
+          // then we should return null.
+          if (args.all { it == null } && !constructor.parameters.all { it.type.isMarkedNullable } ) return null
```
swankjesse, line 539:
```suggestion
          if (args.all { it == null } && constructor.parameters.none { it.type.isMarkedNullable } ) return null
```

misk-hibernate/src/main/kotlin/misk/hibernate/ReflectionQueryFactory.kt
```diff
@@ -528,11 +527,37 @@ internal class ReflectionQuery<T : DbEntity<T>>(
...
+          // Otherwise, we should check if we *can* assign the null values to the constructor.
+          checkNullability(constructor, args)
+          // Now it's (probably!) safe to call the constructor.
+          constructor.call(*args)
+        }
+      }
+    }
+
+    private fun checkNullability(constructor: KFunction<*>, args: Array<*>) {
```
swankjesse, line 548: Finding a way to remember this information would be really nice. This isn’t a particularly efficient operation and doing it once per row is bad.

https://publicobject.com/2016/03/24/reflection-machines/

misk-hibernate/src/test/kotlin/misk/hibernate/AggregationQueryTest.kt
```diff
@@ -223,6 +223,36 @@ class AggregationQueryTest {
     )
   }
 
+  @Test fun `aggregations work when there are no matching rows`() {
```
swankjesse, line 226: You write the best tests

swankjesse (approved):

misk-hibernate/src/main/kotlin/misk/hibernate/ReflectionQueryFactory.kt
keeferrourke: Thanks for pointing out the performance impact here! I will give this some thought!

Strictly speaking, I don't actually think we'll run into this very often (maybe only if the mysql server is running in a weird mode and someone forgets to group their aggegration?). 

I could just remove this and let an NPE occur in the rare instance that this happens.

PR #3252, opened 2024-05-01
Update Job interface to expose delay functionality

din-cashapp:
Context: https://cash.slack.com/archives/C01M5M297T7/p1714144374605119
Summary: We want to expose backOffDelay() functionality to all users of Misk. In order to do that we need to update the Job interface and update FakeJobQueue.

### DIFF
--- a/misk-jobqueue/src/main/kotlin/misk/jobqueue/Job.kt
+++ b/misk-jobqueue/src/main/kotlin/misk/jobqueue/Job.kt
@@ -29,4 +29,7 @@ interface Job {
 
   /** Moves the job from the main queue onto the associated dead letter queue. May perform an RPC */
   fun deadLetter()
+
+  /** Assign a visibility timeout for the job that increases with exp. backoff */
+  fun backOffDelay()
 }
--- a/misk-jobqueue/src/main/kotlin/misk/jobqueue/Job.kt
+++ b/misk-jobqueue/src/main/kotlin/misk/jobqueue/Job.kt
@@ -30,6 +30,9 @@ interface Job {
   /** Moves the job from the main queue onto the associated dead letter queue. May perform an RPC */
   fun deadLetter()
 
-  /** Assign a visibility timeout for the job that increases with exp. backoff */
-  fun backOffDelay()
+  /**
+   * Assigns and applies a visibility timeout for the job by making it unreachable for some time duration.
+   * The timeout increases after each retry. May perform an RPC.
+   * */
+  fun delayWithBackoff()
 }
--- a/misk-jobqueue/src/testFixtures/kotlin/misk/jobqueue/FakeJobQueue.kt
+++ b/misk-jobqueue/src/testFixtures/kotlin/misk/jobqueue/FakeJobQueue.kt
@@ -17,6 +17,7 @@ import jakarta.inject.Singleton
 import java.lang.Long.max
 import java.util.concurrent.LinkedBlockingQueue
 import kotlin.jvm.Throws
+import kotlin.math.min
 
 /**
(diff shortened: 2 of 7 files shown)


misk-jobqueue/src/testFixtures/kotlin/misk/jobqueue/FakeJobQueue.kt
```diff
@@ -268,13 +290,34 @@ data class FakeJob(
...
+    internal set
+
   override fun acknowledge() {
     acknowledged = true
   }
 
   override fun deadLetter() {
     deadLettered = true
   }
+  override fun backOffDelay() {
```
din-cashapp, line 306: i honestly don't like this implementation this much, so would be happy to hear ideas/thoughts on the alternatives!

misk-jobqueue/src/testFixtures/kotlin/misk/jobqueue/FakeJobQueue.kt
```diff
@@ -268,13 +290,34 @@ data class FakeJob(
   var deadLettered: Boolean = false
     internal set
 
+  var backoffDelayedTime: Instant? = null
+    internal set
+
+  var delayDuration: Long? = null
+    internal set
```
tgregory-block, line 297: Instead of using these two tracking vars, what about:
1. Add `var delayedForBackoff: Boolean = false internal set`
2. In `backOffDelay()`, set `delayedForBackoff = true` and set `deliveryDelay` to the new delivery time
3. In `FakeJobQueue` put the job into `jobsToQueueBack` if `delayedForBackoff == true`
4. In `FakeJobQueue`, set `delayedForBackoff = false` on each job after re-enqueueing it

misk-jobqueue/src/testFixtures/kotlin/misk/jobqueue/FakeJobQueue.kt
```diff
@@ -268,13 +290,34 @@ data class FakeJob(
...
+    delayDuration = if (delayDuration == null) {
+      1000L
+    }  else {
+      max(delayDuration!! * 2, 10_000L)
+    }
+
+    val deliveryTime = if (clock.instant().isAfter(deliverAt)) {
+      clock.instant()
+    } else {
+      deliverAt
```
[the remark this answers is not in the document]
din-cashapp: ohh true yea, we should expect that deliverAt would already be in by the time we run backoffDelay()

misk-jobqueue/src/testFixtures/kotlin/misk/jobqueue/FakeJobQueue.kt
din-cashapp: oh yes this sounds much NICER!
updated the code (had to keep the delayDuration as we needed something to keep on increasing)

misk-jobqueue/src/main/kotlin/misk/jobqueue/Job.kt
```diff
@@ -29,4 +29,7 @@ interface Job {
 
   /** Moves the job from the main queue onto the associated dead letter queue. May perform an RPC */
   fun deadLetter()
+
+  /** Assign a visibility timeout for the job that increases with exp. backoff */
+  fun backOffDelay()
```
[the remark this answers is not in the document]
din-cashapp: sg updated the comment!

tgregory-block (approved):

frojasg (approved):
LGTM. I would like to get the review from @hamdanjaveed 🙏🏽

hamdanjaveed (approved):
LGTM

adrw (approved):

PR #3254, opened 2024-05-03
Update TurboStream.kt with extra Actions

SeabertYu:
The prev Enum is missing some functionalities so adding them in

### DIFF
--- a/misk-hotwire/src/main/kotlin/misk/turbo/TurboStream.kt
+++ b/misk-hotwire/src/main/kotlin/misk/turbo/TurboStream.kt
@@ -43,5 +43,7 @@ enum class TurboStreamAction {
   UPDATE,
   REMOVE,
   BEFORE,
-  AFTER
+  AFTER,
+  REFRESH,
+  MORPH,
 }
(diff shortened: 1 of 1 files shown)


misk-hotwire/src/main/kotlin/misk/turbo/TurboStream.kt
```diff
@@ -43,5 +43,10 @@ enum class TurboStreamAction {
...
   REMOVE,
   BEFORE,
-  AFTER
+  AFTER,
+  REFRESH,
+  MORPH,
 }
+
+// Content Type header needed for https://turbo.hotwired.dev/handbook/drive#streaming-after-a-form-submission
+val TURBO_STREAM_CONTENT_TYPE = "text/vnd.turbo-stream.html"
```
adrw, line 52: This is already in `MediaTypes` in Misk so this is duplicative.

adrw (approved):

adrw:
You'll probably need `gradle apiDump` and potentially other commands from the error logs.

PR #3255, opened 2024-05-04
Create Template.kt for turbo stream

SeabertYu:
The template tag was missing, causing the turbo stream is not usable.


adrw (approved):

adrw:
Thanks for adding this!

adrw:
You'll probably need `gradle apiDump` and potentially other commands from the error logs.

PR #3258, opened 2024-05-07
Add breaking changes log file

damar-block:
This is best effort for now, but the idea is to capture major breaking changes in this file (similar to how we had change log in the past).


BREAKING_CHANGES.md
```diff
@@ -0,0 +1,7 @@
+Breaking Changes
```
damar-block, line 1: Alternatively, can just use the existing change log? Although I don't see us populating it manually for every change. We might want to automate it during the publishing step in that case.

adrw: I'd lean towards doing it in the existing changelog, I have an open PR for docs site fixes that adds a description that we won't update changelog for every version but will try to for breaking changes so I like this addition.

adrw (approved):

frojasg (approved):

PR #3260, opened 2024-05-09
Disable tab links if unauthorized

adrw:
Currently, even if a caller is unauthorized to access a tab, they can click the link in the menu and navigate to it. When they do, they encounter the non-UI Misk standard text `unauthorized`. When they try to navigate back, it does not force a full page load due to the Turbo pre-load enhancements on the base links.

This PR disables the links (they route to the dashboard home which may have access information) and provides on hover tooltip text to explain why the link is not clickable. This will avoid this hard edge case while preserving the pre-loading benefits of Turbo links for tabs that the caller is authorized to access.

Screenshot 2024-05-16 at 09 38 52


uayyaz (approved):
lgtm

PR #3263, opened 2024-05-14
Update replica fallback logic

aparajon:
This PR does the following:
1. Removes Vitess from the `canRecoverOnReplica` path. This is currently causing an issue where a replica gets mistakenly used for performing writes.
3. Update error logs for when the data source fails to initialize. The current logs seem misleading since they indicate that a replica will always be attempted to fallback to, despite the fallback only applying to a subset of DB types.


frojasg (approved):
LGTM

PR #3272, opened 2024-05-22
Add @ExperimentalMiskApi

ean5533:
Shamelessly cribbed from [ExperimentalOkHttpApi](https://github.com/square/okhttp/blob/54238b4c713080c3fd32fb1a070fb5d6814c9a09/okhttp/src/main/kotlin/okhttp3/ExperimentalOkHttpApi.kt), this new annotation will allow us to better communicate the unstable nature of new public classes and methods. This will allow us to more rapidly iterate on these features without worrying as much about breaking changes, which is good both for code authors and reviewers. It's also good for Misk users who will have a clearer signal about what is and isn't safe to start using.

This PR also demonstrates usage of the annotation by applying it to `TaggedLogger`, a class which was created recently and is being iterated on [right now](https://github.com/cashapp/misk/pull/3270).

### DIFF
--- a/misk-api/api/misk-api.api
+++ b/misk-api/api/misk-api.api
@@ -1,3 +1,6 @@
+public abstract interface annotation class misk/annotation/ExperimentalMiskApi : java/lang/annotation/Annotation {
+}
+
 public abstract interface class misk/api/HttpRequest {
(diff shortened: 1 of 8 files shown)


adrw (approved):
Great addition, thanks for landing this!

Could you add a note on this to the `changelog.md` since it could surface as errors in some people's builds?

PR #3276, opened 2024-05-24
Allow isolation levels to be set a jooq transaction

Sethuraman:
This change allows a transaction isolation level to be set for a jooq transaction. 

The DSL context creation has also moved inside the JooqTransacter class in order to allow the isolation level to be set.


misk-jooq/src/main/kotlin/misk/jooq/JooqTransacter.kt
```diff
@@ -61,11 +83,12 @@ class JooqTransacter(
   }
 
   private fun <RETURN_TYPE> createDSLContextAndCallback(
+    options: TransacterOptions,
     callback: (jooqSession: JooqSession) -> RETURN_TYPE
   ): RETURN_TYPE {
     var jooqSession: JooqSession? = null
     return try {
-      dslContext.value.transactionResult { configuration ->
+      dslContext(dataSourceService, clock, dataSourceConfig, options).transactionResult { configuration ->
```
davidgarth, line 91: is this doing extra work on each transaction and is there a chance it could slow things down? one thing that looks unnecessary is creating the settlings and mappings.

davidgarth (approved):

misk-jooq/src/main/kotlin/misk/jooq/JooqTransacter.kt
Sethuraman: This actually hasn't changed much. Previously the creation of the DSL context was 'lazy'. The DSL context was being created every time before as well.

davidgarth: we discussed this offline, looks like it is doing no additional work here.

misk-jooq/src/main/kotlin/misk/jooq/JooqTransacter.kt
```diff
@@ -9,14 +9,36 @@ import com.github.michaelbull.retry.policy.limitAttempts
...
+import org.jooq.impl.DataSourceConnectionProvider
+import org.jooq.impl.DefaultExecuteListenerProvider
+import org.jooq.impl.DefaultTransactionProvider
 import wisp.logging.getLogger
+import java.time.Clock
 import kotlin.coroutines.coroutineContext
 
 class JooqTransacter(
-  private val dslContext: Lazy<DSLContext>
+  private val dataSourceService: DataSourceService,
```
adrw, line 35: Is this change backwards compatible? Is JooqTransacter used downstream anywhere? Looks like those call sites would break with this change.

Sethuraman: The JooqTransacter should be Injected in. I don't imagine anyone is instantiating a JooqTranasacter. In the off chance that someone is, this breaking change could / should force them to reconsider how they are using the lib.

PR #3277, opened 2024-05-24
Add `ResourceLoader.bytes`

mericson-square:
Add support for binary resources by adding `ResourceLoader.bytes` and `ResourceLoader.requireBytes`.


adrw (approved):
Thanks for upstreaming this!

adrw (approved):

PR #3280, opened 2024-05-30
Add support for src and loading=lazy attributes to TurboFrame

mslogar-squareup:
See https://turbo.hotwired.dev/handbook/frames for documentation on these features


keeferrourke (approved):
Let's goo

adrw (approved):
Nice!

PR #3285, opened 2024-06-05
Pretty-print the service graph upon start-up

keeferrourke:
It can at times be difficult to reason about what the dependency graph is in larger misk services. This change logs the ServiceGraph when the ServiceManager is created during service start-up, in order to allow service operators to better understand their constituent service dependencies.

By default, no graph is printed on service start-up, but this can be opted into by passing a `ServiceManagerConfig` to `MiskRealServiceModule`.

### DIFF
--- a/misk-service/src/main/kotlin/misk/ServiceGraphBuilder.kt
+++ b/misk-service/src/main/kotlin/misk/ServiceGraphBuilder.kt
@@ -72,7 +72,7 @@ internal class ServiceGraphBuilder {
    */
   private fun linkDependencies() {
     for ((key, service) in serviceMap) {
-      val dependencies = dependencyMap[key]?.map { serviceMap[it]!! } ?: listOf()
+      val dependencies = dependencyMap[key].map { serviceMap[it]!! }
       service.addDependentServices(*dependencies.toTypedArray())
     }
   }
--- a/misk-service/src/main/kotlin/misk/ServiceGraphBuilder.kt
+++ b/misk-service/src/main/kotlin/misk/ServiceGraphBuilder.kt
@@ -15,7 +15,7 @@ internal class ServiceGraphBuilder {
   private val serviceMap = mutableMapOf<Key<*>, CoordinatedService>()
   private val serviceNames = mutableMapOf<Key<*>, String>()
 
-  // A map of downstream services -> their upstreams.
+  /** A map of downstream services -> their upstreams. */
   private val dependencyMap = LinkedHashMultimap.create<Key<*>, Key<*>>()
 
   /**
--- a/misk-service/src/main/kotlin/misk/ServiceGraphBuilder.kt
+++ b/misk-service/src/main/kotlin/misk/ServiceGraphBuilder.kt
@@ -12,7 +12,10 @@ import com.google.inject.Provider
  * services are ready.
  */
 internal class ServiceGraphBuilder {
-  private var serviceMap = mutableMapOf<Key<*>, CoordinatedService>()
+  private val serviceMap = mutableMapOf<Key<*>, CoordinatedService>()
+  private val serviceNames = mutableMapOf<Key<*>, String>()
+
+  // A map of downstream services -> their upstreams.
   private val dependencyMap = LinkedHashMultimap.create<Key<*>, Key<*>>()
 
   /**
@@ -22,14 +25,15 @@ internal class ServiceGraphBuilder {
    * Keys must be unique. If a key is reused, then the original key-service pair will be replaced.
    */
   fun addService(key: Key<*>, service: Service) {
-    addService(key) { service }
+    addService(key, service::class.qualifiedName) { service }
   }
 
-  fun addService(key: Key<*>, serviceProvider: Provider<out Service>) {
+  fun addService(key: Key<*>, serviceName: String?, serviceProvider: Provider<out Service>) {
     check(serviceMap[key] == null) {
       "Service $key cannot be registered more than once"
     }
     serviceMap[key] = CoordinatedService(serviceProvider)
+    serviceNames[key] = serviceName ?: "Anonymous Service(${key.typeLiteral.type.typeName})"
   }
 
   /**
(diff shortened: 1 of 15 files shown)


misk-service/src/main/kotlin/misk/ServiceManagerModule.kt
```diff
@@ -21,15 +20,13 @@ class ServiceManagerModule : KAbstractModule() {
     newMultibinder<Service>()
     newMultibinder<ServiceManager.Listener>()
 
-    multibind<ServiceManager.Listener>().toProvider(
-      Provider<ServiceManager.Listener> {
```
adrw, line 25: Did this even work before? Or is the syntax change a no-op?

misk-service/src/main/kotlin/misk/ServiceGraphBuilder.kt
```diff
@@ -6,13 +6,16 @@ import com.google.common.util.concurrent.ServiceManager
...
+import wisp.logging.getLogger
 
 /**
  * Builds a graph of [CoordinatedService]s which defer start up and shut down until their dependent
  * services are ready.
  */
 internal class ServiceGraphBuilder {
   private var serviceMap = mutableMapOf<Key<*>, CoordinatedService>()
+
+  // A map of downstream services -> their upstreams.
```
adrw, line 18:
```suggestion
  /** A map of downstream services -> their upstreams. */
```

Probably can use proper Kotlin docs since this is a val not within a function.

misk-service/src/main/kotlin/misk/ServiceGraphBuilder.kt
```diff
@@ -100,4 +104,51 @@ internal class ServiceGraphBuilder {
...
+      append(if (isLast) " \\__ " else "|__ ")
+    }
+    append(serviceNameOf(serviceKey))
+    append("\n")
+    visited.add(serviceKey)
+
+    val downstreamServices = dependencyMap.asMap().filter { it.value.contains(serviceKey) }.keys.toList()
+    for ((index, downstreamService) in downstreamServices.withIndex()) {
+      val newPrefix = prefix + if (isLast) "    " else "|   "
+      depthFirst(downstreamService, visited, newPrefix, isLast = (index == downstreamServices.size - 1), isRoot = false)
```
adrw, line 140: Could we do named params for all parameters?

misk-service/src/main/kotlin/misk/ServiceGraphBuilder.kt
```diff
@@ -100,4 +104,51 @@ internal class ServiceGraphBuilder {
...
   }
+
+  override fun toString(): String = buildString {
+    val visited = mutableSetOf<Key<*>>()
+    val allServices = serviceMap.keys
+    val serviceRoots = allServices.filterNot { it in dependencyMap.keys() }
+
+    for (root in serviceRoots) {
+      if (root !in visited) {
+        depthFirst(root, visited, "", isLast = true, isRoot = true)
```
adrw, line 115: Could we do named params for all parameters?

adrw (approved):
Thanks for adding this. Not sure how big the graph will be for production services but maybe it'd be worth making this configurable by yaml since it will be log output on every pod startup?

keeferrourke:
>Thanks for adding this. Not sure how big the graph will be for production services but maybe it'd be worth making this configurable by yaml since it will be log output on every pod startup?

I thought about that, and will do!

misk/src/main/kotlin/misk/MiskServiceModule.kt
```diff
@@ -21,28 +21,35 @@ import misk.tokens.TokenGeneratorModule
  * with [MiskTestingServiceModule]. Only bindings that are not suitable for a unit testing
  * environment belong here.
  */
-class MiskRealServiceModule : KAbstractModule() {
+class MiskRealServiceModule(
+  private val serviceManagerConfig: ServiceManagerConfig,
+) : KAbstractModule() {
+  constructor(): this(ServiceManagerConfig())
```
keeferrourke, line 27: These secondary constructors should be good enough for backwards compatibility — anyone who wants to add this behaviour will need to explicitly opt in by providing a ServiceManagerConfig.

adrw: I'd probably lean towards having the config param have a default config value than secondary constructor. A bit easier to reason about where default values are coming from.

misk-service/src/main/kotlin/misk/ServiceManagerModule.kt
```diff
@@ -76,6 +76,9 @@ class ServiceManagerModule : KAbstractModule() {
     }
 
     val serviceManager = builder.build()
+    if (serviceManagerConfig.debug_service_graph) {
```
adrw, line 79: Would this fail if the no-config constructor is used? Since the val wouldn't exist or be initialized?

misk/src/main/kotlin/misk/MiskServiceModule.kt
```diff
@@ -21,28 +21,35 @@ import misk.tokens.TokenGeneratorModule
...
+    install(MiskCommonServiceModule(serviceManagerConfig))
   }
 }
 
 /**
  * This module has common bindings for all environments (both real and testing).
  */
-class MiskCommonServiceModule : KAbstractModule() {
+class MiskCommonServiceModule(
+  private val serviceManagerConfig: ServiceManagerConfig,
```
adrw, line 44: Might want to have this include a default too instead of the secondary constructor? Or once it's rolled out internally, we should remove the secondary constructor.

misk-service/src/main/kotlin/misk/ServiceManagerModule.kt
keeferrourke: It shouldn't, the config data class has a default value of `false` for this property, and the no-arg constructor for this module uses a default construction of the config class.

misk/src/main/kotlin/misk/MiskServiceModule.kt
keeferrourke: Sure I can change this to a single constructor with `@JvmOverloads`.

keeferrourke: This had an implication on the `projectHealth` check (see https://github.com/cashapp/misk/pull/3285/commits/fe2b1892c24c0faf857110b2af6f409afd94c283), but I think it's not a problem, since the dependency is still pulled in — seems like more of a misk samples specific linting error?

PR #3286, opened 2024-06-05
Improves calculateTypes: adds returnTypes, recurses into map value ty…

brunofrts:
…pes and generate Kotlin test classes through Wire.

This PR

1. Introduces `returnTypes` (similar to `responseTypes`) in `WebActionMetadata`. This is useful when the return of a web action is not wrapped (e.g., in a `Response`).
2. Removes the hardcoded Kotlin Wire compiled test classes in `misk-admin` and instead generates them with Gradle as needed. To do this, we duplicate the .proto definitions using different package names so that Wire can process both with different outputs.
3. Implements Map handling in `calculateTypes` similarly to how we parse Lists.


misk-admin/src/test/kotlin/misk/web/metadata/MiskWebFormBuilderTest.kt
```diff
@@ -50,6 +54,10 @@ internal class MiskWebFormBuilderTest {
       )
     )
 
+    // Check map value types
```
alvinsee, line 57: this isnt really testing the map type right?

alvinsee (approved):

misk-admin/src/test/kotlin/misk/web/metadata/MiskWebFormBuilderTest.kt
```diff
@@ -50,6 +54,13 @@ internal class MiskWebFormBuilderTest {
       )
     )
 
+    // Check map value types are included
```
adrw, line 57: Could you add assertions that the key type (in this case int32) is also present in the calculated types?

adrw (approved):
LG

oldergod:
I wrote a `fixes #<this PR>` in the related Wire PR and this closed this issue.... Sorry about it.

PR #3287, opened 2024-06-07
Print repeat subtrees in the service graph

keeferrourke:
Follow-on for #3285.

The previous implementation was overly restrictive and did not print repeat nodes in the service dependency tree. I thought this was fine before, but there can be similar (but different) subtrees, which need to be printed to get a full picture of the service.

This change stops tracking the visited nodes in each tree, and traverses them all anyway. This is safe to do, because we guarantee at this point that the graph is acyclic.

Note: This change will make the debug tree output much longer, and more verbose. We could probably try to be clever to detect identical subtrees to make the output more concise, but I don't feel like that is worth the effort at this time.


adrw (approved):
LG

PR #3310, opened 2024-06-17
Make metadata classes internal

adrw:
Metadata (JSON representation of various parts of the Misk application) is still under active development. Use of `internal` limits public use and prevents potential breaking changes for consumers who inadvertently start to rely on it outside of the exposed API / admin dashboard.


build.gradle.kts
```diff
@@ -236,7 +236,6 @@ subprojects {
       exceptionFormat = TestExceptionFormat.FULL
       showStandardStreams = false
     }
-    dependsOn(":startRedis")
```
adrw, line 239: Starting Redis via Docker before running any test is annoying locally. I think I've added it as a dependency in the modules that require it which has sped up local development already.

misk-admin/src/main/kotlin/misk/web/metadata/config/ConfigMetadataAction.kt
```diff
@@ -11,9 +11,9 @@ import misk.web.dashboard.AdminDashboardAccess
 import misk.web.mediatype.MediaTypes
 
 @Singleton
-class ConfigMetadataAction @Inject constructor(
-  private val metadataProvider: Provider<ConfigMetadata>
-) : WebAction {
+class ConfigMetadataAction @Inject constructor() : WebAction {
+  @Inject private lateinit var metadataProvider: Provider<ConfigMetadata>
```
adrw, line 15: Field injection allows use to make the metadata `internal`.

danieloh0714 (approved):

misk-api/api/misk-api.api
```diff
@@ -85,3 +85,6 @@ public final class misk/web/DispatchMechanism : java/lang/Enum {
 	public static fun values ()[Lmisk/web/DispatchMechanism;
 }
 
+public abstract interface annotation class misk/web/dashboard/AdminDashboardAccess : java/lang/annotation/Annotation {
```
adrw, line 88: Moved this annotation to `misk-api` so that endpoints can use it to set access without pulling in the entire `misk-admin` subproject.

misk/src/main/kotlin/misk/web/metadata/webaction/WebActionMetadataAction.kt
```diff
@@ -11,9 +11,9 @@ import misk.web.dashboard.AdminDashboardAccess
 import misk.web.mediatype.MediaTypes
 
 @Singleton
-class WebActionMetadataAction @Inject constructor(
-  val provider: Provider<WebActionsMetadata>
-) : WebAction {
+class WebActionMetadataAction @Inject constructor() : WebAction {
+  @Inject private lateinit var provider: Provider<WebActionsMetadata>
```
adrw, line 15: Field injection lets us make the metadata `internal`

PR #3312, opened 2024-06-17
Remove Config Misk-Web tab, use Metadata tab instead

adrw:
This removes the existing Misk-Web Config admin dashboard tab in favor of the new Metadata tab which is built with the new UI stack (kotlinx HTML and Hotwire). The Config link will still be in the admin dashboard for UX purposes for a while as people learn about the new Metadata tab and find the same YAML config information they're looking for in the new location.

This leaves the Database tab and Web Actions tab as remaining Misk-Web code to gradually replace in Misk for both security, velocity, functionality, and ongoing maintenance benefits.

- [x] Wait for #3305, #3313, #3314, #3315 to land and rebase
- [x] Refactor tests

Screenshot 2024-06-18 at 01 17 44


misk-admin/src/main/kotlin/misk/web/metadata/config/ConfigMetadataAction.kt
```diff
@@ -1,29 +1,11 @@
...
-import misk.web.ResponseContentType
 import misk.web.actions.WebAction
-import misk.web.dashboard.AdminDashboardAccess
-import misk.web.mediatype.MediaTypes
 
 @Singleton
 class ConfigMetadataAction @Inject constructor() : WebAction {
-  @Inject private lateinit var metadataProvider: Provider<ConfigMetadata>
-
-  @Get("/api/v1/config/metadata")
```
adrw, line 17: API is only used in the Misk-Web tab.

danieloh0714 (approved):

PR #3317, opened 2024-06-18
Delete misk-events-testing which has been replaced by misk-events test fixtures

aerb:
Internal usages have been removed: [query](https://github.com/search?type=code&q=org%3Asquareup+NOT+is%3Aarchived+miskEventsTesting)


adrw (approved):
Love to see it, exciting progress.

adrw:
Probably need to delete it from the settings Gradle too so it isn't included in the build?

aerb:
> Probably need to delete it from the settings Gradle too so it isn't included in the build?

Ah yes, thanks!

staktrace (approved):

PR #3318, opened 2024-06-18
Add contract to requireRequest

ean5533:
Allows for more smart compiler checks after `requireRequest` is called


misk-actions/src/main/kotlin/misk/exceptions/Exceptions.kt
```diff
@@ -124,13 +124,17 @@ open class UnsupportedMediaTypeException @JvmOverloads constructor(message: Stri
...
 inline fun requireRequest(check: Boolean, lazyMessage: () -> String) {
+  contract { 
+    returns() implies check
+  }
   if (!check) throw BadRequestException(lazyMessage())
 }
 
 /** Similar to [kotlin.requireNotNull], but throws [BadRequestException] if the check fails */
 @OptIn(ExperimentalContracts::class)
-public inline fun <T : Any> requireRequestNotNull(value: T?, lazyMessage: () -> String): T {
```
ean5533, line 133: `public` is redundant here

adrw (approved):
It'd be cool to see what the new error is using this vs current.

ean5533:
@adrw This should bring Misk's `requireRequest` into alignment with the stdlib's `require`. For example:

image

PR #3326, opened 2024-06-24
Distribute cron tasks

tgregory-block:
Currently Cron tasks are governed by a single lease, so all tasks are executed on whichever pod takes that lease. This change adds a new `CronLeaseBehavior` parameter to `CronModule`, making it possible to take out a lease per-cron-task and distribute the tasks across the deployment. The default behavior remains to take out a single lease

### DIFF
--- a/misk-cron/src/main/kotlin/misk/cron/CronManager.kt
+++ b/misk-cron/src/main/kotlin/misk/cron/CronManager.kt
@@ -7,6 +7,8 @@ import com.cronutils.parser.CronParser
 import jakarta.inject.Inject
 import jakarta.inject.Singleton
 import misk.cron.CronManager.CronEntry.ExecutionTimeMetadata.Companion.toMetadata
+import wisp.lease.Lease
+import wisp.lease.LeaseManager
 import wisp.logging.getLogger
 import java.time.Clock
 import java.time.Instant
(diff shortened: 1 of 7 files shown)


misk-cron/src/main/kotlin/misk/cron/CronManager.kt
```diff
@@ -114,6 +138,45 @@ class CronManager @Inject constructor() {
...
+          "CronJob ${cronEntry.name} was ready at $nextExecutionTime"
+        }
+        runCron(cronEntry)
+        taskLease.release()
+      }
+    }
+  }
+
+  internal fun buildTaskLease(klass: KClass<out Runnable>): Lease {
+    val sanitizedClassName = klass.qualifiedName!!.replace(".", "-")
```
keeferrourke, line 175: There are some gotchas here with inner/anonymous classes. You might end up with some very strange looking lease names. Is this replacement just for readability or are there actual characters/sequences you intend to avoid in the lease name string?

tgregory-block: We mainly weren't sure if the partition key had forbidden characters, and since the dynamo lease implementation lives in `Skim` it wasn't straightforward to write an integration test. Looking more thoroughly at the [dynamo docs](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/HowItWorks.NamingRulesDataTypes.html), it doesn't seem like that's the case though.

tgregory-block: Removed the sanitization, `.` are clearly fine since the original cluster wide lease uses them

PR #3327, opened 2024-06-24
Log x-request-id header

kalyancash:


misk/src/main/kotlin/misk/client/ClientLoggingInterceptor.kt
```diff
@@ -44,6 +44,7 @@ import jakarta.inject.Singleton
       "x-datadog-parent-id",
       "x-datadog-trace-id",
       "x-datadog-span-id",
+      "x-request-id",
```
adrw, line 47: Could you provide some more context to the PR description as to why this is being added?

kalyancash: Added!

frojasg (approved):
LGTM

adrw (approved):

PR #3329, opened 2024-06-25
Add environment variable support to the wisp resource loader

jessejacksonafterpay:
This PR adds a new wisp resource loader for loading environment variables. An example usage might look like this:
```
data_source_clusters:
  afterpay-misk-exemplar-service-001:
    writer:
      type: MYSQL
      username: "environment:DB_USERNAME"
      password: "environment:DB_PASSWORD"
      database: "environment:DB_NAME"
      migrations_resource: "classpath:/migrations"
```

In order to facilitate this new loader, the validation of the `path` component of a resource was separated from the top level validation, as an environment variable does not have a path-like pattern. This is a non-breaking change, as all existing validation works as it previously did.

The JUnit Pioneer library was introduced to the tests in order to facilitate testing of environment variables.


jvmakine (approved):
Nice!

shellz-n-stuff (approved):
MVP

PR #3330, opened 2024-06-25
Install the MetadataModule only once for cron metadata

mmollaverdi:
It's being installed both from the real and fake cron modules, and for some reason the fake module is being installed by the real one, meaning any binding of the real module results in MetadataModule being installed twice and failing with a duplicate key error.

### DIFF
--- a/misk-cron/src/main/kotlin/misk/cron/CronModule.kt
+++ b/misk-cron/src/main/kotlin/misk/cron/CronModule.kt
@@ -14,8 +14,6 @@ import java.time.ZoneId
 import jakarta.inject.Qualifier
 import misk.inject.KAbstractModule
 import misk.inject.KInstallOnceModule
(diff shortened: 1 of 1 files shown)


misk-cron/src/main/kotlin/misk/cron/CronModule.kt
```diff
@@ -24,7 +22,6 @@ class CronModule @JvmOverloads constructor(
   private val dependencies: List<Key<out Service>> = listOf()
 ) : KInstallOnceModule() {
   override fun configure() {
-    install(FakeCronModule(zoneId, threadPoolSize, dependencies))
```
adrw, line 27: I’d suggest removing the MetadataModule install line from the RealMoxule instead. Less likely to cause real breakages.

adrw (approved):
Thanks!

PR #3333, opened 2024-06-26
Allow bypassing of the relay proxy if no base URI is specified

jessejacksonafterpay:
When using the `LaunchDarklyModule` directly, there is no ability to pass in a blank value for the `base_uri` in order to bypass usage of the relay proxy.

This PR allows for a blank URL which doesn't use the relay proxy, and instead uses the standard Launch Darkly config.


misk-launchdarkly/src/main/kotlin/misk/feature/launchdarkly/LaunchDarklyFeatureFlagsModule.kt
```diff
@@ -61,13 +61,17 @@ class LaunchDarklyModule @JvmOverloads constructor(
...
     //   apps who're installing this module but not even using the misk or wisp LaunchDarklyFeatureFlags.
-    val baseUri = URI.create(config.base_uri)
     val ldConfig = LDConfig.Builder()
       // Set wait to 0 to not block here. Block in service initialization instead.
       .startWait(Duration.ofMillis(0))
       .dataSource(Components.streamingDataSource())
       .events(Components.sendEvents())
-      .serviceEndpoints(Components.serviceEndpoints().relayProxy(baseUri))
+
+    if (config.base_uri.isNotBlank()) {
```
adrw, line 70: I wonder if this should be an explicit Boolean rather than magic behavior of passing in a null or empty string? Would it be bad if someone accidentally did this and their service unintentionally wasn't using the relay?

jessejacksonafterpay: Certainly, could introduce a `use_relay_proxy` or similar? I think it's not *too* bad in it's current form,. because it's essentially an override for the base LD config. I think if you're using this directly and didn't provide a value, you wouldn't be surprised by it not using the relay.

jessejacksonafterpay: @adrw have added a new (optional) config to explicitly manage this (and defaulted it to `true` for backward compatibility).

misk-launchdarkly/src/main/kotlin/misk/feature/launchdarkly/LaunchDarklyFeatureFlagsModule.kt
```diff
@@ -89,5 +94,6 @@ data class LaunchDarklyConfig @JvmOverloads constructor(
   @Redact
   val sdk_key: String,
   val base_uri: String,
+  val use_relay_proxy: Boolean? = true,
```
adrw, line 97: What would null signify? Can it just be non-null?

jessejacksonafterpay: I think if I make it non-null, that would force somebody to provide that in the config right? Or have I misunderstood?

I was trying to avoid anybody with existing config from having to provide this.

jessejacksonafterpay: All good - tested this out and was able to make it non-null with a default value to get the same behaviour. The config value can be omitted and will fallback to `true` (the way it works today).

jvmakine (approved):

PR #3335, opened 2024-06-26
Add Guice Metadata that shows DI bindings

adrw:
First PR will add Guice bindings to Misk's Metadata. This will show up
in the Metadata admin dashboard tab and be exposed in the metadata APIs.

Follow up PRs may add a more specialized UI which turns this metadata
into a browsable tree chart or adds search etc.

Screenshot 2024-06-26 at 17 58 53

Screenshot 2024-06-26 at 17 59 10

### DIFF
--- a/misk/src/main/kotlin/misk/web/metadata/guice/GuiceMetadata.kt
+++ b/misk/src/main/kotlin/misk/web/metadata/guice/GuiceMetadata.kt
@@ -3,6 +3,11 @@ package misk.web.metadata.guice
 import com.google.inject.Binding
 import com.google.inject.Injector
 import com.google.inject.Key
+import com.google.inject.spi.ConstructorBinding
+import com.google.inject.spi.InstanceBinding
+import com.google.inject.spi.LinkedKeyBinding
+import com.google.inject.spi.ProviderInstanceBinding
+import com.google.inject.spi.ProviderKeyBinding
 import jakarta.inject.Inject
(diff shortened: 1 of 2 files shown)


misk/src/main/kotlin/misk/web/metadata/guice/GuiceMetadata.kt
```diff
@@ -0,0 +1,163 @@
...
+    .toFormattedJson(guice.grouped),
+  descriptionString = "Direct injection bindings, powered by Guice. This metadata is work-in-progress."
+)
+
+internal class GuiceMetadataProvider : MetadataProvider<GuiceMetadata> {
+  @Inject lateinit var injector: Injector
+
+  override val id = "guice"
+
+  internal data class Metadata(
```
mpawliszyn, line 28: I wonder if this data can ever get too big to a point where we might need to page through?

mpawliszyn (approved):
Looks great though!

damar-block (approved):

misk/src/main/kotlin/misk/web/metadata/guice/GuiceMetadata.kt
adrw: Yeah might be a risk though I did strip a lot of data out of the `prettyPrint` version so that should likely be fine.

I agree though that the Metadata primitive could end up requiring paging, will need to give some thought on how best to do that.

PR #3337, opened 2024-06-27
Don't log the `UnrecognizedPropertyException` for warnings

mmollaverdi:
It adds a fair bit of noise in tests. The error message itself is clear enough.

E.g.
```
15:33:59.994 [Test worker] WARN misk.config.MiskConfig -- xyz-testing.yaml: 'abc' not found in 'XyzConfig', ignoring 

com.fasterxml.jackson.databind.exc.UnrecognizedPropertyException: Unrecognized field "streams" (class com.squareup.cash.xyz.service.XyzConfig), not marked as ignorable (6 known properties: ...])
 at [Source: (StringReader); line: 1, column: 1674] (through reference chain: 
....
```

### DIFF
--- a/misk-config/src/main/kotlin/misk/config/MiskConfig.kt
+++ b/misk-config/src/main/kotlin/misk/config/MiskConfig.kt
@@ -155,7 +155,7 @@ object MiskConfig {
       }
 
       val path = Joiner.on('.').join(e.path.map { it.fieldName ?: it.index })
-      logger.warn(e) {
+      logger.warn {
         "$configFile: '$path' not found in '${configClass.simpleName}', ignoring " +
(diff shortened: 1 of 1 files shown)


misk-config/src/main/kotlin/misk/config/MiskConfig.kt
```diff
@@ -155,7 +155,7 @@ object MiskConfig {
       }
 
       val path = Joiner.on('.').join(e.path.map { it.fieldName ?: it.index })
-      logger.warn(e) {
+      logger.warn {
```
adrw, line 158: On the flip side the full stack trace probably shows where the call site is trying to access the non-existent field?

PR #3346, opened 2024-06-30
[Chore] Google Library Updates

shellz-n-stuff:
## Description
This pull request includes updates to the versions of several dependencies in the `gradle/libs.versions.toml` file. The updated dependencies are `gcpDatastore`, `gcpSpanner`, and `googleApiClient`.

* [`gradle/libs.versions.toml`](diffhunk://#diff-697f70cdd88ba88fe77eebda60c7e143f6ad1286bca75017421e93ad84fb87dfL54-R59): Updated the `gcpDatastore` module from version `1.107.1` to `2.2.10`, the `gcpSpanner` module from version `6.13.0` to `6.23.4`, and the `googleApiClient` module from version `1.32.1` to `1.35.0`.

These updates are aiming to address CVEs in the downstream libraries.

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -51,12 +51,12 @@ flywayGradlePlugin = { module = "org.flywaydb:flyway-gradle-plugin", version.ref
 gax = { module = "com.google.api:gax", version = "2.5.0" }
 gcpCloudCore = { module = "com.google.cloud:google-cloud-core", version = "1.95.4" }
 gcpCloudStorage = { module = "com.google.cloud:google-cloud-storage", version = "1.117.1" }
-gcpDatastore = { module = "com.google.cloud:google-cloud-datastore", version = "1.107.1" }
+gcpDatastore = { module = "com.google.cloud:google-cloud-datastore", version = "2.2.10" }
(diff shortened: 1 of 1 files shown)


gradle/libs.versions.toml
```diff
@@ -51,12 +51,12 @@ flywayGradlePlugin = { module = "org.flywaydb:flyway-gradle-plugin", version.ref
 gax = { module = "com.google.api:gax", version = "2.5.0" }
 gcpCloudCore = { module = "com.google.cloud:google-cloud-core", version = "1.95.4" }
 gcpCloudStorage = { module = "com.google.cloud:google-cloud-storage", version = "1.117.1" }
-gcpDatastore = { module = "com.google.cloud:google-cloud-datastore", version = "1.107.1" }
+gcpDatastore = { module = "com.google.cloud:google-cloud-datastore", version = "2.2.10" }
```
adrw, line 54: This is a big version jump, any migrations required?

adrw (approved):

gradle/libs.versions.toml
shellz-n-stuff: The API Safety check seemed to pass for this. My local setup was having issues running some of the DB connectivity tests but those occurred on main too.

I think for safety it would be great to just gradle build off someone else's machine to validate if these are env specific or if they are related to this change.

PR #3348, opened 2024-07-01
CCPOKR-14318 | [Security][High] - Incomplete string escaping or encoding

katukota:
Example string: "Enum<app.cash.backfila.BackfillType,ISOLATED,PARALLEL>"


misk-admin/web/tabs/web-actions/src/rewrite/WebActionSendRequestFormComponents.tsx
```diff
@@ -126,7 +126,7 @@ function FormEnumComponent({
   onChange
 }: FormComponentProps<string>) {
   const enumValues = field.type
-    .replace(">", "")
+    .replace(/>/g, "")
```
adrw, line 129: I'm not familiar with this syntax and the tests for the admin dashboard tabs are insufficient to merge confidently. 

Have you tried building the tab locally and then running the exemplar service locally to confirm it still works as expected?

katukota: Just ran this locally and it works fine. 

Without the change:
Screenshot 2024-07-03 at 4 34 27 PM

With change:
Screenshot 2024-07-03 at 4 48 45 PM

PR #3349, opened 2024-07-01
[SQS] Add BatchJobHandler

Rizziepit:


misk-aws/src/main/kotlin/misk/jobqueue/sqs/SqsJobConsumer.kt
```diff
@@ -88,50 +99,40 @@ internal class SqsJobConsumer @Inject internal constructor(
...
 
-  internal inner class QueueReceiver(
-    queueName: QueueName,
-    private val handler: JobHandler
-  ) {
-    private val queue = queues.getForReceiving(queueName)
-    private val shouldKeepRunning = AtomicBoolean(false)
+  internal abstract inner class QueueReceiver(queueName: QueueName) {
+    val queue = queues.getForReceiving(queueName)
+    private val shouldKeepRunning = AtomicBoolean(true)
```
Rizziepit, line 104: I'm pretty sure `AtomicBoolean(false)` is a bug, but we didn't notice the bug because there's another bug in `QueueReceiver::run` below that means we never return `Status.NO_SCHEDULE`. I'm fixing both.

Safe shutdown was added in https://github.com/cashapp/misk/pull/2758/

misk-aws/src/main/kotlin/misk/jobqueue/sqs/SqsJobConsumer.kt
```diff
@@ -166,11 +167,37 @@ internal class SqsJobConsumer @Inject internal constructor(
...
+          .fold(Status.NO_WORK) { finalStatus, status ->
+            when {
+              status == Status.FAILED -> status
+              status == Status.OK && finalStatus != Status.FAILED -> status
+              status == Status.NO_WORK && finalStatus == Status.NO_WORK -> status
+              else -> finalStatus
+            }
+          }
+      }.join()
+    }
```
Rizziepit, line 197: This logic was moved as-is

misk-aws/src/test/kotlin/misk/jobqueue/sqs/DockerSqs.kt
```diff
@@ -85,7 +86,7 @@ internal object DockerSqs : ExternalDependency {
       try {
         client.getQueueUrl("does not exist")
       } catch (e: Exception) {
-        if (e is QueueDoesNotExistException) {
+        if (e is QueueDoesNotExistException || (e is AmazonSQSException && e.statusCode == 400)) {
```
Rizziepit, line 89: My tests were failing locally because the SQS API was not throwing `QueueDoesNotExistException`. Instead it threw `AmazonSQSException` with an error message "AWS.SimpleQueueService.NonExistentQueue; ...<more words>". The error code was not set to something useful. Should we keep this workaround?

misk-aws/src/main/kotlin/misk/jobqueue/sqs/SqsJobConsumer.kt
```diff
@@ -231,6 +253,71 @@ internal class SqsJobConsumer @Inject internal constructor(
...
+
+    @OptIn(ExperimentalMiskApi::class)
+    private fun handleMessages(messages: List<SqsJob>): Status {
+      metrics.jobsReceived.labels(queue.queueName, queue.queueName).inc(messages.size.toDouble())
+
+      return tracer.traceWithNewRootSpan("handle-job-${queue.queueName}") { span ->
+        // Run the handler and record timing
+        try {
+          messages.forEachIndexed { index, message ->
+            MDC.put("$SQS_JOB_IDS_STRUCTURED_MDC.$index", message.id)
```
Rizziepit, line 295: How can I log a list of IDs? DD supports lists in log attributes but I don't know what key format to use.

misk-aws/src/main/kotlin/misk/jobqueue/sqs/SqsJobConsumer.kt
```diff
@@ -231,6 +253,71 @@ internal class SqsJobConsumer @Inject internal constructor(
...
+
+    override fun receive(): List<Status> {
+      // Repeatedly long poll for messages until we have batchSize messages.
+      // If there are no messages available, short circuit further requests.
+      val cutoff = TimeSource.Monotonic.markNow() + 20.seconds // TODO make configurable
+      val batch: MutableList<SqsJob> = mutableListOf()
+      while (batch.size < batchSize() && !cutoff.hasPassedNow()) {
+        val messages = fetchMessages(
+          sqsBatchSize = min(10, batchSize() - batch.size),
+          waitTimeSeconds = 5
```
Rizziepit, line 272: I need to change this to `min(secondsPerBatch, secondsUntilCutoff)`. That achieves two things:
* We need to limit `secondsPerBatch` such that we issue at least 5 requests for batchSize 50, 4 requests for batchSize 40, etc. before we reach the cutoff.
* We need `secondsUntilCutoff` to avoid exceeding the max total wait time. It also has the benefit of enabling short polling by setting max total wait time to 0s.

misk-aws/src/main/kotlin/misk/jobqueue/sqs/SqsJobConsumer.kt
```diff
@@ -231,6 +253,71 @@ internal class SqsJobConsumer @Inject internal constructor(
...
+    queueName: QueueName,
+    private val handler: BatchJobHandler
+  ) : QueueReceiver(queueName) {
+
+    override fun receive(): List<Status> {
+      // Repeatedly long poll for messages until we have batchSize messages.
+      // If there are no messages available, short circuit further requests.
+      val cutoff = TimeSource.Monotonic.markNow() + 20.seconds // TODO make configurable
+      val batch: MutableList<SqsJob> = mutableListOf()
+      while (batch.size < batchSize() && !cutoff.hasPassedNow()) {
```
tgregory-block, line 269: There will be `SQS_CONSUMERS_PER_POD` threads doing this long poll, waiting until they get the desired batch size, and then each thread will handle a batch of that size. What if we override `run()` in `BatchQueueReceiver` to instead have `SQS_CONSUMERS_PER_POD` poll and combine their results to reach `batchSize()` more quickly? 

I'm not sure how surprising that would be to a user, I'm mostly thinking that for largish batch sizes, it will take many RPCs to reach the threshold, and making the threads build the batch cooperatively could alleviate that

misk-aws/src/main/kotlin/misk/jobqueue/sqs/SqsJobConsumer.kt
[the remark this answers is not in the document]
Rizziepit: Hmm... that's a good point. I don't know the default timeouts. I was planning on changing `waitTimeSeconds` to [this](https://github.com/cashapp/misk/pull/3349#discussion_r1662976928). I can add a `maxSecondsPerBatch` limit. So `waitTimeSeconds = min(secondsPerBatch, secondsUntilCutoff, maxSecondsPerBatch)`

misk-aws/src/main/kotlin/misk/jobqueue/sqs/SqsJobConsumer.kt
tgregory-block: I was thinking of the case where the desired batch size is "large" compared to SQS's enforced limit of 10 messages per call, and there are many messages in the queue (e.g. my service has a few million messages it needs to process some days). In that case, there wouldn't be contention among the threads to get enough jobs, so the threads could cooperatively build a batch. However, perhaps it doesn't matter in this case, since there will be enough jobs to hydrate each thread's batch, and it's probably an anti-pattern to try to build a batch that's 10x or more the SQS read batch limit. 

In the case where messages are sparse though, I think it's suboptimal to have multiple threads consuming on the same pod. In the worst case, their batches would starve each other and always end up waiting `cutoff` seconds, decreasing both latency and throughput. I think that could just be called out in documentation though (i.e. if a service wants to batch process messages from a queue with low messages per sec it should reduce SQS_CONSUMERS_PER_POD to 1)

adrw:
Closing since this merged in https://github.com/cashapp/misk/commit/285aac9c035e7a1ac21f9382ef9037c2f3e4dd86

PR #3352, opened 2024-07-03
Make JobqueueMetadata install once

adrw:
The new dedicated module ensures install once semantics since it is installed as part of the AwsSqsJobHandlerModule module which is installed multiple times.


frojasg (approved):
LGTM

rainecp (approved):

PR #3356, opened 2024-07-08
Add Hikari `keepaliveTime` option

aparajon:
This PR adds support for `keepaliveTime`, which is a configuration option in Hikari. We use this config option in some internal services that run on a different framework, and enabling it may aid in refreshing "stuck" closed connections (also recommended by the Hikari author in https://github.com/brettwooldridge/HikariCP/issues/1288).

Setting a default value of 0 to match the Hikari default, which makes it disabled.


misk-jdbc/src/main/kotlin/misk/jdbc/DataSourceConfig.kt
```diff
@@ -59,6 +59,7 @@ data class DataSourceConfig @JvmOverloads constructor(
   val connection_idle_timeout: Duration? = null,
   val connection_max_lifetime: Duration = Duration.ofMinutes(1),
   val query_timeout: Duration? = Duration.ofMinutes(1),
+  val keepalive_time: Duration = Duration.ofSeconds(0),
```
mpawliszyn, line 62: Is zero infinite? Is that the default?

aparajon: 0 is the default, which translates to disabled (sourced from the README in https://github.com/brettwooldridge/HikariCP)

adrw (approved):
LG

frojasg (approved):

PR #3358, opened 2024-07-09
CCPOKR-14316 | [Security][High] - Arbitrary file access during archive extraction

katukota:
Use the normalized path instead of the actual file name from JAR

### DIFF
--- a/wisp/wisp-resource-loader/src/main/kotlin/wisp/resources/ClasspathResourceLoaderBackend.kt
+++ b/wisp/wisp-resource-loader/src/main/kotlin/wisp/resources/ClasspathResourceLoaderBackend.kt
@@ -4,6 +4,7 @@ import okio.BufferedSource
 import okio.buffer
 import okio.source
 import java.io.File
+import java.nio.file.Paths
 import java.util.jar.JarFile
 
 /**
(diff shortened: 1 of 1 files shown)


wisp/wisp-resource-loader/src/main/kotlin/wisp/resources/ClasspathResourceLoaderBackend.kt
```diff
@@ -79,10 +79,11 @@ object ClasspathResourceLoaderBackend : ResourceLoader.Backend() {
...
                 if (!childFilePath.startsWith(Paths.get(pathPrefix))) continue
 
-                var endIndex = entry.name.indexOf("/", pathPrefix.length)
-                if (endIndex == -1) endIndex = entry.name.length
+                val childFilePathStr = childFilePath.toString()
+                var endIndex = childFilePathStr.indexOf("/", pathPrefix.length)
+                if (endIndex == -1) endIndex = childFilePathStr.length
 
-                result += entry.name.substring(pathPrefix.length, endIndex)
+                result += childFilePathStr.substring(pathPrefix.length, endIndex)
```
staktrace, line 86: this is fine but it might be better to do instead of string manipulation:
```
val prefix = Paths.get(pathPrefix) // this can be extracted from line 80
if (childFilePath.getNameCount() > prefix.getNameCount()) {
  result += childFilePath.name(prefix.getNameCount()).toString()
}
```

staktrace (approved):

wisp/wisp-resource-loader/src/main/kotlin/wisp/resources/ClasspathResourceLoaderBackend.kt
katukota: This looks so much cleaner! Updated the code.

PR #3361, opened 2024-07-10
Improve gradle test logging

rainecp:
This adds a new property `misk.test.logging` to enable or disable certain logging features. By default, it will be set to `false` to reduce the amount of logs.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -2,6 +2,10 @@ import com.vanniktech.maven.publish.SonatypeHost
 import io.gitlab.arturbosch.detekt.Detekt
 import io.gitlab.arturbosch.detekt.extensions.DetektExtension
 import org.gradle.api.tasks.testing.logging.TestExceptionFormat
+import org.gradle.api.tasks.testing.logging.TestLogEvent.FAILED
+import org.gradle.api.tasks.testing.logging.TestLogEvent.PASSED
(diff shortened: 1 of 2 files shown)


build.gradle.kts
```diff
@@ -246,10 +250,21 @@ subprojects {
 
   tasks.withType<Test> {
     useJUnitPlatform()
-    testLogging {
-      events("started", "passed", "skipped", "failed")
-      exceptionFormat = TestExceptionFormat.FULL
-      showStandardStreams = false
+
+    val enableLogging = project.findProperty("misk.test.logging")?.toString()?.toBoolean() ?: true
```
adrw, line 254:
```suggestion
    val enableLogging = project.findProperty("misk.test.logging")?.toString().toBoolean()
```

If we're already setting this to false for the project, we can probably not set true as the null default. `toBoolean` can also be called on null and will return false and be simpler to maintain.

adrw (approved):

gradle.properties
```diff
@@ -4,3 +4,4 @@ org.gradle.caching=true
 org.gradle.vfs.watch=true
 org.gradle.configureondemand=true
 org.gradle.parallel=true
+misk.test.logging=false
```
adrw, line 7: Removing the new line in this file broke publishing, fix incoming.

PR #3365, opened 2024-08-01
Add deprecated annotation to TaggedLogger

afkelsall:
We are refactoring the TaggedLogger within misk to be less opinionated and complicated to use. Under the new implementation, services will still be able to access the original functionality of this logger to add tags to logs and include them on underlying web exceptions thrown by misk.

This PR is just to mark this implementation as deprecated to hint that it shouldn't be used yet.


ean5533 (approved):

adrw (approved):
Thanks!

adrw:
Merged in https://github.com/cashapp/misk/commit/d13f8e10fa64c45c5a0a9574730ec02bb642e85f

PR #3366, opened 2024-08-02
Introduce simpler `withSmartTags` in place of the `TaggedLogger`

afkelsall:
The original implementation of `TaggedLogger` became heavy, stateful and opinionated. We've since reconsidered this approach and instead simplified it greatly to just a function call `withSmartTags` doing away with all the stateful builders etc.

For this draft PR I have left the tests as similar as possible by introducing a `TestTaggedLogger` at the bottom allowing them to remain largely unchanged. I wanted to do this to prove the new approach satisfied those tests.

I'll convert these tests over to `withSmartTags` before merging this PR.


misk-aws/src/test/kotlin/misk/jobqueue/sqs/SmartTagsJobQueueTest.kt
afkelsall, line 1: This is not actually a new file. The tests remain mostly the same as they were but added a nested inner class to show the different approaches in action.

wisp/wisp-logging/src/main/kotlin/wisp/logging/SmartTagsThreadLocalHandler.kt
```diff
@@ -0,0 +1,41 @@
...
+    }
+  }
+
+  private data class ThreadLocalTaggedLoggerMdcContext(
+    val triggeringThrowable: Throwable,
+    val tags: Set<Tag>
+  ) {
+    fun wasTriggeredBy(throwable: Throwable): Boolean {
+      return triggeringThrowable == throwable
+    }
```
ean5533, line 39: Do we want to handle the case where an original exception was caught and wrapped up in a new exception? We could do that by recursively checking against the `.cause`

wisp/wisp-logging/src/main/kotlin/wisp/logging/Logging.kt
```diff
@@ -113,3 +113,82 @@ fun withTags(vararg tags: Tag, f: () -> Unit) {
...
+  }
+
+  try {
+    return f().also {
+      // Exiting this block gracefully: Lets do some cleanup to keep the ThreadLocal clear.
+      // The scenario here is that when nested `withSmartTags` threw an exception and it was
+      // caught and handled by this `withSmartTags`, it should clean up the unused and unneeded context.
+      SmartTagsThreadLocalHandler.clear()
+    }
+  } catch (th: Throwable) {
```
ean5533, line 184: Catching `Throwable` always raises an eyebrow for me, as it includes things like `OutOfMemoryError` and `ThreadDeath` that need careful treatment. It might be ok here since this block is just doing a little bit of work and then rethrowing anyway, but unless you're sure you need to handle those scenarios, I would suggest catching `Exception` instead

wisp/wisp-logging/src/main/kotlin/wisp/logging/Logging.kt
```diff
@@ -113,3 +113,82 @@ fun withTags(vararg tags: Tag, f: () -> Unit) {
...
+    // Calls to `withSmartTags` can be nested - only set if there is not already a context set
+    // This will be cleared upon logging of the exception within misk or if the thrown exception
+    // is handled by a higher level `withSmartTags`
```
ean5533, line 187: What will happen if someone uses `WithSmartTags` in a place that never pops them automatically (say, in a Kafka consumer)? Could tags leak from one exceptional call to another if they share a thread?

ean5533 (approved):

wisp/wisp-logging/src/main/kotlin/wisp/logging/Logging.kt
```diff
@@ -113,3 +113,82 @@ fun withTags(vararg tags: Tag, f: () -> Unit) {
...
+ * ```
+ *   Log MDC context: [] Log message: "Received request"
+ *   Log MDC context: [processValue: "PV_123", contextToken: "contextTokenValue"] Log message: "Processing request"
+ *   Log MDC context: [processValue: "PV_123", contextToken: "contextTokenValue"] Log message: "Start Process"
+ *   Log MDC context: [processValue: "PV_123", contextToken: "contextTokenValue"] Log message: "unexpected error dispatching to ServiceAction" // This log would not normally include the MDC context
+ * ```
+ */
+fun <T> withSmartTags(vararg tags: Tag, f: () -> T): T {
+  // Establish MDC, saving prior MDC
+  val priorMDC = tags.map { (k, v) ->
```
adrw, line 171:
```suggestion
  val priorMDC = tags.map { (key, value) ->
```

Generally for better readability, a proper word is preferred for idiomatic Kotlin.

wisp/wisp-logging/src/main/kotlin/wisp/logging/Logging.kt
```diff
@@ -113,3 +113,82 @@ fun withTags(vararg tags: Tag, f: () -> Unit) {
...
+ *
+ * Logging result:
+ * ```
+ *   Log MDC context: [] Log message: "Received request"
+ *   Log MDC context: [processValue: "PV_123", contextToken: "contextTokenValue"] Log message: "Processing request"
+ *   Log MDC context: [processValue: "PV_123", contextToken: "contextTokenValue"] Log message: "Start Process"
+ *   Log MDC context: [processValue: "PV_123", contextToken: "contextTokenValue"] Log message: "unexpected error dispatching to ServiceAction" // This log would not normally include the MDC context
+ * ```
+ */
+fun <T> withSmartTags(vararg tags: Tag, f: () -> T): T {
```
adrw, line 169:
```suggestion
fun <T> withSmartTags(vararg tags: Tag, block: () -> T): T {
```

Preferred for better readability.

wisp/wisp-logging/src/main/kotlin/wisp/logging/Logging.kt
adrw: In this case you could also just leave it as impliciy `it` and then the callsites are equivalently readable `it.key to it.value.toString()`

afkelsall: Yep no problem. This is the code what was originally present prior to my changes which I've moved around but not modified :) 
I agree with the changes though and am happy to push them through.

afkelsall: The type from MDC is a `Pair<String, String>` however which means this would read as `it.first` as key and `it.second` as value. I'm going to explicitly name them `key` and `value` here.

adrw:
Closing since this was merged in https://github.com/cashapp/misk/commit/8cffeac185c4b16bee379acfd112670656588d36

PR #3371, opened 2024-08-16
Adding deployment version on Cluster.Member

din-cashapp:
Updating Cluster.Member data class to contain the information on the deployment version. 

This should work for services that use KubernetesCluster however because of us having 2 different versions of a cluster: Dynamo and Kubernetes based, we won't be able to retrieve deployment version for Dynamo (unless we change the yaml configuration for how we set the data in pods).


adrw:
Please re-open this internally in the cash-server repo, message me on Slack if you need help importing it.

PR #3372, opened 2024-08-16
Allow overrides for ActionScopedProviders to be passed in when calling ActionScope.create

szabado-faire:
This allows provider overrides to be passed in when entering a new action scope.

## Use case

I'm working on making sure that our cache loading functions cannot access data in the `ActionScope` that is not part of the cache key. The easiest way to prevent access to providers that were not accounted for as part of the cache key is to override them to providers that throw exceptions when accessed.


misk-action-scopes/src/main/kotlin/misk/scope/ActionScope.kt
```diff
@@ -97,15 +97,18 @@ class ActionScope @Inject internal constructor(
...
+  ): Instance {
     check(!inScope()) {
       "cannot create an ActionScope.Instance on a thread that is already running in an action scope"
     }
 
     val immediateValues = seedData.mapValues { (_, value) -> ImmediateLazy(value) }
 
     val lazyValues = providers.mapValues { (key, _) ->
-      SynchronizedLazy(providerFor(key))
+      SynchronizedLazy(providerFor(key, providerOverrides))
```
JGulbronson, line 111: Instead of passing the overrides, can we just update line 110 to be `(providers + providerOverrides).mapValues` instead?

JGulbronson (approved):
This seems reasonable to me, effectively a way to pass a lazy value when creating an action scope instead of having to seed it with already computed values.

@adrw what do you think?

misk-action-scopes/src/main/kotlin/misk/scope/ActionScope.kt
szabado-faire: Good push. I added another block that contains `lazyOverrides` because `providers` contains `Provider<ActionScopedProvider<*>>` so the types won't mesh

frojasg:
This change makes sense to me. We recently changed the way to incorporate change to misk. Let me chat with @adrw to make this happen.

adrw (approved):
LG

frojasg:
I have been thinking more about this. Could you add a small test illustrating what you are trying to do there? Why can't these be normal providers and need to be in the seed?

JGulbronson (approved):

frojasg:
This was merged here https://github.com/cashapp/misk/compare/7b4a53a60212...90de8d27e20a

szabado-faire:
Fantastic, thank you @frojasg!

szabado-faire:
Closing as this is no longer necessary

PR #3375, opened 2024-11-01
Allow setting Jetty direct byte buffer usage

JGulbronson:
We're seeing a large amount of direct memory allocation, which prevents us from rolling out ZGC internally at Faire. This will allow us to toggle the usage of direct byte buffers, and let the garbage collector manage it instead.


swankjesse (approved):
LGTM 

I think our merge process is new.

adrw:
This has been merged in https://github.com/cashapp/misk/commit/caab39337351600ef7bfa05f2474b6b5dcfc37f4

PR #3385, opened 2025-03-06
Prevent leaking stacktrace when unable to parse URI query params

JGulbronson:
## Description

When calling `JettyWebSocketServlet.service`, Jetty can throw when trying to extract the query params. This is a [known issue](https://stackoverflow.com/questions/44345128/jetty-java-lang-illegalargumentexception-not-valid-encoding), and results in a stacktrace being sent in the response. Sending the stacktrace by default seems like an unwanted behavior from a security perspective, so I'd made a change that preserves the status code and message, but removes the stacktrace to avoid leaking unwanted information. I didn't feel it was necessary to add a flag for this to `WebConfig` because in almost all cases I believe this to be the behavior that users would want.

### DIFF
--- a/misk/src/main/kotlin/misk/web/jetty/WebActionsServlet.kt
+++ b/misk/src/main/kotlin/misk/web/jetty/WebActionsServlet.kt
@@ -20,6 +20,7 @@ import okio.BufferedSink
 import okio.buffer
 import okio.sink
 import okio.source
+import org.eclipse.jetty.http.BadMessageException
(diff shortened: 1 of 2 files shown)


JGulbronson:
> LG, going to run it against the internal test suite now.

@adrw how'd it go running against the internal test suite?

katukota:
LG!

JGulbronson:
Added in https://github.com/cashapp/misk/commit/9752697cb098f3862da84140e0d0fd4d843dd637

PR #3387, opened 2025-03-19
Gzip compress gRPC responses

zpingcai:
## Description

Currently, all gRPC responses are not compressed. This is not optimal for endpoints with huge responses as compression may save us a decent amount of network resource. This PR adds support to compress gRPC responses using gzip, which can be enabled by setting `WebConfig.grpcGzip` to true.

## Testing Strategy

- Added test cases to `GrpcSourceSinkTest`.
- `GrpcConnectivityTest` doesn't run in CI due to https://github.com/cashapp/misk/issues/1853. So I ran it locally and confirmed all test cases passed.

## Related Work
- Commit https://github.com/square/wire/commit/17a3d9e129dc3ee905a2bdc8fb8267dcaa234c9f adds gzip support to wire-grpc client.

- `WebConfig.grpcGzip` is false by default, so users won't see surprises when they upgrade misk. They can manually turn it on if needed.

## Checklist


misk/src/main/kotlin/misk/grpc/GrpcMessageSink.kt
zpingcai, line 1: Simply copies the changes from https://github.com/square/wire/commit/17a3d9e129dc3ee905a2bdc8fb8267dcaa234c9f

misk/src/main/kotlin/misk/web/exceptions/ExceptionHandlingInterceptor.kt
```diff
@@ -100,7 +100,7 @@ class ExceptionHandlingInterceptor(
     httpCall.setResponseTrailer("grpc-status-details-bin", response.toEncodedStatusProto)
     httpCall.setResponseTrailer("grpc-message", response.message ?: response.status.name)
     httpCall.takeResponseBody()?.use { responseBody: BufferedSink ->
-      GrpcMessageSink(responseBody, ProtoAdapter.BYTES, grpcEncoding = "identity")
+      GrpcMessageSink(responseBody, 0, ProtoAdapter.BYTES, grpcEncoding = "identity")
```
zpingcai, line 103: Doesn't make sense to compress an empty response message.

misk/src/test/kotlin/misk/grpc/GrpcConnectivityTest.kt
```diff
@@ -169,8 +169,8 @@ class GrpcConnectivityTest {
   interface GreeterSayHello : Service {
     @WireRpc(
       path = "/helloworld.Greeter/SayHello",
-      requestAdapter = "com.squareup.protos.test.grpc.HelloRequest.ADAPTER",
-      responseAdapter = "com.squareup.protos.test.grpc.HelloReply.ADAPTER"
+      requestAdapter = "com.squareup.protos.test.grpc.HelloRequest#ADAPTER",
+      responseAdapter = "com.squareup.protos.test.grpc.HelloReply#ADAPTER"
```
zpingcai, line 173: drive-by: fix the test setup

misk/src/main/kotlin/misk/web/WebConfig.kt
```diff
@@ -114,10 +114,15 @@ data class WebConfig @JvmOverloads constructor(
...
   /**
-   * If true responses which are larger than the minGzipSize will be compressed.
+   * If true non-gRPC responses which are larger than the minGzipSize will be compressed.
    */
   val gzip: Boolean = true,
 
+  /**
+   * If true gRPC responses which are larger than the minGzipSize will be compressed.
+   */
+  val grpcGzip: Boolean = false,
```
Nava2, line 124: Should this instead be an enum to allow for the other wire-supported encoding? 

```suggestion
  val grpcEncoding: GrpcEncoding = GrpcEncoding.IDENTITY,
```

adrw: Yes, I'd lean towards the encoding enum, not just a boolean

misk/src/main/kotlin/misk/web/exceptions/ExceptionHandlingInterceptor.kt
adrw: Can you add named params for all of the constructor parameters, not clear what 0 references

misk/src/test/kotlin/misk/grpc/GrpcSourceSinkTest.kt
```diff
@@ -39,10 +39,55 @@ class GrpcSourceSinkTest {
   @Test
   fun grpcMessageSinkHelloReply() {
     val buffer = Buffer()
-    val writer = GrpcMessageSink(buffer, HelloReply.ADAPTER, "identity")
+    val writer = GrpcMessageSink(buffer, 0, HelloReply.ADAPTER, "identity")
```
adrw, line 42: named params for all pls

misk/src/test/kotlin/misk/grpc/GrpcSourceSinkTest.kt
```diff
@@ -39,10 +39,55 @@ class GrpcSourceSinkTest {
...
+
+  @Test
+  fun messageLargerThanMinimumSizeIsCompressed() {
+    val buffer = Buffer()
+    val writer = GrpcMessageSink(buffer, 10, HelloRequest.ADAPTER, "gzip")
+
+    writer.write(HelloRequest("localhost"))
+    writer.close()
+
+    assertEquals(
```
adrw, line 74: Can you add an assertion on the min size and the actual size? Not just a snapshot assertion on expected output (still keep that but add the extra assertion).

misk/src/test/kotlin/misk/grpc/GrpcSourceSinkTest.kt
```diff
@@ -39,10 +39,55 @@ class GrpcSourceSinkTest {
...
+      "010000001f1f8b0800000000000000e3e2ccc94f4eccc9c82f2e01002fdef60d0b000000".decodeHex(),
+      buffer.readByteString()
+    )
+  }
+
+  @Test
+  fun messageSmallerThanMinimumSizeIsNotCompressed() {
```
adrw, line 81: Also here, pls add another assertion on the size difference

misk/src/main/kotlin/misk/web/WebConfig.kt
zpingcai: Not sure if it's actually worth it:
1. The client only supports `identity` and `gzip`: https://github.com/square/wire/blob/master/wire-grpc-client/src/commonMain/kotlin/com/squareup/wire/internal/GrpcDecoder.kt
2. The http response compression only supports gzip: https://github.com/cashapp/misk/blob/4da16a5c6492d5fa5b0364d95686cdaef6d692e5/misk/src/main/kotlin/misk/web/WebConfig.kt#L119
3. We also need to introduce `WebConfig.minGrpcMessageToCompress` instead of reusing `WebConfig.minGzipSize`.

misk/src/test/kotlin/misk/grpc/GrpcSourceSinkTest.kt
zpingcai: The actual size is asserted in `helloRequestSize`, but I can move it here if that makes the test more readable.

misk/src/main/kotlin/misk/web/exceptions/ExceptionHandlingInterceptor.kt
```diff
@@ -124,7 +128,11 @@ class ExceptionHandlingInterceptor(
     val mapper = mapperResolver.mapperFor(th)
     if (mapper != null) {
       if (!suppressLog) {
-        log.log(mapper.loggingLevel(th), th, *mdcTags.toTypedArray()) { "exception dispatching to $actionName" }
+        log.log(
+          level = mapper.loggingLevel(th),
+          th = th,
+          tags = *mdcTags.toTypedArray(),
+        ) { "exception dispatching to $actionName" }
```
zpingcai, line 135: IDE auto formatted this because the line is too long.

misk/src/main/kotlin/misk/web/exceptions/ExceptionHandlingInterceptor.kt
zpingcai: yes, added to all.

misk/src/main/kotlin/misk/web/WebConfig.kt
zpingcai: I think the current implementation makes more sense to the user:

1. Why do you configure `grpcEncoding` to anything that's not supported?
2. Why do you want different configs for min compression size for http vs grpc over http?

cc @Hexcles

Hexcles: Personally, I'd say @zpingcai 's current impl is better in terms of consistency.

`grpcGzip` makes a lot of sense along with `gzip`. If we want `grpcEncoding`, I'd expect to have an enum for plain old HTTP, too. And in fact, not just an enum but a config object as different compression algos have different knobs (`minGzipSize` obviously doesn't make sense for non-gzip compression algos).

zpingcai:
merged as https://github.com/cashapp/misk/commit/cd10157e06761b3e62d172dc2ac132766df178dd

PR #3395, opened 2025-07-10
Allow binding custom StringConverters

JGulbronson:
## Description

Currently params had to be one of the supported types, or expose a `valueOf` function. However at Faire, we want to offer better type validation for our parameters. For example, we have `Token<T>` where `T` could be something like `Brand` or `Product`. Each of these token types has a unique prefix, so even if we have a `valueOf` function exposed we lack the ability to validate the type is correct. With this change we can inspect the type of the parameter, and validate it's the correct type before converting the `String`.

Note that this is technically a breaking change because I've changed `StringConverter` to an interface, and updated the public constructor of `FeatureBindingFactory`, but they're small enough changes that I don't think people will have any issues upgrading.

## Testing Strategy

I updated `QueryParamFeatureBindingTest` to show that a custom `StringConverter` can be used to properly parse a string to a value class, which otherwise would fail. I didn't update all the other binding tests because the code path is the same, but can update them if people feel strongly.

### DIFF
--- a/misk/src/main/kotlin/misk/web/FeatureBinding.kt
+++ b/misk/src/main/kotlin/misk/web/FeatureBinding.kt
@@ -2,6 +2,7 @@ package misk.web
 
 import misk.Action
 import misk.web.actions.WebAction
+import misk.web.extractors.StringConverter
 import okio.BufferedSink
 import okio.BufferedSource
 import java.util.regex.Matcher
@@ -64,7 +65,8 @@ interface FeatureBinding {
     fun create(
       action: Action,
       pathPattern: PathPattern,
-      claimer: Claimer
+      claimer: Claimer,
+      stringConverterFactories: List<StringConverter.Factory> = listOf(),
     ): FeatureBinding?
   }
 
--- a/misk/src/main/kotlin/misk/web/extractors/StringConverter.kt
+++ b/misk/src/main/kotlin/misk/web/extractors/StringConverter.kt
@@ -1,6 +1,7 @@
 package misk.web.extractors
 
 import com.squareup.moshi.rawType
+import kotlin.jvm.Throws
 import kotlin.reflect.KType
 import kotlin.reflect.full.createType
 import kotlin.reflect.full.isSubtypeOf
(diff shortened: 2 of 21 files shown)


misk/src/main/kotlin/misk/web/FeatureBinding.kt
```diff
@@ -64,7 +65,8 @@ interface FeatureBinding {
     fun create(
       action: Action,
       pathPattern: PathPattern,
-      claimer: Claimer
+      claimer: Claimer,
+      stringConverterFactories: List<StringConverter.Factory>
```
ean5533: Could we make this default to empty list to reduce breakage?

misk/src/main/kotlin/misk/web/extractors/StringConverter.kt
```diff
@@ -19,29 +20,51 @@ private val doubleTypeNullable: KType = Double::class.createType(nullable = true
...
+  interface Factory {
+    fun create(kType: KType): StringConverter?
   }
 }
 
-private fun numberConversionWrapper(wrappedFunc: StringConverter): StringConverter {
-  return {
+internal fun converterFor(
+  type: KType,
+  factories: List<StringConverter.Factory>,
```
ean5533: Likewise, this could remain public and have a default value on the new param to reduce breakage

ean5533 (commented):
This looks pretty reasonable to me, but I think we we can minimize backwards compat concerns with some default param values

misk/src/main/kotlin/misk/web/FeatureBinding.kt
JGulbronson: Sounds good, can always drop the default in a later release

misk/src/main/kotlin/misk/web/extractors/StringConverter.kt
JGulbronson: Sure

ean5533 (approved):

frojasg:
Merged in https://github.com/cashapp/misk/commit/7d20fd8200299e160fc21eea691fb8d3576b0666. Thanks for the improvement

Issue #3401, opened 2025-08-21
`ReflectionQuery` does not clone all fields


mgulbronson:
See [code](https://github.com/cashapp/misk/blob/master/misk-hibernate/src/main/kotlin/misk/hibernate/ReflectionQueryFactory.kt#L80):
```kotlin
if (copy.maxRows != -1) {
  copy.maxRows = this.maxRows
}
if (copy.firstResult != 0) {
  copy.firstResult = this.firstResult
}
```

When `copy` is created it sets `maxRows = -1` and `firstResult = 0` by default [[code](https://github.com/cashapp/misk/blob/3fed8ea8748bb1da575324879e33eed19f9c5209/misk-hibernate/src/main/kotlin/misk/hibernate/ReflectionQueryFactory.kt#L49)].  Then the `clone` function checks `copy.maxRows` and `copy.firstResult` to determine whether to set the value on the copy the same as the original.

My guess is that this was a typo that was not caught because [the test](https://github.com/cashapp/misk/blob/3fed8ea8748bb1da575324879e33eed19f9c5209/misk-hibernate/src/test/kotlin/misk/hibernate/ReflectionQueryFactoryTest.kt#L1124) for the `clone` function does not check that all fields are cloned. IMO there are two ways to solve this:
1. Check `this.maxRows != -1` (and same for `firstResult`) <- this was presumably the intention
2. Get rid of the `if` altogether <- IMO this is better, no need to check against the default values when we can just trust the thing being cloned

PR #3402, opened 2025-08-21
Clone all fields in `ReflectionQuery

mgulbronson:
Addresses https://github.com/cashapp/misk/issues/3401

Currently these fields are not copied to the new object when `clone` is called. This PR fixes that and updates an existing test for the `clone` function to check these fields as well.


misk-hibernate/src/test/kotlin/misk/hibernate/ReflectionQueryFactoryTest.kt
```diff
@@ -1131,33 +1131,50 @@ class ReflectionQueryFactoryTest {
...
     assertThat(
       transacter.transaction { session ->
-        original.count(session)
+        original.listAsNames(session)
       }
-    ).isEqualTo(2)
+    ).containsExactlyInAnyOrder(
+      "Jurassic Park",
+      "Jurassic Park: The Lost World",
+    )
```
mgulbronson, line 1195: Changing these assertions to check the names rather than just counting since `count(session)` complains if `maxRows` or `firstResult` are set but I need to set them to assert that they're cloned properly

adrw (approved):
LG

JGulbronson (approved):

PR #3403, opened 2025-08-22
Improve handling of required query parameters

emilydespas-faire:
## Description

Currently if a request does not populate a required (non-nullable / non-default)  query parameter, misk will try to set a null value into the parameter and cause an exception like:
```
java.lang.NullPointerException: Parameter specified as non-null is null: method <name>, parameter <name>
```

This change allows us to continue returning null to any nullable/defaulted parameters, while throwing bad request if a required query param is not passed.

## Testing Strategy

Confirmed the issue via tests. added additional unit tests to cover the not present and default param cases

## Related Work
    - Link any relevant PRs, tickets, or documentation for additional context

## Communication Plan
  - was discussed with @adrw in #misk-external slack channel

should be no material impact, previous behaviour was an internal service exception, will now be a bad request

## Definition of Done

Misk based services should now return HTTP 400 when a required param is not provided rather than HTTP 500
    
Monitor any endpoints currently throwing HTTP 500 due to missing required query params and confirm those requests return HTTP 400

No mitigation required, behaviour should be unchanged other than a different exception type

## Checklist


emilydespas-faire:
cc @adrw

adrw:
Merged in https://github.com/cashapp/misk/commit/355cc3bfcb7063c2f0a404805facad946223af3b I did try to put Git co-author but unfortunately it got stripped by Copybara, I'll see if I can add it back. Thanks for the contribution!

PR #3405, opened 2025-09-25
Deprecate Wisp types

katukota:
deprecate the remaining Wisp Types
DynamicConfig
FeatureFlags
FakeFeatureFlags
LaunchDarklyDynamicConfig
LaunchDarklyFeatureFlags


misk-feature/src/main/kotlin/misk/feature/DynamicConfig.kt
```diff
@@ -37,6 +37,12 @@ interface DynamicConfig {
    */
   fun <T> getJson(feature: Feature, clazz: Class<T>): T
 
+
+  /**
+   * Returns the value of a JSON string value of the dynamic flag.
+   */
+  fun getJsonString(feature: Feature): String
```
adrw, line 44: This is to match the Wisp APIs or why is this API being added?

adrw (approved):
LG

misk-feature/src/main/kotlin/misk/feature/DynamicConfig.kt
katukota: yes to match the Wisp APIs.

PR #3406, opened 2025-09-25
Add distributed cron execution support to Misk

din-cashapp:
## Summary
  Adds distributed cron execution to Misk, allowing tasks to run concurrently across pods instead of using a single cluster-wide lease.

  ## Changes
  - New `CronCoordinator` interface with `SingleLeaseCronCoordinator` (default) and `DistributedCronCoordinator`
  - Added `useDistributedExecution` parameter to `CronModule` (defaults to false)
  - Updated `CronTask` and `CronManager` to use coordinator abstraction
  - Comprehensive test coverage for both modes

  ## Usage
  ```kotlin
  install(CronModule(
    zoneId = ZoneId.of("America/Los_Angeles"),
    useDistributedExecution = true
  ))
```
 ## Benefits

  - Better resource utilization and fault tolerance
  - Eliminates single point of failure
  - Zero breaking changes - backward compatible (as long as the team didn't switch the useDistributedExecution to true)

### DIFF
--- a/misk-cron/README.md
+++ b/misk-cron/README.md
@@ -32,13 +32,14 @@ Note that there are **not** strong guarantees on task execution; tasks can be de
 - Only one pod in the cluster executes cron tasks at a time
 - **Best when:**
    - Tasks idempotency is uncertain or unverified
-   - Tasks have execution order dependencies
+   - You have only a small number of tasks with non-overlapping schedules 
+   - You want a slightly simpler operational model (all tasks run on a single pod)
 
 ### Distributed mode (multiple lease execution)
 - Each cron task has its own lease
 - Multiple pods can run tasks concurrently, increasing resource utilization and fault tolerance
 - **Best when:**
-   - You have multiple independent tasks
+   - You have multiple independent tasks with overlapping schedules
    - Tasks are idempotent
    - You want to avoid a single point of failure
 
--- a/misk-cron/src/main/kotlin/misk/cron/CronModule.kt
+++ b/misk-cron/src/main/kotlin/misk/cron/CronModule.kt
@@ -16,13 +16,22 @@ import misk.tasks.RepeatedTaskQueueFactory
 import java.time.ZoneId
 import wisp.lease.LeaseManager
 
+/**
+ * Provides cron scheduling functionality for Misk services.
+ *
+ * @param useMultipleLeases Controls lease coordination strategy. Changing this value may cause
+ *   overlapping task execution during deployments.
+ *
+ *   Example: switching false→true means old pods use cluster-wide leases while new pods use
+ *   per-task leases, potentially running the same task on both.
+ *
+ *   Deploy during downtime or ensure tasks are idempotent.
+ */
 class CronModule @JvmOverloads constructor(
   private val zoneId: ZoneId,
   private val threadPoolSize: Int = 10,
   private val dependencies: List<Key<out Service>> = listOf(),
   private val installDashboardTab: Boolean = true,
-  // NOTE: `useMultipleLeases = true` may cause overlapping execution during deployments.
-  // Deploy during downtime or ensure tasks are idempotent.
   private val useMultipleLeases: Boolean = false
 ) : KInstallOnceModule() {
   override fun configure() {

### CODE REVIEW

File: misk-cron/src/main/kotlin/misk/cron/CronTask.kt [hunk 1]
```diff
@@ -27,16 +27,8 @@ internal class CronTask @Inject constructor() : AbstractIdleService() {
...
-
       val now = clock.instant()
-      var leaseHeld = lease.checkHeld()
-      if (!leaseHeld) {
-        leaseHeld = lease.acquire()
-      }
-      if (leaseHeld) {
-        cronManager.runReadyCrons(lastRun)
-      }
+      cronManager.runReadyCrons(lastRun, cronCoordinator)
```
lcabancla: Can we just inject `CronCoordinator` directly into `CronManager`?

misk-cron/src/main/kotlin/misk/cron/CronCoordinator.kt
```diff
@@ -0,0 +1,30 @@
...
+    val lease = leaseManager.requestLease(CRON_CLUSTER_LEASE_NAME)
+    return lease.checkHeld() || lease.acquire()
+  }
+
+  companion object {
+    const val CRON_CLUSTER_LEASE_NAME = "misk.cron.lease"
+  }
+}
+
+class DistributedCronCoordinator @Inject constructor(
```
lcabancla: Maybe a better name is `MultipleLeaseCronCoordinator` to avoid confusion with distributed leases

misk-cron/src/main/kotlin/misk/cron/CronModule.kt
```diff
@@ -51,6 +53,7 @@ class FakeCronModule @JvmOverloads constructor(
   private val threadPoolSize: Int = 10,
   private val dependencies: List<Key<out Service>> = listOf(),
   private val installDashboardTab: Boolean = false,
+  private val useDistributedExecution: Boolean = false,
```
lcabancla: I would add a comment/warning here regarding idempotency.

misk-cron/src/main/kotlin/misk/cron/CronTask.kt
din-cashapp: oh yea great point!

misk-cron/src/main/kotlin/misk/cron/CronModule.kt
```diff
@@ -51,6 +55,7 @@ class FakeCronModule @JvmOverloads constructor(
...
```
lcabancla: I would rename this to `useMultipleLeases` or something similar.

misk-cron/src/main/kotlin/misk/cron/CronManager.kt
```diff
@@ -21,6 +21,7 @@ class CronManager @Inject constructor() {
   @Inject private lateinit var clock: Clock
   @Inject @ForMiskCron private lateinit var executorService: ExecutorService
   @Inject @ForMiskCron private lateinit var zoneId: ZoneId
+  @Inject private lateinit var cronCoordinator: CronCoordinator
```
adrw: coordinator isn't scoped to the task queue ie. `@ForMiskCron`?

misk-cron/src/main/kotlin/misk/cron/CronModule.kt
```diff
@@ -73,6 +78,15 @@ class FakeCronModule @JvmOverloads constructor(
       install(CronDashboardTabModule())
     }
   }
+
+  @Provides
+  @Singleton
+  fun cronCoordinator(leaseManager: wisp.lease.LeaseManager): CronCoordinator =
```
adrw: add import for `wisp.lease.LeaseManager`?

misk-cron/src/main/kotlin/misk/cron/CronManager.kt
din-cashapp: oO good catch!

misk-cron/src/main/kotlin/misk/cron/CronModule.kt
```diff
@@ -14,12 +14,16 @@ import misk.inject.toKey
...
 import java.time.ZoneId
+import wisp.lease.LeaseManager
 
 class CronModule @JvmOverloads constructor(
   private val zoneId: ZoneId,
   private val threadPoolSize: Int = 10,
   private val dependencies: List<Key<out Service>> = listOf(),
   private val installDashboardTab: Boolean = true,
+  // NOTE: `useMultipleLeases = true` may cause overlapping execution during deployments.
+  // Deploy during downtime or ensure tasks are idempotent.
```
tgregory-block: nit - maybe move this to a kdoc string on the class, and mention that it can happen any time we flip the value, not just when setting it to `true`

misk-cron/README.md
```diff
@@ -23,3 +24,60 @@ Note that there are **not** strong guarantees on task execution; tasks can be de
...
     }
     ```
+
+## Execution Models
+
+### Single-lease mode (default)
+- One cluster-wide lease controls all cron tasks
+- Only one pod in the cluster executes cron tasks at a time
+- **Best when:**
+   - Tasks idempotency is uncertain or unverified
```
tgregory-block, line 34: I don't know that this is true, running the tasks sequentially doesn't really have anything to do with idempotence. It is possible to have a split brain scenario with leases where multiple pods think they have the same lease, e.g. if a pod acquires the lease and does a gc pause after it's last check and when the lease is about to expire. This is true with both the single and distributed lease approach

din-cashapp: good points! updated the README

tgregory-block (approved):

PR #3408, opened 2025-09-28
Kevink/rst stream config

kevink-sq:
## Description

This change exposes Jetty's frame limiter config for DDoS mitigation (CVE-2023-44487, CVE-2025-5115).

Jetty 10.0.26 introduced HTTP/2 frame limiting for server side RST_STREAM and WINDOW_UPDATE events and is causing issues for some services due to the default 128 events/second being too low. When Jetty rate limits a stream, it kills the whole multiplexed HTTP/2 connection (**cash-inc-6253**)

Misk's implementation is the same as Jettys' [WindowRateControl](https://github.com/jetty/jetty.project/blob/jetty-10.0.x/jetty-http2/http2-common/src/main/java/org/eclipse/jetty/http2/parser/WindowRateControl.java) but with observability. Using peakGauge against the windowed events to accurate capture max values without sampling. Also emitting rate limiting metrics to catch issues when rolling out Jetty 10.0.26+.

Setting `jetty_http2_max_events_per_second: -1` will not rate limit but still emit metrics. Intending to set a higher value in skim.

## Testing Strategy

Unit test to make sure `MeasuredWindowRateControl` behaves as expected. Jetty's internal configuration can be compared [AbstractHTTP2ServerConnectionFactory](https://github.com/jetty/jetty.project/blob/jetty-10.0.x/jetty-http2/http2-server/src/main/java/org/eclipse/jetty/http2/server/AbstractHTTP2ServerConnectionFactory.java#L66)

The current rate limit value of 128 is the same as default Jetty so no changes expected.

## Checklist

### DIFF
--- a/misk-feature/api/misk-feature.api
+++ b/misk-feature/api/misk-feature.api
@@ -109,7 +109,9 @@ public final class misk/feature/TrackerReferenceKt {
 public final class misk/feature/testing/FakeFeatureFlags : com/google/common/util/concurrent/AbstractIdleService, misk/feature/DynamicConfig, misk/feature/FeatureFlags, misk/feature/FeatureService {
 	public static final field Companion Lmisk/feature/testing/FakeFeatureFlags$Companion;
 	public static final field KEY Ljava/lang/String;
+	public fun <init> ()V
 	public fun <init> (Lwisp/feature/testing/FakeFeatureFlags;)V
+	public synthetic fun <init> (Lwisp/feature/testing/FakeFeatureFlags;ILkotlin/jvm/internal/DefaultConstructorMarker;)V
 	public fun get (Lwisp/feature/BooleanFeatureFlag;)Z
 	public fun get (Lwisp/feature/DoubleFeatureFlag;)D
 	public fun get (Lwisp/feature/EnumFeatureFlag;)Ljava/lang/Enum;
--- a/misk-feature/src/testFixtures/kotlin/misk/feature/testing/FakeFeatureFlags.kt
+++ b/misk-feature/src/testFixtures/kotlin/misk/feature/testing/FakeFeatureFlags.kt
@@ -23,7 +23,8 @@ import jakarta.inject.Singleton
  */
 @Singleton
 class FakeFeatureFlags @Inject constructor(
-  val delegate: wisp.feature.testing.FakeFeatureFlags
+  // TODO remove default parameter once callsites are migrated
+  val delegate: wisp.feature.testing.FakeFeatureFlags = wisp.feature.testing.FakeFeatureFlags()
 ) : AbstractIdleService(),
   FeatureFlags,
   FeatureService,


misk/src/main/kotlin/misk/web/jetty/MeasuredWindowRateControl.kt
```diff
@@ -0,0 +1,62 @@
+package misk.web.jetty
```
squarejesse, line 13: Don’t forget to include Jetty’s copyright header for code derived from that project!

```suggestion
//
// ========================================================================
// Copyright (c) 1995 Mort Bay Consulting Pty Ltd and others.
//
// This program and the accompanying materials are made available under the
// terms of the Eclipse Public License v. 2.0 which is available at
// https://www.eclipse.org/legal/epl-2.0, or the Apache License, Version 2.0
// which is available at https://www.apache.org/licenses/LICENSE-2.0.
//
// SPDX-License-Identifier: EPL-2.0 OR Apache-2.0
// ========================================================================
//
package misk.web.jetty
```

squarejesse: I copied from this file:
https://github.com/jetty/jetty.project/blob/41d8f51edd33e81f6658e025488f13fad94d95a6/jetty-core/jetty-http2/jetty-http2-common/src/main/java/org/eclipse/jetty/http2/WindowRateControl.java#L33

misk/src/main/kotlin/misk/web/jetty/MeasuredWindowRateControl.kt
```diff
@@ -0,0 +1,62 @@
...
+    "Count of rate limited events"
+  )
+
+  private val events = ConcurrentLinkedQueue<Long>()
+  private val size = AtomicInteger()
+  private val window = Duration.ofSeconds(1).toNanos()
+
+  override fun onEvent(event: Any?): Boolean {
+    val now = NanoTime.now()
+    while (true) {
```
squarejesse, line 47: I believe this loop is attempting to prune obsolete events from the queue, keeping `size` consistent with the size of the queue?

misk/src/main/kotlin/misk/web/jetty/MeasuredWindowRateControl.kt
```diff
@@ -0,0 +1,62 @@
...
+      val time = events.peek() ?: break
+      if (NanoTime.isBefore(now, time)) break
+      if (events.remove(time)) {
```
squarejesse, line 50: This is surprising . . . if it loses the race this will do a linear scan through a window worth of events to do nothing

misk/src/main/kotlin/misk/web/jetty/MeasuredWindowRateControl.kt
```diff
@@ -0,0 +1,62 @@
...
+internal class MeasuredWindowRateControl internal constructor(
+  private val metrics: Metrics,
+  private val maxEvents: Int,
+) : RateControl {
+
+  private val gauge = metrics.peakGauge(
+    "jetty_http2_rate_control_events_peak",
+    "Peak gauge of observed events per second"
+  )
+  private val counter = metrics.counter(
```
squarejesse: Rename to `rateLimitedEventCounter` ?

squarejesse (approved):

misk/src/main/kotlin/misk/web/jetty/MeasuredWindowRateControl.kt
squarejesse: (Keeping it consistent with Jetty is good though.)

misk/src/main/kotlin/misk/web/jetty/MeasuredWindowRateControl.kt
kevink-sq: Yes the `size` constant should be the queue size although I'm not sure though why `events.size()` was not used instead. As for the loop, even though it looks O(n), amortized cost will behave like O(1).

misk/src/main/kotlin/misk/web/jetty/MeasuredWindowRateControl.kt
kevink-sq: I believe amortized cost will still be O(1) with the queue implementation (each event is either added or removed once from the queue).

misk/src/main/kotlin/misk/web/jetty/MeasuredWindowRateControl.kt
ThePumpingLemma: @kevink-sq From the docs:

> Beware that, unlike in most collections, the size method is NOT a constant-time operation. Because of the asynchronous nature of these queues, determining the current number of elements requires a traversal of the elements, and so may report inaccurate results if this collection is modified during traversal.


> I believe this loop is attempting to prune obsolete events from the queue, keeping size consistent with the size of the queue?

Yes, that's exactly what it's doing. It's required because `size()` is not a constant time operation.

misk/src/main/kotlin/misk/web/jetty/MeasuredWindowRateControl.kt
ThePumpingLemma: @kevink-sq The problem is if two threads call this handler concurrently, its possible that one thread pops the item and the second thread tries to pop it but it has to search though all elements first, which is a linear operation.

We could emit a counter increment here every time we lose the race and that would tell is of it's a practical concern.

misk/src/main/kotlin/misk/web/WebConfig.kt
```diff
@@ -209,6 +209,13 @@ data class WebConfig @JvmOverloads constructor(
   /** Config used by client and server interceptors installed by DeadlinePropagationModule
    * Only applies if DeadlinePropagationModule is installed, ignored otherwise  */
   val request_deadlines: RequestDeadlinesConfig = RequestDeadlinesConfig(),
+
+  /**
+   * Exposed configuration for Jetty's HTTP/2 frame rate limiter to avoid DDoS (CVE-2023-44487, CVE-2025-5115).
+   * -1 will effectively disable rate limiting but will still emit metrics on the rate limiter queue.
+   * See https://jetty.org/docs/jetty/10/operations-guide/modules/standard.html#http2
+   */
+  val jetty_http2_max_events_per_second: Int = 128,
```
ThePumpingLemma, line 218: Should we disable by default so as to not change the existing behavior?

misk/src/main/kotlin/misk/web/jetty/MeasuredWindowRateControl.kt
kevink-sq: I thought Jesse's comment was race against time. The thread clarification makes sense.

misk/src/main/kotlin/misk/web/WebConfig.kt
ThePumpingLemma: We chatted and:

> I don't think so. Jetty already has the frame rate limit limiter with 128/s. 10.0.26 though makes it so the rate limiter applies to server side event frames

kevink-sq: The current value is the default behavior (see [10.0.24](https://github.com/jetty/jetty.project/blob/dbb94514dc9d3fb21fe92080f57c314e7e06a148/jetty-http2/http2-server/src/main/java/org/eclipse/jetty/http2/server/AbstractHTTP2ServerConnectionFactory.java#L66)). My intent was to keep the existing Jetty behavior and modify in skim.

misk/src/main/kotlin/misk/web/jetty/MeasuredWindowRateControl.kt
kevink-sq: Found the [comment](https://github.com/jetty/jetty.project/pull/3996#discussion_r315080906) regarding this:
> I think this is good, as remove(Object) will always operate on the oldest event and only in case of a race run through the whole queue without finding it, and the queue is limited to a small number (20 by default).

Although now it is 128

PR #3409, opened 2025-09-29
misk-hibernate: use single connection in Vitess shard targeting and improve interface

aparajon:
## Description

Update the shard targeting logic to use a single connection throughout the entire lifecycle and align the logic with an internal version of this component that we use on a different service framework that's more battle tested for Vitess. This is necessary to prevent issues where connections don't properly get reset after shard targeting operations, which pollutes the underlying connection pool and can lead to queries picking up a connection that is targeting an unintended destination.

Additionally, introduce a new interface `session.target(destination: Destination, block: (session: Session) -> T`, which allows for more explicit routing than the current interface which only accepts `Shard`, as Destination can also include  `@primary` or `@replica`.

Also moved over the shard targeting tests in `TransacterTest` into a new suite `VitessShardTargetIntegrationTest` for better organization. Also added new tests there as well which test the new `session.target(Destination...)` interface.

Lastly, removed `fun <T> Transacter.failSafeRead(shard: Shard, block: (session: Session) -> T): T`, which is not in use.

## Testing Strategy

Existing shard targeting tests were moved to `VitessShardTargetIntegrationTest`

Canary tested in a service

## Definition of Done
Shard targeting correctly works in Misk service consumers and connection reset issues don't happen

## Checklist

### DIFF
--- a/misk-hibernate/src/main/kotlin/misk/hibernate/Hibernate.kt
+++ b/misk-hibernate/src/main/kotlin/misk/hibernate/Hibernate.kt
@@ -38,11 +38,11 @@ internal val Metadata.allProperties: Multimap<Class<*>, Property>
         if (value is Component) {
           for (subProperty in value.propertyIterator) {
             if (subProperty is Property) {
-              result.put(Class.forName(value.componentClass.name), subProperty)
+              result.put(Class.forName(value.componentClass.name, false, Thread.currentThread().contextClassLoader), subProperty)
             }
           }
         } else {
-          result.put(Class.forName(entityBinding.className), property)
+          result.put(Class.forName(entityBinding.className, false, Thread.currentThread().contextClassLoader), property)
         }
       }
     }
@@ -70,7 +70,7 @@ internal fun Properties.getField(name: String): Field? {
   val declaringClassName = getProperty("${name}DeclaringClass") ?: return null
   val fieldName = getProperty("${name}Name") ?: return null
 
-  val entityClass = Class.forName(declaringClassName)
+  val entityClass = Class.forName(declaringClassName, false, Thread.currentThread().contextClassLoader)
   return entityClass.getDeclaredField(fieldName)
 }
 
--- a/misk-hibernate/src/main/kotlin/misk/hibernate/SessionFactoryService.kt
+++ b/misk-hibernate/src/main/kotlin/misk/hibernate/SessionFactoryService.kt
@@ -256,7 +256,7 @@ internal class SessionFactoryService(
       "float" -> Float::class
       "double" -> Double::class
       "materialized_clob" -> String::class
-      else -> Class.forName(name).kotlin
+      else -> Class.forName(name, false, Thread.currentThread().contextClassLoader).kotlin
     }
   }
 
--- a/misk-redis/src/main/kotlin/misk/redis/FakeRedis.kt
+++ b/misk-redis/src/main/kotlin/misk/redis/FakeRedis.kt
@@ -44,10 +44,11 @@ class FakeRedis : Redis {
 
   @Synchronized
   override fun del(key: String): Boolean {
-    if (!keyValueStore.containsKey(key)) {
-      return false
-    }
-    return keyValueStore.remove(key) != null
+    var deleted = false
+    if (keyValueStore.remove(key) != null) deleted = true
+    if (hKeyValueStore.remove(key) != null) deleted = true
+    if (lKeyValueStore.remove(key) != null) deleted = true
+    return deleted
   }
 
   @Synchronized
--- a/misk-redis/src/test/kotlin/misk/redis/AbstractRedisTest.kt
+++ b/misk-redis/src/test/kotlin/misk/redis/AbstractRedisTest.kt
@@ -255,6 +255,78 @@ abstract class AbstractRedisTest {
     listOf(*keysToInsert, key3).forEach { assertNull(redis[it], "Key should have been deleted") }
   }
 
+  @Test fun deleteHashKey() {
+    val key = "hashKey"
+    val field = "field1"
+    val value = "value".encodeUtf8()
+
+    // Set hash field
+    redis.hset(key, field, value)
+    assertEquals(value, redis.hget(key, field), "Hash field should have been set")
+    assertTrue(redis.exists(key), "Hash key should exist")
+
+    // Delete hash key
+    assertTrue(redis.del(key), "Hash key should have been deleted")
+    assertNull(redis.hget(key, field), "Hash field should be gone after key deletion")
+    assertFalse(redis.exists(key), "Hash key should not exist after deletion")
+
+    // Try deleting the same key again
+    assertFalse(redis.del(key), "Should not have deleted anything")
+  }
+
+  @Test fun deleteListKey() {
+    val key = "listKey"
+    val elements = listOf("element1", "element2").map { it.encodeUtf8() }
+
+    // Push elements to list
+    redis.lpush(key, *elements.toTypedArray())
+    assertEquals(2L, redis.llen(key), "List should have 2 elements")
+    assertTrue(redis.exists(key), "List key should exist")
+
+    // Delete list key
+    assertTrue(redis.del(key), "List key should have been deleted")
+    assertEquals(0L, redis.llen(key), "List should be empty after key deletion")
+    assertFalse(redis.exists(key), "List key should not exist after deletion")
+
+    // Try deleting the same key again
+    assertFalse(redis.del(key), "Should not have deleted anything")
+  }
+
+  @Test fun deleteMixedDataTypes() {
+    val stringKey = "stringKey"
+    val hashKey = "hashKey"
+    val listKey = "listKey"
+    val nonExistentKey = "nonExistentKey"
+
+    val stringValue = "stringValue".encodeUtf8()
+    val hashField = "field1"
+    val hashValue = "hashValue".encodeUtf8()
+    val listElements = listOf("element1", "element2").map { it.encodeUtf8() }
+
+    // Set up different data types
+    redis[stringKey] = stringValue
+    redis.hset(hashKey, hashField, hashValue)
+    redis.lpush(listKey, *listElements.toTypedArray())
+
+    // Verify all keys exist
+    assertTrue(redis.exists(stringKey), "String key should exist")
+    assertTrue(redis.exists(hashKey), "Hash key should exist")
+    assertTrue(redis.exists(listKey), "List key should exist")
+    assertFalse(redis.exists(nonExistentKey), "Non-existent key should not exist")
+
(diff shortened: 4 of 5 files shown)


misk-hibernate/src/main/kotlin/misk/hibernate/RealTransacter.kt
```diff
@@ -262,33 +266,31 @@ private constructor(
...
-          session.preCommit()
-          transaction.commit()
-          session.postCommit()
-          result
-        } catch (e: Throwable) {
-          var rethrow = e
-          if (config.type == DataSourceType.TIDB) {
-            rethrow = mapTidbException(e)
-          }
+      val transaction = session.hibernateSession.beginTransaction()!!
```
yyuan-squareup, line 298: I guess the intention here is to shift the responsibility of whether to target into callers of `transactionInternalSession`.
I saw it is called by transaction() at line 148. Would it change its behavior?

misk-hibernate/src/main/kotlin/misk/hibernate/Session.kt
```diff
@@ -44,4 +59,4 @@ inline fun <R : DbRoot<R>, reified S : DbSharded<R, S>> Session.loadSharded(gid:
     gid, S::class
   )
 
-inline fun <reified T : DbEntity<T>> Session.loadOrNull(id: Id<T>): T? = loadOrNull(id, T::class)
+inline fun <reified T : DbEntity<T>> Session.loadOrNull(id: Id<T>): T? = loadOrNull(id, T::class)
```
yyuan-squareup: nit: new line

misk-hibernate/src/main/kotlin/misk/hibernate/Session.kt
```diff
@@ -19,7 +19,22 @@ interface Session: misk.jdbc.Session {
...
+   *
+   * The destination is merged with the current destination, allowing for composable targeting:
+   * - In a regular transaction: target(Destination(shard)) routes to that shard
+   * - In a replica read: target(Destination(shard)) routes to that shard's replica
+   * - Can explicitly target: target(Destination(shard, TabletType.REPLICA))
+   * 
+   * @param destination The destination to target (shard, tablet type, or both)
+   * @param block Function that receives the session and executes within the targeted context
+   */
+  fun <T> target(destination: Destination, block: (session: Session) -> T): T
```
yyuan-squareup, line 32: In the interface Transacter.kt
```
fun <T> Transacter.transaction(shard: Shard, block: (session: Session) -> T): T =
  transaction { it.target(shard) { block(it) } }
```
Should it be migrated to use this version of target()? Not sure how much effort is needed.

aparajon: Yep that's a good find, now's a good time to migrate this call site as well

misk-hibernate/src/main/kotlin/misk/hibernate/RealTransacter.kt
aparajon: > I guess the intention here is to shift the responsibility of whether to target into callers of transactionInternalSession.

Yep that's correct. My rationale for this change: 
- We don't do this in our internal version of this construct
- For non-Vitess usages, it resolves to a no-op, which I think adds unneeded complexity
- It can potentially mask connection reset issues, since it forcefully sets the `Destination` to `Primary`. In theory, this change should help us get faster feedback that the connection reset issues are resolved

> I saw it is called by transaction() at line 148. Would it change its behavior?

I don't anticipate behavior changes since the only change here is removing the outer `session.target(Destination(TabletType.PRIMARY)` scope (which resolves to a no-op for non-Vitess). All the other original logic is left untouched

misk-hibernate/src/main/kotlin/misk/hibernate/Session.kt
aparajon: Added in https://github.com/cashapp/misk/pull/3409/commits/98af70bf33026b1ca9e0867d6708819c70ba8a7c

misk-hibernate/src/main/kotlin/misk/hibernate/RealTransacter.kt
```diff
@@ -156,7 +156,11 @@ private constructor(
     }
     return transactionWithRetriesInternal {
       replicaReadWithoutTransactionInternalSession { session ->
-        session.target(Destination(TabletType.REPLICA)) { block(session) }
+        if (config.type.isVitess && config.database != Destination.replica().toString()) {
+          session.target(Destination.replica(), block)
+        } else {
+          block(session)
```
chenlily, line 160: I think we should remove this condition: `&& config.database != Destination.replica().toString()`
If I understand `session.target` correctly, the current target could include a specific shard (for ex. `loanstar/-80@replica`), in which case we probably want to update the connection to `@replica`

https://github.com/cashapp/misk/blob/094c7d101b205cff7171802f638ac4ae6666a6be/misk-hibernate/src/main/kotlin/misk/hibernate/RealTransacter.kt#L570-L571

misk-hibernate/src/main/kotlin/misk/hibernate/RealTransacter.kt
```diff
@@ -555,71 +557,61 @@ private constructor(
...
-              use(previous)
-            }.onFailure { e ->
-              logger.error(e) {
-                "Exception restoring destination, previous = $previous, destination = $destination," +
-                "cause = ${e.message}"
+          useConnection { connection ->
+            val previous = currentTarget(connection)
+            // Merge Destinations to support composition (e.g., targeting a shard with TabletType not specified within a replica read)
+            // This will prefer values from the new Destination when merging.
+            val mergedDestination = previous.mergedWith(destination)
```
chenlily: I thought we had identified that this could be an issue if we had previously targeted a specific shard with the database lock. I think it could make sense to merge in TabletType in the replica read case, but it I'm not convinced it's the right thing to merge with the old keyspace and shard targets

aparajon: Ah I wasn't thinking about how this could interact with advisory locking, good call. I originally resurrected the `merge` logic because our internal `target` method still has it, but we may not actually need it. Tests are passing without it 👍 

If we want to be truly safe, I do think we should convert callers of `target` usages to the new interface, which allows for explicitly setting a `tabletType`, e.g. `session.target(Destination(shard = shard.xx, tabletType = TabletType.REPLICA), ...)`

Not specifying a `tabletType` is basically the gap that the `merge` logic was addressing, since a caller could specify `session.target(shard, ...)` without a `tabletType` and inherit what was used previously from its parent. But if we pass in a `tabletType` explicitly, that would have overridden `merge` logic anyway.

TLDR: I removed the `merge` code, and using the new `session.target` interface should help reduce unintended targeting

misk-hibernate/src/main/kotlin/misk/hibernate/RealTransacter.kt
aparajon: I added a comment here and a test case in https://github.com/cashapp/misk/pull/3409/commits/76b9daf4cc5b3ae80350202f1836722a1a83c3a8. This logic helps avoid an unnecessary `target` call if the connection is already pointing to `@replica`, e.g. from a `readerTransacter.replicaRead` call

The test cases in `VitessShardTargetIntegrationTest` prove that shard targeting still works within a `replicaRead` scope. Otherwise, callers can also do (pseudo-code):

```
 session.target(Destination(shard = shard(loanstar/-80), tabletType = TabletType.REPLICA)) {
     // query with session
 }
```

chenlily: I think the shard targeting works if you specify a shard, but what if you're assuming that a read is simply hitting `@replica`? Here's an example of the case I'm thinking of: 
```
// 1. initial example from test, target a specific shard 
transacter.replicaRead { session ->
      session.target(jp.shard(session)) {
        assertThat(
          queryFactory.newQuery<MovieQuery>().allowTableScan()
            .name("Jurassic Park").uniqueResult(session)
        ).isNotNull
        assertThat(
          queryFactory.newQuery<MovieQuery>().allowTableScan()
            .name("Star Wars").uniqueResult(session)
        ).isNull()
}

// 2. subsequent ReplicaRead. Let's say the caller is assuming that since 
// shard targeting was not explicitly stated,  that the shard targeting
// will be determined in VTgate. 
transacter.replicaRead { session ->
      // not calling session.target here, and this would fail because we are not updating the destination to just
      // @replica
     assertThat(
          queryFactory.newQuery<MovieQuery>().allowTableScan()
            .name("Star Wars").uniqueResult(session)
        ).isNotNull()
}
```

misk-hibernate/src/main/kotlin/misk/hibernate/RealTransacter.kt
chenlily: This part LGTM, thanks for updating!

misk-hibernate/src/main/kotlin/misk/hibernate/RealTransacter.kt
aparajon: I think the second case should still work since `session.target(Destination.replica(), block)` would activate for Vitess and for the typical case of database = `@primary`, which should set the destination to `@replica`. I'll add an explicit test for this to verify

aparajon: Do the tests in https://github.com/cashapp/misk/pull/3409/commits/2ea00cf2366445d738d25278c630a51c1c1edd4d address case 2?

chenlily: Sorry for all the back and forth, I'm still not sure if the tests address the case I'm thinking about, and lmk if you want to hop on a huddle if that's easier :/ 

My understanding is that `config.database ` comes from the data source config, and if we are using the reader transacter, it will be set to `@replica`. Assume that we're only talking about the reader transacter in the above example, such that `config.database != Destination.replica().toString()` is true throughout.

If we happened to set shard targeting while in step 1, won't 2 enter the `else` block and execute with the shard set in step 1?  I guess this is not the typical case, where we generally expect not to set the shard targeting in the reader, but there's nothing explicitly guarding against this right?

aparajon: As we tested this scenario out, this fortunately exposed out an edge case that restoration was not happening properly for shard targeting inside `replicaRead`, since the underlying connection would get closed before reaching the `finally` block in `target` where the restoration was supposed to happen.

This should now be fixed in https://github.com/cashapp/misk/pull/3409/commits/6e601c636493a29101a477b8d346c7184fe7c53c!

chenlily (approved):

PR #3410, opened 2025-09-29
Remove usage of deprecated wisp config

katukota:
Remove deprecated wisp config usage

### DIFF
--- a/misk-config/src/main/kotlin/misk/config/ConfigModule.kt
+++ b/misk-config/src/main/kotlin/misk/config/ConfigModule.kt
@@ -2,7 +2,7 @@ package misk.config
 
 import misk.inject.KAbstractModule
 
-class ConfigModule<T : wisp.config.Config>(
+class ConfigModule<T : Config>(
   private val configClass: Class<T>,
   private val appName: String,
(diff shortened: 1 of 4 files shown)


misk-config/src/main/kotlin/misk/config/ConfigModule.kt
```diff
@@ -11,11 +11,11 @@ class ConfigModule<T : wisp.config.Config>(
   override fun configure() {
     install(AppNameModule(appName))
     bind(configClass).toInstance(config)
-    bind<wisp.config.Config>().toInstance(config)
```
adrw, line 14: This is probably a breaking change if there are existing callsites depending on the wisp config on their injection? If we go update all callsites promptly we can still land it as is.

adrw (approved):
LG

misk-config/src/main/kotlin/misk/config/ConfigModule.kt
katukota: yes, cash-server PR is already ready and the robot change PR is already ready so I should be able to move all existing callsites in next couple of days.

PR #3421, opened 2025-10-01
Revert "fix: add missing rate control to uds (#3420)"

kevink-sq:
This reverts commit 1c293df97d021cf653719e0cf9e62efbba236ee7.

Running into issues:

```
        Caused by:
        java.io.IOException: FRAME_SIZE_ERROR: 4740180
            at okhttp3.internal.http2.Http2Reader.nextFrame(Http2Reader.kt:112)
            at okhttp3.internal.http2.Http2Reader.readConnectionPreface(Http2Reader.kt:75)
            at okhttp3.internal.http2.Http2Connection$ReaderRunnable.invoke(Http2Connection.kt:640)
            at okhttp3.internal.http2.Http2Connection$ReaderRunnable.invoke(Http2Connection.kt:631)
            at okhttp3.internal.concurrent.TaskQueue$execute$1.runOnce(TaskQueue.kt:112)
            at okhttp3.internal.concurrent.TaskRunner$runnable$1.run(TaskRunner.kt:81)
            at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1144)
            at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:642)
            at java.base/java.lang.Thread.run(Thread.java:1583)
```

PR #3422, opened 2025-10-02
feat: reattempt in introducing configurable http/2 control frame limiter with observability

kevink-sq:
Previous attempts (https://github.com/cashapp/misk/pull/3408, https://github.com/cashapp/misk/pull/3420) had to be reverted in [here](https://github.com/cashapp/misk/pull/3421) due to failing integration tests and in-memory misk servers. It was found the metrics object instantiation for every http/2 connection added overhead resulting in client errors. Fixed by instantiating metrics in the factory once.

### DIFF
--- a/misk/src/test/kotlin/misk/web/jetty/MeasuredWindowRateControlTest.kt
+++ b/misk/src/test/kotlin/misk/web/jetty/MeasuredWindowRateControlTest.kt
@@ -5,7 +5,9 @@ import misk.testing.MiskTest
 import misk.testing.MiskTestModule
 import misk.web.jetty.MeasuredWindowRateControl
 import org.assertj.core.api.Assertions.assertThat
+import org.eclipse.jetty.io.EndPoint
 import org.junit.jupiter.api.Test
+import org.mockito.Mockito.mock
 
 @MiskTest(startService = false)
 class MeasuredWindowRateControlTest {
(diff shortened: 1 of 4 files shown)


misk/src/test/kotlin/misk/web/jetty/MeasuredWindowRateControlTest.kt
```diff
@@ -24,7 +30,10 @@ class MeasuredWindowRateControlTest {
 
   @Test
   fun `blocks events when over maxEvents limit`() {
-    val rateControl = MeasuredWindowRateControl(metrics, maxEvents = 2)
+    val rateControl = MeasuredWindowRateControl.Factory(
+      metrics,
+      maxEventRate = 2
+    ).newRateControl(mock(EndPoint::class.java))
```
adrw: Why do we need to mock this?

adrw (approved):
LG

misk/src/test/kotlin/misk/web/jetty/MeasuredWindowRateControlTest.kt
ThePumpingLemma: Probably because the function signature is `override fun newRateControl(endPoint: EndPoint): RateControl` instead of `override fun newRateControl(endPoint: EndPoint?): RateControl`? In Jetty's Java interface for rate control factories, it's not clear if it needs to be non-null or not.

kevink-sq: It is not really needed. I updated the function signature to `endPoint: EndPoint?` and removed cleaned up the test dependency.

PR #3435, opened 2025-10-03
Misk servlet embeddable to an external servlet containers

jvmakine:
This allows running Misk servlet in a separately managed servlet container. 

The first commit allows us to disable JettyService from running with Misk.

The second commit exposes the Servlet as `MiskServlet` in the Guice injector, makes the underlying infrastructure work with generic servlet API, and supports mounting the servlet to a servlet path.

### DIFF
--- a/misk/src/main/kotlin/misk/web/WebConfig.kt
+++ b/misk/src/main/kotlin/misk/web/WebConfig.kt
@@ -217,7 +217,7 @@ data class WebConfig @JvmOverloads constructor(
    */
   val jetty_http2_max_events_per_second: Int = 128,
 
-  /** If true, disables the embedded Jetty server */
+  /** If true, disables the embedded Jetty server. */
   val disable_jetty: Boolean = false,
 ) : Config
 
--- a/misk/src/main/kotlin/misk/web/WebConfig.kt
+++ b/misk/src/main/kotlin/misk/web/WebConfig.kt
@@ -216,6 +216,9 @@ data class WebConfig @JvmOverloads constructor(
    * See https://jetty.org/docs/jetty/10/operations-guide/modules/standard.html#http2
    */
   val jetty_http2_max_events_per_second: Int = 128,
+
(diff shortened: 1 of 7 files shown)


misk/src/main/kotlin/misk/web/WebConfig.kt
```diff
@@ -216,6 +216,9 @@ data class WebConfig @JvmOverloads constructor(
    * See https://jetty.org/docs/jetty/10/operations-guide/modules/standard.html#http2
    */
   val jetty_http2_max_events_per_second: Int = 128,
+
+  /** If true, disables the embedded Jetty server */
```
adrw:
```suggestion
  /** If true, disables the embedded Jetty server. */
```

misk/src/main/kotlin/misk/web/jetty/WebActionsServlet.kt
```diff
@@ -131,26 +131,13 @@ internal class WebActionsServlet @Inject constructor(
...
-              (this.connector as ServerConnector).localPort
-            )
-
-            else -> throw IllegalStateException("Unknown socket connector.")
-          }
-        },
+        linkLayerLocalAddress = extractLinkLayerLocalAddress(request),
         dispatchMechanism = dispatchMechanism,
-        upstreamResponse = JettyServletUpstreamResponse(response as Response),
+        upstreamResponse = if (response is Response) {
```
adrw, line 136: So the Jetty Servelet can technically have non-Jetty typed responses and that's what will allow this?

adrw (approved):
LG

misk/src/main/kotlin/misk/web/jetty/WebActionsServlet.kt
jvmakine: I think JSC does some wrapping and the responses there are not Jetty specific, thus this is needed. Now, it should work in both Jetty and Generic servlets.

PR #3439, opened 2025-10-03
Add mcp_tool_handler_latency metric

ean5533:
## Description

This will allow misk-mcp users to see how long individual tool calls are taking, rather than all tools bundled together per MCP server action. It also adds visibility into how tools are ending (success result, error result, or threw an exception).

## Checklist

### DIFF
--- a/misk-mcp/src/main/kotlin/misk/mcp/MiskMcpServer.kt
+++ b/misk-mcp/src/main/kotlin/misk/mcp/MiskMcpServer.kt
@@ -25,6 +25,7 @@ import misk.mcp.config.asResources
 import misk.mcp.config.asTools
 import misk.mcp.internal.MiskServerTransport
 import kotlin.time.Duration.Companion.nanoseconds
+import kotlin.time.TimeSource
 
 /**
  * Misk implementation of a Model Context Protocol (MCP) server.
@@ -177,7 +178,7 @@ class MiskMcpServer internal constructor(
     handler: suspend (CallToolRequest) -> CallToolResult
   ): suspend (CallToolRequest) -> CallToolResult {
     return { request ->
-      val start = System.nanoTime()
+      val mark =  TimeSource.Monotonic.markNow()
       var outcome = McpMetrics.ToolCallOutcome.Success
       
       try {
@@ -190,7 +191,7 @@ class MiskMcpServer internal constructor(
         outcome = McpMetrics.ToolCallOutcome.Exception
         throw ex
       } finally {
-        val duration = (System.nanoTime() - start).nanoseconds
+        val duration = mark.elapsedNow()
         mcpMetrics.mcpToolHandlerLatency(duration, name, toolName, outcome)
       }
     }


misk-mcp/src/main/kotlin/misk/mcp/McpMetrics.kt
```diff
@@ -0,0 +1,27 @@
+package misk.mcp
+
+import jakarta.inject.Inject
+import jakarta.inject.Singleton
+import misk.metrics.v2.Metrics
+import kotlin.time.Duration
+
+@Singleton
```
jclyne, line 8: nit: this doesn't need to be a singleton

misk-mcp/src/main/kotlin/misk/mcp/McpMetrics.kt
```diff
@@ -0,0 +1,27 @@
...
+  fun mcpToolHandlerLatency(duration: Duration, serverName: String, toolName: String, outcome: ToolCallOutcome) {
+    mcpToolHandlerLatency
+      .labels(serverName, toolName, outcome.name)
+      .observe(duration.inWholeMilliseconds.toDouble())
+  }
+  
+  enum class ToolCallOutcome {
+    Success,
+    Error,
+    Exception,
```
jclyne, line 25: nit: should this be Failure?

misk-mcp/src/main/kotlin/misk/mcp/MiskMcpServer.kt
```diff
@@ -160,11 +164,35 @@ class MiskMcpServer internal constructor(
...
       )
     }
   }
 
+  private fun metricReportingHandler(
+    toolName: String,
+    handler: suspend (CallToolRequest) -> CallToolResult
+  ): suspend (CallToolRequest) -> CallToolResult {
+    return { request ->
+      val start = System.nanoTime()
```
jclyne: to be more idiomatic, this can be a time mark:
`val mark =  TimeSource.Monotonic.markNow()`

Then later you can get the duration from the mark with:
`mark.elapsedNow()`

misk-mcp/src/main/kotlin/misk/mcp/McpMetrics.kt
ean5533: Hmm, are you sure? At a minimum it would create multiple Collectors for the same metric which would be a waste of resources. But a quick walk through the prometheus code suggests that registering multiple collectors for the same name would throw:

```
  public void register(Collector m) {
    List<String> names = collectorNames(m);
    assertNoDuplicateNames(m, names);
    synchronized (namesCollectorsLock) {
      for (String name : names) {
        if (namesToCollectors.containsKey(name)) {
          throw new IllegalArgumentException("Failed to register Collector of type " + m.getClass().getSimpleName()
                  + ": " + name + " is already in use by another Collector of type "
                  + namesToCollectors.get(name).getClass().getSimpleName());
        }
      }
      for (String name : names) {
        namesToCollectors.put(name, m);
      }
      collectorsToNames.put(m, names);
    }
  }
```

misk-mcp/src/main/kotlin/misk/mcp/McpMetrics.kt
ean5533: I think "Failure" is kind of ambiguous; if I saw it I wouldn't know what it means compared to "Error", but "Exception" is pretty self-explanatory that an exception was thrown

misk-mcp/src/main/kotlin/misk/mcp/MiskMcpServer.kt
ean5533: TIL, thanks!

misk-mcp/src/main/kotlin/misk/mcp/McpMetrics.kt
jclyne: Yeah, you are right. This is prometheus now, not micrometer. Micrometer has an idempotent register call.

PR #3447, opened 2025-10-06
Support getChatPermalink and threaded responses on slack api

hirzamitz:
## Description
Add support for `getChatPermalink` and sending responses to message threads in Slack api

## Testing Strategy
Tested changes locally

## Checklist


misk-slack/src/main/kotlin/misk/slack/webapi/RealSlackClient.kt
```diff
@@ -41,6 +42,10 @@ class RealSlackClient @Inject constructor(
     return callSlack { slackApi.updateUserGroup(request).execute() }
   }
 
+  fun getChatPermalink(channel: String, ts: String): GetChatPermalinkResponse {
```
adrw: Can you write out fully what ts is? thread something? We don't usually do abbreviated / acronym variable names because the cost to maintainability and readability isn't worth it.

misk-slack/src/main/kotlin/misk/slack/webapi/helpers/SlackJson.kt
```diff
@@ -9,6 +9,7 @@ data class PostMessageRequest @JvmOverloads constructor(
   val channel: String,
   val response_type: String? = "in_channel",
   val blocks: List<Any>,
+  val thread_ts: String? = null,
```
adrw: What does ts stand for? Can we write it out in full?

adrw (approved):
LG

misk-slack/src/main/kotlin/misk/slack/webapi/RealSlackClient.kt
hirzamitz: It says `thread_ts` and `message_ts` on the Slack docs, but it's a unique ID based on the timestamp. Updated variable names to `thread_timestamp` and `message_timestamp `

misk-slack/src/main/kotlin/misk/slack/webapi/helpers/SlackJson.kt
hirzamitz: Updated to `thread_timestamp`

PR #3452, opened 2025-10-07
SRE-627: match hibernate exception handling in jooq

vitaly-block:
# Hoist hibernate-like exception handling to misk-jdbc to cover other providers with the same logic

## Description

Refactored the TransactionRetryHandler to make database exception retry logic specific to the database type being used.
Database-Specific Retry Logic: Retryable exceptions are now only checked for the configured database type:
* Connection closed exceptions remain retryable for all databases
* Vitess-specific exceptions only retry when DataSourceType.VITESS_MYSQL is configured
* CockroachDB-specific exceptions only retry when DataSourceType.COCKROACHDB is configured
* TiDB-specific exceptions only retry when DataSourceType.TIDB is configured

## API Changes
New Constructor: Added convenient TransactionRetryHandler(qualifierName, databaseType) constructor for easy database type-specific configuration
Backward Compatibility: Existing constructors and APIs remain unchanged

```kotlin
// Before: All database exceptions checked regardless of database type
isMessageRetryable(th)

// After: Database type-specific checking
isConnectionClosed(th) ||
  (databaseType == DataSourceType.VITESS_MYSQL && isRecoverableVitessException(th)) ||
  (databaseType == DataSourceType.COCKROACHDB && isRecoverableCockroachException(th)) ||
  (databaseType == DataSourceType.TIDB && isRecoverableTidbException(th))
```

There should be no breaking changes, existing code should work:

```kotlin
// Old (still works)
TransactionRetryHandler("mydb")

// New (recommended)
TransactionRetryHandler("mydb", DataSourceType.VITESS_MYSQL)
```

## Testing Strategy

New test cases included, https://gradle.com/s/7wbp6xzb5nt4q

## Checklist

### DIFF
--- a/misk-hibernate/src/main/kotlin/misk/hibernate/HibernateExceptionClassifier.kt
+++ b/misk-hibernate/src/main/kotlin/misk/hibernate/HibernateExceptionClassifier.kt
@@ -1,7 +1,7 @@
 package misk.hibernate
 
-import misk.jdbc.DefaultExceptionClassifier
-import misk.jdbc.RetryTransactionException
+import misk.jdbc.retry.DefaultExceptionClassifier
+import misk.jdbc.retry.RetryTransactionException
 import javax.persistence.OptimisticLockException
 import org.hibernate.StaleObjectStateException
 import org.hibernate.exception.LockAcquisitionException
--- a/misk-hibernate/src/main/kotlin/misk/hibernate/RealTransacter.kt
+++ b/misk-hibernate/src/main/kotlin/misk/hibernate/RealTransacter.kt
@@ -17,8 +17,10 @@ import java.util.concurrent.Future
 import java.util.concurrent.TimeUnit
 import javax.persistence.OptimisticLockException
 import kotlin.reflect.KClass
+import misk.backoff.ExponentialBackoff
+import misk.backoff.RetryConfig
+import misk.backoff.retry
 import misk.concurrent.ExecutorServiceFactory
-import misk.jdbc.TransactionRetryHandler
 import misk.hibernate.advisorylocks.tryAcquireLock
 import misk.hibernate.advisorylocks.tryReleaseLock
 import misk.jdbc.CheckDisabler
@@ -58,11 +60,10 @@ private constructor(
   private val shardListFetcher: ShardListFetcher,
   private val hibernateEntities: Set<HibernateEntity>,
 ) : Transacter {
-  
-  private val retryHandler = TransactionRetryHandler(
-    qualifierName = qualifier.simpleName ?: "hibernate",
-    exceptionClassifier = HibernateExceptionClassifier()
-  )
+
+  private val qualifierName = qualifier.simpleName ?: "hibernate"
+  private val exceptionClassifier = HibernateExceptionClassifier()
+
   constructor(
     qualifier: KClass<out Annotation>,
     sessionFactoryService: SessionFactoryService,
@@ -255,14 +256,16 @@ private constructor(
   }
 
   private fun <T> transactionWithRetriesInternal(block: () -> T): T {
-    return retryHandler.executeWithRetries(
-      maxAttempts = options.maxAttempts,
-      minRetryDelayMillis = options.minRetryDelayMillis,
-      maxRetryDelayMillis = options.maxRetryDelayMillis,
-      retryJitterMillis = options.retryJitterMillis
-    ) {
-      block()
-    }
+    val backoff = ExponentialBackoff(
+      baseDelay = Duration.ofMillis(options.minRetryDelayMillis),
+      maxDelay = Duration.ofMillis(options.maxRetryDelayMillis),
+      jitter = Duration.ofMillis(options.retryJitterMillis)
+    )
+    val retryConfig = RetryConfig.Builder(options.maxAttempts, backoff)
+      .shouldRetry { exceptionClassifier.isRetryable(it) }
+      .onRetry { attempt, e -> logger.info(e) { "$qualifierName transaction failed, retrying (attempt $attempt)" } }
+      .build()
+    return retry(retryConfig) { block() }
   }
 
   private fun <T> transactionInternalSession(block: (session: RealSession) -> T): T {
--- a/misk-hibernate/src/test/kotlin/misk/hibernate/TransacterTest.kt
+++ b/misk-hibernate/src/test/kotlin/misk/hibernate/TransacterTest.kt
@@ -6,7 +6,7 @@ import misk.hibernate.VitessTestExtensions.createInSeparateShard
 import misk.hibernate.VitessTestExtensions.save
 import misk.hibernate.VitessTestExtensions.shard
 import misk.jdbc.DataSourceType
-import misk.jdbc.RetryTransactionException
+import misk.jdbc.retry.RetryTransactionException
 import misk.jdbc.uniqueString
 import misk.testing.MiskExternalDependency
 import misk.testing.MiskTest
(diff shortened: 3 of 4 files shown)


misk-jooq/src/main/kotlin/misk/jooq/JooqTransacter.kt
```diff
@@ -152,9 +168,73 @@ class JooqTransacter @JvmOverloads constructor(
...
+          // For DataAccessException without SQLException cause, always retry to match original behavior
+          true
+        }
+      }
+      is SQLException -> if (isMessageRetryable(th)) true else isCauseRetryable(th)
+      else -> isCauseRetryable(th)
+    }
+  }
+
+  private fun isMessageRetryable(th: SQLException) =
```
ThePumpingLemma: Can we hoist these checks into `misk-jdbc` (or some other common library) and share between hibernate and jooq?

misk-jooq/src/main/kotlin/misk/jooq/JooqTransacter.kt
```diff
@@ -20,8 +20,12 @@ import org.jooq.impl.DataSourceConnectionProvider
 import org.jooq.impl.DefaultExecuteListenerProvider
 import org.jooq.impl.DefaultTransactionProvider
 import misk.logging.getLogger
+import java.sql.SQLException
+import java.sql.SQLRecoverableException
+import java.sql.SQLTransientException
 import java.time.Clock
 import java.time.Duration
+import javax.persistence.OptimisticLockException
```
ThePumpingLemma: We talked offline that there may be a better jooq-specific exception, see: https://www.jooq.org/doc/latest/manual/sql-execution/crud-with-updatablerecords/optimistic-locking/

misk-jooq/src/main/kotlin/misk/jooq/JooqTransacter.kt
```diff
@@ -152,9 +168,73 @@ class JooqTransacter @JvmOverloads constructor(
     DataSourceType.VITESS_MYSQL -> SQLDialect.MYSQL
     DataSourceType.POSTGRESQL -> SQLDialect.POSTGRES
     DataSourceType.TIDB -> SQLDialect.MYSQL
+    DataSourceType.COCKROACHDB -> throw IllegalArgumentException("CockroachDB is not supported in JOOQ")
```
ThePumpingLemma: Doesn't the `else` already cover this case? We were discussing deleting support from Misk entirely but I think we can do this in a separate PR.

vitaly-block: I thought it would be nice for it to be more explicit. Sure, I can go away with else

misk-jooq/src/main/kotlin/misk/jooq/JooqTransacter.kt
vitaly-block: yup yup yup

misk-jooq/src/main/kotlin/misk/jooq/JooqTransacter.kt
```diff
@@ -152,9 +168,73 @@ class JooqTransacter @JvmOverloads constructor(
...
     else -> throw IllegalArgumentException("no SQLDialect for " + this.name)
   }
 
+  /**
+   * Determines if an exception should trigger a transaction retry.
+   * Adapted from Hibernate RealTransacter exception handling logic (without CockroachDB support).
+   */
+  private fun isRetryable(th: Throwable): Boolean {
+    return when (th) {
```
aparajon: From our internal JSC component, we may also want to consider adding  `OptimisticLockException` from `javax.persistence`, which is ORM agnostic

Edit: ah I missed that you explicitly changed this in https://github.com/cashapp/misk/pull/3452/commits/8c43db6952c97aace3372a123b7305855a7f8253. Perhaps `OptimisticLockException`  can be included in the common ORM agnostic exceptions that are shared between jOOQ and Hibernate

misk-jooq/src/main/kotlin/misk/jooq/JooqTransacter.kt
```diff
@@ -152,9 +168,73 @@ class JooqTransacter @JvmOverloads constructor(
...
+        }
+      }
+      is SQLException -> if (isMessageRetryable(th)) true else isCauseRetryable(th)
+      else -> isCauseRetryable(th)
+    }
+  }
+
+  private fun isMessageRetryable(th: SQLException) =
+    isConnectionClosed(th) ||
+      isVitessTransactionNotFound(th) ||
```
aparajon: There are quite a bit of other retryable message strings that we may want to include here that are in our JSC component. Perhaps worth a follow-up PR if we want to keep this PR simpler

misk-jooq/src/main/kotlin/misk/jooq/JooqTransacter.kt
```diff
@@ -152,9 +168,73 @@ class JooqTransacter @JvmOverloads constructor(
...
+      isTidbWriteConflict(th)
```
aparajon: Also following our JSC component, I think we may want to gate database specific retryable exceptions by DB type, e.g.,:

```
isConnectionClosed(th) 
|| (databaseType == DatabaseType.VITESS_MYSQL && isRecoverableVitessException(th)) 
|| (databaseType == DatabaseType.TIDB && isRecoverableTidbException(t))
```

This could also be a future follow-up if we don't want to add too much complexity to this PR

misk-jooq/src/main/kotlin/misk/jooq/JooqTransacter.kt
aparajon: Ok scratch that, jOOQ is not a JPA provider

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -0,0 +1,7 @@
+distributionBase=GRADLE_USER_HOME
```
adrw: This shouldn't be checked in, your local setup seems broken since it should be using Hermit?

gradlew
```diff
@@ -0,0 +1,251 @@
+#!/bin/sh
```
adrw: delete pls

misk-jdbc/src/main/kotlin/misk/jdbc/TransactionRetryHandler.kt
```diff
@@ -0,0 +1,187 @@
...
+ * Interface for classifying exceptions as retryable or non-retryable.
+ */
+interface ExceptionClassifier {
+  fun isRetryable(throwable: Throwable): Boolean
+}
+
+/**
+ * Default exception classifier that handles common database retry scenarios.
+ */
+open class DefaultExceptionClassifier @JvmOverloads constructor(
```
tgregory-block: Can we pull this, the interface, and the `RetryTransactionException` into their own files? Might make sense to put them into like an `exceptionmapper` subpackage or similar as well

misk-jdbc/src/main/kotlin/misk/jdbc/TransactionRetryHandler.kt
```diff
@@ -0,0 +1,187 @@
...
+  constructor(
+    qualifierName: String = "database",
+    databaseType: DataSourceType? = null
+  ) : this(qualifierName, DefaultExceptionClassifier(databaseType), databaseType)
+  
+  /**
+   * Executes a block with retry logic for transient database failures.
+   */
+  @JvmOverloads
+ fun <T> executeWithRetries(
```
tgregory-block: If we pass a logger into this method, can we inline it?

misk-jdbc/src/main/kotlin/misk/jdbc/TransactionRetryHandler.kt
```diff
@@ -0,0 +1,187 @@
...
+    maxAttempts: Int = 3,
+    minRetryDelayMillis: Long = 100,
+    maxRetryDelayMillis: Long = 500,
+    retryJitterMillis: Long = 400,
+    block: () -> T
+  ): T {
+    require(maxAttempts > 0) { "maxAttempts must be positive" }
+    
+    val backoff = ExponentialBackoff(
```
tgregory-block: Since `misk-backoff` is already pulled in, could this retry logic be achieved with its `retry` function and `RetryConfig`?

PR #3453, opened 2025-10-08
Add http_response_header_size to WebConfig

Hexcles:
## Description

Jetty defaults to an 8KB limit on both the request and response header sizes. Misk provides a config to change the request header size limit but not the response. This PR adds `http_response_header_size` to fill in that gap.

## Testing Strategy

The existing `http_request_header_size` doesn't seem to have any test so I'm doing the same.

This is pretty low risk. The default is a no-op.

## Checklist

- [ ] I have added tests to have confidence my changes work as expected.


adrw (approved):
LG

PR #3459, opened 2025-10-10
New AsyncKAbstractModule to allow dynamic toggling of async

adrw:
In some cases, we want an easy way to disable async functionality within a deployment of an application. By extending from the new `AsyncKAbstractModule` we can filter centrally at Guice injector creation based on env var or feature flag to disable async and swap in a non-async module such that the application can still be constructed and boots up as expected.


misk-aws/src/main/kotlin/misk/jobqueue/sqs/AwsSqsJobHandlerModule.kt
```diff
@@ -37,6 +35,22 @@ class AwsSqsJobHandlerModule<T : JobHandler> private constructor(
     )
   }
 
+  private class CommonModule<T : JobHandler>(
```
katukota, line 40: Should we also override moduleWhenAsyncDisabled() method here?

PR #3463, opened 2025-10-15
Redis to support parallel tests by using different DBs per test process

mmollaverdi:
## Description

There is the existing support for running misk tests in parallel processes for faster build times, with isolation across processes for MySQL and DynamoDB (the former using separate DB schemas and the latter using namespaced table names). This PR adds support for another external data source used commonly in Misk services, namely Redis. This implementation uses different Redis databases, which Redis (in the non-cluster mode) supports ranging from 0 to 15.    

## Checklist

### DIFF
--- a/misk-redis/src/testFixtures/kotlin/misk/redis/testing/RealRedisTestModule.kt
+++ b/misk-redis/src/testFixtures/kotlin/misk/redis/testing/RealRedisTestModule.kt
@@ -1,0 +1,52 @@
+package misk.redis.testing
+
+import com.google.inject.Provides
+import com.google.inject.util.Modules
+import jakarta.inject.Singleton
+import misk.inject.KAbstractModule
+import misk.redis.JedisPooledWithMetrics
+import misk.redis.RedisClientMetrics
+import misk.redis.RedisModule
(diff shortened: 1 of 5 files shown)


misk-redis/src/testFixtures/kotlin/misk/redis/testing/RealRedisTestModule.kt
```diff
@@ -0,0 +1,52 @@
...
+import misk.redis.RedisReplicationGroupConfig
+import misk.testing.parallelTestIndex
+import redis.clients.jedis.ConnectionPoolConfig
+import redis.clients.jedis.UnifiedJedis
+import wisp.deployment.Deployment
+
+/**
+ * Installs a real redis for testing with support for parallel tests.
+ */
+class RealRedisTestModule(
```
mmollaverdi, line 19: Adding this separate test module in the testFixtures source dir to avoid the need for changes to the public APIs for RedisModule which is used in production code.
This also allowed me to have more sensible defaults for testing, such as `useSsl = false` and `requiresPassword = false`.

jvmakine (approved):
Nice!

PR #3466, opened 2025-10-16
Support larger McpTool hirarchies

ean5533:
## Description

This PR allows McpTools to implement (Structured)McpTool transitively instead of directly. Previously this would throw an exception while generating the input/output schemas because the generic types couldn't be resolved.

## Checklist


misk-mcp/src/main/kotlin/misk/mcp/util/resolveTypeArgument.kt
```diff
@@ -0,0 +1,94 @@
...
+import kotlin.reflect.KClass
+import kotlin.reflect.KType
+import kotlin.reflect.KTypeParameter
+import kotlin.reflect.KTypeProjection
+import kotlin.reflect.full.createType
+
+/**
+ * Resolves the type argument at [argumentIndex] of [targetSuper] as used by [concreteClass].
+ */
+internal fun resolveTypeArgument(
```
ean5533: Could I have written this method myself? Yes, I (eventually) could have.

Did I get AI to generate it for me instead? Yes, I did.

misk-mcp/src/main/kotlin/misk/mcp/McpTool.kt
```diff
@@ -543,13 +540,6 @@ abstract class StructuredMcpTool<I : Any, O : Any> : McpTool<I>() {
...
   private val outputClass: KClass<O> by lazy {
-    @Suppress("UNCHECKED_CAST")
-    this::class.supertypes
-      .first { type ->
-        (type.classifier as? KClass<*>)?.simpleName?.let { simpleName ->
-          simpleName == StructuredMcpTool::class.simpleName
-        } ?: false
-      }
-      .arguments.last().type!!.classifier as KClass<O>
+    resolveTypeArgument(this::class, StructuredMcpTool::class, 1).classifier as KClass<O>
```
adrw: very nice, even better if we take the 2nd param (McpTool/StructuredMcpTool) as an inline reified parameter to cut out some of the `::class`

adrw (approved):
LG

misk-mcp/src/main/kotlin/misk/mcp/McpTool.kt
ean5533: Done, and also converted the first param to a receiver, so now it looks like:

```
this::class.resolveTypeArgument<McpTool<*>>(0).classifier as KClass<I>
```

misk-mcp/src/main/kotlin/misk/mcp/McpTool.kt
```diff
@@ -543,13 +536,6 @@ abstract class StructuredMcpTool<I : Any, O : Any> : McpTool<I>() {
   }
 
   private val outputClass: KClass<O> by lazy {
-    @Suppress("UNCHECKED_CAST")
-    this::class.supertypes
```
jclyne, line 547: The bug is actually that this should be `this::class.allSupertypes` This literally does what resolveTypeArgument does

misk-mcp/src/main/kotlin/misk/mcp/McpTool.kt
ean5533: Edit: I've deleted this whole thing in favor of `allSupertypes`, see below

misk-mcp/src/main/kotlin/misk/mcp/McpTool.kt
ean5533: TIL [allSupertypes](https://kotlinlang.org/api/core/kotlin-reflect/kotlin.reflect.full/all-supertypes.html)! It's an extension in the kotlin-reflect-full module instead of the stdlib, but we're apparently already bringing that in anyway.

Updated to use that instead.

ean5533: Quick note: it looks like this only works correctly on Kotlin 1.9+. On older versions, `allSupertypes` didn't re-bind generic types through complex hierarchies.

jclyne (approved):
🚢

PR #3469, opened 2025-10-16
Allow overriding of configureClient in AwsSqsJobQueueModule again

mmollaverdi:
## Description

[This previous PR](https://github.com/cashapp/misk/pull/3459) added a `configureClient` methods to the new CommonModule class and the one in `AwsSqsJobQueueModule` which was `open` for overrides is no longer used. But that method is overriden in some cash-app services to customize the AWS auth logic for integration tests and those tests are failing on a Misk version upgrade. 

https://github.com/cashapp/misk/pull/3459

This PR makes sure that method is used again.


misk-aws/src/main/kotlin/misk/jobqueue/sqs/AwsSqsJobQueueModule.kt
```diff
@@ -182,8 +188,6 @@ open class AwsSqsJobQueueModule(
     }
   }
 
-  open fun <BuilderT : AwsClientBuilder<BuilderT, ClientT>, ClientT> configureClient(builder: BuilderT) {}
```
mmollaverdi, line 185: Moved it further up so that it's clear that it's part of the outer module class.

adrw (approved):
LG

PR #3475, opened 2025-10-20
Flush only the current DB when resetting redis between tests

mmollaverdi:
## Description

[This previous PR](https://github.com/cashapp/misk/pull/3463) which added support for using Redis in parallel tests, caused some flakiness in tests I believe due to the current implementation for resetting Redis in between tests, using `flushAll` which flushes all DBs. Switching to `flushDB` instead in this PR which should address the flakiness by flushing data only from the database the current unified jedis client is conneced to.

### DIFF
--- a/misk-redis/src/testFixtures/kotlin/misk/redis/testing/RealRedisTestModule.kt
+++ b/misk-redis/src/testFixtures/kotlin/misk/redis/testing/RealRedisTestModule.kt
@@ -27,6 +27,7 @@ class RealRedisTestModule(
         RedisModule(redisReplicationGroupConfig, connectionPoolConfig, useSsl)
       ).with(TestUnifiedJedisModule(connectionPoolConfig, useSsl))
     )
+    install(RedisTestFlushModule())
   }
 }
 
--- a/misk-redis/src/testFixtures/kotlin/misk/redis/testing/RedisFlushService.kt
+++ b/misk-redis/src/testFixtures/kotlin/misk/redis/testing/RedisFlushService.kt
@@ -4,9 +4,9 @@ import com.google.common.util.concurrent.AbstractIdleService
 import jakarta.inject.Inject
 import jakarta.inject.Provider
 import jakarta.inject.Singleton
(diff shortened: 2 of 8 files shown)


misk-redis/src/testFixtures/kotlin/misk/redis/testing/RealRedisTestModule.kt
```diff
@@ -27,6 +27,7 @@ class RealRedisTestModule(
         RedisModule(redisReplicationGroupConfig, connectionPoolConfig, useSsl)
       ).with(TestUnifiedJedisModule(connectionPoolConfig, useSsl))
     )
+    install(RedisTestFlushModule())
```
mmollaverdi, line 30: This is being manually installed, whereas it's almost always needed when real Redis is used in tests, so installingit here makes more sense.

misk-redis/src/testFixtures/kotlin/misk/redis/testing/RedisFlushService.kt
```diff
@@ -20,19 +20,18 @@ class RedisFlushService @Inject constructor() : AbstractIdleService(), TestFixtu
...
   override fun reset() {
-    flushAll()
+    flush()
   }
 
-  private fun flushAll() {
+  private fun flush() {
     logger.info("Flushing Redis")
-    redis.flushAll()
-    logger.info("Flushed Redis")
```
mmollaverdi, line 35: This second log line seemed excessive, so I removed it.

misk-redis/src/testFixtures/kotlin/misk/redis/testing/RedisFlushService.kt
```diff
@@ -20,19 +20,18 @@ class RedisFlushService @Inject constructor() : AbstractIdleService(), TestFixtu
...
+    redis.flushDB()
```
mmollaverdi, line 34: The main change.

jvmakine (approved):
Nice! Love the flushing in this PR!

PR #3485, opened 2025-10-22
Set data source properties for AWS secret fetching

katukota:
## Description

The "user" property needs to be set with the secret value instead of as a query param: https://github.com/aws/aws-secretsmanager-jdbc?tab=readme-ov-file#usage-example

## Checklist

### DIFF
--- a/misk-jdbc/src/main/kotlin/misk/jdbc/DataSourceConfig.kt
+++ b/misk-jdbc/src/main/kotlin/misk/jdbc/DataSourceConfig.kt
@@ -6,7 +6,6 @@ import misk.containers.ContainerUtil
 import wisp.deployment.Deployment
 import java.time.Duration
 import java.util.Properties
-import kotlin.collections.set
 
 /** Defines a type of datasource */
 enum class DataSourceType(


misk-jdbc/src/main/kotlin/misk/jdbc/PingDatabaseService.kt
```diff
@@ -65,7 +64,9 @@ class PingDatabaseService @Inject constructor(
 
   private fun createDataSource(jdbcUrl: String): DriverDataSource {
     return DriverDataSource(
-            jdbcUrl, config.getDriverClassName(), Properties(), config.username, config.password)
+      jdbcUrl, config.getDriverClassName(), config.getDataSourceProperties(),
```
adrw, line 67: named params?

misk-jdbc/src/main/kotlin/misk/jdbc/DataSourceConfig.kt
```diff
@@ -5,6 +5,8 @@ import misk.config.Config
 import misk.containers.ContainerUtil
 import wisp.deployment.Deployment
 import java.time.Duration
+import java.util.Properties
+import kotlin.collections.set
```
adrw: unused?
```suggestion
```

misk-jdbc/src/main/kotlin/misk/jdbc/PingDatabaseService.kt
katukota: Named params are prohibited for non-Kotlin functions :/ DriverDataSource is a java function.

PR #3490, opened 2025-10-23
return the result without unnecessary overhead of supplyAsync

yyuan-squareup:
## Description

Noticing this while working on this file.
.thenApply { true } does the same as before, but without the overhead of creating a new CompletableFuture<Boolean> by .supplyAsync().

## Testing Strategy

Regression test

The code does not change the behavior, just a bit more efficient

## Checklist


mateuszmrozewski (approved):
LGTM

PR #3495, opened 2025-10-23
Add tool descriptor metadata support

andrewjenkins:
## Description

Adds support for tool descriptor metadata to misk-mcp. This is required by the OpenAI Apps SDK: https://developers.openai.com/apps-sdk/reference

This is a temporary solution until https://github.com/modelcontextprotocol/kotlin-sdk/pull/339 can be pulled in.

## Testing Strategy

Unit testing
Manual testing locally with the cash-server monorepo, verified it worked with our gptstore service as expected

## Checklist

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -27,7 +27,7 @@ prometheus = "0.16.0"
 protobuf = "3.25.8"
 retrofit = "3.0.0"
 sqldelight = "2.1.0"
-tempest = "2025.10.21.114346-63954e1"
+tempest = "2025.10.23.165355-269d2e7"
 wire = "5.4.0"
 
 [libraries]


misk-mcp/src/main/kotlin/misk/mcp/McpTool.kt
```diff
@@ -202,6 +202,12 @@ abstract class McpTool<I : Any> {
    */
   open val openWorldHint: Boolean = true
 
+  /**
+   * Metadata included in the tool descriptor. Not officially part of the MCP spec, but relied
+   * upon by some clients (e.g., OpenAI Apps SDK).
+   */
+  open val metadata: JsonObject? = null
```
jclyne, line 211: This is going to be a breaking change. Im intending on creating a tool level field like this 
`open val meta : Any? = null`
That way it can be a Serializable data class when defined. This would allow defining something like a class OpenAIMeta() that could set to this field. 

If you could take that approach, we wouldn't need to make a breaking change here later, it would be the same interface when its supported in the framework

jclyne (commented):
Looks good

misk-mcp/src/main/kotlin/misk/mcp/McpTool.kt
jclyne: You would need to serialize the data class into a JsonObject when modifying the ListToolsResult

jclyne (approved):

misk-mcp/src/main/kotlin/misk/mcp/McpTool.kt
andrewjenkins: Discussed offline - this solution won't quite work as described due to kotlinx serialization not knowing how to serialize the `Any` type.

We decided to pursue the current approach for now, and later we can make a breaking change if there's a better approach discovered.

PR #3499, opened 2025-10-27
[proposal] Always reset all test fixtures

mgulbronson:
## Description

Currently if the injector is not reused no test fixtures are reset, presumably because it's assumed the test fixtures are scoped to the injector and will therefore be reconstructed for each test when the injector is created. But test fixtures that are Kotlin objects aren't scoped to the injector, they're scoped to the JVM classloader and therefore persist across tests/injectors. So if the injector is not reused, those test fixtures never get reset.

Practically this means that binding a test fixture doesn't do anything. This PR makes it so all test fixtures are always reset, making it possible to implement test fixtures as objects.

***

Alternative approach that I did originally was to reflect on the test fixtures to check if they're an object or not. For test fixtures that aren't objects no change from the current behaviour is needed. For test fixtures that _are_ objects they need to be reset every time (regardless of whether the injector is being reused). But realistically the cost of reflecting over the test fixtures is more expensive than just resetting all test fixtures.

### DIFF
--- a/docs/actions.md
+++ b/docs/actions.md
@@ -352,6 +352,80 @@ class MyTest {
 }
 ```
 
+### Verifying action registration
+
+A common mistake is implementing a `WebAction` but forgetting to register it via 
+`WebActionModule.create<YourAction>()`. Without registration, the action won't be exposed as an 
+HTTP endpoint.
+
+Misk provides `WebActionRegistrationTester` to catch this. It scans your service's packages for 
+`WebAction` implementations and verifies each has a corresponding `WebActionEntry` registration.
+
+**Using the base class:**
+
+```kotlin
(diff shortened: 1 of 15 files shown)


adrw (approved):
Makes sense, going to wait for one more review and then merge.

mmollaverdi:
Makes sense

mmollaverdi (approved):

mgulbronson:
> @mgulbronson Not sure why but some test seems to keep timing out in the testShardNonHibernate, I've triggered a rebuild a few times so it might not be a test flake but a regression from your code change. Do you mind taking a look?

@adrw I think I see the issue — will try to get a fix up today but may wait until after the holidays

PR #3510, opened 2025-10-30
feat: add support for binding multiple DynamoDB modules via qualifiers

ThePumpingLemma:
## Description
This allows us to have separate client configurations for different
sets of tables (e.g. application vs framework cluster leasing)

### DIFF
--- a/misk-aws2-dynamodb/src/main/kotlin/misk/aws2/dynamodb/RealDynamoDbModule.kt
+++ b/misk-aws2-dynamodb/src/main/kotlin/misk/aws2/dynamodb/RealDynamoDbModule.kt
@@ -1,12 +1,15 @@
 package misk.aws2.dynamodb
 
-import com.google.inject.Provides
-import jakarta.inject.Singleton
+import com.google.inject.Injector
+import com.google.inject.Provider
+import jakarta.inject.Inject
 import misk.ReadyService
 import misk.ServiceModule
 import misk.cloud.aws.AwsRegion
 import misk.exceptions.dynamodb.DynamoDbExceptionMapperModule
 import misk.inject.KAbstractModule
+import misk.inject.asSingleton
+import misk.inject.keyOf
 import software.amazon.awssdk.auth.credentials.AwsCredentialsProvider
 import software.amazon.awssdk.core.client.config.ClientOverrideConfiguration
 import software.amazon.awssdk.regions.Region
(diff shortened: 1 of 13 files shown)


misk-aws2-dynamodb/src/main/kotlin/misk/aws2/dynamodb/RealDynamoDbModule.kt
```diff
@@ -15,27 +17,69 @@ import software.amazon.awssdk.services.dynamodb.DynamoDbClientBuilder
...
+open class RealDynamoDbModule
+@JvmOverloads constructor(
+  private val qualifier: KClass<out Annotation>? = null,
+  private val clientOverrideConfig: ClientOverrideConfiguration = ClientOverrideConfiguration.builder().build(),
   private val requiredTables: List<RequiredDynamoDbTable> = listOf(),
   private val endpointOverride: URI? = null,
 ) : KAbstractModule() {
+  // Backward-compatible constructor (unqualified)
+  @JvmOverloads
+  constructor(
```
valerio86, line 53: is this necessary given the previous constructor with defaults?

ThePumpingLemma: We need a version without the `qualifier` parameter for backwards compatibility.

misk-aws2-dynamodb/src/main/kotlin/misk/aws2/dynamodb/RealDynamoDbService.kt
```diff
@@ -1,16 +1,42 @@
...
-class RealDynamoDbService @Inject constructor(
-  private val dynamoDb: DynamoDbClient,
-  private val requiredTables: List<RequiredDynamoDbTable>,
-) : AbstractIdleService(), DynamoDbService {
+class RealDynamoDbService : AbstractIdleService, DynamoDbService {
+  private val dynamoDb: DynamoDbClient
+  private val requiredTables: List<RequiredDynamoDbTable>
+
+  constructor(
+    injector: Injector,
```
adrw, line 18: Not great to use the Injector directly but I don't think there's another way so we can leave it.

ThePumpingLemma: Yeah, I had the same gripe but I don't think we can avoid it.

PR #3516, opened 2025-11-05
feat: Add dynamic debug logging module

lcabancla:
## Description

Module that enables runtime log level control via feature flags for debugging issues.

This allows you to temporarily increase log verbosity for specific loggers without restarting the service, useful for
troubleshooting issues in staging/production environments.

### 1. Install the module

```kotlin
install(DynamicLogLevelModule(DynamicLoggingConfig(enabled=true, featureFlag="dynamic-log-level")))
```

### 2. Set the feature flag
Set the feature flag value to a comma-separated string of `logger=LEVEL` pairs:
```
com.amazonaws.services.dynamodbv2.AmazonDynamoDBLockClient=TRACE,com.squareup.cash.dynamodb.lease=DEBUG
```

Supported levels: `TRACE`, `DEBUG` (other levels are ignored for safety)

To disable: Set the feature flag to an empty string `""` to remove all dynamic log level overrides

Logger names can be:
- Fully qualified class name: `com.example.MyClass`
- Package prefix: `com.example` (applies to all classes in that package)

The service polls the feature flag every 20 seconds and applies changes automatically.

## Important: Logging Framework Bridges

This module controls Logback log levels. If the logger you want to control uses a different logging framework, you
must add the appropriate SLF4J bridge to route logs through Logback:
```kotlin
// In your build.gradle.kts
implementation(libs.slf4jJclBridge) // Bridge Commons Logging (e.g., AmazonDynamoDBLockClient)
implementation(libs.slf4jJulBridge) // Bridge Java Util Logging
```

Without the bridges, log level changes won't affect loggers using those frameworks.

## Testing Strategy

The module was tested in a service.

## Checklist


misk/src/main/kotlin/misk/logging/DynamicLogLevelModule.kt
```diff
@@ -0,0 +1,69 @@
...
+ *
+ * This allows you to temporarily increase log verbosity for specific loggers without restarting the service, useful for
+ * troubleshooting issues in staging/production environments.
+ *
+ * ## Configuration
+ *
+ * ### 1. Install the module
+ *
+ * ```kotlin
+ * install(DynamicLogLevelModule(DynamicLoggingConfig(enabled=true, featureFlag="dynamic-log-level")))
```
adrw:
```suggestion
 * install(DynamicLogLevelModule(DynamicLoggingConfig(enabled=true, feature_flag_name="dynamic-log-level")))
```

misk/src/main/kotlin/misk/logging/DynamicLogLevelService.kt
```diff
@@ -0,0 +1,56 @@
...
+import java.time.Duration
+
+@Singleton
+class DynamicLogLevelService
+@Inject
+constructor(
+  private val featureFlags: FeatureFlags,
+  private val config: DynamicLoggingConfig,
+  @DynamicLogLevel private val repeatedTaskQueue: RepeatedTaskQueue,
+
```
adrw, line 20:
```suggestion
```

misk/src/main/kotlin/misk/logging/DynamicLogLevelController.kt
```diff
@@ -0,0 +1,58 @@
...
+  }
+
+  fun clearAll() = removeFilter()
+
+  private fun removeFilter() {
+    synchronized(this) {
+      currentFilter?.let {
+        context.turboFilterList.remove(it)
+        logger.info("Removed DynamicDebugFilter (dynamic log levels cleared).")
+      } ?: logger.debug("No DynamicDebugFilter found to remove.")
```
ean5533: Feels like this should be the same log message as the one on line 21

misk/src/main/kotlin/misk/logging/DynamicLogLevelController.kt
```diff
@@ -0,0 +1,58 @@
...
+
+  fun refresh(commaDelimitedPairs: String) {
+    val newLevels = DynamicLogParser.parseLoggerToLevelPairs(commaDelimitedPairs)
+
+    // Empty → remove filter
+    if (newLevels.isEmpty()) {
+      if (currentFilter != null) {
+        removeFilter()
+      } else {
+        logger.debug("Dynamic log levels remain disabled (no filter to remove).")
```
ean5533: Technically this case would be covered by lines 27-30. Is it worth having this more specific "log levels _remain disabled_" phrasing, or should we just let it fall through to that more general check instead (or maybe move that check up above this block)?

misk/src/main/kotlin/misk/logging/DynamicLogLevelController.kt
```diff
@@ -0,0 +1,58 @@
...
+      }
+      return
+    }
+
+    // Skip if identical contents
+    if (newLevels == currentLevels) {
+      logger.debug("Dynamic log levels unchanged; skipping replacement.")
```
ean5533: Is it worth logging the log levels here like we do when they change, just for clarity?

misk/src/main/kotlin/misk/logging/DynamicLogLevelService.kt
```diff
@@ -0,0 +1,56 @@
...
+  fun updateLoggingLevels() {
+    val commaDelimitedPairs = featureFlags.getString(feature, "test",
+      Attributes(mapOf("pod_name" to podName))
+    )
+    controller.refresh(commaDelimitedPairs)
+  }
+
+  companion object {
+    private val logger = getLogger<DynamicLogLevelService>()
+    const val UNSET = "UNSET"
```
ean5533: Unused?

misk/src/main/kotlin/misk/logging/DynamicLogLevelService.kt
```diff
@@ -0,0 +1,56 @@
...
+    const val POD_NAME = "MY_POD_NAME"
```
ean5533: I recommend `POD_NAME_ENV_VAR` -- I was confused why the pod name would be hard-coded.

misk/src/main/kotlin/misk/logging/LogLevelFilter.kt
```diff
@@ -0,0 +1,33 @@
+package misk.logging
+
+import ch.qos.logback.classic.Level
+import ch.qos.logback.classic.turbo.TurboFilter
+import ch.qos.logback.core.spi.FilterReply
+import org.slf4j.Marker
+
+class LogbackLevelFilter(val dynamicLevels: Map<String, Level>) : TurboFilter() {
```
ean5533: File name `LogLevelFilter` doesn't quite match class name

misk/src/main/kotlin/misk/logging/LogLevelFilter.kt
```diff
@@ -0,0 +1,33 @@
...
+
+    // Try exact match first
+    dynamicLevels[logger.name]?.let { target ->
+      return if (level.isGreaterOrEqual(target)) FilterReply.ACCEPT else FilterReply.NEUTRAL
+    }
+
+    // Try prefix match (find most specific matching prefix)
+    val target =
+      dynamicLevels.entries.filter { logger.name.startsWith(it.key) }.maxByOrNull { it.key.length }?.value
+        ?: return FilterReply.NEUTRAL
```
ean5533, line 32: This prefix matching logic will probably catch things that the author didn't intend them to catch. For example, a filter like `com.example.Foo=DEBUG` would end up applying to both `com.example.Foo` and `com.example.FooFactory`.

To disambiguate, it might be best to use an explicit wildcard syntax like `com.example.*=DEBUG` for prefix matching.

misk/src/test/kotlin/misk/logging/DynamicLogLevelIntegrationTest.kt
```diff
@@ -0,0 +1,293 @@
...
+    assertThat(filter1).isSameAs(filter2)
+
+    // Verify dynamicLevels contents remain the same
+    val dynamicLevels2 = getDynamicLevelsFromFilter(context)
+    assertThat(dynamicLevels2).hasSize(1)
+    assertThat(dynamicLevels2["com.test"]).isEqualTo(ch.qos.logback.classic.Level.DEBUG)
+  }
+
+  @Test
+  fun `service processes feature flag with whitespace`() {
```
ean5533: This test seems redundant to `DynamicLogParser.handles whitespace around entries`

ean5533 (approved):

misk/src/main/kotlin/misk/logging/DynamicLogLevelController.kt
lcabancla: Will remove the "log levels remain disabled" log.

misk/src/main/kotlin/misk/logging/LogLevelFilter.kt
lcabancla: Will change.

misk/src/test/kotlin/misk/logging/DynamicLogLevelIntegrationTest.kt
lcabancla: Good catch. Will remove the test.

misk/src/main/kotlin/misk/logging/DynamicLogLevelController.kt
lcabancla: Will add.

misk/src/main/kotlin/misk/logging/DynamicLogLevelService.kt
lcabancla: Will remove this constant. It is not necessary.

misk/src/main/kotlin/misk/logging/DynamicLogLevelService.kt
lcabancla: Will remove.

misk/src/main/kotlin/misk/logging/DynamicLogLevelController.kt
lcabancla: Cleaned up the code to remove redundant logs.

misk/src/main/kotlin/misk/logging/LogLevelFilter.kt
lcabancla: Good catch. Will implement dot-boundary matching instead to prevent this.

PR #3518, opened 2025-11-06
fix: make FakeCronModule also KInstallOnceModule

stuartwdouglas:
This mirrors the behaviour of the module it is faking for, preventing issues when async is disabled.


adrw (approved):
LG

PR #3524, opened 2025-11-07
Fixes transitive ktor dependencies from kotlin-sdk

jclyne:
## Description

The latest kotlin-sdk switched to using the ktor-bom for version constraints on the ktor dependencies. These are not propagated when used as in an included build causing missing versions. Since misk-mcp doesn't use these, it doesn't declare them as a dependency.  This excludes the server dependencies and makes the client dependencies explicit and API

## Checklist


adrw (approved):
LG

PR #3532, opened 2025-11-10
Add annotation for logging interceptor entry/exit

peckb1:
## Description

Services are reporting issues where the call _seems_ to hit the service but the web action never reports entry. The thought is that there is an interceptor causing an error / failing the request, but there are no logs indicating which of the interceptors it would be.

This adds an annotation that can be placed on WebActions to allow the RealNetworkChain to log when it's about to kick off an interceptor, as well as a log when an interceptor comes back from handling the chain.

---

*WARNING* this causes a decent chunk of logs for every call.

For Example the `An AnnotatedAction with a failing intercept still logs` test has 11 logs:
```
Interceptor about to process: misk.web.interceptors.GunzipRequestBodyInterceptor@6afa6740
Interceptor about to process: misk.web.interceptors.InternalErrorInterceptorFactory$Companion$INTERCEPTOR$1@2bb77efd
Interceptor about to process: misk.web.interceptors.RequestLogContextInterceptor@365d08c5
Interceptor about to process: misk.web.interceptors.MetricsInterceptor@15c993c1
Interceptor about to process: misk.web.exceptions.ExceptionHandlingInterceptor@1bc2e83b

Interceptor about to process: misk.web.interceptors.FailableNetworkInterceptor@4afc1a23

Interceptor finished processing: misk.web.exceptions.ExceptionHandlingInterceptor@1bc2e83b
Interceptor finished processing: misk.web.interceptors.MetricsInterceptor@15c993c1
Interceptor finished processing: misk.web.interceptors.RequestLogContextInterceptor@365d08c5
Interceptor finished processing: misk.web.interceptors.InternalErrorInterceptorFactory$Companion$INTERCEPTOR$1@2bb77efd
Interceptor finished processing: misk.web.interceptors.GunzipRequestBodyInterceptor@6afa6740

```

The Interceptor that returns the 400 doesn't have a `finished processing` log

## Concerns:
- ~~[ ] log level is INFO, allow for configurable level?~~
  - [x] keep default to not bring in new LogLevel dependency
- [x] enter and exit logs both always happen, allow for one or the other to be enabled/disabled.

## Testing Strategy

Unit Test for the annotation

### DIFF
--- a/misk/src/main/kotlin/misk/web/RealNetworkChain.kt
+++ b/misk/src/main/kotlin/misk/web/RealNetworkChain.kt
@@ -28,14 +28,14 @@ internal class RealNetworkChain(
 
   private fun logCallingInterceptor(interceptor: NetworkInterceptor) {
     val loggingAnnotation = webAction::class.findAnnotation<LogNetworkInterceptorChain>()
-    if (loggingAnnotation != null) {
+    if (loggingAnnotation != null && loggingAnnotation.logBefore) {
       logger.info("Interceptor about to process: {}", interceptor)
     }
   }
 
   private fun logInterceptorReturned(interceptor: NetworkInterceptor) {
     val loggingAnnotation = webAction::class.findAnnotation<LogNetworkInterceptorChain>()
-    if (loggingAnnotation != null) {
+    if (loggingAnnotation != null && loggingAnnotation.logAfter) {
       logger.info("Interceptor finished processing: {}", interceptor)
     }
   }
--- a/misk-actions/api/misk-actions.api
+++ b/misk-actions/api/misk-actions.api
@@ -307,6 +307,8 @@ public final class misk/web/dashboard/ValidWebEntry$Companion {
 }
 
 public abstract interface annotation class misk/web/interceptors/LogNetworkInterceptorChain : java/lang/annotation/Annotation {
+	public abstract fun logAfter ()Z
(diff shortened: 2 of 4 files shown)


misk/src/main/kotlin/misk/web/RealNetworkChain.kt
```diff
@@ -12,7 +15,32 @@ internal class RealNetworkChain(
...
+    val result = interceptor.intercept(next)
+    logInterceptorReturned(interceptor)
+
+    return result
+  }
+
+  private fun logCallingInterceptor(interceptor: NetworkInterceptor) {
+    val loggingAnnotation = webAction::class.findAnnotation<LogNetworkInterceptorChain>()
+    if (loggingAnnotation != null && loggingAnnotation.logBefore) {
+      logger.info("Interceptor about to process: {}", interceptor)
```
adrw, line 32: Does the empty {} signify something or the logger does the find/replace for you? I'd do `${interceptor}` Kotlin string template instead.

peckb1: what it's _supposed_ to do it lazy evaluation avoid calling the `.toString()` on the parameter object unless the log is accepted for output. Which for `info` in my service is probably gonna be always, but for like debug/trace it could save things or services which turn off info logs for a bit.

PR #3536, opened 2025-11-11
Make `SqsClientFactory` an interface so it can be faked for tests

mgulbronson:
## Description

Makes `SqsClientFactory` an interface so that it can be overridden in test environments.

We ran into this issue when trying to upgrade our Misk version. In real environments it works fine (since we want a real SQS client), in most test environments it works fine (since we use `FakeJobEnqueuer` so don't bother with `SqsJobEnqueuer` or `SqsJobConsumer`). But in local development we're trying to mimic the job publish/consume flow as closely as possible and just want to fake the call to SQS. Previously we could do this by binding a fake implementation of `SqsAsyncClient` but that no longer works with `SqsClientFactory`. 

With this change we can bind a fake implementation of `SqsClientFactory` that returns our in-memory fake for testing.

- No functional change so no risk — the exact same implementation of `SqsClientFactory` is provided, we're just slightly changing the API to expose it as an interface

### DIFF
--- a/misk-aws2-sqs/api/misk-aws2-sqs.api
+++ b/misk-aws2-sqs/api/misk-aws2-sqs.api
@@ -13,11 +13,8 @@ public final class misk/aws2/sqs/jobqueue/DefaultDeadLetterQueueProvider : misk/
 	public fun deadLetterQueueFor (Lmisk/jobqueue/QueueName;)Lmisk/jobqueue/QueueName;
 }
 
-public final class misk/aws2/sqs/jobqueue/SqsClientFactory {
-	public fun <init> (Lsoftware/amazon/awssdk/auth/credentials/AwsCredentialsProvider;)V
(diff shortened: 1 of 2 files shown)


misk-aws2-sqs/src/main/kotlin/misk/aws2/sqs/jobqueue/SqsJobQueueModule.kt
```diff
@@ -59,7 +62,18 @@ open class SqsJobQueueModule @JvmOverloads constructor(
     fun sqsClientClientFactory(
       credentialsProvider: AwsCredentialsProvider,
     ): SqsClientFactory {
-      return SqsClientFactory(credentialsProvider, configureClient)
+      return SqsClientFactory { region ->
+        val clients = ConcurrentHashMap<String, SqsAsyncClient>()
```
adrw: Could we pull this out into a `RealSqsClientFactory` class vs. inline?

adrw (approved):
LG

mateuszmrozewski (approved):

PR #3538, opened 2025-11-11
Migrate AsyncModule to AsyncSwitch

adrw:
Migrate AsyncModule to AsyncSwitch. 

The previous AsyncModule approach in practice had issues with preserving Provides methods on the module and the API for fallback module leading to more complicated refactoring than necessary. In most cases, swapping the real for a no-op Guava service was the only change needed to disable async functionality on a pod so a more narrower abstraction for an `AsyncSwitch` with tight integration with `ServiceModule` is being proposed instead.

Spitting this into two to land more safely, first adding the new APIs, then using them.
- [x] https://github.com/cashapp/misk/pull/3562
- [x] https://github.com/cashapp/misk/pull/3563
- [x] using them PR: this PR
- [x] don't merge until internal version bumped to latest


adrw:
Closing since it landed in separate PRs

PR #3542, opened 2025-11-14
Extend misk slack api to react to message

qnguyen-block:
I want to extend the misk slack api to support reacting to thread.


adrw (approved):
LG

PR #3549, opened 2025-11-19
Add tokens for slack socket mode

srhjang:
## Description

I want to extend the Slack web api config so that I can use Slack socket mode to make an update in the cash release dashboard service when a reaction is added to a slack thread. This sets up the config so that I can add the app token and enable socket mode.


adrw (approved):
LG

PR #3550, opened 2025-11-20
fix: Span isn't a DDSpan and original_trace_id isn't being set in the _jobqueue-metadata

nssherpa:
## Description

While testing some sqs changes we found that `JOBQUEUE_METADATA_ORIGINAL_TRACE_ID` isn't being set when testing some changes rlated to sqs integration work with playpen. Looks like the Span is never a DDSpan but an OTSpan.

## Testing Strategy

Tested using playpen before and after the change in cash-playpen-testbed app.

### Before
`original_trace_id` not present in the `_jobqueue-metadata` MessageAttribute.
Screenshot 2025-11-19 at 4 42 12 PM

### After
`original_trace_id` now present in the `_jobqueue-metadata` MessageAttribute.
Screenshot 2025-11-19 at 4 04 16 PM

## Related Work

https://sq-block.slack.com/archives/C096YVAPX7G/p1763502983981289?thread_ts=1763434406.968859&cid=C096YVAPX7G

## Checklist

- [ ] I have manually testing this via playpen in one of the app (`cash-playpen-testbed`) by including local changes from misk.

### DIFF
--- a/misk-aws/src/main/kotlin/misk/jobqueue/sqs/SqsJobQueue.kt
+++ b/misk-aws/src/main/kotlin/misk/jobqueue/sqs/SqsJobQueue.kt
@@ -141,9 +141,8 @@ internal class SqsJobQueue @Inject internal constructor(
     )
 
     // Preserve original trace id, if available.
-    (span as? DDSpan)?.let {
-      val traceId = it.context().traceId.toString()
-      metadata[SqsJob.JOBQUEUE_METADATA_ORIGINAL_TRACE_ID] = traceId
+    span.context().toTraceId()?.takeIf { it.isNotBlank() }?.let {
+      metadata[SqsJob.JOBQUEUE_METADATA_ORIGINAL_TRACE_ID] = it
     }
 
     return MessageAttributeValue()
--- a/misk-aws/build.gradle.kts
+++ b/misk-aws/build.gradle.kts
@@ -27,7 +27,6 @@ dependencies {
   implementation(libs.openTracingDatadog)
   implementation(libs.prometheusClient)
   implementation(libs.slf4jApi)
-  implementation(libs.tracingDatadog)
   implementation(project(":misk-api"))
   implementation(project(":misk-moshi"))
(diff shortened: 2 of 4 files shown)


misk-aws/src/main/kotlin/misk/jobqueue/sqs/SqsJobQueue.kt
```diff
@@ -141,10 +141,7 @@ internal class SqsJobQueue @Inject internal constructor(
     )
 
     // Preserve original trace id, if available.
-    (span as? DDSpan)?.let {
-      val traceId = it.context().traceId.toString()
-      metadata[SqsJob.JOBQUEUE_METADATA_ORIGINAL_TRACE_ID] = traceId
-    }
+    metadata[SqsJob.JOBQUEUE_METADATA_ORIGINAL_TRACE_ID] = span.context().toTraceId()
```
nssherpa: Doesn't seem like span or context would be null.. but can add a check if necessary.

mpeyper: We _may_ want some safety around an empty/null traceID as I think if there isn't an active tracer, this will get a NoopSpan (or something like that) and could publish an empty value.

Something like:

```suggestion
    span.context().toTraceId()?.takeIf { it.isNotBlank() }?.let {
        metadata[SqsJob.JOBQUEUE_METADATA_ORIGINAL_TRACE_ID] = it
    }
```

mpeyper: Yeah, I don't think the span or context can be `null` (the OT library will return a NoopSpan or something like that), but see my comment [here](https://github.com/cashapp/misk/pull/3550/files#r2543991249) and some additional logic around empty trace IDs.

nssherpa: Good idea. Done!

PR #3553, opened 2025-11-21
misk-mcp: Rewrite JsonSchemaExtensions to use serial descriptor

andrewjenkins:
## Description

Rewrites the JsonSchemaExtensions to rely on using the serial descriptor, rather than reflection on the primary constructor.

This fixes some shortcomings with the current implementation:
* Objects without primary constructors cannot be included in the JsonSchema, since it raises an exception.
* Objects with custom serializers have their custom serializer ignored when it comes to schema generation

## Testing Strategy

Unit tests

## Related Work
This unblocks https://github.com/squareup/cash-server/pull/46242

## Communication Plan
I'll update any callers of misk-mcp that are impacted by this change.

## Definition of Done
Confirm that this works across all consumers of misk-mcp

## Checklist

### DIFF
--- a/misk-admin/src/main/kotlin/misk/web/metadata/guice/GuiceDashboardTabModule.kt
+++ b/misk-admin/src/main/kotlin/misk/web/metadata/guice/GuiceDashboardTabModule.kt
@@ -1,6 +1,5 @@
 package misk.web.metadata.guice
 
-import com.google.inject.multibindings.OptionalBinder
 import misk.inject.KAbstractModule
 import misk.web.WebActionModule
 import misk.web.dashboard.AdminDashboard
@@ -16,7 +15,6 @@ class GuiceDashboardTabModule: KAbstractModule() {
       menuCategory = "Container Admin",
       menuLabel = "Guice",
     ))
-    OptionalBinder.newOptionalBinder(binder(), GuiceSourceUrlProvider::class.java)
-      .setDefault().to(GitHubSourceUrlProvider::class.java)
+    bindOptionalDefault<GuiceSourceUrlProvider>().to<GitHubSourceUrlProvider>()
   }
 }
--- a/misk-admin/web/tabs/database/package-lock.json
+++ b/misk-admin/web/tabs/database/package-lock.json
@@ -45,6 +45,7 @@
       "resolved": "https://registry.npmjs.org/@babel/core/-/core-7.15.0.tgz",
       "integrity": "sha512-tXtmTminrze5HEUPn/a0JtOzzfp0nk+UEXQ/tqIJo3WDGypl/2OFQEMll/zSFU8f/lfmfLXvTaORHF3cfXIQMw==",
       "dev": true,
+      "peer": true,
       "dependencies": {
         "@babel/code-frame": "^7.14.5",
         "@babel/generator": "^7.15.0",
@@ -1879,6 +1880,7 @@
       "version": "10.0.28",
       "resolved": "https://registry.npmjs.org/@emotion/core/-/core-10.0.28.tgz",
       "integrity": "sha512-pH8UueKYO5jgg0Iq+AmCLxBsvuGtvlmiDCOuv8fGNYn3cowFpLN98L8zO56U0H1PjDIyAlXymgL3Wu7u7v6hbA==",
+      "peer": true,
       "dependencies": {
         "@babel/runtime": "^7.5.5",
         "@emotion/cache": "^10.0.27",
@@ -3715,6 +3717,7 @@
       "version": "16.14.11",
       "resolved": "https://registry.npmjs.org/@types/react/-/react-16.14.11.tgz",
       "integrity": "sha512-Don0MtsZZ3fjwTJ2BsoqkyOy7e176KplEAKOpr/4XDdzinlyJBn9yfsKn5mcSgn4kh1B22+3tBnzBC1z63ybtQ==",
+      "peer": true,
       "dependencies": {
         "@types/prop-types": "*",
         "@types/scheduler": "*",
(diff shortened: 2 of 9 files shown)


misk-mcp/src/main/kotlin/misk/mcp/Description.kt
andrewjenkins, line 1: I needed to change the annotation target from `VALUE_PARAMETER` -> `PROPERTY` in order for the kotlinx serialization compiler plugin to be able to pick up this annotation.

See docs for details https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-core/kotlinx.serialization/-serial-info/

andrewjenkins: This is technically a breaking change, but I suspect in practice it won't actually break anyone. If it does, I'll make the necessary updates.

misk-mcp/src/main/kotlin/misk/mcp/internal/JsonSchemaExtensions.kt
```diff
@@ -1,139 +1,144 @@
...
-    val name = param.name ?: continue
-    val description = param.findAnnotation<Description>()?.value
-    properties[name] = param.type.generateJsonSchema(level, description)
-
-    if (!param.isOptional && !param.type.isMarkedNullable) {
-      required.add(name)
-    }
-  }
+    val serializer =
+        serializerOrNull() ?: throw IllegalArgumentException("No serializer found for class ${this.qualifiedName}")
```
jclyne: nit: maybe add something like "did you add a @Serializable annotation?"

misk-mcp/src/main/kotlin/misk/mcp/internal/JsonSchemaExtensions.kt
```diff
@@ -1,139 +1,144 @@
...
     description?.let {
-      put("description", JsonPrimitive(it))
+        put("description", JsonPrimitive(it))
     }
-    put("properties", JsonObject(properties))
-    if (level == 1) {
-      put("required", JsonArray(required.map { JsonPrimitive(it) }))
+    // JsonObjects and Elements have infinite serial descriptor depth. Prevent recursing into them.
+    if (this@generateJsonSchema in setOf(JsonObject.serializer().descriptor, JsonElement.serializer().descriptor)) {
+        put("additionalProperties", JsonPrimitive(true))
```
jclyne: Not exactly sure what this does and how it interacts with the additionalProperties field. Also this only sets it to true and the default its true

misk-mcp/src/main/kotlin/misk/mcp/internal/JsonSchemaExtensions.kt
```diff
@@ -1,139 +1,144 @@
...
-          }
-          put("enum", JsonArray(enumValues.map { JsonPrimitive(it) }))
+            put("properties", JsonObject(properties))
+            if (level == 1) {
+                put("required", JsonArray(required.map { JsonPrimitive(it) }))
+            }
+        }
+
+        SerialKind.CONTEXTUAL -> {
+            put("additionalProperties", JsonPrimitive(true))
```
jclyne: Same question here about additionalProperties

misk-mcp/src/main/kotlin/misk/mcp/internal/JsonSchemaExtensions.kt
jclyne: Makes sense now with the added test, assume it can anything inside a JsonObject or JsonElement

misk-mcp/src/main/kotlin/misk/mcp/internal/JsonSchemaExtensions.kt
```diff
@@ -1,139 +1,144 @@
...
-    if (level == 1) {
-      put("required", JsonArray(required.map { JsonPrimitive(it) }))
+    // JsonObjects and Elements have infinite serial descriptor depth. Prevent recursing into them.
+    if (this@generateJsonSchema in setOf(JsonObject.serializer().descriptor, JsonElement.serializer().descriptor)) {
+        put("additionalProperties", JsonPrimitive(true))
+        return@buildJsonObject
     }
-  }
-}
+    when (kind) {
```
jclyne: sealed type is great here

misk-mcp/src/main/kotlin/misk/mcp/internal/JsonSchemaExtensions.kt
```diff
@@ -1,139 +1,144 @@
...
-        put("description", JsonPrimitive(it))
-      }
-    }
+        SerialKind.ENUM -> {
+            val enumValues = (0 until elementsCount).map { index ->
+                val enumName = getElementName(index)
+                // Check for @SerialName annotation
+                val serialNameAnnotation =
+                    getElementDescriptor(index).annotations.filterIsInstance<SerialName>().firstOrNull()
+                serialNameAnnotation?.value ?: enumName
```
jclyne: Do we still need to handle this explicitly if we are using the SerialDescriptor, I bet the names are already taken from the SerialName

misk-mcp/src/main/kotlin/misk/mcp/internal/JsonSchemaExtensions.kt
andrewjenkins: This just means that any additional properties that aren't defined in the schema are allowed. But it looks like the default is true, so we can leave it out: https://json-schema.org/understanding-json-schema/reference/object#additionalproperties

misk-mcp/src/main/kotlin/misk/mcp/internal/JsonSchemaExtensions.kt
andrewjenkins: Oh 100%, good call!

PR #3558, opened 2025-11-23
feat: introduce `isHeld()` method for `Lease`

ThePumpingLemma:
## Description
This method works similarly to `checkHeld()` save that it should never make a network call.

Additionally, deprecate:
* `checkHeld()`, since it is the same as `acquire()`
* `checkHeldElsewhere()`, since there is no guarantee that the lease is actually held somewhere else

### DIFF
--- a/wisp/wisp-lease-testing/src/main/kotlin/wisp/lease/FakeLease.kt
+++ b/wisp/wisp-lease-testing/src/main/kotlin/wisp/lease/FakeLease.kt
@@ -8,7 +8,9 @@ class FakeLease(
 
     override fun shouldHold(): Boolean = true
 
-    override fun checkHeld() = manager.isLeaseHeld(name)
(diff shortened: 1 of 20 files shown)


wisp/wisp-lease-testing/src/main/kotlin/wisp/lease/FakeLease.kt
```diff
@@ -19,19 +21,19 @@ class FakeLease(
      * @return true if this process acquires the lease.
      */
     override fun acquire(): Boolean {
-        val result = checkHeld()
-        if (checkHeld()) {
+        val result = isHeld()
+        if (isHeld()) {
```
moshegood: if(result)

ThePumpingLemma: fixed!

adrw (approved):
LG

PR #3564, opened 2025-11-26
misk-clustering-dynamodb: disable cluster membership when DISABLE_ASYNC_TASKS is set

moshegood:
## Description

Playpen pods can potentially run with DISABLE_ASYNC_TASKS=true
When this is set, these pods should not be part of the clustering group membership. This way, none of the load-balanced leases are assigned to this pod. This is necessary, as these pods will not be doing any of the work that the leases are for.

## Checklist

- [ ] I have added tests to have confidence my changes work as expected.

### DIFF
--- a/misk-admin/web-actions/package-lock.json
+++ b/misk-admin/web-actions/package-lock.json
@@ -87,6 +87,7 @@
       "integrity": "sha512-vMqyb7XCDMPvJFFOaT9kxtiRh42GwlZEg1/uIgtZshS5a/8OaduUfCi7kynKgc3Tw/6Uo2D+db9qBttghhmxwQ==",
       "dev": true,
       "license": "MIT",
+      "peer": true,
       "dependencies": {
         "@ampproject/remapping": "^2.2.0",
         "@babel/code-frame": "^7.26.2",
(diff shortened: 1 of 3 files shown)


doctown (approved):

misk-clustering-dynamodb/src/main/kotlin/misk/clustering/dynamo/DynamoClusterModule.kt
```diff
@@ -21,8 +23,10 @@ class DynamoClusterModule @JvmOverloads constructor(private val config: DynamoCl
     bind<DefaultCluster>().toInstance(defaultCluster)
     bind<ClusterService>().toInstance(defaultCluster)
     install(ServiceModule<ClusterService>())
+    install(DefaultAsyncSwitchModule())
```
nssherpa, line 26: New to this code, but looks like this binds an `AlwaysEnabledSwitch` where [isEnabled](https://github.com/cashapp/misk/blob/9435a31ec93f3d8fa0fc0fe1166bde40ed2867d6/misk-inject/src/main/kotlin/misk/inject/AsyncSwitch.kt#L69-L73) returns true regardless, so is the intention to always run `DynamoClusterWatcherTask`? Or not run `DynamoClusterWatcherTask` when `DISABLE_ASYNC_TASKS` is set?

adrw: Internally, we'll bind a real switch which overrides the always enabled one, but we bind this optional one so that any existing tests and code paths still work as before. https://github.com/squareup/cash-server/pull/47262

PR #3576, opened 2025-12-01
Add delete method to WebTestClient

qnguyen-block:
## Description

My service has a delete method and we want to add this method to the WebTestClient so that we can write integration tests for our method

## Testing Strategy
Added integration test for the delete method. 
This change is a low to no risk as it only affect the usage of WebTestClient. 


adrw (approved):
LG

PR #3579, opened 2025-12-02
fix: Populate attributes map by bringing parity to the v1 version of SqsJob

nssherpa:
## Description

This PR exposes `message.attributes` into the top level `Job#attributes` map so that consumers can access any specific attributes they might need from the `SqsJob` via this api. 

The V1 version does the same [here](https://github.com/cashapp/misk/blob/49ac24ad3f55f6f175b681f177051fba4f813a5b/misk-aws/src/main/kotlin/misk/jobqueue/sqs/SqsJob.kt#L28-L33) and this logic seems to missing from the v2 version of SqsJob. Among other things this is helpful accessing various attributes via the Job#attributes api, and this is particularly useful for baggage propagation and piggybacking on AWSTraceHeader which serializes the baggage headers, which is only available in message.attributes map. See related [PR and testing](https://github.com/squareup/cash-server/pull/46267#discussion_r2579582066) on this.

## Testing Strategy

Tested this with cash-playpen-testbed and using the aws sqs V2 modules to pull AWSTraceHeaders from the serialized message.attributes. Prior to this change `Job#attributes` has an empty map.

### Before this change

image

### After this change
Screenshot 2025-12-03 at 10 44 05 AM

## Checklist

### DIFF
--- a/misk-admin/web/tabs/database/package-lock.json
+++ b/misk-admin/web/tabs/database/package-lock.json
@@ -45,6 +45,7 @@
       "resolved": "https://registry.npmjs.org/@babel/core/-/core-7.15.0.tgz",
       "integrity": "sha512-tXtmTminrze5HEUPn/a0JtOzzfp0nk+UEXQ/tqIJo3WDGypl/2OFQEMll/zSFU8f/lfmfLXvTaORHF3cfXIQMw==",
       "dev": true,
+      "peer": true,
       "dependencies": {
         "@babel/code-frame": "^7.14.5",
         "@babel/generator": "^7.15.0",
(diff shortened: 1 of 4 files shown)


misk-aws2-sqs/src/main/kotlin/misk/aws2/sqs/jobqueue/SqsJob.kt
```diff
@@ -14,8 +14,12 @@ class SqsJob(
...
 ) : Job {
   override val body: String = message.body()
-  override val attributes: Map<String, String>
-    get() = emptyMap()
+  override val attributes: Map<String, String> by lazy {
+    message.messageAttributes()
+      .filter { (key, _) -> key != JOBQUEUE_METADATA_ATTR }
+      .map { (key, value) -> key to value.stringValue() }.toMap()
+      .plus(message.attributes().map {(key, value) -> key.toString() to value })
+  }
```
nssherpa, line 22: The V1 version does the same [here](https://github.com/cashapp/misk/blob/49ac24ad3f55f6f175b681f177051fba4f813a5b/misk-aws/src/main/kotlin/misk/jobqueue/sqs/SqsJob.kt#L28-L33) and this logic seems to missing from the v2 version of `SqsJob`. Among other things this is helpful in accessing various attributes via the [Job#attributes ](https://github.com/cashapp/misk/blob/49ac24ad3f55f6f175b681f177051fba4f813a5b/misk-jobqueue/src/main/kotlin/misk/jobqueue/Job.kt#L21-L22) api, and this is particularly useful for baggage propagation and piggybacking on `AWSTraceHeader` which serializes the baggage headers, and it is only available in `message.attributes` map. See related [PR and testing ](https://github.com/squareup/cash-server/pull/46267#discussion_r2579582066)on this.

PR #3581, opened 2025-12-03
Fix ServiceModule duplicate binding failures

adrw:


jclyne (approved):

misk-service/src/main/kotlin/misk/ServiceModule.kt
```diff
@@ -186,26 +186,6 @@ constructor(
   @JvmOverloads
   inline fun <reified T : Service> enhancedBy(qualifier: KClass<out Annotation>? = null) =
     enhancedBy(T::class.toKey(qualifier))
-
-  override fun equals(other: Any?): Boolean {
```
adrw, line 190: this wasn't needed so removing

PR #3585, opened 2025-12-09
Add enqueueBufferedAsync for automatic batching and parallelization

yyuan-squareup:
## Description

`SqsAsyncBatchManager` is a replacement for `AmazonSQSBufferedAsyncClient` in SDK v2.

This PR leverage `SqsAsyncBatchManager` to make it easy for client code to send requests to SQS without the need to chunk into small batches of 10 messages. The client code can just send one message at at time, and SqsAsyncBatchManager will handle batching and parallelizing.

This should give a higher throughput compared to the batching API without parallelization, since `SqsAsyncBatchManager` will send batches async in parallel.

As added benefit, `enqueueBufferedAsync` returns a future to the client code, allowing the client code to keep track of the success / failure for each message.

## Testing Strategy
Unit test

## Related Work
    Outbox for SQS

### DIFF
--- a/misk-aws2-sqs/api/misk-aws2-sqs.api
+++ b/misk-aws2-sqs/api/misk-aws2-sqs.api
@@ -73,6 +73,7 @@ public final class misk/aws2/sqs/jobqueue/SqsJobEnqueuer : misk/jobqueue/v2/JobE
 	public fun enqueueAsync (Lmisk/jobqueue/QueueName;Ljava/lang/String;Ljava/lang/String;Ljava/time/Duration;Ljava/util/Map;)Ljava/util/concurrent/CompletableFuture;
 	public fun enqueueBlocking (Lmisk/jobqueue/QueueName;Ljava/lang/String;Ljava/lang/String;Ljava/time/Duration;Ljava/util/Map;)V
 	public fun enqueueBuffered (Lmisk/jobqueue/QueueName;Ljava/lang/String;Ljava/lang/String;Ljava/time/Duration;Ljava/util/Map;Lkotlin/coroutines/Continuation;)Ljava/lang/Object;
+	public fun enqueueBuffered (Lmisk/jobqueue/QueueName;Lmisk/jobqueue/v2/JobEnqueuer$JobRequest;Lkotlin/coroutines/Continuation;)Ljava/lang/Object;
 	public fun enqueueBufferedAsync (Lmisk/jobqueue/QueueName;Ljava/lang/String;Ljava/lang/String;Ljava/time/Duration;Ljava/util/Map;)Ljava/util/concurrent/CompletableFuture;
 	public fun enqueueBufferedAsync (Lmisk/jobqueue/QueueName;Lmisk/jobqueue/v2/JobEnqueuer$JobRequest;)Ljava/util/concurrent/CompletableFuture;
 }
--- a/misk-jobqueue/api/misk-jobqueue.api
+++ b/misk-jobqueue/api/misk-jobqueue.api
@@ -250,6 +250,7 @@ public final class misk/jobqueue/v2/FakeJobEnqueuer : misk/testing/FakeFixture, 
 	public fun enqueueAsync (Lmisk/jobqueue/QueueName;Ljava/lang/String;Ljava/lang/String;Ljava/time/Duration;Ljava/util/Map;)Ljava/util/concurrent/CompletableFuture;
 	public fun enqueueBlocking (Lmisk/jobqueue/QueueName;Ljava/lang/String;Ljava/lang/String;Ljava/time/Duration;Ljava/util/Map;)V
 	public fun enqueueBuffered (Lmisk/jobqueue/QueueName;Ljava/lang/String;Ljava/lang/String;Ljava/time/Duration;Ljava/util/Map;Lkotlin/coroutines/Continuation;)Ljava/lang/Object;
+	public fun enqueueBuffered (Lmisk/jobqueue/QueueName;Lmisk/jobqueue/v2/JobEnqueuer$JobRequest;Lkotlin/coroutines/Continuation;)Ljava/lang/Object;
 	public fun enqueueBufferedAsync (Lmisk/jobqueue/QueueName;Ljava/lang/String;Ljava/lang/String;Ljava/time/Duration;Ljava/util/Map;)Ljava/util/concurrent/CompletableFuture;
 	public fun enqueueBufferedAsync (Lmisk/jobqueue/QueueName;Lmisk/jobqueue/v2/JobEnqueuer$JobRequest;)Ljava/util/concurrent/CompletableFuture;
 	public final fun handleJob (Lmisk/jobqueue/v2/Job;ZI)Z
@@ -304,6 +305,7 @@ public abstract interface class misk/jobqueue/v2/JobEnqueuer {
 	public abstract fun enqueueAsync (Lmisk/jobqueue/QueueName;Ljava/lang/String;Ljava/lang/String;Ljava/time/Duration;Ljava/util/Map;)Ljava/util/concurrent/CompletableFuture;
 	public abstract fun enqueueBlocking (Lmisk/jobqueue/QueueName;Ljava/lang/String;Ljava/lang/String;Ljava/time/Duration;Ljava/util/Map;)V
 	public abstract fun enqueueBuffered (Lmisk/jobqueue/QueueName;Ljava/lang/String;Ljava/lang/String;Ljava/time/Duration;Ljava/util/Map;Lkotlin/coroutines/Continuation;)Ljava/lang/Object;
+	public abstract fun enqueueBuffered (Lmisk/jobqueue/QueueName;Lmisk/jobqueue/v2/JobEnqueuer$JobRequest;Lkotlin/coroutines/Continuation;)Ljava/lang/Object;
 	public abstract fun enqueueBufferedAsync (Lmisk/jobqueue/QueueName;Ljava/lang/String;Ljava/lang/String;Ljava/time/Duration;Ljava/util/Map;)Ljava/util/concurrent/CompletableFuture;
 	public abstract fun enqueueBufferedAsync (Lmisk/jobqueue/QueueName;Lmisk/jobqueue/v2/JobEnqueuer$JobRequest;)Ljava/util/concurrent/CompletableFuture;
 }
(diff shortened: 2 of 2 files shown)


misk-jobqueue/src/main/kotlin/misk/jobqueue/v2/JobEnqueuer.kt
```diff
@@ -81,6 +80,45 @@ interface JobEnqueuer {
...
+   *
+   * Messages are buffered client-side and sent in batches automatically when:
+   * - The batch size reaches 10 messages, or
+   * - The send frequency timeout is reached (50 ms, see RealSqsBatchManagerFactory)
+   *
+   * Callers should check the returned future if they need to verify the message was successfully sent.
+   *
+   * @return CompletableFuture that completes with true when the message is successfully sent
+   */
+  fun enqueueBufferedAsync(
```
yyuan-squareup, line 111: this API is designed to be called for a large batch of requests, no need to have a blocking or suspend version.

misk-aws2-sqs/src/main/kotlin/misk/aws2/sqs/jobqueue/SqsJobQueueModule.kt
```diff
@@ -52,4 +53,14 @@ open class SqsJobQueueModule @JvmOverloads constructor(
...
     credentialsProvider: AwsCredentialsProvider,
   ): SqsClientFactory = RealSqsClientFactory(credentialsProvider, configureClient)
+
+  @Provides
+  @Singleton
+  fun sqsBatchManagerFactory(
+    sqsClientFactory: SqsClientFactory,
+  ): SqsBatchManagerFactory {
+    // Pool size of 2 provides fault isolation between regions if one has an outage.
+    val scheduledExecutor = Executors.newScheduledThreadPool(2)
```
yyuan-squareup: this executor only handles flushing the buffer every 10 messages or when reaching 50ms. Flushing the buffer  is efficient and should not be blocking, so using a single thread scheduled executor is probably enough. Using a pool of size 2 (one for each region) for added safety.

misk-aws2-sqs/src/main/kotlin/misk/aws2/sqs/jobqueue/RealSqsBatchManagerFactory.kt
```diff
@@ -0,0 +1,36 @@
...
+  private val sendRequestFrequency: Duration = Duration.ofMillis(50),
+) : SqsBatchManagerFactory {
+  private val batchManagers = ConcurrentHashMap<String, SqsAsyncBatchManager>()
+
+  override fun get(region: String): SqsAsyncBatchManager {
+    return batchManagers.computeIfAbsent(region) {
+      val client = sqsClientFactory.get(region)
+      SqsAsyncBatchManager.builder()
+        .client(client)
+        .scheduledExecutor(scheduledExecutor)
```
yyuan-squareup: share a scheduled executor for both regions. See more comment about the scheduledExecutor below

misk-aws2-sqs/src/main/kotlin/misk/aws2/sqs/jobqueue/SqsJobEnqueuer.kt
```diff
@@ -177,6 +177,57 @@ class SqsJobEnqueuer @Inject constructor(
...
+        .messageBody(body)
+        .delaySeconds(deliveryDelay?.toSeconds()?.toInt())
+        .messageAttributes(attrs)
+        .build()
+
+      val startTimeMs = clock.millis()
+      try {
+        val region = sqsConfig.getQueueConfig(queueName).region!!
+        val batchManager = sqsBatchManagerFactory.get(region)
+        val response = batchManager.sendMessage(request)
```
yyuan-squareup, line 180: the main part is here. It just calls `SqsAsyncBatchManager.sendMessage` to add the message to the buffer. Everything else is the same as `enqueueAsync`

adrw: If everything else is the same, should we pull out a common method that both delegate to and avoid the code duplication?

misk-jobqueue/src/main/kotlin/misk/jobqueue/v2/JobEnqueuer.kt
adrw: The main advantage would be exposing a suspend version too which works well for services using coroutines. Might be worth considering. General philosophy of these new libraries has been to expose blocking and suspending versions of each API to give teams the choice, and then delegate implementation to essentially the same code within the library for easy maintenance.

misk-aws2-sqs/src/main/kotlin/misk/aws2/sqs/jobqueue/SqsJobEnqueuer.kt
yyuan-squareup: yes, good point, refactored

misk-jobqueue/src/main/kotlin/misk/jobqueue/v2/JobEnqueuer.kt
yyuan-squareup: Thought about it being used from coroutine, having a suspend version makes sense. Added.

misk-aws2-sqs/src/main/kotlin/misk/aws2/sqs/jobqueue/RealSqsBatchManagerFactory.kt
```diff
@@ -0,0 +1,31 @@
...
+import software.amazon.awssdk.services.sqs.batchmanager.SqsAsyncBatchManager
+import java.time.Duration
+import java.util.concurrent.ConcurrentHashMap
+import java.util.concurrent.ScheduledExecutorService
+
+internal class RealSqsBatchManagerFactory(
+  private val sqsClientFactory: SqsClientFactory,
+  private val scheduledExecutor: ScheduledExecutorService,
+  // 50ms flush frequency (vs 200ms default) for higher throughput at the cost of slightly more API calls.
+  private val sendRequestFrequency: Duration = Duration.ofMillis(50),
```
mateuszmrozewski: This should go to configuration.

misk-aws2-sqs/src/main/kotlin/misk/aws2/sqs/jobqueue/SqsJobQueueModule.kt
mateuszmrozewski: I am confused by this. SQS is regional. The only situation when you need more clients is if you explicitly use queue from different regions (via configuration). This is handled by having separate clients. I don't see how hardcoding this thread pool to 2 helps with that.

mateuszmrozewski: I think separate executors per region would be better with a configurable amount of threads.

mateuszmrozewski (changes requested):

misk-aws2-sqs/src/main/kotlin/misk/aws2/sqs/jobqueue/RealSqsBatchManagerFactory.kt
```diff
@@ -0,0 +1,67 @@
...
+) : SqsBatchManagerFactory, AbstractIdleService() {
+  private val sendRequestFrequency: Duration = Duration.ofMillis(sqsConfig.buffered_batch_flush_frequency_ms)
+  private val batchManagers = ConcurrentHashMap<String, SqsAsyncBatchManager>()
+  private val scheduledExecutors = ConcurrentHashMap<String, ScheduledExecutorService>()
+
+  override fun get(region: String): SqsAsyncBatchManager {
+    return batchManagers.computeIfAbsent(region) {
+      val client = sqsClientFactory.get(region)
+      val executor = Executors.newSingleThreadScheduledExecutor { runnable ->
+        Thread(runnable, "sqs-async-batch-manager-$region").apply { isDaemon = true }
```
yyuan-squareup, line 27: set it as daemon, so that JVM does not block on it when shutting down. Not strictly needed here since guice will manage the shutdown, just being defensive.

misk-jobqueue/src/main/kotlin/misk/jobqueue/v2/JobEnqueuer.kt
keeferrourke: Should there also be a `enqueueBufferedBlocking` method like the following?

```kt
fun enqueueBufferedBlocking(
    queueName: QueueName,
    body: String,
    idempotencyKey: String? = null,
    deliveryDelay: Duration? = Duration.ZERO,
    attributes: Map<String, String> = emptyMap(),
): Boolean = enqueueBufferedAsync(queueName, body, idempotencyKey, deliveryDelay, attributes).join()
```

misk-jobqueue/src/main/kotlin/misk/jobqueue/v2/JobEnqueuer.kt
```diff
@@ -81,6 +80,62 @@ interface JobEnqueuer {
...
+  ): CompletableFuture<Boolean>
+
+  /**
+   * Enqueue a job using automatic batching for high-throughput scenarios.
+   *
+   * This is a convenience overload that accepts a [JobRequest] instead of individual parameters.
+   *
+   * @see enqueueBufferedAsync
+   */
+  fun enqueueBufferedAsync(
```
keeferrourke, line 140: should there be `suspend` and Blocking variants of this API as well?

misk-jobqueue/src/main/kotlin/misk/jobqueue/v2/JobEnqueuer.kt
yyuan-squareup: No, I intentionally left out the blocking version to prevent people from misusing it. `SqsAsyncBatchManager` will not send the requests out to SQS until it accumulates 10 messages or it reaches the configured timeout. If we provide the blocking version, in the worst scenario, the caller will be blocked until another 9 messages are enqueued and the batch is sent before getting a response.

misk-jobqueue/src/main/kotlin/misk/jobqueue/v2/JobEnqueuer.kt
yyuan-squareup: good catch, added a suspend version.

misk-aws2-sqs/src/main/kotlin/misk/aws2/sqs/jobqueue/RealSqsBatchManagerFactory.kt
yyuan-squareup: Done, updated the PR

misk-aws2-sqs/src/main/kotlin/misk/aws2/sqs/jobqueue/SqsJobQueueModule.kt
yyuan-squareup: implemented

PR #3588, opened 2025-12-10
Add extension function to extract typed result from ToolResult

jclyne:
## Description

Added a new extension function `result()` to `ToolResult` that allows for type-safe extraction of results from `StructuredToolResult`. This function provides a convenient way to cast the result to the expected type with proper error handling when the types don't match. This is useful mainly for unit testing to extract the result for verification

## Checklist

### DIFF
--- a/misk-mcp/src/main/kotlin/misk/mcp/McpTool.kt
+++ b/misk-mcp/src/main/kotlin/misk/mcp/McpTool.kt
@@ -581,3 +581,55 @@ abstract class StructuredMcpTool<I : Any, O : Any> : McpTool<I>() {
       .arguments.last().type!!
   }
 }
+
+/**
+ * Extracts the typed result from a [McpTool.ToolResult].
+ *
+ * This extension function provides a convenient way to extract the strongly-typed result
+ * from a [StructuredMcpTool.StructuredToolResult]. It is primarily useful in testing scenarios
+ * where you want to verify the output of a [StructuredMcpTool] without dealing with the
+ * sealed interface directly.
+ *
+ * ## Usage
+ *
+ * ```kotlin
+ * @Test
+ * fun `test calculator tool returns correct result`() = runBlocking {
+ *   val tool = CalculatorTool()
+ *   val toolResult = tool.handle(CalculatorToolInput("ADD", 5, 3))
+ *
+ *   // Extract the typed result
+ *   val output: CalculatorToolOutput = toolResult.result()
+ *   assertEquals(8, output.result)
+ * }
+ * ```
+ *
+ * ## Type Safety
+ *
+ * The function uses reified type parameters to provide compile-time type checking.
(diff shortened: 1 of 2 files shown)


jclyne:
* **#3588** <a href="https://app.graphite.com/github/pr/cashapp/misk/3588?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/cashapp/misk/3588?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

This stack of pull requests is managed by <a href="https://graphite.dev?utm-source=stack-comment"><b>Graphite</b></a>. Learn more about <a href="https://stacking.dev/?utm_source=stack-comment">stacking</a>.

misk-mcp/src/main/kotlin/misk/mcp/McpTool.kt
```diff
@@ -581,3 +581,55 @@ abstract class StructuredMcpTool<I : Any, O : Any> : McpTool<I>() {
...
+ *   [StructuredMcpTool] that produced this result.
+ * @return The typed result extracted from the [misk.mcp.StructuredMcpTool.StructuredToolResult].
+ * @throws IllegalStateException If this [misk.mcp.McpTool.ToolResult] is not a [misk.mcp.StructuredMcpTool.StructuredToolResult],
+ *   or if the result cannot be cast to type [T].
+ *
+ * @see StructuredMcpTool for creating tools with structured output
+ * @see StructuredMcpTool.StructuredToolResult for the result type this function operates on
+ */
+@OptIn(ExperimentalMiskApi::class)
+inline fun <reified T: Any> McpTool.ToolResult.result(): T = when (this) {
```
ean5533, line 623: Two comments that are completely unrelated to each other:

1. Should this method just extend `StructuredToolResult` instead? What is the use case for a caller having a `ToolResult` that it doesn't know if it's actually `StructuredToolResult` or not?

2. It might be useful to also offer a `resultOrNull()` so that callers can do something like `assertNull(myToolResult.resultOrNull<MyType>())` without needing to handle an exception.

jclyne: This is mainly for testing and it removes the need to cast ToolResult to a StructuredToolResult and perform an unchecked cast on the type parameter. The expectation is that the caller knows what they are trying to get. Tool authors just need to build the result, the mcp framework will just serialize it.

jclyne: There is no need to extend this to StructuredToolResult the result is already available and properly typed. This just wraps the casting of the ToolResult that comes back from tool.handle(). This was done in a bunch of places in Activity so I think it makes sense to make it easier.

PR #3598, opened 2025-12-16
Update slack client with new endpoint

srhjang:
## Description

Adding a new endpoint for the misk Slack client, which allows making GET request for user object. Need this to complete a webhook for another service, which will get the user's LDAP from a Slack message thread.

## Related Work
Related to [this ticket ](https://linear.app/squareup/issue/RATS-149/look-into-adding-webhook-so-slack-emojis-will-update-ccb-statuses)


adrw (approved):
LG

PR #3599, opened 2025-12-17
Implement equals and hashcode on GrpcClientModule

andrewjenkins:
## Description
Implements equals() and hashCode() methods on the GrpcClientModule.

This means that the same GrpcClientModule can be safely installed multiple times since Guice will de-dupe the module installations if they are equal to each other.

## Checklist


misk/src/main/kotlin/misk/client/GrpcClientModule.kt
adrw, line 36:
```suggestion
) : KInstallOnceModule() {
```

We should do this instead.

Or just fix the callsite errors?

PR #3600, opened 2025-12-17
fix: restore calls to checkHeld()

ThePumpingLemma:
## Description

An internal implementation of `checkHeld()` has a side effect of dropping the lock, which is a dependency for these uses to work with load-balanced leasing. For now, add back the calls to `checkHeld()` to restore the previous behavior.

The long-term fix is to change the API to require a release with every call to acquire (e.g. via try-with-resource or a use block) so that callers are agnostic to how the underlying lease implementation manages its lifetime.


adrw (approved):
LG

PR #3601, opened 2025-12-17
Cache jOOQ Configurations

tgregory-block:
## Description

1. jOOQ's `Configuration` object is where it maintains its reflection cache, which it uses to cache field lookups when mapping records into various other objects (e.g. when using `into()`). 
2. The current `JooqTransacter` implementation creates a new `Configuration` on every transaction, which means services are always running with an empty reflection cache.
3. jOOQ [recommends sharing the `Configuration`](https://www.jooq.org/doc/latest/manual/sql-building/dsl-context/thread-safety/)

This PR solves this by caching the configuration objects. Currently, users can modify the session by passing in `TransactionOptions` to each transaction call site; however, only one parameter of this object, the transaction isolation level, actually affects the `Configuration`. We leverage this by using the isolation level as the cache key, and returning the cached `Configuration` for the given isolation level. 

## Testing Strategy

Added an integration tests which asserts that the new config and old configs are identical, and that the implementation is thread safe

## Communication Plan
  * Will tag folks in internal channels. This should not affect service owners unless they are manually constructing `JooqTransacter`, which should not be done regardless (would ideally like to make the constructor internal in a followup, but it's API breaking)

## Checklist

### DIFF
--- a/misk-jooq/src/main/kotlin/misk/jooq/config/CachedConfigurationFactory.kt
+++ b/misk-jooq/src/main/kotlin/misk/jooq/config/CachedConfigurationFactory.kt
@@ -1,21 +1,21 @@
 package misk.jooq.config
 
 import com.google.common.annotations.VisibleForTesting
+import java.time.Clock
+import java.util.concurrent.ConcurrentHashMap
 import misk.jdbc.DataSourceConfig
 import misk.jdbc.DataSourceService
 import misk.jooq.JooqTransacter
 import misk.jooq.TransactionIsolationLevel
 import misk.jooq.listeners.JooqTimestampRecordListenerOptions
 import org.jooq.Configuration
-import java.time.Clock
-import java.util.concurrent.ConcurrentHashMap
 
 /**
  * This [ConfigurationFactory] generates configurations and caches them by their associated [TransactionIsolationLevel].
  *
  * This is doable because the only user-modifiable parameter that can change the [Configuration] after the Guice
- * injector is built (i.e. after JooqModule is installed), is the transaction isolation level. It is important to
- * cache the [Configuration] and share it among transactions because [it contains the reflection cache](
+ * injector is built (i.e. after JooqModule is installed), is the transaction isolation level. It is important to cache
+ * the [Configuration] and share it among transactions because [it contains the reflection cache](
  * https://www.jooq.org/doc/latest/manual/sql-building/dsl-context/thread-safety/).
  */
 internal class CachedConfigurationFactory(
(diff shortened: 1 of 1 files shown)


misk-jooq/src/main/kotlin/misk/jooq/JooqModule.kt
```diff
@@ -10,96 +11,128 @@ import misk.inject.toKey
...
-  private val dataSourceClusterConfig: DataSourceClusterConfig,
-  private val jooqCodeGenSchemaName: String,
-  private val databasePool: DatabasePool = RealDatabasePool,
-  private val readerQualifier: KClass<out Annotation>? = null,
-  private val jooqTimestampRecordListenerOptions: JooqTimestampRecordListenerOptions =
-    JooqTimestampRecordListenerOptions(install = false),
-  private val installHealthChecks: Boolean = true,
-  private val installSchemaMigrator: Boolean = true,
-  private val jooqConfigExtension: Configuration.() -> Unit = {},
+    private val qualifier: KClass<out Annotation>,
```
adrw: Can you reverse the formatting differences? Seems to have moved from 2 spaces to 4.

misk-jooq/src/main/kotlin/misk/jooq/JooqModule.kt
```diff
@@ -10,96 +11,128 @@ import misk.inject.toKey
...
-        dataSourceClusterConfig.reader,
-        databasePool,
-        installHealthChecks,
-        installSchemaMigrator
-      )
-    )
+    override fun configure() {
+        install(
+            JdbcModule(
+                qualifier,
```
adrw: can we add named params?

misk-jooq/src/main/kotlin/misk/jooq/JooqModule.kt
```diff
@@ -10,96 +11,128 @@ import misk.inject.toKey
...
+                databasePool,
+                installHealthChecks,
+                installSchemaMigrator
+            )
+        )
+
+        bind<ConfigurationFactory>().annotatedWith(qualifier.java)
+            .toProvider(
+                CachedConfigurationFactoryProvider(
+                    dataSourceClusterConfig.writer,
```
adrw: can we add named params?

misk-jooq/src/main/kotlin/misk/jooq/JooqModule.kt
```diff
@@ -10,96 +11,128 @@ import misk.inject.toKey
...
+                    jooqConfigExtension,
+                    qualifier
+                )
+            ).asSingleton()
+        bindTransacter(qualifier)
+        if (readerQualifier != null && dataSourceClusterConfig.reader != null) {
+            bind<ConfigurationFactory>().annotatedWith(readerQualifier.java)
+                .toProvider(
+                    CachedConfigurationFactoryProvider(
+                        dataSourceClusterConfig.reader!!,
```
adrw: can we add named params?

misk-jooq/src/main/kotlin/misk/jooq/JooqTransacter.kt
```diff
@@ -86,7 +63,7 @@ class JooqTransacter @JvmOverloads constructor(
   ): RETURN_TYPE {
     var jooqSession: JooqSession? = null
     return try {
-      dslContext(dataSourceService, clock, dataSourceConfig, options).transactionResult { configuration ->
+      DSL.using(configurationFactory(options)).transactionResult { configuration ->
```
adrw, line 66: Nice catch to properly leverage the cache

PR #3602, opened 2025-12-17
Add kotlin-formatter to misk

tgregory-block:
## Description

We'd like to standardize formatting within `misk`. To that end, this PR:

1. Installs `kotlin-formatter` and `lefthook` so engineers can easily setup the bundled pre-commit hook
2. Adds a root `.editorconfig` that approximates the formatting given by `kotlin-formatter`
3. Updates the getting started doc to explain how to use these tools, as well as the `kotlin-formatter` intellij plugin

## Communication Plan

Will notify internally about this change after merging this PR to give time to merge outstanding PRs prior to the batch formatting update PR that will follow


bin/activate-hermit.fish
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/env fish
```
adrw, line 1: This some new part of Hermit? Misk already has the activate scripts? Why is it in fish?

docs/getting-started.md
```diff
@@ -12,6 +12,17 @@ Please follow [this](https://cashapp.github.io/hermit/) to install and activate
 
 Misk uses [Gradle][gradle] to build and run test locally.
 
+### Formatting
+
+Misk uses [`kotlin-formatter`](https://github.com/block/kotlin-formatter) for formatting. You can integrate it with your
+workflow by:
+
+1. Invoking `./bin/lefthook install` to install the formatting pre-commit hook.
+2. Installing the [gradle plugin](https://plugins.jetbrains.com/plugin/26482-kotlin-formatter).
```
adrw, line 21:
```suggestion
2. Installing the [IDE plugin](https://plugins.jetbrains.com/plugin/26482-kotlin-formatter).
```

bin/activate-hermit.fish
tgregory-block: I'm not sure when, but at some point `fish` support was added to hermit. It's sufficiently different from other shells that it needs a separate script for activation, and the change to hermit was made in such a way that downloading the script isn't automatic if hermit was already installed in a repo prior to the change; you have to rerun `hermit init`. I use `fish`, so I figured I'd go ahead and do that so my shell automatically activates the misk hermit when changing into the repo

PR #3606, opened 2025-12-18
Format all source files with block/kotlin-formatter

tgregory-block:
## Description

1. Applies the formatting added in #3602 to all source files
2. Adds a `.git-blame-ignore-revs`, and instructions for using it, for excluding this commit from `git-blame`. Will need a follow up to use the SHA from the merge commit, the existing one is just me testing that it worked. 

## Testing Strategy

I tested that this SHA is excluded from intellij git blame gutters. The test suite should continue to pass as normal. GitHub also respects a rev ignore file named `.git-blame-ignore-revs` so diff/blame viewed in github should ignore this as well. 

## Related Work
	* #3602

## Communication Plan
    - Will notify internal teams that they may wish to merge pending PRs before this goes in


adrw (approved):
LG

PR #3620, opened 2026-01-07
feat: remove `linux/amd64` platform hardcoding

kevink-sq:
## Description

Looking to introduce an arm64 build of `vttestserver` so need to remove the platform config as it prevents the arm image from starting up. Docker by default should look up the platform automatically on the host.

## Testing Strategy

Will follow up with testing internally since this is a test package.

### DIFF
--- a/misk-vitess/src/testFixtures/kotlin/misk/vitess/testing/internal/VitessDockerContainer.kt
+++ b/misk-vitess/src/testFixtures/kotlin/misk/vitess/testing/internal/VitessDockerContainer.kt
@@ -682,7 +682,6 @@ internal class VitessDockerContainer(
         .createContainerCmd(vitessImage)
         .withName(containerName)
         .withHostConfig(hostConfig)
-        .withPlatform("linux/amd64")
         .withHealthcheck(healthCheck)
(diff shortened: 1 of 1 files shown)


misk-vitess/src/testFixtures/kotlin/misk/vitess/testing/internal/VitessDockerContainer.kt
```diff
@@ -682,7 +682,6 @@ internal class VitessDockerContainer(
         .createContainerCmd(vitessImage)
         .withName(containerName)
         .withHostConfig(hostConfig)
-        .withPlatform("linux/amd64")
```
adrw, line 685: What happens if this is removed? Does it fall back to some default ie. the platform that it's currently run on?

adrw (approved):
LG

misk-vitess/src/testFixtures/kotlin/misk/vitess/testing/internal/VitessDockerContainer.kt
kevink-sq: Yes

PR #3779, opened 2026-04-14
feat: Support HEAD requests

sqchigozirim:
## Description

HTTP HEAD requests are rejected by Misk (404) because `dispatchMechanism()` doesn't recognize the HEAD method. This change maps HEAD requests to `DispatchMechanism.GET`, allowing them to be routed to the matching GET action. Jetty already handles [stripping](https://github.com/jetty/jetty.project/blob/8fc0958d5aff34151dd10e71eba3e3d7dbbf1862/jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/internal/HttpConnection.java#L911) the response body for HEAD requests, so no additional logic is needed beyond the routing change.

## Testing Strategy

Added a test in SupportedHttpMethodsTest that sends a HEAD request to a GET endpoint and verifies it returns a successful response with an empty body.
Built and test with my project (it receives and handles HEAD requests vs returning a 404 previously)


shellz-n-stuff (approved):
🔥

PR #3782, opened 2026-04-15
Preserve MULTI transaction mode through onSessionClose hooks

aparajon:
## Why

When `allowCrossShardTransactions()` resets `transaction_mode` to UNSPECIFIED in `afterCompletion`, the Hibernate session may still have pending `onSessionClose` hooks that need cross-shard access. These hooks reuse the thread-local Hibernate session, which is now in SINGLE mode.

Additionally, vtgate's reserved connection mode (triggered by `SET transaction_mode`) retains stale shard sessions across COMMIT ([vitessio/vitess#11616](https://github.com/vitessio/vitess/issues/11616)). Without re-setting to MULTI, subsequent WRITE operations on the same pooled connection fail in SINGLE mode with "multi-db transaction attempted".

## What

After resetting to UNSPECIFIED in `afterCompletion`, check if the Hibernate session is still open. If so, re-set to MULTI so close hooks and subsequent pooled connection users operate correctly.

## Test plan

- Added regression test `onSessionClose hooks can use cross-shard transactions after allowCrossShardTransactions` that:
  - Opens a transaction with `allowCrossShardTransactions()`
  - Registers an `onSessionClose` hook that does a cross-shard read
  - Verifies the hook succeeds
  - **Fails without the fix, passes with it** (verified locally)

Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/misk-hibernate/src/main/kotlin/misk/hibernate/Session.kt
+++ b/misk-hibernate/src/main/kotlin/misk/hibernate/Session.kt
@@ -74,12 +74,23 @@ fun Session.allowCrossShardTransactions() {
   // Reset to UNSPECIFIED after the transaction completes (commit or rollback) so it doesn't leak
   // to subsequent transactions via connection pool reuse. UNSPECIFIED falls back to whatever the
   // vtgate-level default is, so this is safe regardless of the vtgate's configured transaction mode.
+  //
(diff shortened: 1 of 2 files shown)


aparajon:
Not needed with v23 — read-only multi-shard transactions work in SINGLE mode, so onSessionClose hooks don't need MULTI re-set.

PR #3789, opened 2026-04-27
misk-testing: bound reused injector cache via MISK_TEST_REUSE_LRU_SIZE

staktrace:
The injector reuse cache could grow unboundedly in suites with many distinct module sets. Setting MISK_TEST_REUSE_LRU_SIZE to a positive integer caps the cache; evicted entries have their services stopped.

### DIFF
--- a/misk-testing/src/main/kotlin/misk/testing/MiskTestExtension.kt
+++ b/misk-testing/src/main/kotlin/misk/testing/MiskTestExtension.kt
@@ -3,10 +3,13 @@ package misk.testing
 import com.google.common.util.concurrent.ServiceManager
 import com.google.inject.Guice
 import com.google.inject.Injector
+import com.google.inject.Key
 import com.google.inject.Module
 import com.google.inject.testing.fieldbinder.BoundFieldModule
 import jakarta.inject.Inject
 import jakarta.inject.Singleton
+import java.util.Collections
+import java.util.LinkedHashMap
 import java.util.concurrent.ConcurrentHashMap
 import java.util.concurrent.TimeUnit
 import kotlin.concurrent.thread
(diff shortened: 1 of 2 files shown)


misk-testing/src/main/kotlin/misk/testing/MiskTestExtension.kt
```diff
@@ -24,8 +27,36 @@ internal class MiskTestExtension : BeforeEachCallback, AfterEachCallback {
...
+
+    private val maxLruSize: Int? =
+      System.getenv("MISK_TEST_REUSE_LRU_SIZE")?.toIntOrNull()?.takeIf { it > 0 }
+
+    // Access-order LinkedHashMap so the least-recently-used entry is the eviction candidate when
+    // MISK_TEST_REUSE_LRU_SIZE is set. Callers must hold the map's monitor for compound ops
+    // (e.g. getOrPut) since access-order updates the map on read.
+    private val injectedModules: MutableMap<List<Module>, Injector> = Collections.synchronizedMap(
+      object : LinkedHashMap<List<Module>, Injector>(16, 0.75f, true) {
+        override fun removeEldestEntry(eldest: Map.Entry<List<Module>, Injector>): Boolean {
```
frojasg: Did you consider using [Guava caches](https://github.com/google/guava/wiki/cachesexplained)? We already use it in Misk. They have the feature you need and it's using a ConcurrentMap to avoid the additional `synchronized` call.

ThePumpingLemma: Side note, we should replace all Guava caches with https://github.com/ben-manes/caffeine

staktrace: Updated to use guava caching. Presumably we can switch to caffeine across all of misk at some point, but I didn't want to introduce it here

ThePumpingLemma (approved):

PR #3797, opened 2026-05-01
fix(vitess):  prevent evicting connections on application errors

aparajon:
## Summary

- Fix `VitessExceptionHandler` incorrectly evicting healthy connections from the HikariCP pool on application-level SQL errors
- Vitess error messages typically include "vttablet" in the text (e.g. `target: keyspace.shard.primary: vttablet: rpc error: ...`), which matches the existing `Regex("vttable")` check. This causes `MUST_EVICT` for all Vitess errors, including benign ones like duplicate key (1062) and deadlock (1213) that don't indicate connection problems
- Add an `applicationErrors` set that short-circuits to `MUST_NOT_EVICT` for known application-level error codes before the regex check runs
- Downgrade log level from `error` to `warn` for application errors to reduce noise
- Add unit tests for the exception handler


aparajon:
Taken over by https://github.com/cashapp/misk/pull/3799

PR #3799, opened 2026-05-01
Rework the Vitess error handler to mostly rely on the actual error codes

eeSeeGee:
This change is mostly to be smarter about when you force a connection reset, to avoid churning connections for no good reason. Ideally this should be entirely based on the error codes, but I want to be a little conservative with this this change initially.

Co-authored by Claude etc etc

### DIFF
--- a/misk-jdbc/src/main/kotlin/misk/jdbc/VitessExceptionHandler.kt
+++ b/misk-jdbc/src/main/kotlin/misk/jdbc/VitessExceptionHandler.kt
@@ -21,11 +21,35 @@ internal class VitessExceptionHandler(registry: CollectorRegistry? = null) : SQL
   private val probablyBadState =
     setOf(
       // VT05005: typical of a connection error if we can't find the keyspace. Abort the connection
-      KnownErrors(sqlState = "42S02", errorCode = 1146),
-      KnownErrors(sqlState = "hy000", errorCode = 1105),
+      KnownErrors(sqlState = "42S02", errorCode = 1146), // VT05005: keyspace not found
+      KnownErrors(sqlState = "HY000", errorCode = 1105), // general Vitess internal error
+      KnownErrors(sqlState = "HY000", errorCode = 2013), // Lost connection to server during query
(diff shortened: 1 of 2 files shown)


misk-jdbc/src/main/kotlin/misk/jdbc/VitessExceptionHandler.kt
```diff
@@ -40,22 +64,31 @@ internal class VitessExceptionHandler(registry: CollectorRegistry? = null) : SQL
...
     return when {
       sqlException.toKnownErrors() in probablyBadState -> SQLExceptionOverride.Override.MUST_EVICT
+      sqlException.toKnownErrors() in probablySafeState -> SQLExceptionOverride.Override.DO_NOT_EVICT
       containsKnownBadErrorMessage(sqlException.message) -> SQLExceptionOverride.Override.MUST_EVICT
+      sqlException.sqlState?.uppercase()?.startsWith("HY") == true -> SQLExceptionOverride.Override.MUST_EVICT
+      sqlException.sqlState?.startsWith("21") == true -> SQLExceptionOverride.Override.DO_NOT_EVICT
+      sqlException.sqlState?.startsWith("22") == true -> SQLExceptionOverride.Override.DO_NOT_EVICT
+      sqlException.sqlState?.startsWith("23") == true -> SQLExceptionOverride.Override.DO_NOT_EVICT
+      sqlException.sqlState?.startsWith("42") == true -> SQLExceptionOverride.Override.CONTINUE_EVICT
+      sqlException.sqlState?.startsWith("08") == true -> SQLExceptionOverride.Override.MUST_EVICT
```
tgregory-block, line 74: Is this more broad than we want? Wondering why we have this in addition to the flat `probablySafeState` / `probablyBadState` strings

eeSeeGee: It's sorta lazily collected from https://vitess.io/docs/24.0/reference/errors/query-serving/. As I understand it normal mysql sql states can come through as well, so the idea is to treat this section as a "mysql catch-all"

tgregory-block (approved):

PR #3826, opened 2026-05-25
misk-jooq: add transactionOrAmbient for nested transaction reuse

tsaraiva-block:
## Summary

- Adds `JooqTransacter.transactionOrAmbient(ambient: JooqSession? = null, block)`: when `ambient` is non-null the block runs inside the supplied session (no new transaction, no retry); when `ambient` is null it delegates to `transaction` with default options.
- Callers thread the outer `JooqSession` through the call chain, so repository methods can run standalone or participate in an outer transaction's commit boundary without opening a nested one.
- `JooqTransacter.transaction` semantics are unchanged — this is strictly additive.

Coroutine-safe by construction: the active session is passed as an explicit parameter rather than tracked in a `ThreadLocal`, so suspension and dispatcher hops cannot desynchronize the lookup from the underlying JDBC connection.

## Test plan

- [x] `bin/gradle :misk-jooq:test --tests "misk.jooq.JooqTransactionContextTest"` — 5/5 pass
- [x] `bin/gradle :misk-jooq:test` — green
- [x] `bin/gradle :misk-jooq:apiDump` — only additive diff (`transactionOrAmbient` overloads on `JooqTransacter`)

Test coverage:
- Opens a new transaction when no ambient session is supplied
- Nested call reuses the supplied ambient session
- Nested writes participate in the outer transaction's commit
- Outer rollback discards nested writes
- Omitting ambient opens a new independent transaction


aparajon (approved):

jvmakine (approved):
Amazing!

aparajon (approved):

PR #3830, opened 2026-06-03
Update misk to jetty 11

staktrace:
This updates misk from jetty 10 to jetty 11. Jetty 11 is functionally the same as Jetty 10, but uses jakarta.servlet instead of javax.servlet APIs.


tgregory-block (approved):

adrw (approved):
LG

PR #3833, opened 2026-06-04
Upgrade compile target to jvm 17

staktrace:
## Description

In order to migrate misk to use newer (non-EOL) versions of jetty, we need to update the compile target to be JVM 17 or higher.

## Testing Strategy

Passing in CI

## Communication Plan

- Will announce in the `#misk-external` slack channel. Note that is a breaking change; any consumers still targeting JVM 11 will be unable to use misk from this version onwards.


adrw (approved):
LG

PR #3837, opened 2026-06-08
Fix cron log collector startup race

ThePumpingLemma:
## Why
`CronModuleTest.dependentServicesStartUpBeforeCron()` can miss the dependent service startup log if the log collector has not attached its appender yet. This makes the log-based ordering assertion flaky even when service startup ordering is correct.

## What
- Make the test-only `DependentService` depend on `LogCollectorService`
- Keep production cron service wiring unchanged

## Risk Assessment
Low — this only changes test service wiring for a single cron test to make its log capture deterministic.

## References
- Focused local test passed: `bin/gradle :misk-cron:test --tests "misk.cron.CronModuleTest.dependentServicesStartUpBeforeCron" --warn --console=plain --stacktrace --no-scan`

Generated with Amp


0x1b-xyz (approved):
GTG

PR #3840, opened 2026-06-09
feat: support Kotlin data classes in MiskWebFormBuilder

salamagd:
### Summary

Extends `MiskWebFormBuilder.calculateTypes()` to introspect Kotlin data class request types, not just Wire proto `Message` subclasses. Previously, the Web Actions admin dashboard tab only pre-populated request JSON fields for actions whose request type extended `com.squareup.wire.Message` — actions with Kotlin data class request types got an empty `types` map and the UI showed no form fields.

### What changed

**`misk/src/main/kotlin/misk/web/MiskWebFormBuilder.kt`**

- After the existing Wire `Message` check, added a branch for `requestClass.isData` that delegates to a new `calculateDataClassTypes()` method
- `calculateDataClassTypes()` mirrors the existing stack-based traversal pattern: it walks the type graph starting from the root class, using each data class's primary constructor parameters to discover fields
- `handleDataClassField()` maps Kotlin types to the same primitive type names the frontend already understands:
  - Primitives: `String`, `Int`, `Long`, `Double`, `Float`, `Boolean`, `Byte`, `Short`, `Char`
  - Binary: `ByteString`
  - Temporal: `Instant`, `Duration`, `LocalDate`
  - Enums: reuses the existing `createEnumField` helper to produce `Enum<qualified.Name,VALUE1,VALUE2>` format
  - `List<T>`: unwraps element type and marks field as `repeated`
  - `Map<K, V>`: unwraps value type and marks field as `repeated` (same as Wire map handling)
  - Nested data classes: adds field with canonical class name and pushes onto the traversal stack
  - Nullable types: handled transparently — `KType.classifier` already strips nullability

**`misk-admin/src/test/kotlin/misk/web/metadata/MiskWebFormBuilderTest.kt`**

Added 6 new test cases:
- Simple data class with primitive fields (`String`, `Int`, `Double`, `Boolean`, `Long`)
- Nested data classes (verifies recursive type discovery)
- Data class with `List` and `Map` fields (verifies repeated field handling)
- Data class with enum fields (verifies `Enum<...>` type string)
- Data class with nullable fields (verifies nullable types are treated as their non-null counterparts)
- Non-data-class, non-Wire type still returns empty map (existing behavior preserved)

### What didn't change

- All existing Wire proto handling is untouched — the data class branch only triggers when the request type is *not* a `Message` subclass
- The `Type` and `Field` data classes are unchanged
- The frontend consumes the same `Map<String, Type>` structure, so no UI changes are needed

Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>

### DIFF
--- a/misk-admin/src/test/kotlin/misk/web/metadata/MiskWebFormBuilderTest.kt
+++ b/misk-admin/src/test/kotlin/misk/web/metadata/MiskWebFormBuilderTest.kt
@@ -12,6 +12,36 @@ import misk.web.MiskWebFormBuilder.Field
 import org.assertj.core.api.Assertions.assertThat
 import org.junit.jupiter.api.Test
 
+enum class TestColor {
+  RED,
+  GREEN,
+  BLUE,
+}
+
+data class SimpleDataClass(val name: String, val age: Int, val score: Double, val active: Boolean, val count: Long)
+
+data class NestedInner(val value: String, val count: Int)
+
+data class NestedDataClass(val label: String, val inner: NestedInner)
+
+data class DataClassWithCollections(
+  val tags: List<String>,
+  val counts: List<Int>,
+  val metadata: Map<String, String>,
+  val nested: List<NestedInner>,
+)
+
+data class DataClassWithEnum(val name: String, val color: TestColor)
+
+data class DataClassWithNullables(
+  val required: String,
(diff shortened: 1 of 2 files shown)


frojasg (approved):
LGTM

PR #3842, opened 2026-06-10
Update misk to use okHttp 5.4.0

staktrace:
The new version contains additional interface elements that break consumers implementing those interfaces. A number of misk consumers in particular end up implementing Interceptor.Chain and will now need to bloat their implementations with extra stubs when all they really want to do is delegate to a misk ClientNetworkChain. This PR adds a new open base class that they can extend to facilitate that.


swankjesse (approved):
This is great. Sorry for the break.

I'd be slightly tempted to replace ‘TODO("not implemented")‘ with ‘error("not implemented")‘  so nobody expects an implementation is coming.

ThePumpingLemma (approved):

adrw (approved):
LG

PR #3853, opened 2026-06-24
Define ActionScope.Instance.withOverrides

szabado-faire:
## Description

Overriding a few key values in an action scope is a pattern we use in several places within Faire, and it'd be a substantial migration to update all the different use cases.

https://github.com/cashapp/misk/pull/3691 removed the deprecated `snapshotActionScope()` and `asImmediateValues()` methods, which together enabled the `actionScope.create(actionScope.snapshotActionScope() + overrides).inScope(block)` mechanism we used for this. The new `Instance`-based API doesn't offer an alternative approach right now

This PR adds `ActionScope.Instance.withOverrides(seedData, providerOverrides)`, which returns a new `Instance` derived from the receiver with the given overrides layered on. Merge order matches `ActionScope.create()`: existing lazy values < provider overrides < seed data. Already-initialized lazy values from the source instance are preserved, so calling
`withOverrides` on a mid-flight snapshot does not re-compute providers that have already run.

Usage:

```kotlin
actionScope.snapshotActionScopeInstance().withOverrides(seedData).inScope(block)
```

## Testing Strategy

Added four unit tests in `ActionScopedTest` covering:
- replacing seed data via `withOverrides`
- replacing an `ActionScopedProvider` via `withOverrides`
- `withOverrides()` with no arguments — equivalent to the receiver
- precedence — `seedData` wins over `providerOverrides` when both map the same key

`./bin/gradle :misk-action-scopes:test --tests 'misk.scope.ActionScopedTest'` passes locally.

## Related Work

- https://github.com/cashapp/misk/pull/3691

## Screenshots

N/A — no visual changes; pure backend API addition in `misk-action-scopes`.

## Communication Plan

- This is a strict, additive API change in `misk-action-scopes` - no existing callers are affected. No broadcast announcement planned.

## Definition of Done

- A new release is cut with this feature.

- No rollout planned. This is opt in logic.

## Checklist

### DIFF
--- a/misk-action-scopes/src/main/kotlin/misk/scope/ActionScope.kt
+++ b/misk-action-scopes/src/main/kotlin/misk/scope/ActionScope.kt
@@ -89,7 +89,9 @@ internal constructor(
 
     val immediateValues = seedData.mapValues { (_, value) -> ImmediateLazy(value) }
 
-    val lazyValues = providers.mapValues { (key, _) -> SynchronizedLazy(providerFor(key)) }
+    val lazyValues = providers.mapValues { (key, _) ->
+      SynchronizedLazy(providerFor(key)) { providerFor(key) }
+    }
 
     val lazyOverrides = providerOverrides.mapValues { (_, provider) -> SynchronizedLazy(provider) }
 
--- a/misk-action-scopes/src/main/kotlin/misk/scope/ActionScope.kt
+++ b/misk-action-scopes/src/main/kotlin/misk/scope/ActionScope.kt
@@ -210,18 +210,23 @@ internal constructor(
 
     /**
      * Returns a new [Instance] derived from this one with additional seed data and/or provider
-     * overrides layered on top. Already-initialized lazy values from this instance are preserved;
-     * entries in [seedData] and [providerOverrides] override matching keys. [seedData] takes
-     * precedence over [providerOverrides], matching the merge order of [ActionScope.create].
+     * overrides layered on top. Already-initialized lazy values from this instance are preserved
+     * and propagate to the returned instance; uncomputed lazy cells are copied so that computing
+     * them in the returned instance does not write back into this one. Entries in [seedData] and
+     * [providerOverrides] override matching keys, with [seedData] taking precedence over
+     * [providerOverrides], matching the merge order of [ActionScope.create].
      */
     @JvmOverloads
     fun withOverrides(
       seedData: Map<Key<*>, Any?> = emptyMap(),
       providerOverrides: Map<Key<*>, ActionScopedProvider<*>> = emptyMap(),
     ): Instance {
+      val isolatedLazyValues = lazyValues.mapValues { (_, lazy) ->
+        if (lazy is SynchronizedLazy && !lazy.isInitialized()) lazy.copy() else lazy
+      }
       val immediateValues = seedData.mapValues { (_, value) -> ImmediateLazy(value) }
       val lazyOverrides = providerOverrides.mapValues { (_, provider) -> SynchronizedLazy(provider) }
-      return Instance(lazyValues + lazyOverrides + immediateValues, scope)
+      return Instance(isolatedLazyValues + lazyOverrides + immediateValues, scope)
     }
   }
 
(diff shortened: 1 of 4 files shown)


misk-action-scopes/src/main/kotlin/misk/scope/ActionScope.kt
```diff
@@ -207,6 +207,22 @@ internal constructor(
...
+     * precedence over [providerOverrides], matching the merge order of [ActionScope.create].
+     */
+    @JvmOverloads
+    fun withOverrides(
+      seedData: Map<Key<*>, Any?> = emptyMap(),
+      providerOverrides: Map<Key<*>, ActionScopedProvider<*>> = emptyMap(),
+    ): Instance {
+      val immediateValues = seedData.mapValues { (_, value) -> ImmediateLazy(value) }
+      val lazyOverrides = providerOverrides.mapValues { (_, provider) -> SynchronizedLazy(provider) }
+      return Instance(lazyValues + lazyOverrides + immediateValues, scope)
```
oldergod: This reuses the receiver’s `lazyValues` objects in the new Instance. That copies the map, but not the `SynchronizedLazy` cache cells. If an override instance is the first one to read a non-overridden provider, that provider’s value is cached into the original instance too.
The new tests exercise the override instance only, so they miss the leakage back into the source/sibling instances.

szabado-faire: Very good point. I was thinking in the context of https://github.com/cashapp/misk/pull/3182 that this was desirable behaviour - I didn't think about the problematic corner cases where incorrect values bubble back up.

misk-action-scopes/src/test/kotlin/misk/scope/ActionScopedTest.kt
```diff
@@ -314,4 +314,94 @@ internal class ActionScopedTest {
...
+        override fun get(): String = "from-provider"
+      }
+
+    scope.create(seedData)
+      .withOverrides(
+        seedData = mapOf(keyOf<String>(Names.named("from-seed")) to "from-seed-direct"),
+        providerOverrides = mapOf(keyOf<String>(Names.named("from-seed")) to seedKeyProviderOverride),
+      )
+      .inScope { assertThat(foo.get()).isEqualTo("from-seed-direct and bar and foo!") }
+  }
```
oldergod, line 406:
```suggestion
  }

  @Test
  fun `withOverrides gives uncomputed providers independent state from the receiver`() {
    val injector = Guice.createInjector(TestActionScopedProviderModule())
    injector.injectMembers(this)

    val parent = scope.create(emptyMap())
    val child = parent.withOverrides()

    child.inScope { assertThat(countingString.get()).isEqualTo("Called CountingProvider 1 time(s)") }
    parent.inScope { assertThat(countingString.get()).isEqualTo("Called CountingProvider 1 time(s)") }
  }
```

oldergod (commented):
Let's get this test to pass as well please

misk-action-scopes/src/main/kotlin/misk/scope/ActionScope.kt
```diff
@@ -207,6 +209,27 @@ internal constructor(
...
+     * [providerOverrides] override matching keys, with [seedData] taking precedence over
+     * [providerOverrides], matching the merge order of [ActionScope.create].
+     */
+    @JvmOverloads
+    fun withOverrides(
+      seedData: Map<Key<*>, Any?> = emptyMap(),
+      providerOverrides: Map<Key<*>, ActionScopedProvider<*>> = emptyMap(),
+    ): Instance {
+      val isolatedLazyValues = lazyValues.mapValues { (_, lazy) ->
+        if (lazy is SynchronizedLazy && !lazy.isInitialized()) lazy.copy() else lazy
```
adrw:
```suggestion
        if (lazy is SynchronizedLazy && !lazy.isInitialized()) { lazy.copy() } else { lazy }
```

easier to read and less risky if the line gets longer and wraps at some point

adrw (approved):

jvmakine (approved):

PR #3854, opened 2026-06-25
feat: add per-transaction isolation option to misk.jdbc.Transacter

mpeyper:
## What

Adds a fluent `Transacter.isolationLevel(level)` that returns a transacter running its transactions at the given JDBC isolation level. The level is applied per transaction — set on the connection after it's acquired and before the transaction begins, then restored when the transaction returns — so it never leaks to other transactions sharing the pool.

This complements pool-level `transaction_isolation` (#3844) with an opt-in choice, and mirrors the fluent `isolationLevel()` on misk-jooq's transacter as well as misk-jdbc's own `retries()` / `noRetries()`.

### Motivation

An outbox-forwarder poll (`SELECT … FOR UPDATE SKIP LOCKED`) must run at `READ COMMITTED`. At `REPEATABLE READ` (the inherited MySQL server default) its range scan takes next-key/gap locks that can starve writer inserts and cause a same-partition-key reorder (a committed front-of-segment row gets skipped by `SKIP LOCKED`). The isolation must be set **before** the transaction begins, because MySQL's `setTransactionIsolation` issues `SET SESSION …`, which is a no-op for an already-started transaction. This replaces the fragile in-lambda `setTransactionIsolation` approach.

## Changes

- **`TransactionIsolationLevel`** gains `jdbcLevel` (the `java.sql.Connection.TRANSACTION_*` constant) used to set the level on a connection.
- **New `Transacter.isolationLevel(level: TransactionIsolationLevel)`** returning a reconfigured transacter. It's a default interface method (binary-compatible — external implementers don't have to implement it) that **throws `UnsupportedOperationException` unless overridden**, so an implementation that doesn't support per-transaction isolation fails fast rather than silently ignoring the request. The base (un-decorated) transacter inherits the pool/server default.
- **`RealTransacter`** stores the level in its `TransacterOptions` (like `maxAttempts` for `retries()`); per transaction it captures the connection's prior isolation, sets the requested level before any work runs (misk begins transactions lazily, so this lands before the txn starts), and restores it in a `finally`. Correct whether or not the pool configures `transaction_isolation`; HikariCP's reset-on-return is only a complementary safety net.

Usage: `transacter.isolationLevel(READ_COMMITTED).transactionWithSession { … }`

## Compatibility

`misk-jdbc.api` was regenerated — additive only (`isolationLevel` default interface method + `Transacter$DefaultImpls`, and `TransactionIsolationLevel.jdbcLevel`), so this is binary-backwards-compatible.

## Testing

- `DataSourceTransactionIsolationTest` — **6/6 pass**, including 3 new cases: on a single-connection pool with **no** `transaction_isolation` configured (so HikariCP's reset-on-return never fires) the level is applied, restored on return, and doesn't leak to later transactions; plus override-and-restore over a pool-configured level.
- `MySQLRealTransacterTest` — **18/18 pass**, confirming the `transactionInternal` change preserved existing behavior.

### DIFF
--- a/misk-jdbc/src/main/kotlin/misk/jdbc/Transacter.kt
+++ b/misk-jdbc/src/main/kotlin/misk/jdbc/Transacter.kt
@@ -36,18 +36,50 @@ interface Transacter {
   fun <T> transactionWithSession(work: (session: JDBCSession) -> T): T
 
   /**
+   * Like [transactionWithSession], but applies the per-transaction [options] to this single transaction only.
+   *
+   * The default implementation ignores [options] and behaves identically to [transactionWithSession]; implementations
+   * that support per-transaction options (such as [RealTransacter]) override it. See [TransactionOptions].
+   */
+  fun <T> transactionWithSession(options: TransactionOptions, work: (session: JDBCSession) -> T): T =
+    transactionWithSession(work)
+
+  /**
    * Returns a new transacter configured to retry transactions up to [maxAttempts] times on retryable exceptions.
    * Default is 3 attempts.
    */
   fun retries(maxAttempts: Int): Transacter
 
-  /**
-   * Returns a new transacter configured to not retry transactions.
-   */
+  /** Returns a new transacter configured to not retry transactions. */
   fun noRetries(): Transacter
 }
 
-class RealTransacter private constructor(
+/**
+ * Options that apply to a single [Transacter.transactionWithSession] call.
+ *
+ * Unlike [Transacter.retries]/[Transacter.noRetries], which return a reconfigured transacter, these options are scoped
+ * to the one transaction they are passed to and never affect other transactions sharing the pool.
+ */
+data class TransactionOptions
+@JvmOverloads
+constructor(
+  /**
+   * Isolation level to apply to this transaction only.
+   *
+   * It is set on the connection after it is acquired and before the transaction begins. Misk begins transactions lazily
+   * — the database transaction starts on the first statement — so the level takes effect before the transaction begins,
+   * which matters on MySQL where `setTransactionIsolation` issues `SET SESSION ...` and is a no-op for an
+   * already-started transaction. The connection's previous level is restored before it is returned to the pool, so the
+   * change never leaks to the next borrower, regardless of whether the pool configures
+   * [DataSourceConfig.transaction_isolation].
+   *
+   * When `null` (the default) the connection's isolation is left unchanged.
+   */
+  val isolationLevel: TransactionIsolationLevel? = null
+)
+
+class RealTransacter
+private constructor(
   private val dataSourceService: DataSourceService,
   private val config: DataSourceConfig?,
   private val options: TransacterOptions,
@@ -55,17 +87,14 @@ class RealTransacter private constructor(
   private val transacting = ThreadLocal.withInitial { false }
   private val exceptionClassifier = DefaultExceptionClassifier(config?.type)
 
-  constructor(dataSourceService: DataSourceService) : this(
-    dataSourceService = dataSourceService,
-    config = null,
-    options = TransacterOptions(),
-  )
+  constructor(
+    dataSourceService: DataSourceService
+  ) : this(dataSourceService = dataSourceService, config = null, options = TransacterOptions())
 
-  constructor(dataSourceService: DataSourceService, config: DataSourceConfig) : this(
-    dataSourceService = dataSourceService,
-    config = config,
-    options = TransacterOptions(),
-  )
+  constructor(
+    dataSourceService: DataSourceService,
+    config: DataSourceConfig,
+  ) : this(dataSourceService = dataSourceService, config = config, options = TransacterOptions())
 
   override val inTransaction: Boolean
     get() = transacting.get()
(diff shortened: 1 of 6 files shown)


misk-jdbc/src/main/kotlin/misk/jdbc/Transacter.kt
```diff
@@ -35,37 +35,66 @@ interface Transacter {
...
   fun <T> transactionWithSession(work: (session: JDBCSession) -> T): T
 
+  /**
+   * Like [transactionWithSession], but applies the per-transaction [options] to this single transaction only.
+   *
+   * The default implementation ignores [options] and behaves identically to [transactionWithSession]; implementations
+   * that support per-transaction options (such as [RealTransacter]) override it. See [TransactionOptions].
+   */
+  fun <T> transactionWithSession(options: TransactionOptions, work: (session: JDBCSession) -> T): T =
+    transactionWithSession(work)
```
mpeyper: Not sure if I like this or would rather other implementations implement this if they want a passthrough. It does mean the change is non-breaking though, which is nice.

misk-jdbc/src/main/kotlin/misk/jdbc/Transacter.kt
```diff
@@ -75,24 +104,29 @@ class RealTransacter private constructor(
...
+      ExponentialBackoff(
+        baseDelay = Duration.ofMillis(options.minRetryDelayMillis),
+        maxDelay = Duration.ofMillis(options.maxRetryDelayMillis),
+        jitter = Duration.ofMillis(options.retryJitterMillis),
+      )
+    val retryConfig =
+      RetryConfig.Builder(options.maxAttempts, backoff)
+        .shouldRetry { exceptionClassifier.isRetryable(it) }
+        .onRetry { attempt, e -> logger.info(e) { "JDBC transaction failed, retrying (attempt $attempt)" } }
+        .build()
```
mpeyper, line 99: Auto-formatting

misk-jdbc/src/main/kotlin/misk/jdbc/Transacter.kt
```diff
@@ -105,35 +139,51 @@ class RealTransacter private constructor(
...
-    dataSourceService = dataSourceService,
-    config = config,
-    options = options.copy(maxAttempts = maxAttempts),
-  )
+  override fun retries(maxAttempts: Int): Transacter =
+    RealTransacter(
+      dataSourceService = dataSourceService,
+      config = config,
+      options = options.copy(maxAttempts = maxAttempts),
+    )
```
mpeyper, line 157: Auto-formatting

mpeyper:
Perhaps not for this PR, but should `misk.hibernate.Transacter` also be updated for feature parity with `misk.jdbc.Transactor` (this PR) and `misk.jooq.JooqTransacter` (existing)?

ThePumpingLemma (approved):

misk-jdbc/src/main/kotlin/misk/jdbc/Transacter.kt
```diff
@@ -35,37 +35,66 @@ interface Transacter {
    */
   fun <T> transactionWithSession(work: (session: JDBCSession) -> T): T
 
+  /**
+   * Like [transactionWithSession], but applies the per-transaction [options] to this single transaction only.
+   *
+   * The default implementation ignores [options] and behaves identically to [transactionWithSession]; implementations
+   * that support per-transaction options (such as [RealTransacter]) override it. See [TransactionOptions].
+   */
+  fun <T> transactionWithSession(options: TransactionOptions, work: (session: JDBCSession) -> T): T =
```
mpeyper: should this have been a fluent function instead of an options arg like `retries`/`noRetries`?

afkelsall (approved):
Nice!

misk-jdbc/src/main/kotlin/misk/jdbc/Transacter.kt
```diff
@@ -41,31 +41,35 @@ interface Transacter {
...
+
   /**
-   * Returns a new transacter configured to not retry transactions.
+   * Returns a transacter that runs its transactions at the given isolation [level], restoring the connection's prior
+   * level afterwards so it never leaks to others sharing the pool. A `null` [level] leaves isolation unchanged.
+   *
+   * The default implementation is a no-op that returns `this`.
    */
-  fun noRetries(): Transacter
+  fun isolationLevel(level: TransactionIsolationLevel?): Transacter = this
```
mpeyper: Default implementation provided as a no-op for backwards compatibility in the API if there are any external implementations. Ideally this would not be a default but we wanted to avoid a breaking change on this one.

misk-jdbc/src/main/kotlin/misk/jdbc/Transacter.kt
```diff
@@ -41,31 +41,35 @@ interface Transacter {
...
 
-  constructor(dataSourceService: DataSourceService, config: DataSourceConfig) : this(
-    dataSourceService = dataSourceService,
-    config = config,
-    options = TransacterOptions(),
-  )
+  constructor(
+    dataSourceService: DataSourceService,
+    config: DataSourceConfig,
+  ) : this(dataSourceService = dataSourceService, config = config, options = TransacterOptions())
```
mpeyper, line 74: More auto-formatting (~or just claude formatting?~ I asked claude and it says it's a pre-commit hook doing the formatting here)

mmollaverdi (approved):

ThePumpingLemma (approved):

PR #3861, opened 2026-07-08
Add servlet filter multibind for pre-routing request modification

qnguyen-block:
## Summary

Adds a `Set<jakarta.servlet.Filter>` Guice multibind that allows Misk applications to register servlet filters that run **before** `WebActionsServlet` dispatches requests to matched routes.

### Problem

Misk's `NetworkInterceptor` runs **after** route matching — the `NetworkInterceptor.Factory.create(action: Action)` signature requires a matched action. This means interceptors cannot modify the request URL before route matching occurs.

For use cases like trailing-slash normalization (`/developers/` → `/developers`), the URL must be rewritten before Misk's `WebActionsServlet` attempts to match it against registered `@Get`/`@Post` routes. Without this, `/developers/` doesn't match `@Get("/developers")` and falls through to a catch-all 404 route.

### Solution

- Add `newMultibinder<jakarta.servlet.Filter>()` in `MiskWebModule` to initialize the empty set
- Inject `Set<jakarta.servlet.Filter>` in `JettyService`
- Register each filter on the `ServletContextHandler` with `/*` mapping and `DispatcherType.REQUEST`

This follows the same pattern already used for CORS filters in `JettyService`, which are registered via `servletContextHandler.addFilter()`.

### Usage

Applications can bind servlet filters in their Guice modules:

```kotlin
class MyModule : KAbstractModule() {
  override fun configure() {
    multibind<jakarta.servlet.Filter>().to<TrailingSlashFilter>()
  }
}
```

Where `TrailingSlashFilter` is a standard `jakarta.servlet.Filter` that wraps the request with an `HttpServletRequestWrapper` to modify the request URI before it reaches `WebActionsServlet`.

## Test plan
- [x] Verify existing Misk tests pass (no behavioral change when no filters are bound)

### DIFF
--- a/misk/src/main/kotlin/misk/web/MiskWebModule.kt
+++ b/misk/src/main/kotlin/misk/web/MiskWebModule.kt
@@ -176,6 +176,7 @@ constructor(private val config: WebConfig, private val jettyDependsOn: List<Key<
     newMultibinder<ApplicationInterceptor.Factory>()
     newMultibinder<StaticResourceEntry>()
     newMultibinder<WebProxyEntry>()
+    newMultibinder<jakarta.servlet.Filter>()
     val logContextProviderBinder = MapBinder.newMapBinder(binder(), String::class.java, LogContextProvider::class.java)
     logContextProviderBinder
       .addBinding(RequestLogContextInterceptor.MDC_HTTP_METHOD)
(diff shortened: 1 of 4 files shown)


jvmakine (approved):
Nice!

PR #3863, opened 2026-07-16
Wire 7.0.0-alpha05

oldergod:
Bumping !

https://github.com/square/wire/blob/master/CHANGELOG.md

Biggest things are one new feature with FieldMask, and some behavior changes on edge cases when unique fields would receive multiple values; should not concern normal users.


adrw (approved):

staktrace:
This is causing kotlin stdlib conflicts in downstream consumers of misk and we might need to revert it.

PR #3867, opened 2026-07-24
Keep SQS subscribers running after failures

simmons-square:
## Description

Prevent AWS2 SQS v2 subscriber coroutines from terminating when applying a returned `JobStatus` or fetching messages fails. Cancellation still propagates normally, and handler-failure metrics remain limited to handler exceptions.

## Testing Strategy

- `bin/gradle :misk-aws2-sqs:test --tests 'misk.aws2.sqs.jobqueue.SubscriberTest' --warn` (2 tests pass)
- Kotlin formatter and `git diff --check`

Medium risk because this changes failure handling in a shared subscriber. The regression tests recreate a stale receipt handle and a failed receive, then verify the same coroutine continues processing.

## Checklist

- [x] Tests cover both recovered exception paths
- [x] Review polling retry behavior described in the PR comment

Generated with Amp

### DIFF
--- a/misk-aws2-sqs/src/main/kotlin/misk/aws2/sqs/jobqueue/Subscriber.kt
+++ b/misk-aws2-sqs/src/main/kotlin/misk/aws2/sqs/jobqueue/Subscriber.kt
@@ -5,6 +5,7 @@ import io.opentracing.Tracer
 import io.opentracing.tag.Tags
 import java.time.Clock
 import java.util.concurrent.CompletableFuture
+import kotlinx.coroutines.CancellationException
 import kotlinx.coroutines.channels.Channel
 import kotlinx.coroutines.delay
 import kotlinx.coroutines.flow.flow
@@ -69,21 +70,29 @@ class Subscriber(
               }
             sqsMetrics.handlerDispatchTime.labels(queueName.value).observe((clock.millis() - startTime).toDouble())
             result
+          } catch (e: CancellationException) {
+            throw e
           } catch (e: Exception) {
             logger.warn(e) { "Handler failed for job ${job.id} from queue ${job.queueName.value}" }
             sqsMetrics.handlerFailures.labels(queueName.value).inc()
             return@withSpan
           }
-        when (result) {
-          JobStatus.OK -> deleteMessage(job)
-          JobStatus.DEAD_LETTER -> {
-            deadLetterMessage(job)
-            deleteMessage(job)
+        try {
+          when (result) {
+            JobStatus.OK -> deleteMessage(job)
+            JobStatus.DEAD_LETTER -> {
+              deadLetterMessage(job)
+              deleteMessage(job)
+            }
+            JobStatus.RETRY_WITH_BACKOFF -> retryWithBackoff(job)
+            JobStatus.RETRY_LATER -> {
+              /* no-op, will be retried after visibility timeout passes */
+            }
           }
-          JobStatus.RETRY_WITH_BACKOFF -> retryWithBackoff(job)
-          JobStatus.RETRY_LATER -> {
-            /* no-op, will be retried after visibility timeout passes */
-          }
+        } catch (e: CancellationException) {
+          throw e
+        } catch (e: Exception) {
+          logger.warn(e) { "Failed to apply status $result to job ${job.id} from queue ${job.queueName.value}" }
         }
       }
     }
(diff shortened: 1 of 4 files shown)


simmons-square:
**Polling behavior question:** v1 only handles expected long-poll timeouts inside `fetchMessages()`; other receive exceptions bubble to `RepeatedTaskQueue.scheduleWithBackoff()`, which applies its failure backoff. V2 owns the `messageFlow` loop directly, so this PR's catch immediately retries and repeated receive failures could hot-loop.

This PR intentionally limits its scope to preventing the polling coroutine from terminating. Should v2 adopt v1-equivalent receive-failure backoff in this PR, or should that be handled separately?

keeferrourke (approved):
Seems reasonable

misk-aws2-sqs/src/main/kotlin/misk/aws2/sqs/jobqueue/Subscriber.kt
```diff
@@ -69,21 +70,29 @@ class Subscriber(
...
+            }
+            JobStatus.RETRY_WITH_BACKOFF -> retryWithBackoff(job)
+            JobStatus.RETRY_LATER -> {
+              /* no-op, will be retried after visibility timeout passes */
+            }
           }
+        } catch (e: CancellationException) {
+          throw e
+        } catch (e: Exception) {
+          logger.warn(e) { "Failed to apply status $result to job ${job.id} from queue ${job.queueName.value}" }
```
mateuszmrozewski: This should also record metrics

misk-aws2-sqs/src/main/kotlin/misk/aws2/sqs/jobqueue/Subscriber.kt
```diff
@@ -69,21 +70,29 @@ class Subscriber(
...
+              deadLetterMessage(job)
+              deleteMessage(job)
+            }
+            JobStatus.RETRY_WITH_BACKOFF -> retryWithBackoff(job)
+            JobStatus.RETRY_LATER -> {
+              /* no-op, will be retried after visibility timeout passes */
+            }
           }
+        } catch (e: CancellationException) {
+          throw e
```
mateuszmrozewski: What will this ultimately lead to? Subscriber is started as Guice service. Will that cause that specific service to stop? Or will it lead to complete Misk app stop? 

Notice that in the previous try-catch when handler fails, we just jump back to the top of the loop and keep working. I think we should do the same thing here.

simmons-square: Based on what I've read about coroutines and cancellations, I don't think it's appropriate to simply "keep working" when we receive a CancellationException.

First, CancellationException happens when `cancel()` is called on the coroutine. In this case, we call cancel when the app has been signaled to stop: https://github.com/cashapp/misk/blob/3da343c2a945ed7a1b03bf134f5fe1e0b95ab678/misk-aws2-sqs/src/main/kotlin/misk/aws2/sqs/jobqueue/SqsJobConsumer.kt#L105-L109

To continue in that case means the Subscriber continues processing knowing the app is shutting down. In my experience, that's error-prone as other dependencies may also have stopped.

Normally CancellationExceptions go uncaught and propagate through to the coroutine hierarchy correctly. We're only catching and rethrowing it here because (a) we need to catch the blanket `Exception` type for AWS SDK errors and (b) we need ensure the cancellation propagates through the hierarchy correctly as noted [here](https://kotlinlang.org/docs/cancellation-and-timeouts.html#cancel-coroutines).

If you're saying we don't need to catch it *at this point in the code*, we can push these try/catch blocks down to each of the methods. In fact, that may make it easier to tabulate metrics as we'll know specifically what operation failed. `deleteMessage` already handles it, but `deadLetterMessage` and `retryWithBackoff` (the method I saw the issue with) do not. I'm happy to make that alternative change if you like.

Second, according to the `await` [documentation](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/-deferred/await.html), there are 2 reasons we may receive a CancellationException: 
1) the thread we were awaiting was cancelled
2) the coroutine was cancelled.

I can agree that it makes sense to continue if we receive a CancellationException from the thread being cancelled, so I'll add the lines from the documentation to ensure we propagate the CancellationException appropriately in this scenario.

misk-aws2-sqs/src/main/kotlin/misk/aws2/sqs/jobqueue/Subscriber.kt
simmons-square: Done

misk-aws2-sqs/src/main/kotlin/misk/aws2/sqs/jobqueue/Subscriber.kt
simmons-square: Updated

mateuszmrozewski (approved):

keeferrourke (approved):

keeferrourke (approved):

PR #3872, opened 2026-08-01
Migrate to Jackson 3.2.1

staktrace:
Bumps the Jackson BOM from 2.21.2 to 3.2.1 and moves the five Jackson-using
source files onto the `tools.jackson` packages. `jackson-annotations` is
deliberately left on `com.fasterxml` — it is not renamed in 3.x and the 3.2.1
BOM pins it to 2.22, which is what lets Jackson 2 and 3 coexist on a classpath.
`jackson-datatype-jsr310` is dropped; java.time support is folded into
databind in 3.x.

Five changes are behavioural rather than mechanical:

- Mappers are immutable in 3.x, but `SecretDeserializer` and
  `ResourceAwareDeserializer` parse nested documents with the very mapper they
  are registered on. They now take a `() -> ObjectMapper` supplier that
  resolves once `builder.build()` returns. `SecretJacksonModule` keeps its
  `ObjectMapper` constructor and exposes `mapper` as a computed property, so
  its source shape is unchanged.

- `FAIL_ON_UNKNOWN_PROPERTIES` defaults to false in 3.x. Left alone,
  MiskConfig's "'x' not found in Config, did you mean...?" warning would
  silently never fire again and config typos would be ignored. It is now
  explicitly enabled for the first parse attempt; the retry path used to
  relax the mapper in place and instead rebuilds one.

- `SORT_PROPERTIES_ALPHABETICALLY` and `EnumFeature.READ/WRITE_ENUMS_USING_TO_STRING`
  default on in 3.x and are pinned off. Sorting would reshuffle every
  service's redacted config dashboard, and config enums are matched by name
  while `toString()` is frequently overridden for display.

- `KotlinFeature.StrictNullChecks` defaults on in 3.x and is pinned off. With
  it on a null element of a collection nested inside a map is rejected even
  when that element type is declared nullable, so `Map<String, Set<String?>>`
  fails where a top-level `Set<String?>` is accepted. Config that loads today
  would stop loading.

- A `Map` field with an enum key is deserialized into an `EnumMap` in 3.x
  (databind#1853), which iterates in enum declaration order rather than in the
  order the keys appear in the YAML. There is no feature flag for it, but the
  rewrite only fires when the declared raw type is exactly `Map` and abstract
  type resolution runs first, so naming `LinkedHashMap` as the implementation
  settles it beforehand. Iteration order of a config map is observable and this
  would have changed it silently.

This is a breaking change for consumers. The affected ABI, confirmed by the
regenerated api dumps, is limited to: the `MiskConfig.load` overloads taking
`JsonNode`/`ValueDeserializerModifier`, the three `SimpleModule` subclasses in
misk-config, and `BackwardsCompatibleClientsConfigConverter`. Both misk-config
and misk expose Jackson via `api(...)`, so consumers relying on the transitive
dependency inherit Jackson 3.

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -12,7 +12,7 @@ googleAuth = "1.39.1"
 googleHttp = "2.0.0"
 guava = "33.5.0-jre"
 hoplite = "2.7.5"
-jackson = "2.21.2"
+jackson = "3.2.1"
 jooq = "3.19.29"
 junit = "5.14.2"
 kotest = "6.0.7"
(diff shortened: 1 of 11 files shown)


staktrace:
- [x] fix https://github.com/cashapp/misk/issues/3876 first, or we get silent misbehaviour

adrw (approved):
LG assuming internal test suite has passed.

staktrace:
@adrw I pushed another small update since your last review, can see it here: https://github.com/cashapp/misk/compare/6da3b4a712cd36b13ec6d8d60335e28e4f036e12..cacde3b0ccc6d64137979a996b52e6e717249904.

Assuming it looks good I'll merge this early next week

staktrace:
And one more push to update to 3.2.1 instead of 3.1.5. 3.1 is an LTS release but with renovate keeping us constantly updated it's easier to just be on the latest version.

adrw (approved):

PR #3874, opened 2026-08-03
Fix five sorted-set defects in the Redis testFixtures fake

sanket1729:
## Description

Five defects in `misk.redis.testing.FakeRedis` made it disagree with a real server, so a fake-backed test could pass on behaviour production would not reproduce: `zadd` deleted a member instead of moving it when the destination score already had a bucket, `zremRangeByRank` discarded the key's TTL, `getMinMaxIndex` clamped an empty range into a one-member range so a no-op trim removed a member, the `+inf`/`-inf` score markers were compared as raw doubles, and pipelined commands were lazy and non-memoizing.

Each fix has a contract test in `AbstractRedisTest`, which runs against both the fake and a real server. The TTL fix is the exception: `Redis` exposes no TTL getter, so observing expiry needs the fake's `FakeClock`, and its test lives in `FakeRedisTest`.


frojasg (approved):
Thanks for the change. It seems the TTL fixes are missing test coverage. Could you add test for that as well? 

Also, this change might break existing clients that rely on the broken behavior of the fake. It would be good keep an eye on the job that update the library internally

sanket1729:
Added a followup to not invalidate this ACK

frojasg (approved):
still LGTM

PR #3873, opened 2026-08-03
Add zremRangeByScore to the Redis interface

sanket1729:
## Description

Adds a typed `zremRangeByScore` to `Redis` and `DeferredRedis`, so a caller can drop sorted-set members by score range without knowing the member values stored under them.

PR #3875, opened 2026-08-04
Cover zremRangeByRank's expiry in the fake

sanket1729:
Adds the missing test coverage for the expiry fix in #3874: trimming a sorted set rebuilt the stored value with `expiryInstant = Instant.MAX`, so the key outlived the deadline its writer had set.

This can't go in `AbstractRedisTest` because `Redis` exposes no TTL getter — the only way to observe expiry is the fake's `FakeClock`, so the assertions are fake-only by necessity.

Stacked on #3873; review only the last commit.


sanket1729:
🤖 Folded into #3874, so the expiry test sits with the fix it covers. Closing.

PR #3877, opened 2026-08-05
Fail loudly when a config scalar can't be coerced to its field type

### LINKED ISSUE #3876: MiskConfig silently coerces unparseable config scalars to 0/false via ResourceAwareDeserializer
### Summary

`MiskConfig`'s `ResourceAwareDeserializer` is registered for `Int`, `Long`, `Float`, and `Boolean`. For any scalar that isn't a `${scheme:path}` reference it falls through to `JsonParser.getValueAsX()`, which is contractually lenient — it returns `0`/`false` and never throws. This replaces Jackson's strict coercion with silent data corruption: a config value that can't be converted to the target type becomes `0` or `false` instead of failing the service startup.

### The code path

[`ResourceAwareJacksonModule`](https://github.com/cashapp/misk/blob/f63011a5d9406451b3ea9daa7de09770bb4f62eb/misk-config/src/main/kotlin/misk/config/MiskConfig.kt#L335-L358) installs the deserializer on the primitive types:

```kotlin
addDeserializer(Long::class.java, ResourceAwareDeserializer<Long>(resourceLoader, mapper))
addDeserializer(Boolean::class.java, ResourceAwareDeserializer<Boolean>(resourceLoader, mapper))
// ... Int, Float, and boxed forms
```

When the scalar is not a resource reference, [`deserialize` falls through](https://github.com/cashapp/misk/blob/f63011a5d9406451b3ea9daa7de09770bb4f62eb/misk-config/src/main/kotlin/misk/config/MiskConfig.kt#L419) to the value computed at the top of the method:

```kotlin
} ?: valueAsType as? T? // Not a resource reference, return the type as is.
```

And [`valueAsTypeOrNull`](https://github.com/cashapp/misk/blob/f63011a5d9406451b3ea9daa7de09770bb4f62eb/misk-config/src/main/kotlin/misk/config/MiskConfig.kt#L486-L505) uses the lenient accessors:

```kotlin
Long::class.java,
java.lang.Long::class.java -> valueAsLong
Boolean::class.java,
java.lang.Boolean::class.java -> valueAsBoolean
```

Per jackson-core's own javadoc for `getValueAsLong()`:

> If representation cannot be converted to a long (including structured type markers like start/end Object/Array) default value of **0L** will be returned; **no exceptions are thrown**.

So a `VALUE_STRING` token that isn't parseable as a number yields `0`, the deserializer returns it as if it were the real value, and nothing surfaces. `FAIL_ON_NULL_FOR_PRIMITIVES` does not help — there's no null token involved.

### Reproduction

Parser-level, showing exactly what `valueAsTypeOrNull` sees (jackson 3.1.5, `YAMLMapper`):

```
field     token               valueAsString   valueAsLong   valueAsBoolean
num       VALUE_STRING        '5_000'         0             false
flag      VALUE_STRING        'yes'           0             false
off_flag  VALUE_STRING        'off'           0             false
good_num  VALUE_NUMBER_INT    '5000'          5000          true
good_flag VALUE_TRUE          'true'          1             true
```

For contrast, a plain Jackson mapper binding the same value **does** fail loudly, which is the behaviour these deserializers are suppressing:

```
tools.jackson.databind.exc.InvalidFormatException:
  Cannot deserialize value of type `long` from String "5_000": not a valid `long` value
```

I verified that separately against a bare `YAMLMapper`, against Misk's own tree → `toString()` → JSON-rebind path, with and without `@JvmOverloads` on the config class, and against the real config class — all four throw. The exception only disappears once `ResourceAwareDeserializer` is in play.

### Why this is worth fixing now

The bug is **pre-existing** — `valueAsTypeOrNull` is unchanged between `main` and the Jackson 3 branch, so #3872 does not introduce it. But Jackson 3 makes it far more *reachable*.

Jackson 3's YAML backend is snakeyaml-engine, which resolves per the YAML 1.2 core schema instead of YAML 1.1. Two categories of scalar that used to arrive as typed tokens now arrive as `VALUE_STRING`:

- **Underscore digit separators.** `5_000` was an integer under YAML 1.1; under 1.2 it's a string. This is not configurable — `YAMLParser._yamlResolver` is `protected final` with no `YAMLFactoryBuilder` hook.
- **`yes` / `no` / `on` / `off`.** Booleans under YAML 1.1, strings under 1.2. (Note single-letter `Y`/`N` are unaffected — SnakeYAML's `BOOL` pattern never matched them.)

Before Jackson 3, these took the `VALUE_NUMBER_INT` / `VALUE_TRUE` path and never touched the lenient accessors. After, they land in exactly the code above.

The boolean case is the one I'd worry about most: **`enabled: yes` silently becomes `false`.** A feature flag or kill switch quietly inverting is worse than a zeroed timeout, and no test or type will catch it.

### How this was found

Migrating cash-server to Jackson 3 against the #3872 snapshot. `slow_call_duration_threshold_ms: 5_000` became `0`, which surfaced as:

```
IllegalArgumentException: slowCallDurationThreshold must be at least 1[ns]
  at io.github.resilience4j.circuitbreaker.CircuitBreakerConfig$Builder.slowCallDurationThreshold
```

It surfaced **only** because resilience4j happens to validate its inputs. A repo-wide scan then found 61 underscore-separated values across 31 service configs — circuit-breaker timeouts, connection-pool sizes, a wire-transfer limit — and most of those fields have no validation, so they would have taken the zero in production without a word. We've since stripped the separators repo-wide (a no-op under YAML 1.1), but that only fixes our configs, not the failure mode.

### Suggested fix

Have `valueAsTypeOrNull` use the strict accessors (`longValue()`, `booleanValue()`, …), or return a sentinel so `deserialize` can delegate to the standard deserializer instead of swallowing the conversion failure. Either way the goal is that an unconvertible config value fails startup rather than silently becoming `0`/`false`.

Happy to put up a PR if you'd like — wanted to file the analysis first since the fix will turn some currently-"working" configs into hard startup failures, which is a deliberate call to make.

---
🤖 Filed by Kartikaya's AI agent.

staktrace:
Fixes #3876.

## The problem

`ResourceAwareDeserializer` is registered for `Int`, `Long`, `Float`, `Double`, `Byte` and `Boolean`, and when the scalar isn't a resource reference it fell back to the `JsonParser.getValueAsX()` family. Those accessors are documented to answer `0`/`false` for anything they can't parse and to **never throw**, so a malformed config value silently became a plausible-looking one:

```yaml
slow_call_duration_threshold_ms: "5_000"   # -> 0
enabled: "yes"                             # -> false
```

Plain Jackson binding rejects both (`InvalidFormatException: Cannot deserialize value of type 'long' from String "5_000"`). Only the Misk deserializer accepts them, and nothing is logged, so the service starts up with the wrong value and fails later somewhere unrelated. In our case a resilience4j circuit breaker rejected its own threshold at injector-creation time, which is a long way from the config line that caused it.

`String.toTypeOrNull`, used for environment-variable references, had the same hole via `String.toBoolean()` — that returns `false` for anything other than `"true"`, so `MY_FLAG=yes` quietly disabled the flag. The numeric branches next to it already threw.

## Why now

This is pre-existing and reproduces byte-identically on `master` today. YAML 1.2 makes it much easier to hit. snakeyaml-engine (the parser behind the Jackson 3 YAML backend, see #3872) implements the YAML 1.2 core schema, which drops YAML 1.1's underscore digit separators and its `yes`/`no`/`on`/`off` booleans. Scalars that used to arrive already typed now arrive as text and land on exactly this path. We found it migrating cash-server, where 61 config lines across 31 services silently became `0`.

## The fix

- Text is converted strictly (`toIntOrNull()` and friends); an unconvertible value raises `InvalidFormatException`. Non-text tokens — a real YAML number on a numeric field — keep their existing behaviour, so nothing that works today changes.
- `MiskConfig.load` catches `InvalidFormatException` ahead of `MismatchedInputException` and reports the property path and the offending value. Without that it falls into the "could not find property" branch, which points readers at the wrong problem: the property is present, it just can't be represented.
- Boolean failures name the likely cause, since `yes` was a boolean under YAML 1.1.
- `String.toTypeOrNull` rejects non-`true`/`false` text instead of coercing it to `false`.

One subtlety: making the conversion strict also required making it **lazy**. It was evaluated eagerly, before the resource-reference check. `"${environment:PORT}"` is not a valid `Int`, so strict-and-eager would reject references on every non-String field. Moving it after the reference check fixes that, and the pre-existing `configLoadsValuesFromEnvironmentVariables` test covers it.

## Jackson 3

Written to survive #3872: `currentToken()` is called as a method (Jackson 3 removed the `getCurrentToken()` getter), and `InvalidFormatException.from(...)` / `originalMessage` exist in both 2 and 3 under the renamed package. The test fixtures quote their values so they produce a `VALUE_STRING` token under both YAML 1.1 and 1.2 — an unquoted `5_000` is still an integer on Jackson 2 and would not reproduce.

## Verification

Reverting only `MiskConfig.kt` and re-running `:misk-config:test` gives 72 completed, 4 failed — exactly the four new tests, with no pre-existing breakage.

`testShardNonHibernate` is green: 44 test tasks executed, 0 failures. That was the run I actually wanted, since this converts previously-silent values into hard startup failures and could have broken unrelated modules' fixtures.

A side finding from the reverted run: `numbersAndBooleansWrittenAsStringsStillParse` fails on `"TRUE"` today, because Jackson 2's `getValueAsBoolean()` accepts only `"true"`/`"True"`. This PR makes it parse, matching YAML 1.2's own boolean spelling.

## Reviewer note

This is a deliberate behaviour change: config that loads today with a wrong value will now refuse to start. That's the point, but it means a service whose config has a latent bad value will fail on deploy rather than limp along. Worth deciding whether that wants a release note.


adrw (approved):
Good catch

PR #3878, opened 2026-08-05
Tolerate MySQL index/column visibility attributes in declarative schema validation

aparajon:
## Summary

`DeclarativeSchemaMigrator.requireAll()` parses declarative `.sql` files with JSqlParser to extract expected table and column names. JSqlParser (5.3, the latest release) rejects MySQL 8 index/column visibility attributes, so a declarative table file containing a bare invisible index —

```sql
KEY `idx_status` (`status`) INVISIBLE
```

— failed boot-time validation with `Failed to parse SQL`, even though MySQL and the local apply path (Spirit) both accept it. Making an index invisible before dropping it is the standard way to retire an index safely, so declarative files need to be able to express it; today the only workaround is the conditional-comment spelling `/*!80000 INVISIBLE */`, which JSqlParser skips as a comment.

Since this validation only compares table and column *names* — which visibility never affects — the fix strips bare `VISIBLE`/`INVISIBLE` attribute keywords before parsing. Backtick-quoted identifiers are left untouched, so columns named `visible`/`invisible` still parse correctly.

Added `passRequireAllWithInvisibleIndexes`, which applies a table carrying both spellings via `applyAll()` (Spirit → real MySQL) and then validates it via `requireAll()`.

Note: this is more of a short-term workaround. Raised a long-term fix in this PR: https://github.com/JSQLParser/JSqlParser/pull/2449

### DIFF
--- a/misk-jdbc/src/main/kotlin/misk/jdbc/DeclarativeSchemaMigrator.kt
+++ b/misk-jdbc/src/main/kotlin/misk/jdbc/DeclarativeSchemaMigrator.kt
@@ -114,7 +114,7 @@ internal class DeclarativeSchemaMigrator(
 
       try {
         // Parse the file content
-        val statement = CCJSqlParserUtil.parse(fileContent)
+        val statement = CCJSqlParserUtil.parse(stripVisibilityAttributes(fileContent))
 
         // Check if the parsed statement is a CREATE TABLE statement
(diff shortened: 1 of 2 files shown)


JashLal (approved):

frojasg (approved):
LGTM

PR #3892, opened 2026-08-11
Add graceful shutdown draining to aws2-sqs consumer

jayjanssen:
## The problem

When a misk service using the aws2-sqs consumer shuts down (deploy, pod eviction, node drain), `SqsJobConsumer` cancels its handling scope immediately. Jobs that are mid-handler are killed on the spot: the work is lost, the messages stay invisible until their visibility timeout expires, and on queues with a low `maxReceiveCount` they go straight to the DLQ. For a service that deploys frequently, every rollout systematically abandons whatever was in flight — in Cash App's Rey service this is the dominant source of dead-lettered jobs (`maxReceiveCount: 1`, 30-minute visibility timeout).

## What this adds

A new **optional** per-queue setting, `drain_timeout_ms`:

```yaml
aws_sqs:
  per_queue_overrides:
    my_queue:
      drain_timeout_ms: 15000
```

When set to a positive value, consumer shutdown becomes a drain:

1. Polling stops issuing new receive requests.
2. Jobs already received keep their handlers and get up to the deadline to finish and acknowledge.
3. Only work still unfinished at the deadline is cancelled.

**Default behavior is unchanged.** The field defaults to `null`; null, zero, or negative keeps the legacy immediate-cancel path, as do subscriptions whose handling scope was already cancelled (unsubscribe/reset). The constructor is `@JvmOverloads` and the binary-compat API dump is regenerated, so existing callers and configs are untouched.

The drain deadline is enforced from a dedicated single-slot dispatcher view so it cannot be starved by work saturating the shared IO dispatcher. Operators should size the value below the pod's remaining termination grace after SIGTERM (accounting for any preStop hook) and below the queue's visibility timeout.

One caveat: stopping new receives is best effort. A receive request in flight when the drain starts is cancelled, but SQS may have assigned messages to it server-side; those become receivable again after their visibility timeout. Draining removes the systematic abandonment of in-flight work; it does not make shutdown atomic with a racing receive.

This PR also hardens a pre-existing race the new tests made reproducible: the async on/off switch only gated *issuing* a receive, so a long poll in flight when the switch was disabled would still hand its messages to handlers. Receives completing after disable now release their messages back to the queue via `ChangeMessageVisibility(0)` instead.

## Metrics

New counters/histograms for observing drain behavior: drains started, in-flight jobs at drain start, jobs completed during drain (labeled by handler result), jobs cancelled at deadline, drain duration (labeled completed/timeout), receive attempts after drain, and drain-scoped ack successes/failures.

## Testing

- `SqsJobConsumerDrainTest`: 8 integration tests against a real SQS container (drain completes early, deadline enforcement, opt-out queues keep legacy behavior, unsubscribe path, mixed-queue services, metrics).
- `SubscriberTest`: unit coverage of the drain path and the async-switch race (no handoff after disable, visibility zeroed).
- `SqsConfigTest`: per-queue override resolution for the new field.
- README documentation and regenerated API dump included.

### DIFF
--- a/misk-aws2-sqs/README.md
+++ b/misk-aws2-sqs/README.md
@@ -271,6 +271,24 @@ Refer to the "threading model" section below for in-depth description.
     the caller's account. This setting is only needed for queues defined outside of your AWS account 
     (external queues)
 
+### Shutdown settings
+
+* `drain_timeout_ms`
+  * default value: null
+  * when set to a positive value, consumer shutdown drains the queue's in-flight work gracefully:
+    polling stops issuing new receive requests (cancelling any receive already in flight), jobs that
+    were already handed off keep their handlers and may finish and acknowledge within this deadline,
+    and only after the drain completes or times out does the service report stopped
+  * when null, zero, or negative, shutdown cancels polling and handling immediately (the legacy
+    behavior); in-flight messages stay invisible until their visibility timeout expires and are then
+    redelivered (or dead-lettered once the queue's maximum receive count is exhausted)
+  * choose a value lower than the termination grace your process actually has when it receives
+    SIGTERM (account for any Kubernetes `preStop` hook that runs first) and lower than the queue's
+    `visibility_timeout`; visibility is not extended while draining
+  * stopping new receives is best effort: SQS may have assigned messages to a receive request that
+    was cancelled at the drain instant; those messages become receivable again after their
+    visibility timeout
+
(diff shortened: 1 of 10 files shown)


jayjanssen:
🤖 Closing: superseded by #3895 (now merged), which incorporated this PR's in-flight receive cancellation behind a shutdown grace period. Real-SQS validation of the merged approach: https://github.com/squareup/jayj-Notebook/tree/main/2026-08/real-sqs-validation

PR #3893, opened 2026-08-11
Add atomic Redis transaction blocks

sanket1729:
## Why
misk-redis exposes Jedis transactions whose pooled connection lifetime is unsafe, while its test fixture fake cannot exercise atomic multi-command updates.

## What
- Add a Misk-owned `transaction { ... }` API that retains one connection through `MULTI/EXEC`
- Reuse deferred Redis commands for transaction blocks and preserve deferred replies
- Accept only known Jedis `Pipeline`, `ClusterPipeline`, and `Transaction` queue types
- Queue fake transaction commands and commit them atomically, discarding them if block construction fails

## Risk Assessment
Low — this adds an opt-in standalone Redis API. Existing pipeline behavior is unchanged, unknown queue types fail fast, and Redis Cluster transactions remain unsupported.

Generated with Codex

---
**Update Aug 11, 14:08 PDT:** Added a shared transaction contract for real and fake Redis
- Run identical order-book replacement and failed-block tests against both implementations
- Risk remains low: this follow-up changes test structure only.


jvmakine (approved):
Nice!

PR #3894, opened 2026-08-12
Drain aws2-sqs consumers on shutdown by stopping the polling loops

jayjanssen:
## What this is

A simpler alternative to #3892, following @mateuszmrozewski's feedback in [#misk-discuss](https://sq-block.slack.com/archives/C01M5M297T7/p1786572865121529?thread_ts=1786456653.714679&cid=C01M5M297T7): rather than cancelling scopes and spinning up new ones to track in-flight work, shutdown now *breaks the loops* and waits for the coroutines to finish, in the spirit of the v1 `SqsJobConsumer`'s `ExecutorService.shutdown()`/await approach.

If maintainers prefer this approach I'll close #3892 in favor of this PR.

## The problem (same as #3892)

When a misk service using the aws2-sqs consumer shuts down (deploy, pod eviction, node drain), `SqsJobConsumer` cancels its scopes immediately. Jobs mid-handler are killed: the work is lost, the messages stay invisible until their visibility timeout expires, and on queues with a low `maxReceiveCount` they go straight to the DLQ. In Cash App's Rey service this is the dominant source of dead-lettered jobs (`maxReceiveCount: 1`, 30-minute visibility timeout), and it also causes re-processing that burns scarce external API rate limits.

## How it works

The config surface is the same as #3892 — an optional per-queue `drain_timeout_ms`, defaulting to null (legacy immediate cancellation):

```yaml
aws_sqs:
  per_queue_overrides:
    my_queue:
      drain_timeout_ms: 15000
```

The mechanism is now just the natural completion of the existing loops:

1. `doStop()` sets a flag on each draining subscriber (`Subscriber.stopPolling()`).
2. The `messageFlow` loops check the flag instead of looping forever (`while (!stopPollingRequested)`). A receive already in flight completes normally and its messages are still handed to the handlers — the "at least one more roundtrip" — rather than being left stranded in the visibility window.
3. When the flows complete, `poll()` closes the channel.
4. `run()` handles whatever the channel still holds and returns when the channel is closed and empty (`receiveCatching`).
5. `doStop()` joins the polling and handling jobs with `withTimeoutOrNull(drain_timeout_ms)`; only work still unfinished at the deadline is cancelled. `notifyStopped()` fires once all queues have drained or timed out, so `ServiceManager` ordering still holds.

No new scopes are launched while others are cancelled, no in-flight tracking structures, no drain metrics; the only bookkeeping added is keeping the `Job` handles that `subscribe()` already creates so `doStop()` can join them.

**Default behavior is unchanged**: queues without `drain_timeout_ms` (and unsubscribed/reset ones) keep the immediate-cancel path, and `notifyStopped()` is synchronous when nothing drains.

## Trade-offs vs #3892

- The in-flight receive is allowed to finish rather than being cancelled, so draining an *idle* queue can take up to `wait_timeout` (bounded by `drain_timeout_ms`). Services that deploy often may want a lower `wait_timeout` alongside the drain; the README documents this.
- Because receives are never cancelled mid-drain, the racy "receive completes after shutdown" window that #3892 patched with `ChangeMessageVisibility(0)` doesn't arise during a drain — those messages are simply handled.
- Draining only helps on orderly shutdown; a crashed process still waits out visibility timeouts (inherent to SQS semantics).

## Testing

- `SqsJobConsumerDrainTest` (real SQS container): in-flight job finishes and is acknowledged with nothing left in the visibility window; deadline is enforced and shutdown terminates predictably; no message is stranded invisibly mid-handoff; default keeps legacy behavior; mixed drain/legacy subscriptions; idle drain terminates promptly; stop after unsubscribe doesn't burn the deadline.
- `SubscriberTest`: `stopPolling()` lets the in-flight receive finish, hands off its messages, and closes the channel; `run()` drains the channel and completes on close.
- `SqsConfigTest`: `drain_timeout_ms` default and per-queue override/inheritance.
- README documentation and regenerated API dump included.

## Adversarial review

An adversarial model review of this branch surfaced two issues that are now fixed: `subscribe()` after shutdown began would create a subscription invisible to `doStop()` (now rejected with `IllegalStateException`), and the no-new-receives test could pass even with post-shutdown receives (now deterministic with exact assertions).

One acknowledged trade-off it flagged: adding a field to the `SqsQueueConfig` data class changes the JVM descriptors of Kotlin's `copy`/`copy$default` and the synthetic defaults constructor. `@JvmOverloads` preserves the positional constructors, and this matches how fields have previously been added to these config classes (the regenerated API dump reflects it), but callers compiled against the old `copy` signature would need a recompile.

### DIFF
--- a/misk-aws2-sqs/README.md
+++ b/misk-aws2-sqs/README.md
@@ -254,6 +254,14 @@ Refer to the "threading model" section below for in-depth description.
     If no value is provided, queue settings will be used
   * note: if the value is smaller than typical processing time, it may result in the job being processed
     multiple times
+* `drain_timeout_ms`
+  * default value: null
+  * when set to a positive value, consumer shutdown drains the queue's in-flight work instead of cancelling
+    it: polling stops issuing new receives, jobs already received get up to this deadline to finish and
+    acknowledge, and only work still unfinished at the deadline is cancelled
+  * when null (the default), zero or negative, shutdown cancels all in-flight work immediately (legacy behavior),
+    leaving the messages invisible until their visibility timeout expires
+  * see "Graceful shutdown draining" below for sizing guidance
 
 ### Connection settings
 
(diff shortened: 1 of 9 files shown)


jayjanssen:
🤖 Fixed the flaky `drain stops new receives and leaves no message invisibly stranded` test (b40953faa). The test assumed the poll pipeline parks at exactly 2 of 3 messages received, but CI and a run against real AWS SQS both showed the pipeline can legitimately prefetch the third message: the `merge` over (queue, retryQueue) decouples the receive loop from the rendezvous handoff, and the drain deliberately lets an in-flight receive complete. The test now asserts scheduling-independent invariants instead of an exact split: nothing left stranded in the visibility window, conservation (every message either fully handled or still visible for the replacement pod, handled >= 2), and proof that polling stopped (a still-visible message remains visible well past the receive wait_timeout). No production code changes.

jayjanssen:
🤖 Closing in favor of #3895, which merges the approaches from this PR and #3892 (structured-concurrency drain plus bounded in-flight receive cancellation). Real-SQS validation results for both branches: https://github.com/squareup/jayj-Notebook/tree/main/2026-08/real-sqs-validation

PR #3895, opened 2026-08-14
Graceful sqs shutdown

mateuszmrozewski:
Shuts the SQS job consumer down gracefully instead of cancelling it out from under in-flight work.

Co-authored with @jayjanssen

## The problem

`doStop()` used to `scope.cancel()` the polling scope and cancel every handling scope. Jobs that a handler had already picked up were killed mid-flight: not acknowledged, not dead-lettered, just abandoned until their visibility timeout expired and SQS redelivered them. Every deploy paid for that in duplicate work and latency.

## What changes

**Draining shutdown.** `SqsJobConsumer` now tracks each subscription's polling job and handling jobs. `doStop()` tells the subscriber to stop polling, waits for the poller to finish, then joins the handlers so jobs already in progress are acknowledged or dead-lettered normally before the service goes down.

**Cancelling in-flight receives.** Stopping the poller only prevented *new* receives from being issued. A `ReceiveMessage` already parked in `await()` kept the poller alive until its long poll expired — up to 20s, paid per queue, since `doStop()` joins subscriptions in turn. `Subscriber` now tracks the in-flight receive futures and exposes `cancelInFlightReceives()` to abort them (the AWS SDK propagates cancellation down to an HTTP request abort).

Cancelling is the escalation, not the first move. `doStop()` gives the poller a 1s grace period to wind down on its own and only cancels if it overruns. A long poll that is holding messages returns immediately, so the grace period covers the receive that is about to deliver; one still waiting out its `wait_timeout` has nothing to lose by being cancelled. Cancelling a receive that had already been served would leave those messages invisible until their visibility timeout expired.

Two supporting fixes fell out of this:

- `isRunning` is now `@Volatile`. It is written from the shutdown thread and read on the polling dispatcher, with no barrier between them.
- The receive path catches `CancellationException` explicitly. `java.util.concurrent.CancellationException` extends `IllegalStateException`, so the existing `catch (Exception)` would have logged a warning and incremented `sqsReceiveFailures` on every shutdown. Genuine coroutine cancellation still propagates via `ensureActive()`, and an unexpected cancel while still running is still counted as a failure and retried.

## Tests

`SubscriberTest` covers the three paths: `stop()` alone lets an in-flight fetch deliver its messages and then closes the channel; `cancelInFlightReceives()` ends polling without cancelling the coroutine and without recording a receive failure; and a fetch issued after a sweep is cancelled immediately, which is the registration race.

The Docker-backed integration tests exercise the escalation path against a real SQS endpoint, and they are back on 20s long polling now that shutdown no longer has to wait it out. Full module suite: 69 tests, 0 failures, ~1m17s.

### DIFF
--- a/misk-aws2-sqs/src/main/kotlin/misk/aws2/sqs/jobqueue/SqsJobConsumer.kt
+++ b/misk-aws2-sqs/src/main/kotlin/misk/aws2/sqs/jobqueue/SqsJobConsumer.kt
@@ -16,6 +16,7 @@ import kotlinx.coroutines.channels.Channel
 import kotlinx.coroutines.joinAll
 import kotlinx.coroutines.launch
 import kotlinx.coroutines.runBlocking
+import kotlinx.coroutines.withTimeoutOrNull
 import misk.aws2.sqs.jobqueue.config.SqsQueueConfig
 import misk.inject.AsyncSwitch
 import misk.jobqueue.QueueName
@@ -23,6 +24,7 @@ import misk.jobqueue.v2.JobConsumer
 import misk.jobqueue.v2.JobHandler
 import misk.logging.getLogger
 import misk.testing.TestFixture
+import kotlin.time.Duration.Companion.seconds
 
 /**
  * Instruments queue consumption.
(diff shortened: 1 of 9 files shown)


misk-aws2-sqs/src/main/kotlin/misk/aws2/sqs/jobqueue/SqsJobConsumer.kt
```diff
@@ -83,31 +86,50 @@ constructor(
...
 
   override fun doStart() {
     notifyStarted()
   }
 
   override fun doStop() {
-    scope.cancel()
-    handlingScopes.values.forEach { it.cancel() }
+    logger.info("Stopping job consumer")
+    runBlocking(scope.coroutineContext) {
```
mateuszmrozewski, line 112: This blocks the current thread which allows us to wait for the polling and handling jobs to finish

misk-aws2-sqs/src/main/kotlin/misk/aws2/sqs/jobqueue/Subscriber.kt
```diff
@@ -192,12 +197,15 @@ class Subscriber(
       } else {
         messageFlow(queueName)
       }
-      .collect { received -> channel.send(received) }
+      .collect { received ->
+        channel.send(received)
+      }
+    channel.close()
```
mateuszmrozewski, line 219: Closes the channel when the polling loop is stopped.

misk-aws2-sqs/src/main/kotlin/misk/aws2/sqs/jobqueue/Subscriber.kt
```diff
@@ -211,6 +219,7 @@ class Subscriber(
         wasDisabled = false
       }
       val startTime = clock.millis()
+      logger.info("receiving")
```
mateuszmrozewski: Just a debug for now, but proves that the client hangs on the long poll, up to 20 seconds with default settings.

keeferrourke (approved):
LGTM

PR #3906, opened 2026-08-27
misk-jooq: fix render mapping using wrong database with test pools

ThinTim:
## Description

When a DatabasePool renames the database for test isolation (e.g. my_db → my_db__20260827__1), jOOQ's render mapping was still using the original config database name. This caused queries to reference the wrong schema, making jOOQ incompatible with SHARED_TEST_DATABASE_POOL.

## Testing Strategy

Added unit tests verifying mismatch in database name used by jOOQ queries when shared pool is used, then changed config source to correct it.

## Checklist


hugomd (approved):
Nice fix!

jvmakine (approved):

PR #3910, opened 2026-09-01
[HOOD-4502] misk-admin: make Web Actions tab deeplinkable via URL hash slug

canrozanes:
## Summary

The `/_admin/web-actions/` tab keeps the selected action only in component state, so a selection cannot be shared as a link. This change makes each action deeplinkable:

- When you select an action, the app writes a readable slug to the URL hash of the top window (e.g. `/_admin/web-actions/#MyAction`) via `history.replaceState`.
- When the page loads with a hash, the app resolves the slug and selects the matching action.
- When the hash changes on an open page (a URL-bar edit or back/forward navigation), a `hashchange` listener selects the matching action live, without a reload.

## Slug format

The slug is the action's simple class name. If two routes share a class name, the HTTP method is appended (`MyAction:GET`). If they still collide, the path is appended (`MyAction:GET:/my/path`). The slug index is built deterministically from the fetched metadata, so read and write always agree.

## Why the top window

The React app is served from `/_tab/web-actions/index.html` inside a same-origin iframe on the dashboard, so the address bar belongs to the top window. The code reads and writes `window.top.location`, and falls back to the current window if the top window is cross-origin. No server-side (Kotlin) change is needed.

## Testing

- Added unit tests for the slug index (unique names, method collision, path collision).
- `npm test`: 20 passed. `eslint` and `prettier` pass on the changed files. `webpack --mode production` builds.
- Verified in a browser against the `samples/exemplar` service dashboard at `localhost:8080`:
  - Loading `/_admin/web-actions/#DevCheckReloadAction` auto-selects the action.
  - Selecting a different action updates the URL to `#CronTabRunAction`.
  - Reloading with that URL restores the selection.
  - Editing the hash in the URL bar on an open page switches the selection live.
  - The browser back button restores the previous selection.

## Demo
https://github.com/user-attachments/assets/9d09c451-2985-495d-aee9-cf504e3e0584

### DIFF
--- a/misk-admin/web-actions/src/web-actions/ui/EndpointSelection.tsx
+++ b/misk-admin/web-actions/src/web-actions/ui/EndpointSelection.tsx
@@ -10,6 +10,7 @@ import RealMetadataClient from '@web-actions/api/RealMetadataClient';
 import { appEvents, APP_EVENTS } from '@web-actions/events/appEvents';
 import {
   buildSlugIndex,
+  clearSlugFromUrl,
   onSlugChange,
   readSlugFromUrl,
   writeSlugToUrl,
@@ -181,6 +182,8 @@ export default class EndpointSelection extends React.Component<Props, State> {
       const slug = this.slugIndex?.slugByRoute.get(value.value);
       if (slug !== undefined) {
         writeSlugToUrl(slug);
+      } else {
+        clearSlugFromUrl();
       }
       appEvents.emit(APP_EVENTS.ENDPOINT_SELECTED, value.value);
     }
--- a/misk-admin/web-actions/src/web-actions/ui/EndpointSelection.tsx
+++ b/misk-admin/web-actions/src/web-actions/ui/EndpointSelection.tsx
@@ -8,6 +8,12 @@ import Select, {
 import { MiskRoute } from '@web-actions/api/responseTypes';
 import RealMetadataClient from '@web-actions/api/RealMetadataClient';
 import { appEvents, APP_EVENTS } from '@web-actions/events/appEvents';
+import {
+  buildSlugIndex,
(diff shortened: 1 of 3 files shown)


misk-admin/web-actions/src/web-actions/ui/EndpointSelection.tsx
```diff
@@ -127,6 +178,10 @@ export default class EndpointSelection extends React.Component<Props, State> {
 
   handleChange = (value: OnChangeValue<EndpointOption, false>) => {
     if (value) {
+      const slug = this.slugIndex?.slugByRoute.get(value.value);
```
salamagd, line 182: 🤖 [P2] Clear the hash when the selected route has no slug. In the ambiguous-route case that `buildSlugIndex` intentionally leaves unindexed, this branch does nothing, so any hash from the previously selected action remains in the address bar. Copying or reloading that URL then points to the previous action rather than the current one. Please remove the existing fragment in the `slug === undefined` path.

salamagd (approved):
🤖 Reviewed the deeplink lifecycle, slug disambiguation, top-window history handling, and tests. Approved with one P2 inline comment about clearing a stale hash when an ambiguous route has no slug.

misk-admin/web-actions/src/web-actions/ui/EndpointSelection.tsx
canrozanes: 🤖 Updated by Codex 🤖

salamagd (approved):
🤖 Re-reviewed the latest change. The stale-hash issue is resolved: selections without a unique slug now clear the fragment while preserving the path, query string, and existing history state. The focused test covers the regression.

PR #3916, opened 2026-09-03
Prepopulate web-actions request body from last submit

canrozanes:
## Why
In the admin Web Actions tab, the request editor always starts from a generated skeleton. Users who call the same action repeatedly retype or re-paste the same JSON body each time.

## What
- Save the last JSON body for each action (keyed by HTTP method + path) to localStorage after a submit
- Prepopulate the editor with the saved body when the action is selected again; fall back to the generated skeleton
- Add a reset button to the editor that restores the skeleton and clears the saved entry
- Skip saving for GET actions, empty bodies, and bodies larger than 100 KB; invalid JSON never reaches the save because validation throws first

## Risk Assessment
Low. The change is limited to the admin dashboard Web Actions UI; localStorage failures are caught and fall back to current behavior.

Linear: HOOD-4592

Generated with [Amp](https://ampcode.com)

### DIFF
--- a/misk-admin/web-actions/src/web-actions/hooks/useSubmitRequest.ts
+++ b/misk-admin/web-actions/src/web-actions/hooks/useSubmitRequest.ts
@@ -5,6 +5,7 @@ import {
   JsonValidationError,
 } from '@web-actions/services/ApiService';
 import { APP_EVENTS, appEvents } from '@web-actions/events/appEvents';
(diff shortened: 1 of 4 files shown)


frojasg (approved):
LGTM

not sure if this can break someone flow that depends on the default values being set, but happy to try it out

PR #3917, opened 2026-09-03
Scope saved web-actions request bodies by action and expire after 7 days

canrozanes:
## Why
Review follow-up from #3916. The storage key used only HTTP method + path, but Misk distinguishes routes by method + path + action name, so two actions on the same path could overwrite each other's saved bodies. Saved bodies also stayed in localStorage indefinitely, which retains sensitive test data longer than needed.

## What
- Include the qualified action name in the storage key
- Wrap the stored value in `{savedAt, body}` and expire entries after 7 days; expired or corrupt entries are removed on read
- Bump the key prefix to `v2` and remove orphaned `v1` entries on save
- Add tests for key collision, expiry, legacy cleanup, and corrupt entries

## Risk Assessment
Low. The change is limited to the admin dashboard Web Actions UI. The worst failure mode is a lost saved body, which falls back to the generated skeleton.

Linear: HOOD-4592

Generated with [Amp](https://ampcode.com)

### DIFF
--- a/misk-admin/web-actions/src/web-actions/storage/RequestBodyStore.ts
+++ b/misk-admin/web-actions/src/web-actions/storage/RequestBodyStore.ts
@@ -1,21 +1,75 @@
 import { MiskRoute } from '@web-actions/api/responseTypes';
 
-const KEY_PREFIX = 'web-actions.lastBody.v1';
+const KEY_PREFIX = 'web-actions.lastBody.v2';
+const LEGACY_KEY_PREFIXES = ['web-actions.lastBody.v1'];
 const MAX_BODY_LENGTH = 100 * 1024; // 100 KiB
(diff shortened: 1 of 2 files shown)


salamagd (approved):

misk-admin/web-actions/src/web-actions/storage/RequestBodyStore.ts
```diff
@@ -1,30 +1,85 @@
...
+  savedAt: number;
+  body: string;
+}
 
 function storageKeyFor(route: MiskRoute): string {
-  return `${KEY_PREFIX}::${route.httpMethod} ${route.path}`;
+  return `${KEY_PREFIX}::${route.httpMethod} ${route.path} ${route.actionName}`;
+}
+
+function removeStaleEntries() {
```
frojasg, line 17: oof, but I guess it works

frojasg (approved):

PR #3920, opened 2026-09-09
misk-grpc: bound inbound gRPC message size to prevent decompression-bomb OOM

jvmakine:
## Problem

Misk's gRPC decode path streams inbound messages through an **uncapped gzip inflater** (`GzipGrpcDecoder` → bare `okio.GzipSource`) and then lets Wire's `ProtoAdapter.decode` buffer a length-delimited field whose declared length is client-controlled and unbounded. There is no `maxInboundMessageSize` equivalent anywhere on the path (`WebConfig` only caps the HTTP request line/headers).

As a result, a small gzip-compressed gRPC frame (~1.5 MB on the wire) can declare a ~1.5 GiB field backed by inflatable zeros; decoding it forces the server to buffer the whole inflated field into heap, producing an `OutOfMemoryError`. The amplification ratio (~1000:1) is tunable, so a request of roughly `heap / 1000` bytes exhausts any fixed heap.

The decode runs **before** the authorization interceptor:
- `GrpcFeatureBinding.beforeCall` reads/decodes the message (`BoundAction` runs `beforeCall` before the application-interceptor chain, and `AccessInterceptor` is an `ApplicationInterceptor`), so the cost is paid pre-auth on any non-streaming gRPC action.
- The reflection endpoint is auto-bound and `@Unauthenticated`, and calls `requests.read()` directly — reachable anonymously.

Standard gRPC (e.g. grpc-java) defends against exactly this with a default 4 MiB `maxInboundMessageSize` applied to both the on-wire frame and the decompressed message; Misk's hand-rolled gRPC codec had no equivalent.

## Fix

Enforce a configurable maximum inbound message size, defaulting to gRPC's standard **4 MiB**, mirroring grpc-java's two-layer check:

1. **On-the-wire frame cap** — `GrpcMessageSource.read()` rejects a frame whose declared length exceeds the limit *before* buffering it (like grpc-java's `processHeader` `nextFrameSize` check).
2. **Decompressed-size cap** — the decoded stream is wrapped in a new `LimitedSource` that counts bytes as they are produced and aborts with `GrpcMessageTooLargeException` once the decompressed message exceeds the limit (like grpc-java's `SizeEnforcingInputStream`). This bounds heap even when a tiny compressed frame inflates enormously.

Tunable via `WebConfig.grpcMaxInboundMessageBytes`.

## Behavior change

Previously there was **no** limit; now the default is 4 MiB on both the server (inbound requests) and the client (inbound responses), consistent with the gRPC ecosystem default. Services that legitimately send or receive gRPC messages larger than 4 MiB must raise `grpcMaxInboundMessageBytes`. Flagging for reviewer input on the default.

## Testing

- New `GrpcMessageSourceSizeLimitTest`: a within-limit message still decodes; an oversized on-wire frame is rejected before buffering; a small gzip frame that inflates past the limit is rejected while decompressing (the bomb scenario) rather than OOMing.
- Existing `misk-grpc-tests` end-to-end client/server suites pass (11/11).
- Pre-existing `misk-grpc-reflect` failures (`FRAME_SIZE_ERROR` on a ~4.7 MB reflection response) reproduce identically on the base commit and are unrelated to this change.

### DIFF
--- a/misk/src/main/kotlin/misk/grpc/GrpcMessageSource.kt
+++ b/misk/src/main/kotlin/misk/grpc/GrpcMessageSource.kt
@@ -23,6 +23,8 @@ constructor(
   private val source: BufferedSource,
   private val messageAdapter: ProtoAdapter<T>,
   private val grpcEncoding: String? = null,
+  // Defaults to gRPC's standard 4 MiB maximum inbound message size.
+  private val maxMessageBytes: Long = 4L * 1024 * 1024,
 ) : MessageSource<T>, Closeable by source {
   override fun read(): T? {
     if (source.exhausted()) return null
@@ -45,9 +47,17 @@ constructor(
 
     val encodedLength = source.readInt().toLong() and 0xffffffffL
 
+    // Reject an oversized frame before buffering it, and cap the decoded (post-decompression)
+    // message so a small compressed frame cannot inflate into a heap-exhausting message.
+    if (encodedLength > maxMessageBytes) {
+      throw GrpcMessageTooLargeException(maxMessageBytes)
+    }
+
     val encodedMessage = Buffer().write(source, encodedLength)
 
-    return messageDecoding.decode(encodedMessage).buffer().use { messageAdapter.decode(it) }
+    return LimitedSource(messageDecoding.decode(encodedMessage), maxMessageBytes).buffer().use {
+      messageAdapter.decode(it)
+    }
   }
 
   override fun toString() = "GrpcMessageSource"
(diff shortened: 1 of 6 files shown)


frojasg (approved):
LGTM. Thanks for the fix!

misk/src/main/kotlin/misk/grpc/GrpcMessageSource.kt
```diff
@@ -23,6 +23,8 @@ constructor(
   private val source: BufferedSource,
   private val messageAdapter: ProtoAdapter<T>,
   private val grpcEncoding: String? = null,
+  // Defaults to gRPC's standard 4 MiB maximum inbound message size.
```
frojasg, line 26: TIL

PR #3930, opened 2026-09-15
misk-gcp: retry Spanner emulator instance and database creation

shanegautreau:
## TL;DR

Tests that use the Spanner emulator sometimes fail during startup rather than in the
test itself. The emulator answers its first health check before it is actually ready
to create an instance, so setup can fail and take the rest of the test run with it.
This change makes emulator startup wait properly and retry while the emulator is
still coming up.

---

## The failure

`GoogleSpannerEmulator.startUp()` fails intermittently on a cold container:

```
SpannerException: UNAVAILABLE: io exception
  at GoogleSpannerEmulator.createDatabase(GoogleSpannerEmulator.kt:219)
  at GoogleSpannerEmulator.doStart(GoogleSpannerEmulator.kt:189)
  at GoogleSpannerEmulator.startUp(GoogleSpannerEmulator.kt:73)
```

This is a shared test fixture, so JUnit reports the failure against whichever test
case happened to trigger service startup. The blamed case moves from run to run,
which hides the real cause and sends people looking at test bodies.
`GoogleSpannerEmulatorLifecycleTest` is already `@Disabled("Flakey test")`.

## Why it happens

1. `waitUntilHealthy()` treats one successful `listInstances()` call as proof of
   readiness, and retries it with `binaryExponentialBackoff(1L, 5L)`. Those values
   are milliseconds, so the gate can clear in well under a second.
2. `createDatabase()` then calls `createInstance(...).get()` and
   `instance.createDatabase(...)` with no retry. A transient `UNAVAILABLE` there
   propagates straight out of `startUp()`.
3. The `catch (e: SpannerException)` around `getInstance` treats `UNAVAILABLE`
   exactly like a missing instance, so a server that is not ready is read as
   "instance absent, create it", which walks into the call in point 2.
4. `startUp()` stores the exception in `startupFailure` and rethrows it on every
   later call, so one race fails every remaining test in the JVM fork.

An emulator that can answer `listInstances` is not necessarily ready to serve admin
writes. That gap between points 1 and 2 is the bug.

## The change

- Widen the `waitUntilHealthy()` backoff from milliseconds to seconds.
- Retry instance and database creation while the emulator reports itself
  unavailable. Each step reads before it writes, so the block is safe to repeat.
- Rethrow a not-ready failure from the `getInstance` and `getDatabase` catches so
  the retry handles it, instead of reading it as "absent".
- Await the `createDatabase` operation. Its `OperationFuture` was discarded, so a
  database creation that failed after `startUp()` returned escaped the retry
  entirely and left tests pointed at a database that did not exist. `createInstance`
  was already awaited, so this only removes an asymmetry.

Point 4 above needs no change of its own. Once startup stops failing transiently,
there is no failure left to cache.

Two deliberate choices, both worth a reviewer's attention:

- **The catches are not narrowed to `NOT_FOUND`.** Only `UNAVAILABLE` and
  `DEADLINE_EXCEEDED` are singled out for retry. Every other error keeps its
  existing create-path behaviour. I could not confirm which code the emulator
  returns for a missing instance, so the change avoids depending on it.
- **The admin retry is a plain loop rather than `com.github.michaelbull.retry`.**
  That library retries on any exception, and this call site needs to retry only a
  not-ready emulator and let real errors surface on the first attempt.

`isEmulatorUnavailable()` walks the cause chain because `OperationFuture.get` wraps
the Spanner error in an `ExecutionException`.

## Testing

New: `SpannerEmulatorReadinessTest`, 8 tests, no Docker and no timing. The readiness
decision and the retry loop are extracted into `SpannerEmulatorReadiness` so they can
be tested at all. `src/test` sees only the public API of `src/testFixtures`, so the
helper is public and `apiDump` records it. That is the one judgement call in this
commit, and I am happy to hear a better seam.

The tests pin two things worth pinning:

- **The failure shape that actually occurs.** `OperationFuture.get` wraps the Spanner
  error in an `ExecutionException`, so a check reading only the top-level exception
  would miss every real occurrence while still looking correct. There is a test for
  the wrapped case specifically.
- **That real errors are not masked.** A non-readiness failure surfaces on the first
  attempt rather than being retried, and exhausting the attempts throws with the last
  failure as the cause.

Also run:

- `bin/gradle :misk-gcp:compileTestFixturesKotlin` passes.
- `GoogleSpannerEmulatorTest` 4 of 4 against a warm container.
- `GoogleSpannerEmulatorTest` 4 of 4 after `docker rm -f misk-spanner-testing`, so the
  instance and database were both absent and had to be created. That is the path the
  change in point 3 touches and the only path that reaches `createDatabase`, so it
  matters that creation still works. Re-run after the extraction: 12 of 12 across both
  classes.

**What this still does not prove.** The race itself is not reproducible on demand. I
captured it once on a cold Docker daemon; the same command passed with a warm daemon,
and passed again after merely removing the container. The unit tests cover the logic
the fix turns on, not the race. Proving the race end to end would need a fake gRPC
`InstanceAdmin` returning `UNAVAILABLE` on demand, and a timing-based approximation
would risk being flaky itself, which seemed like the wrong thing to add here. If you
have seen Spanner emulator startup flakes, I would value a second opinion on whether
this matches them.


misk-gcp/src/testFixtures/kotlin/misk/cloud/gcp/spanner/SpannerEmulatorReadiness.kt
```diff
@@ -0,0 +1,74 @@
...
+/**
+ * Tells a Spanner emulator that is still coming up apart from a real failure, and retries around it.
+ *
+ * A fresh emulator container answers `listInstances` before it will serve admin writes, so [GoogleSpannerEmulator]
+ * needs to wait out an emulator that is up but not ready yet.
+ *
+ * This is public only so those retries can be tested. Service code has no reason to call it.
+ */
+object SpannerEmulatorReadiness {
+  private val logger = KotlinLogging.logger {}
```
tgregory-block: Is `getLogger` available here? If not, we should name the logger after this class another way

tgregory-block (approved):

misk-gcp/src/testFixtures/kotlin/misk/cloud/gcp/spanner/SpannerEmulatorReadiness.kt
shanegautreau: Yes, it is available here: `misk-gcp` already declares `testFixturesImplementation(project(":misk-logging"))`, and `misk-gcp/src/main` uses `getLogger<T>()` in two places. Switched to `getLogger<SpannerEmulatorReadiness>()` in b4b3ebc44, which names the logger after the class instead of the lambda's synthetic class.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

